PR #130, opened 2018-11-10
Scale down the model slightly to handle when the room is perfectly sn…

jsantell:
…ug with the model without changing framing.

Now we can see the cube:

screenshot from 2018-11-09 19-33-03

### DIFF
--- a/src/three-components/ModelScene.js
+++ b/src/three-components/ModelScene.js
@@ -35,6 +35,13 @@ const ScaleTypeNames = Object.keys(ScaleTypes).map(type => ScaleTypes[type]);
 // would be 20m x 10m if FRAMED_HEIGHT === 10.
 export const FRAMED_HEIGHT = 10;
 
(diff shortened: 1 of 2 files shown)


src/three-components/ModelScene.js
```diff
@@ -35,6 +35,13 @@ const ScaleTypeNames = Object.keys(ScaleTypes).map(type => ScaleTypes[type]);
 // would be 20m x 10m if FRAMED_HEIGHT === 10.
 export const FRAMED_HEIGHT = 10;
 
+// The model is sized to the room, and if too perfect of a fit,
+// the front of the model becomes clipped by the near plane. Rather than
+// change the near plane or camera's position (if we wanted to implement a
+// visible "room" in the future where framing needs to be precise), we increase
+// the room by a little bit so it's always slightly bigger than the model.
```
cdata: 👍

Issue #120, opened 2018-11-09
Changing background-image causes model to render strangely


jsantell:
Could be related to the empty background-image, but even changing to an empty value causes the helmet to vanish. See `background-image.html` example of toggling the image. Via https://github.com/PolymerLabs/model-viewer/pull/113#discussion_r232412248

PR #133, opened 2018-11-10
Render fixes, HDR support, PMREM for correct radiance

### LINKED ISSUE #65: Support HDR environment maps, and HDR-ify default environment map

### LINKED ISSUE #101: Investigate vanishing mesh
The reflective sphere in `background-image.html` will sometimes, but not always, when being resized or interacted with (via orbit controls) will just vanish, leaving behind its shadow.

screenshot from 2018-11-08 19-44-16

screenshot from 2018-11-08 19-47-50

jsantell:
* Fixes #101, #120 render glitches from #94/skysphere depth test change. I think this fixes all the weird rendering issues I've seen.
* Fixes CachedGLTFLoader instances sharing materials (and environment maps) resulting in race conditions to determine what envmap is displayed
* Implements HDR equirectangular loader -- can load `.hdr` files now. Fixes #65
* Adds Prefiltered-Mipmapped Radiance Environment Maps (PMREM) for provided background-image and default environment (images below)
* Add HDR samples to the examples

## Added HDR support for background-image

hdr

## Generated PMREM for LDR background-images

This generates radiance maps so that rough, yet metallic surfaces, do not reflect the environment directly, but a mipmapped approximation -- notice the difference with the spheres on the right

Before, with a series of metalness (1) spheres, with increasing roughness:

ldr

After:

ldr-pmrem

## Generated PMREM for HDR background-image

Before:

hdr2k

After:

hdr-pmrem-spheres

## Generated PMREM for default environment

Most likely less useful and probably need more improvements on the default envmap (and caching/coloring), but this is applied here as well:

Before:

default-env

After:

default-pmrem


jsantell:
Understandable if wanting to hold off landing this, but should pull out the render fixes at least!

src/three-components/CachingGLTFLoader.js
```diff
@@ -37,6 +37,20 @@ export class CachingGLTFLoader {
...
+    const model = gltf.scene ? gltf.scene.clone(true) : null;
+
+    // Materials aren't cloned when cloning meshes; geometry
+    // and materials are copied by reference. This is necessary
+    // for the same model to be used twice with different
+    // environment maps.
+    if (model) {
+      model.traverse(object => {
+        if (object.material) {
+          object.material = object.material.clone();
```
cdata: WDYT if clone let us visit each node and decide how to clone it with a callback @mrdoob ?

Could save us a double traversal, but also might be general enough for other use cases

src/three-components/ModelScene.js
```diff
@@ -261,6 +264,10 @@ export default class ModelScene extends Scene {
    */
   onModelLoad() {
     this.applyRoomSize();
+
+    // Ensure the model renders after the skysphere
+    this.model.traverse(obj => obj.renderOrder = 1000);
```
cdata: Since this is guaranteed to happen, why not do it in the loader traversal where materials are cloned?

jsantell:
Superceded by https://github.com/GoogleWebComponents/model-viewer/pull/237

PR #136, opened 2018-11-11
Update package.json in preparation for publishing

cdata:
NPM: https://www.npmjs.com/package/@google/model-viewer
Webcomponents.org: https://www.webcomponents.org/element/@google/model-viewer
Unpkg.com: https://unpkg.com/@google/model-viewer@0.0.1/

Unfortunately, Webcomponents.org picks up the elements that are defined in tests. I'm going to bug @samuelli about this next week to see if there is anything we can do.


samuelli:
@cdata: I think our preferred method here is that you don't include tests in your published binary/package.

PR #137, opened 2018-11-11
WIP Add Travis CI config

cdata:
This isn't ready for prime time yet, but this PR adds a Travis config.

Track status here: https://travis-ci.org/GoogleWebComponents/model-viewer/builds


jsantell:
Great! This seemed peculiar to me in the travis results:

```
Linux chrome 41.0 (3/0/0)             
Windows 10 firefox 62.0 (58/1/6)      
macOS 10.13 safari 12.0 (63/1/1)      
Windows 10 MicrosoftEdge 17.17134 (61/1/3)
iPhone XR Simulator (63/1/1)          
Android GoogleAPI Emulator (error)   
```

Besides the failures, the Chrome target seems very out of date

cdata:
We use it as a low water mark for Chrome support. ~10% of the Android ecosystem is KitKat and the webview on that version is eternally pegged at Chrome 41. If it becomes a problem we can change it but it's a handy canary for low-end Android device support.

cdata:
(note that we also test latest stable Chrome and Firefox locally on the Travis instance as a prerequisite to testing on Sauce)

src/test/model-viewer-element-base-spec.js
```diff
@@ -93,7 +97,7 @@ suite('ModelViewerElementBase', () => {
         elements.forEach(e => e.remove());
       });
 
-      test('only models visible in the viewport', async () => {
+      test.skip('only models visible in the viewport', async () => {
```
cdata: Try as I might, I could not get this test to pass in Edge during CI. If I do a live test on Sauce, it passes fine. I'm going to propose that we skip it until we have a better understanding of why Intersection Observer behaves the way it does on Edge.

jsantell: :+1:

jsantell:
Looks great :+1:

PR #140, opened 2018-11-11
Flip skysphere's X axis so that it matches the orientation of the environment map applied on models

jsantell:
see the examples in http://localhost:8000/examples/background-image.html and you'll see the skysphere does not match the environment map. This is now fixed:

screenshot from 2018-11-11 09-31-24


cdata:
Nice catch!

PR #149, opened 2018-11-12
Do not apply default when EnvironmentMixin connected to DOM when background-image/color defined. Fixes #141

### LINKED ISSUE #141: Race condition setting background color
It's possible for `connectedCallback` to fire after `attributesChangedCallback`, causing the background color to be reset to white.

jsantell:
The glitch link is down, but tests confirm this should be fixed

### DIFF
--- a/src/features/environment.js
+++ b/src/features/environment.js
@@ -27,6 +27,8 @@ const $currentCubemap = Symbol('currentCubemap');
 const $setEnvironmentImage = Symbol('setEnvironmentImage');
 const $setEnvironmentColor = Symbol('setEnvironmentColor');
 const $envMapGenerator = Symbol('envMapGenerator');
+const $hasBackgroundImage = Symbol('hasBackgroundImage');
+const $hasBackgroundColor = Symbol('hasBackgroundColor');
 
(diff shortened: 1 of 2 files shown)


src/features/environment.js
```diff
@@ -43,9 +45,22 @@ export const EnvironmentMixin = (ModelViewerElement) => {
...
+      // @TODO #76
+      return this.backgroundColor && this.backgroundColor !== 'null';
+    }
+
     connectedCallback() {
       super.connectedCallback();
-      this[$setEnvironmentColor]();
+
+      if (!this[$hasBackgroundImage] && !this[$hasBackgroundColor]) {
+        this[$setEnvironmentColor]();
```
cdata: Two things:

 1. Let's do this check in `update`. `update` is _always_ called at least once (and will be called at `connectedCallback` if not due to attribute change).
 2. I think the method invocation would be easier to read if it didn't rely on implicit side-effects of the implementation. Let's invoke it explicitly with the default color `this[$setEnvironmentColor](DEFAULT_BACKGROUND_COLOR)`, or maybe imagine a different name for the method.

jsantell: :+1:

PR #150, opened 2018-11-12
remove alternate Barramundi formats

smalls:
This removes ~40mb from the repo that we're not using


cdata:
Looks like one test failed due to flakiness. I've seen the failure before. Filed https://github.com/GoogleWebComponents/model-viewer/issues/151

smalls:
Looks like it passed (on the rerun?) so I'll merge.

PR #153, opened 2018-11-12
Fix texture memory leak

### LINKED ISSUE #138: Memory leak in background-image
In the example where we're toggling background-images, memory can grow significantly (discovered when working on HDR textures) by pulling in a giant texture and computing it when swapping out background images. This caused my laptop to hang a few times yesterday until tracking it down.

jsantell:
Convert TextureUtils to TextureManager such that the renderer owns themanager so new texture helper instances do not need to be created, and properly dispose of textures that are no longer being used. Fixes #138, memory leak introduced by textures.

### DIFF
--- a/src/three-components/Renderer.js
+++ b/src/three-components/Renderer.js
@@ -18,6 +18,7 @@ import {EventDispatcher, WebGLRenderer} from 'three';
 import {IS_AR_CANDIDATE} from '../constants.js';
 import {$tick} from '../model-viewer-element-base.js';
 
+import TextureUtils from './TextureUtils.js';
 import {ARRenderer} from './ARRenderer.js';
 
 const GAMMA_FACTOR = 2.2;
@@ -58,6 +59,7 @@ export default class Renderer extends EventDispatcher {
     this.renderer.gammaFactor = GAMMA_FACTOR;
 
     this[$arRenderer] = ARRenderer.fromInlineRenderer(this);
+    this.textureUtils = new TextureUtils(this.renderer);
 
     this.scenes = new Set();
     this.scenesRendered = 0;
@@ -160,5 +162,11 @@ export default class Renderer extends EventDispatcher {
       this.scenesRendered++;
     }
     this.lastTick = t;
+  }
+
+  dispose() {
+    super.dispose();
+    this.textureUtils.dispose();
+    this.textureUtils = null;
   }
 }
--- a/examples/background-image.html
+++ b/examples/background-image.html
@@ -92,8 +92,6 @@
             </template>
         </demo-snippet>
 
-        <!-- Disable demo due to memory leak @TODO #138 -->
-        <!--
         <demo-snippet>
             <h2>Cycling background image</h2>
             <template>
@@ -102,11 +100,10 @@
                     const images = ['assets/equirectangular.png', 'does-not-exist.png', ''];
                     let i = 0;
                     setInterval(() =>
-                        $('#toggle-image').setAttribute('background-image', images[i++ % 3]), 1000);
+                        $('#toggle-image').setAttribute('background-image', images[i++ % 3]), 2000);
                 </script>
             </template>
         </demo-snippet>
-        -->
 
         <div class="attribution">
           Astronaut by <a href="https://poly.google.com/user/4aEd8rQgKu2">Poly</a>,
--- a/src/features/environment.js
+++ b/src/features/environment.js
@@ -16,19 +16,15 @@
 import {Color} from 'three';
 
 import {$needsRender, $onModelLoad, $renderer, $scene, $tick} from '../model-viewer-element-base.js';
-import EnvMapGenerator from '../three-components/EnvMapGenerator.js';
-import {toCubemapAndEquirect} from '../three-components/TextureUtils.js';
-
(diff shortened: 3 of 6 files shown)


src/three-components/TextureManager.js
```diff
@@ -0,0 +1,107 @@
...
+   */
+  constructor(renderer, config) {
+    super();
+    this.config = Object.assign(
+        {},
+        {
+          cubemapSize: 1024,
+          synthesizedEnvmapSize: 512,
+        },
+        config);
```
cdata:
```javascript
const defaultConfig = { cubemapSize: 1024, synthesizedEnvmapSize: 512 };
// ...
constructor(renderer, config = {}) {
  this.config = { ...defaultConfig, ...config }
}
```

src/three-components/TextureManager.js
```diff
@@ -0,0 +1,107 @@
...
+    this.cubemapGenerator = new EquirectangularToCubemap(this.renderer);
+    this.envMapGenerator = new EnvMapGenerator(this.renderer);
+  }
+
+  /**
+   * @param {string} url
+   * @return {Promise<THREE.Texture>}
+   */
+  load(url) {
+    return new Promise((res, rej) => loader.load(url, res, undefined, rej));
```
cdata: Prefer readability over abbreviations

src/three-components/TextureManager.js
```diff
@@ -0,0 +1,107 @@
...
+ * limitations under the License.
+ */
+
+import {TextureLoader, EventDispatcher} from 'three';
+import EquirectangularToCubemap from '../third_party/three.equirectangular-to-cubemap/EquirectangularToCubemap.js';
+import EnvMapGenerator from './EnvMapGenerator.js';
+
+const loader = new TextureLoader();
+
+export default class TextureManager extends EventDispatcher {
```
cdata: Naming nit: I expect a manager to actually manage its namesake. This class doesn't seem to manage textures, but rather seems scoped to texture conversion. The name "manager" would seem more appropriate if the TextureManager also handled disposing of the textures it generates at the appropriate time, but it doesn't seem to do that.

src/three-components/Renderer.js
```diff
@@ -58,6 +59,7 @@ export default class Renderer extends EventDispatcher {
     this.renderer.gammaFactor = GAMMA_FACTOR;
 
     this[$arRenderer] = ARRenderer.fromInlineRenderer(this);
+    this.textureManager = new TextureManager(this.renderer);
```
cdata: This seems orthogonal to the `renderer` in all ways but lifetime. And the `renderer` is never disposed in practice (is that correct? should it be disposed under some set of circumstances?), so the texture manager won't be either.

What I'm driving at: is the texture manager relevant to the `renderer`, or the `environment` feature? Does it belong here?

src/three-components/TextureManager.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Returns a { equirect, cubemap } object with the textures
+   * accordingly, or null if cannot generate a texture from
+   * the URL.
+   *
+   * @see equirectangularToCubemap with regard to the THREE types.
+   * @param {string} url
+   * @return {Promise<Object>}
```
cdata: Can also resolve to `null`

src/three-components/TextureManager.js
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * The texture returned here is from a WebGLRenderCubeTarget,
+   * which is not the same as a THREE.CubeTexture, and just what
+   * the current THREE.CubeCamera uses, and has the same effect
+   * when being used as an environment map.
+   *
+   * @param {THREE.Texture} texture
+   * @param {?number} size
+   * @return {THREE.Texture}
```
cdata: Is this `@return` annotation accurate?

src/three-components/TextureManager.js
jsantell: :+1:

src/three-components/TextureManager.js
jsantell: :+1:

src/three-components/TextureManager.js
jsantell: Agreed -- unfortunately due to how textures work, it'd be difficult to manage their lifetimes here while giving out textures -- there could be a solution in the future where we have our own texture class that allows individual subscribers to relinquish the texture, and only after all consumers have relinquished does it actually dispose (if we were sharing resources).

TextureGenerator?

src/three-components/Renderer.js
jsantell: Correct, renderer is never disposed currently since all elements use the same renderer. We need one TextureGenerator-thing per Renderer, and textures, dispose-wise, are only used and GPU-cached for every WebGLRenderer that uses them

src/three-components/TextureManager.js
jsantell: yeah (explained via proceeding text)

src/three-components/TextureManager.js
cdata: WFM

src/three-components/Renderer.js
cdata: We can split these concepts out later

PR #171, opened 2018-11-13
minor changes to examples and readme

smalls:
I'm not 100% happy with the README text, but I wanted to capture something like that.

With @jsantell's changes to async load the `pbr.html` examples, I don't think we need the warning in `index.html`, so I removed that (and updated the size to reflect the removal of the Flight Helmet sample).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,8 @@
 
 `<model-viewer>` is a web component that makes rendering interactive 3D
 models - optionally in AR - easy to do, without writing any code, on as many
-browsers and devices as possible.
+browsers and devices as possible. This component will strive to have great
+defaults for rendering quality and performance.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -9,7 +9,8 @@
 
 `<model-viewer>` is a web component that makes rendering interactive 3D
 models - optionally in AR - easy to do, without writing any code, on as many
-browsers and devices as possible.
+browsers and devices as possible. This component will strive to have great
```
cdata: "`<model-viewer>` strives to give you great defaults for quality and performance."

Unrelated, but the phrase "without writing any code" is kind of weird. I would be willing to speculate that a lot of the folks who are being targeted with that phrase would consider HTML to be code.

smalls:
Ping (to unstick the cla bot).

PR #172, opened 2018-11-13
add templates and update the contributing guide

### LINKED ISSUE #114: Fix up CONTRIBUTING, add new issue templates, review code of conduct.

smalls:
Fixes #114

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,29 @@
+<!-- Instructions For Filing a Bug: https://github.com/GoogleWebComponents/model-viewer/blob/master/CONTRIBUTING.md#filing-bugs -->
+### Description
+
+Please provide a detailed description of the bug, how to reproduce it, and the
+expected behavior.  Always include a code snippet, screenshot, any errors
+reported in the console, and the model to help us understand and fix the
+problem.
+
+#### Live Demo
+<!-- glitch.me starting point (remix and edit -- must be logged in to persist!) -->
+https://glitch.com/edit/#!/model-viewer
+<!-- ...or provide your own repro URL -->
+
+### Browsers Affected
+<!-- Check all that apply and please include the version tested -->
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,29 @@
...
+<!-- ...or provide your own repro URL -->
+
+### Browsers Affected
+<!-- Check all that apply and please include the version tested -->
+- [ ] Chrome
+- [ ] Firefox
+- [ ] Edge
+- [ ] Safari 11
+- [ ] Safari 10
+- [ ] IE 11
```
cdata: I would recommend that we don't suggest support for Safari < 11. The Polymer team rule is that we support back to ($LATEST_VERSION - 1) of all major browsers, and IE11.

Firefox 62 is an interesting case because it is the last version of Firefox without Web Components.

We also care about iOS / Android browsers as special cases due to AR features. Also Magic Leap's Helios.

CONTRIBUTING.md
```diff
@@ -15,9 +15,32 @@ You generally only need to submit a CLA once, so if you've already submitted one
...
+there's something we've missed and they still need to be addressed.
+
 ## Code reviews
 
 All submissions, including submissions by project members, require review. We
 use GitHub pull requests for this purpose. Consult
 [GitHub Help](https://help.github.com/articles/about-pull-requests/) for more
 information on using pull requests.
+
+Note that we request issues to be filed for all pull requests. To start, open
```
cdata: ❤️

.github/ISSUE_TEMPLATE.md
smalls: Let me drop the versions, and I'll add Helios and an OS picker.

smalls:
Yes, thanks! I've updated to include the feedback re browser versions (by making it more freeform), and I've changed to the standard code of conduct.

PR #190, opened 2018-11-15
Include error message with model load error

gravityrail:
I wasn't able to tell what the issue was with loading models (it was `THREE.GLTFLoader: Unsupported asset. glTF versions >=2.0 are supported. Use LegacyGLTFLoader instead.`). This change helped me debug the error more effectively in the console.

### DIFF
--- a/src/three-components/ModelScene.js
+++ b/src/three-components/ModelScene.js
@@ -131,7 +131,7 @@ export default class ModelScene extends Scene {
     try {
       await this.model.setSource(source);
     } catch (e) {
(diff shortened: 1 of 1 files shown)


src/three-components/ModelScene.js
```diff
@@ -131,7 +131,7 @@ export default class ModelScene extends Scene {
     try {
       await this.model.setSource(source);
     } catch (e) {
-      console.error(`Could not set model source: ${source}`);
```
cdata: Good call, thanks for the suggestion!

Issue #191, opened 2018-11-15
Undefined while building


gravityrail:
### Description

Got this issue while running `npm run build` on Node v10 on Linux:

```
$ npm run build


> @google/model-viewer@0.0.4 build /Users/dan/workspace/a8c/vagrant-local-2/www/wordpress-default/public_html/wp-content/plugins/model-viewer
> tsc && rollup -c


./lib/model-viewer.js → ./dist/model-viewer.js...
(!) Circular dependency: lib/model-viewer-base.js -> lib/three-components/Renderer.js -> lib/model-viewer-base.js
created ./dist/model-viewer.js in 1.8s

./lib/test/index.js → ./dist/unit-tests.js...
(!) Circular dependency: lib/model-viewer-base.js -> lib/three-components/Renderer.js -> lib/model-viewer-base.js
created ./dist/unit-tests.js in 1.6s

./examples/dependencies/index.js → ./examples/built/dependencies.js...
(!) `this` has been rewritten to `undefined`
https://rollupjs.org/guide/en#error-this-is-undefined
node_modules/marked/lib/marked.js
1386:   root.marked = marked;
1387: }
1388: })(this || (typeof window !== 'undefined' ? window : global));
         ^
created ./examples/built/dependencies.js in 464ms
```

cdata:
Yeah, this is somewhat out of our hands as the warning is about code in a third party library. Classically, this was a very popular mechanism to detect the global object in JavaScript, so it is very common to hit such a warning when bundling.

Thankfully it's just a warning. I don't think there is anything actionable for us, other than potentially filing issues against the `marked` library.

gravityrail:
No problem, thanks for the clarification

jsantell:
This was a new configuration feature to me, but rollup's [onwarn](https://rollupjs.org/guide/en#onwarn) may be able to fix this so we can at least hide the spewing for packages that are known to be noisy

PR #219, opened 2018-11-15
Update to three.js r98

### LINKED ISSUE #147: Upgrade to three r98
We're running r94, several months out of date, and have already ran into some things not going as expected, since I was looking at latest docs/possibilities while working on #133. The upgrade started when wanting to use the [EquirectangularToCubeGenerator](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/EquirectangularToCubeGenerator.js) found in three rather than our dependency, which [CubeCamera was only updated 24 days ago](https://github.com/mrdoob/three.js/pull/15086/files) to support additional config needed (and got stuck on material stuff).

Also, with `scene.background` supporting WebGLRenderTargetCube [as of 26 days ago](https://github.com/mrdoob/three.js/pull/15081/files), we may be able to just use the same cube-like texture generated for the model environments, rather than needing to store an equirect and placing it on a skysphere. Also we could remove a lot of skysphere logic. (Partial reason for the complexity at the moment)

A few of the third_party examples API changes that were trivial. StaticShadows were broken, and was wrestling with materials using the different EquirectangularToCubeGenerator that I suspect is unrelated to the HDR stuff, but separately, we'll need to tackle this eventually. Didn't see anything in the [migration guide](https://github.com/mrdoob/three.js/wiki/Migration-Guide) that can explain shadow breakage, possibly the render target API changes?

jsantell:
Updates three.js to r98.

* Changes to StaticShadow from `lookAt` change in r96 https://github.com/mrdoob/three.js/pull/14517
* Updated src/third_party/three/* files from r98
* Updated GLTFLoader pulls in some optional files that may exist, but not in our case via modules. Silenced these warnings in rollup. Can follow up with a fix for #191 probably.

### Reference Issue
Fixes #147


package.json
```diff
@@ -74,7 +74,7 @@
   },
   "dependencies": {
     "@polymer/lit-element": "^0.6.2",
-    "three": "^0.94.0"
+    "three": "^0.98.0"
```
cdata: 🎉

rollup.config.js
```diff
@@ -19,6 +19,28 @@ const rollup = require('rollup');
...
 const cleanup = require('rollup-plugin-cleanup');
 
+// Some third party libraries may have optional modules.
+// Ignore the warnings generated by rollup for expected
+// missing modules.
+const IGNORE_MISSING = {
+  'lib/third_party/three/GLTFLoader.js': [
+    'DDSLoader',
+    'BufferGeometryUtils',
+    'DRACOLoader',
```
cdata: Will we eventually include this for draco support?

rollup.config.js
```diff
@@ -19,6 +19,28 @@ const rollup = require('rollup');
...
+  ],
+};
+
+const onwarn = (warning, warn) => {
```
jsantell: This was new to me! https://rollupjs.org/guide/en#onwarn

src/three-components/ModelScene.js
```diff
@@ -76,21 +78,28 @@ export default class ModelScene extends Scene {
     this.model = new Model();
     this.shadow = new StaticShadow();
     this.light = new AmbientLight(0xffffff, 1);
+    this.light.name = 'AmbientLight';
```
cdata: Was `name` removed from objects at some point?

jsantell: @cdata `name` is default an empty string, so this helps some debugging

rollup.config.js
jsantell: Yeah, if we want to bundle it, but imagine we'd need a different solution. This expects `THREE.DRACOLoader` so to avoid changing the source code that will get clobbered on the next three.js update, the warnings are muted. For DRACO specifically, I think we can set that in CachedGLTFLoader, rather than modifying GLTFLoader outside of the ES Module changes

PR #227, opened 2018-11-19
Memory leak prog

### LINKED ISSUE #220: Use EquirectangularToCubeGenerator from three.js
We should swap out our [three.equirectangular-to-cubemap](https://www.npmjs.com/package/three.equirectangular-to-cubemap) generator with [EquirectangularToCubeGenerator](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/EquirectangularToCubeGenerator.js) from three.js examples collection, ensuring it's in sync with three.js updates.

Dependent on #219, and should be done prior to #133

jsantell:
With the [background-image demo](https://googlewebcomponents.github.io/model-viewer/examples/background-image.html) often open on my machine, after long periods of time, my machine would hang, and had a hunch it had to do with the demo's infinite loading of new textures. In terms of severity, this would only occur when constantly changing textures (not a realistic scenario IMO) over a long period of time as small amounts of resources were being generated and not collected.

The leak wasn't terribly severe tracked leak down to cubemap generator, figured might as well update

three.js creates a [WebGLState](https://github.com/mrdoob/three.js/blob/master/src/renderers/webgl/WebGLState.js) for every camera/scene combo rendered to store light and shadow data, and the [three.equirectangular-to-cubemap](https://github.com/spite/THREE.EquirectangularToCubemap) component we use currently creates a new camera for every texture -- resulting in a new WebGLState created and retained, and does not get disposed unless calling `dispose` on the renderer, which will have other unintended side-effects in our case since we're not yet done with our renderer.

As a part of this change, I figured it'd be a good time to update to three's [EquirectangularToCubeGenerator](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/EquirectangularToCubeGenerator.js) so we can ensure this works with the version of three we're using as a dependency (Fixes #220). Unfortunately, this generator similarly creates a new scene and camera for every texture, and not designed to handle multiple texture conversions, and the workarounds to cache these values are documented in the code.

Will file a follow up with three with changes to EquirectangularToCubeGenerator so that it can handle this scenario without creating extraneous light/shadow data, which would simplify a lot of this code when we next upgrade three.

-- 

Also included is a fuzzer page that, at random intervals, flips some values on the model. The element handles this surprisingly well, but a great test bed to ensure unexpected attribute combinations work well, and shake out any race conditions we have. Using memory tools, we can leave this run for quite some time and see that the memory doesn't grow.


examples/dependencies/index.js
```diff
@@ -15,3 +15,5 @@
 
 import '@polymer/iron-demo-helpers/demo-snippet.js';
 import '@polymer/paper-button/paper-button.js';
+import '@polymer/paper-slider/paper-slider.js';
+import '@polymer/paper-radio-group/paper-radio-group.js';
```
jsantell: I like using these nice components in our demos, but maybe these should be broken out per dependency as-needed in demos (e.g. only the PBR demo and fuzzer use paper-button)

examples/fuzzer.html
```diff
@@ -0,0 +1,205 @@
...
+    for (let attribute of attributes) {
+        if (Math.random() > attribute.probability) {
+            continue;
+        }
+        updateAttribute(attribute.name, attribute.values);
+    }
+
+    updated.style.transition = 'none';
+    updated.style.opacity = 1;
+    requestAnimationFrame(() => {
```
jsantell: This "update" should flash and fade on every change; doesn't work too well for lower values. Could remove

rollup.config.js
```diff
@@ -56,6 +56,7 @@ export default [
     input: './lib/model-viewer.js',
     output: {
       file: './dist/model-viewer.js',
+      sourcemap: true,
```
jsantell: Needed for debugging memory leaks, can remove and open another issue (valuable but we're building a lot of things per `npm run build` currently)

src/three-components/TextureUtils.js
```diff
@@ -35,10 +40,58 @@ export default class TextureManager extends EventDispatcher {
...
+   * The texture returned here is from a WebGLRenderCubeTarget,
+   * which is not the same as a THREE.CubeTexture, and just what
+   * the current THREE.CubeCamera uses, and has the same effect
+   * when being used as an environment map.
+   *
+   * @param {THREE.Texture} texture
+   * @return {THREE.Texture}
+   */
+  equirectangularToCubemap(texture) {
+    // Each instance of EquirectangularToCubeGenerator creates a new camera,
```
jsantell: This special handling could go away if we make upstream fixes in three's generator

cdata:
> In terms of severity, this would only occur when constantly changing textures (not a realistic scenario IMO)

This scenario seems like it would be more common in a WYSIWYG tool that enabled GUI-based configuration of the `<model-viewer>`. So, probably really good to fix regardless.

src/three-components/TextureUtils.js
```diff
@@ -45,21 +50,80 @@ export default class TextureManager extends EventDispatcher {
     super();
     this.config = {...defaultConfig, ...config};
     this.renderer = renderer;
-    this.cubemapGenerator = new EquirectangularToCubemap(this.renderer);
     this.envMapGenerator = new EnvMapGenerator(this.renderer);
```
cdata: Nit: initialize class field `$cubeGenerator`

src/three-components/TextureUtils.js
cdata: Is there an open issue against Three? Maybe link it in the comment with a note about when we can take this out?

src/three-components/TextureUtils.js
```diff
@@ -45,21 +50,80 @@ export default class TextureManager extends EventDispatcher {
...
+      resolution: this.config.cubemapSize,
+    });
+
+    // If this is the first generator, stub out the getShader method
+    // so we don't create unnecessary shaders; we're continuing to
+    // reuse the same camera and scene after all. Necessary to overwrite
+    // the prototype since this is called in the constructor.
+    if (!this[$cubeGenerator]) {
+      const material = generator.boxMesh.material;
+      EquirectangularToCubeGenerator.prototype.getShader = () => material;
```
cdata: Is it possible to just create an extended class rather than monkey patching on the fly here? e.g.,

```javascript
class NonLeakyEquirectangularToCubeGenerator extends EquirectangularToCubeGenerator {
  constructor(sharedShaderMaterial = null) {
    super();
    this.sharedShaderMaterial = sharedShaderMaterial;
  }

  getShader() {
    if (this.sharedShaderMaterial != null) {
      return this.sharedShaderMaterial;
    }
    return super.getShader();
  }
}
```

examples/fuzzer.html
```diff
@@ -0,0 +1,205 @@
...
+    probability: 0.2,
+    values: [],
+}, {
+    name: 'unstable-webxr',
+    probability: 0.2,
+    values: [],
+}];
+attributes.forEach(attr => attr.values.push(null, ''));
+
+function updateAttribute (attr, values) {
```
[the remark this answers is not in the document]
cdata: (No action required, just observing)

examples/fuzzer.html
jsantell: Just to indicate to the viewer that a change occurred (more interesting for longer wait times), since sometimes the attribute changes do not result in a visible change

rollup.config.js
jsantell: :+1: !

src/three-components/TextureUtils.js
jsantell: Just filed https://github.com/mrdoob/three.js/issues/15288

src/third_party/three/EquirectangularToCubeGenerator.js
```diff
@@ -0,0 +1,118 @@
+/**
+* @author Richard M. / https://github.com/richardmonette
+*/
+
+import * as THREE from 'three';
+
+const EquirectangularToCubeGenerator = function ( sourceTexture, options ) {
```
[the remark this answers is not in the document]
jsantell: Not other than the ES6 wrapper stuff, no

src/three-components/TextureUtils.js
jsantell: Good call, trying this

examples/fuzzer.html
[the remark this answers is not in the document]
jsantell: I was thinking of a rolling log of the last 10 changes or so when stopping, to debug a scenario where going from state X to Y causes visual artifact Z

examples/fuzzer.html
jsantell: Going to remove this for now

examples/dependencies/index.js
jsantell: :+1:

PR #233, opened 2018-11-22
Add legacy bundle to test old browser compatibility

### LINKED ISSUE #99: Fix IE11 (possibly as a separate bundled build)
Current status:

image

cdata:
Important details about this change:

 - IE11 is now supported via a "legacy" bundle and tested on CI
 - Splits WCT configuration into three config files due to a bug in how `wct` parses command line options (filed https://github.com/Polymer/tools/issues/798 to track)
 - Configurations cover the following territory in order:
    1. Local Chrome (quick sanity check before running slower sauce tests)
    2. Battery of "modern" browsers via sauce
    3. Battery of "legacy" browsers via sauce, using "legacy" test entry point and bundle
 - Only Chrome is tested "locally," due to Firefox not actually running any tests in this mode on CI 
   - [Headless Firefox does not support WebGL](https://bugzilla.mozilla.org/show_bug.cgi?id=1375585) and it was previously failing in a flakey way just before the 4th test
 - Chrome 41 has been removed from the testing regime, as it was suffering the same issue as headless Firefox in CI only
    - I tried every platform configuration of this browser (Windows, Mac and Linux) and it always fails to create a WebGL context. It works just fine when I run Chrome 41 locally.
 - [`polymer-build`](https://github.com/Polymer/tools/tree/master/packages/build) is used to generate a "legacy"-compatible bundle
    - The amount of configuration to do this ourselves is rather large and complicated by the need for language feature polyfills; `polymer-build` is maintained and updated by a dedicated team of folks so it saves us some of that burden.
 - IE11 doesn't have `remove` on elements, so all usages have been changed to a more conservative check for `parentNode` and then using `removeChild` on that node if it exists.
 - We were previously misconfiguring the test timeout, this has been fixed
 - For a complete build, the following browsers are currently tested:
    - Chrome 70 Desktop
    - Chrome 70 Desktop (tested against "legacy" bundle)
    - Chrome 70 Android
    - Firefox 62 (pre-Web Components)
    - Firefox 63 (latest / post-Web Components)
    - Safari 12 Desktop
    - Safari 12 iOS
    - Safari 11 Desktop
    - Edge 17
    - Internet Explorer 11 (tested against "legacy" bundle)

Fixes #99


.travis.yml
```diff
@@ -9,10 +9,8 @@ env:
 node_js: node
 addons:
   chrome: stable
-  firefox: latest
 script:
-- xvfb-run npm run test
-- if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then wct --plugin sauce; fi
+- xvfb-run npm run test && if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then ./scripts/run-sauce-tests.sh; fi
```
jsantell: Does this only run the sauce tests when a commit lands on master?

scripts/run-sauce-tests.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
```
jsantell: Why do you need to flip the chmod flags?

test.html
```diff
@@ -21,11 +21,13 @@
   <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
```
jsantell: Could we move test.html and test-legacy.html to `test/` maybe? Or `test/index.html` and `test/legacy.html`? Could also contain a README about writing tests (in the future) -- main rational being to minimize things in the root

test.html
```diff
@@ -45,13 +47,13 @@
   <!-- Resize Observer polyfill is required for non-Chrome browsers: -->
   <script src="./node_modules/resize-observer-polyfill/dist/ResizeObserver.js"></script>
 
-  <!-- Fullscreen polyfill is required to support all stable browsers: -->
+  <!-- Fullscreen polyfill is required to support Web XR-based AR in stable browsers that support it: -->
```
jsantell: nit: s/Web XR/WebXR

wct-legacy.conf.json
```diff
@@ -0,0 +1,28 @@
...
+  "plugins": {
+    "sauce": {
+      "browsers": [
+        {
+          "browserName": "internet explorer",
+          "platform": "Windows 10",
+          "version": "11.285"
+        },
+        {
+          "browserName": "safari",
```
jsantell: Does Safari 11.1 need the legacy build?

test-legacy.html
```diff
@@ -0,0 +1,60 @@
...
+<html>
+  <head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
+  <title>&lt;model-viewer&gt; Unit Tests</title>
+
+
+  <!-- The following libraries and polyfills are recommended to maximize browser support -->
+  <!-- Include prismatic.js for Magic Leap support: -->
+  <!--<script src="../node_modules/@magicleap/prismatic/prismatic.min.js"></script>-->
```
jsantell: This should probably be removed here and in `test.html` since it doesn't actually do anything when running tests AFAIK (we don't run tests on Helio or shim prismatic connection)

scripts/run-sauce-tests.sh
cdata: These aren't `chmod` flags. These set flags tell the shell to:

 1. Exit on the first non-zero exit code (otherwise it would invoke all commands
 2. Print all the commands being run by the script as they are invoked

See https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html#The-Set-Builtin

test.html
cdata: 👍

test.html
cdata: Will fix

wct-legacy.conf.json
cdata: Probably not. I'll move it up.

Now that I think about it, it's probably valuable to ensure latest Chrome works against the legacy bundle, as the code is changed very significantly.

test-legacy.html
cdata: Will fix.

cdata: Done

wct-legacy.conf.json
cdata: Done

test.html
cdata: Done

test.html
cdata: Done

cdata:
Added docs about the legacy bundle and Custom Elements ES5 Adapter

PR #234, opened 2018-11-24
Use polling when watching changed files in "npm run watch" on linux

jsantell:
Fixes #53

PR #235, opened 2018-11-24
(WIP) Split up build targets to decrease build times when watching

jsantell:
Dependent on #234


package.json
```diff
@@ -27,14 +27,17 @@
   ],
   "scripts": {
     "clean": "rm -rf ./lib ./dist",
-    "build": "tsc && rollup -c",
+    "build": "npm run build:main && npm run build:test && npm run build:examples",
+    "build:main": "tsc && rollup -c ./build/main.config.js",
+    "build:test": "rollup -c ./build/test.config.js",
+    "build:examples": "rollup -c ./build/examples.config.js",
```
cdata: It seems like both `test` and `examples` depend on `main`. Will the `watch` script always perform the appropriate build as files in test or examples change?

jsantell: `examples` should be standalone AFAIK (just builds some iron-components), but you're right this isn't quite what is expected for test (changing a test file in src/* would trigger a build of only the main component) -- I'll give this component another think

cdata:
I think the main problem here is that if I do `tsc -w` and `rollcup -c -w` in their own terminal windows, `tsc` only recompiles files that change, and then `rollup` only rebuilds the bundles that change. However, if I use `npm run dev`, the entire project is rebuilt every single time any file changes. This means that rebuilds that should take 1-2s can end up taking 10-15s.

Ideally we could make `npm run dev` somehow do the same thing as running the two build tools separately.

cdata:
We have WIP to improve this via #970 , so closing.

Issue #215, opened 2018-11-15
Add PMREM support to renderer


cdata:
[PMREM Example](https://threejs.org/examples/?q=hdr#webgl_materials_envmaps_hdr)

jsantell:
WIP in #133

Issue #200, opened 2018-11-15
Add EXR support


cdata:
This seems like the more efficient, modern version of HDR

PR #237, opened 2018-11-27
(WIP) PMREM+HDR+r99

### LINKED ISSUE #65: Support HDR environment maps, and HDR-ify default environment map

jsantell:
Fixes #65, #196, #215, enables #200
Supercedes #133

I had PMREM+HDR working in #133 again from master, but the envmap was off with the skybox again, so figured lets get this working on three r99dev that added `scene.background` support for cubemaps (https://github.com/mrdoob/three.js/pull/15187, https://github.com/mrdoob/three.js/pull/15186, https://github.com/mrdoob/three.js/pull/15081) so that we can have a good baseline for judging reflectivity correctness rather than keep flipping the skysphere.

I have PMREM+HDR with `scene.background` handling the sky rather than a skysphere mostly working, but there were some issues with cubemap encoding changes when used in `scene.background` (https://github.com/mrdoob/three.js/issues/15325), and the flipping of the reflection found in WebGLRenderTargetCubes (https://github.com/mrdoob/three.js/issues/15301) mentioned earlier. PR out for encoding changes (https://github.com/mrdoob/three.js/pull/15299), and I have a hacky way of fixing the reflection, but my in progress PR is out (https://github.com/mrdoob/three.js/pull/15302).

At this point, can the HDR+PMREM work be decoupled from the `scene.background` change? We could land just the HDR+PMREM changes, and flip the skysphere again, but it'll be inverted. We can't just land the `scene.background` change and remove the skysphere because we can't reuse the same cubemap from the envmap in `scene.background` (or else we get `INVALID_OPERATION: bindTexture: textures can not be used with multiple targets`), and the rational why we have an equirect for the skysphere (although we could deep clone the texture?). Adding PMREMs means we can use cubemaps for both the skysphere and envmap since we have 2 cubemaps per environment now, and can discard the equirects.

There are still a few more caveats for the PMREM feature. The memory leak issue worked around in #227 is also in three's PMREM generators -- we would either have to do the same messy workaround, or can try to fix it upstream in https://github.com/mrdoob/three.js/issues/15288. The leak FWIW would only occur when swapping out background images continuously for an hour, so not a cause for alarm (edit: using HDR images exacerbate the leak; I don't think the texture is being duplicated, but maybe more resources are created). iPhone 6 also has an issue with the PMREM implementation (https://github.com/mrdoob/three.js/issues/15005).

Not quite ready for review yet, still some tests failing, and textures sometimes bleed into other elements (more alarming). These changes of course may not always look great out of the box but will enable the ability to add configuration. If missing my three.js patches, the first environment map encoding used is applied globally to all skyboxes, and the environment map projected on the model will be inversed. Almost there, wanted to document progress so far here.

Some pictures! (More pictures in #133)

LDR radiance vs HDR radiance: (maybe the LDR image should be boosted/mapped or PMREM should not be used for LDR)

screenshot from 2018-11-20 17-38-14

Inverted env map

screenshot from 2018-11-21 15-42-23

PMREM at different sampling levels (top/left clockwise 8, 16, 24, 32), lower samples more noticeable on smooth meshes.

pmrem-samples


jsantell:
My PRs for WebGLBackground texture encoding (https://github.com/mrdoob/three.js/pull/15299), inverted reflection PMREMs for WebGLRenderTargetCubes (https://github.com/mrdoob/three.js/pull/15302), and the PMREM/EquirectangularToCubeGenerator memory leak (https://github.com/mrdoob/three.js/pull/15330) have been merged upstream which will make this PR much easier, as well as remove some of the work arounds in place for the memory leak :tada:

jsantell:
Superceded by https://github.com/GoogleWebComponents/model-viewer/pull/323  and https://github.com/GoogleWebComponents/model-viewer/pull/338

PR #238, opened 2018-11-28
Remove hard dependency on ResizeObserver

### LINKED ISSUE #155: Performance: remove dependency on ResizeObserver API
@sorvell mentioned that `ResizeObserver` polyfill has really poor performance characteristics. We probably shouldn't force this on our users.

cdata:
Some details about this change:

 - `ResizeObserver` is used if (and only if) the API is detected on the page
 - If `ResizeObserver` is not available, the element falls back to updating on global `resize` events
 - If global `resize` events are used, layout measurement is debounced with a 50ms threshold
 - Resize observation for all conditions is now enabled in `connectedCallback` and disabled in `disconnectedCallback`
 - `ResizeObserver` polyfill is no longer loaded in the "legacy" test suite in order to get coverage for the fallback condition
 - Docs have been updated to emphasize that `ResizeObserver` is optional and will be used only if available

Fixes #155

### DIFF
--- a/POLYFILLS.md
+++ b/POLYFILLS.md
@@ -39,10 +39,29 @@ fine. The following is a selection of recommended polyfill implementations:
 Please keep in mind that your mileage may vary depending on the browsers you
 need to support and the fidelity of the polyfills used.
 
-**NOTE:** The Fullscreen API is only necessary for the experimental Web XR
+### Regarding Fullscreen API
+
+The Fullscreen API is only necessary for the experimental Web XR
 Device API-based AR mode. Since this is only available behind a flag in Chrome
 Dev today, it is not necessary to load a Fullscreen API polyfill in production
 scenarios.
+
+### Regarding Resize Observer
+
+If Resize Observer is available in the page, the `<model-element>` will be able
+to automatically recompute the scale and framing of its 3D content in many types
+of scenarios where layout is changing (for example, when its parent container
+changes size due to an animation or transition).
+
+However, it is important to note that Resize Observer is optional because the
+polyfill is known to have performance consequences that might be considered
+unacceptable for some use cases (it uses a Mutation Observer that observes the
+whole document tree).
+
+If Resize Observer is _not_ available, the `<model-element>` will fall back to
+observing global `resize` events. In this condition, you can force the element
+to recompute its internal layout by dispatching a synthetic global `resize`
+event.
 
 ## Usage Example
 
--- a/POLYFILLS.md
+++ b/POLYFILLS.md
@@ -59,8 +59,8 @@ unacceptable for some use cases (it uses a Mutation Observer that observes the
 whole document tree).
 
 If Resize Observer is _not_ available, the `<model-element>` will fall back to
-observing global `resize` events. In this condition, you can force the element
-to recompute its internal layout by dispatching a synthetic global `resize`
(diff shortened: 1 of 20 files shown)


POLYFILLS.md
```diff
@@ -1,21 +1,26 @@
...
+`<model-viewer>`:
 
  - [Custom Elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements) ([CanIUse](https://caniuse.com/#feat=custom-elementsv1), [Platform Status](https://www.chromestatus.com/features/4696261944934400))
  - [Shadow DOM](https://dom.spec.whatwg.org/#shadow-trees) ([CanIUse](https://caniuse.com/#feat=shadowdomv1), [Platform Status](https://www.chromestatus.com/features/4667415417847808))
- - [Resize Observer](https://wicg.github.io/ResizeObserver/) ([CanIUse](https://caniuse.com/#feat=resizeobserver), [Platform Status](https://www.chromestatus.com/features/5705346022637568))
  - [Intersection Observer](https://w3c.github.io/IntersectionObserver/) ([CanIUse](https://caniuse.com/#feat=intersectionobserver), [Platform Status](https://www.chromestatus.com/features/5695342691483648))
  - [Fullscreen API](https://fullscreen.spec.whatwg.org/) ([CanIUse](https://caniuse.com/#feat=fullscreen), [Platform Status](https://www.chromestatus.com/features/6596356319739904))
 
+The following emerging web platform API is optional, and will be used by
+`<model-viewer>` if it is detected on the page:
```
jsantell: It's unclear what happens if I do not have ResizeObserver support, what feature I miss out on, or cost I incur, or info to weigh the decision of "should I include the polyfill"

src/model-viewer-base.js
```diff
@@ -118,8 +127,8 @@ export default class ModelViewerElementBase extends UpdatingElement {
           this[$updateSize](entry.contentRect);
         }
       }
-    });
-    this.resizeObserver.observe(this);
+    }) :
+                                                null;
```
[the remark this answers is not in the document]
cdata: This looks pathological. I'll try to fix it.

POLYFILLS.md
cdata: 👍 will get more details in

src/model-viewer-base.js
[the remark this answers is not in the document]
cdata: Fixed it (looks like `clang-format` strongly prefers that args are wrapped in parens)

POLYFILLS.md
```diff
@@ -34,11 +39,30 @@ fine. The following is a selection of recommended polyfill implementations:
...
+changes size due to an animation or transition).
+
+However, it is important to note that Resize Observer is optional because the
+polyfill is known to have performance consequences that might be considered
+unacceptable for some use cases (it uses a Mutation Observer that observes the
+whole document tree).
+
+If Resize Observer is _not_ available, the `<model-element>` will fall back to
+observing global `resize` events. In this condition, you can force the element
+to recompute its internal layout by dispatching a synthetic global `resize`
```
jsantell: While 'global resize event' is correct, it might be more clear if specified as 'window resize event'

cdata: Tweaked it. Going to go ahead and merge, since the tests are moot at this point.

PR #245, opened 2018-11-30
Implement basic keyboard support for camera orbit

### LINKED ISSUE #222: Camera orientation can be controlled by keyboard input
This rolls into #187 

Look into the built-in support keyboard input and see what it actually offers.

cdata:
This change patches `OrbitControls` to support basic keyboard input for controlling the camera orbit. ~Currently only orbit is supported (notably not zoom).~

Update: zoom support has also been added via page-up and page-down.

I tried out the built-in keyboard input, but it is designed to control panning (not orbit), which we currently don't enable for `<model-viewer>`.

Also, the built-in keyboard input responds to global keyboard events, so it would likely be unsuitable anyway (every key input would control all `<model-viewer>`s at once).

This implementation only listens for events on the `<canvas>`, so the `<model-viewer>` must be focused before it will respond to keyboard input.

Keyboard orbiting is jumpy right now, but as we rewrite controls later this sprint we can look into smoothing this out substantially.

### Orbit

keyboardinput

### Zoom

kbdzoom

Fixes #222

### DIFF
--- a/src/features/controls.js
+++ b/src/features/controls.js
@@ -16,8 +16,8 @@
 import {PerspectiveCamera, Vector3} from 'three';
 
 import {$needsRender, $onModelLoad, $onResize, $scene} from '../model-viewer-base.js';
-import OrbitControls from '../third_party/three/OrbitControls.js';
 import {FRAMED_HEIGHT} from '../three-components/ModelScene.js';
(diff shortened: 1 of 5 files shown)


cdata:
Good call, will fix!

cdata:
Tests added

PR #256, opened 2018-12-03
Create legacy bundles during prepublish

### LINKED ISSUE #254: Distribute legacy bundle
In README, there's a link to the legacy build: https://unpkg.com/@google/model-viewer@0.0.6/dist/model-viewer-legacy.js

The legacy build is not currently being distributed, and the link 404's

cdata:
Fixes #254 

Ensures that the legacy bundles get generated before publish so that they get distributed as part of the published artifacts.


cdata:
I'll rebase on #255 when that lands

PR #258, opened 2018-12-04
Cap DPR when <meta name="viewport"> is missing

### LINKED ISSUE #185: Basic usage can be slow on Android chrome
This user-created demo is very slow (low FPS when orbiting w/ touch input) on Chrome on my Pixel 3: https://s.codepen.io/brianmontanaweb/debug/oQZvZp

Doesn't seem terribly complex on the surface. Flagging for investigation.

cdata:
This change is a candidate to fix #185 . ~**I think this proposal deserves some discussion / debate** as it is an imperfect approach that works in a lot of common cases, but has some known (and probably unknown) failure modes.~ Concisely, this is the strategy being proposed:

 > If the document _does not_ have a meta viewport tag, ~_and_ the ratio of the window dimensions to the screen dimensions (in CSS pixels) is greater than one,~ _then_ cap the effective device pixel ratio to 1.

~Known failure modes include:~

 i. ~**Page without meta viewport viewed on desktop** window-to-screen ratio exceeds 1 due to user manually scaling the window to be wider than their desktop screen.~
 ii.  ~**Page without meta viewport viewed on mobile** is loaded in an iframe, and the iframe-to-screen ratio is <= 1 (see [here](https://lopsided-motion.glitch.me/index-nmv2.html)).~
 iii. ~Any others...?~

EDIT: The demo links are somewhat frozen in time, and use a build that applies a more complex heuristic than will be shipped in this PR.

A slightly patched version of this change can be viewed in the live demo: [https://lopsided-motion.glitch.me/](https://lopsided-motion.glitch.me/).

A version of the same demo where `<meta name="viewport">` is omitted can be found here: [https://lopsided-motion.glitch.me/index-nmv.html](https://lopsided-motion.glitch.me/index-nmv.html).

In both demos, the device pixel ratio being used is printed to the console for your reference.

I tested both pages in the following browsers:

 - iPhone XR Simulator / iOS 12 Safari
 - Pixel 1 / Google Chrome 70
 - Pixel 1 / Firefox 63

For the basic no-`<meta name="viewport">` case, performance is significantly improved without any subjective loss in quality.

Still TODO:
 - [x] Test `resolveDpr` implementation

Fixes #185

### DIFF
--- a/src/three-components/Renderer.js
+++ b/src/three-components/Renderer.js
@@ -17,17 +17,15 @@ import {EventDispatcher, WebGLRenderer} from 'three';
 
 import {IS_AR_CANDIDATE} from '../constants.js';
 import {$tick} from '../model-viewer-base.js';
-import {debounce, resolveDpr} from '../utils.js';
+import {resolveDpr} from '../utils.js';
 
 import {ARRenderer} from './ARRenderer.js';
 import TextureUtils from './TextureUtils.js';
 import * as WebGLUtils from './WebGLUtils.js';
 
 const GAMMA_FACTOR = 2.2;
-const UPDATE_DPR_DEBOUNCE_MS = 50;
 
 export const $arRenderer = Symbol('arRenderer');
-const $updateDpr = Symbol('updateDpr');
 
 /**
  * Registers canvases with Canvas2DRenderingContexts and renders them
@@ -65,15 +63,7 @@ export default class Renderer extends EventDispatcher {
     this.renderer.autoClear = false;
     this.renderer.gammaOutput = true;
     this.renderer.gammaFactor = GAMMA_FACTOR;
-
-    this[$updateDpr] = debounce(() => {
-      this.dpr = resolveDpr();
-      this.renderer.setPixelRatio(this.dpr);
-    }, UPDATE_DPR_DEBOUNCE_MS);
-
-    this.dpr = resolveDpr();
-    this.renderer.setPixelRatio(this.dpr);
-    window.addEventListener('resize', this[$updateDpr]);
+    this.renderer.setPixelRatio(resolveDpr());
 
     this[$arRenderer] = ARRenderer.fromInlineRenderer(this);
     this.textureUtils = new TextureUtils(this.renderer);
@@ -162,8 +152,9 @@ export default class Renderer extends EventDispatcher {
       this.renderer.setViewport(0, 0, width, height);
       this.renderer.render(scene, camera);
 
-      const widthDPR = width * this.dpr;
-      const heightDPR = height * this.dpr;
+      const dpr = resolveDpr();
+      const widthDPR = width * dpr;
+      const heightDPR = height * dpr;
       context.drawImage(
           this.renderer.domElement,
           0,
@@ -185,6 +176,5 @@ export default class Renderer extends EventDispatcher {
     super.dispose();
     this.textureUtils.dispose();
     this.textureUtils = null;
-    window.removeEventListener('resize', this[$updateDpr]);
   }
 }
--- a/src/three-components/Renderer.js
+++ b/src/three-components/Renderer.js
@@ -17,15 +17,17 @@ import {EventDispatcher, WebGLRenderer} from 'three';
 
 import {IS_AR_CANDIDATE} from '../constants.js';
 import {$tick} from '../model-viewer-base.js';
+import {debounce, resolveDpr} from '../utils.js';
 
 import {ARRenderer} from './ARRenderer.js';
 import TextureUtils from './TextureUtils.js';
 import * as WebGLUtils from './WebGLUtils.js';
 
 const GAMMA_FACTOR = 2.2;
-const DPR = window.devicePixelRatio;
+const UPDATE_DPR_DEBOUNCE_MS = 50;
 
 export const $arRenderer = Symbol('arRenderer');
+const $updateDpr = Symbol('updateDpr');
 
 /**
  * Registers canvases with Canvas2DRenderingContexts and renders them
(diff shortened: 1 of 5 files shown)


src/utils.js
```diff
@@ -86,3 +86,79 @@ export const openIOSARQuickLook = url => {
...
+    for (const meta of metas) {
+      if (meta.name === 'viewport') {
+        return true;
+      }
+    }
+
+    return false;
+  })();
+
+  // NOTE(cdata): f true, this suggests that the window contents have been
```
jsantell: nit: 'if'`

src/utils.js
```diff
@@ -86,3 +86,79 @@ export const openIOSARQuickLook = url => {
...
+  // scaled such that the inner width is a multiple of the effective CSS pixel
+  // dimensions of the actual screen. This is assumed to imply one of the
+  // following conditions:
+  //
+  //  1. The window has been manually sized to extent beyond the bounds of
+  // the user's display (probably only possible in desktop scenarios).
+  //  2. There is no meta viewport on a high-DPR device, or else the meta
+  // viewport has scaled the defuault width higher than the screen width of
```
jsantell: nit: 'default'

src/three-components/Renderer.js
```diff
@@ -61,10 +63,18 @@ export default class Renderer extends EventDispatcher {
...
     this.renderer.gammaFactor = GAMMA_FACTOR;
 
+    this[$updateDpr] = debounce(() => {
+      this.dpr = resolveDpr();
+      this.renderer.setPixelRatio(this.dpr);
+    }, UPDATE_DPR_DEBOUNCE_MS);
+
+    this.dpr = resolveDpr();
+    this.renderer.setPixelRatio(this.dpr);
+    window.addEventListener('resize', this[$updateDpr]);
```
jsantell: Is this fired on 'zoom' as well?

src/utils.js
```diff
@@ -86,3 +86,79 @@ export const openIOSARQuickLook = url => {
...
+  // is a meta viewport, we won't cap the DPR.
+  // Condition #2 is usually not intentional but very easy to arrive at (a
+  // user need only omit the meta viewport tag and view their site on a
+  // contemporary mobile device).
+  //
+  // @see https://www.quirksmode.org/dom/w3c_cssom.html#screenview
+  const hasRiskyWindowToScreenRatio = () =>
+      Math.max(
+          window.innerWidth / window.screen.availWidth,
+          window.innerHeight / window.screen.availHeight) > 1;
```
jsantell: It's been awhile since I've used `availWidth|Height`; found this note on MDN:

"Internet Explorer will take into account the zoom setting when reporting the screen width. It will only return the real width of the screen if the zoom is set to 100%."

Just a comment, not sure if this is addressed

src/utils.js
```diff
@@ -86,3 +86,79 @@ export const openIOSARQuickLook = url => {
...
+  // viewport has scaled the defuault width higher than the screen width of
+  // the device.
+  //
+  // Condition #1 is not very common, but certainly possible. As long as there
+  // is a meta viewport, we won't cap the DPR.
+  // Condition #2 is usually not intentional but very easy to arrive at (a
+  // user need only omit the meta viewport tag and view their site on a
+  // contemporary mobile device).
+  //
+  // @see https://www.quirksmode.org/dom/w3c_cssom.html#screenview
```
jsantell: great link

src/utils.js
jsantell: Edit: looks like quirksmode doesn't address IE10+, but unclear what IE the [MDN notes for screen.width](https://developer.mozilla.org/en-US/docs/Web/API/Screen/width#Notes) refers to

cdata:
> For the iframe (ii.) case, we can check the window.top values?

I wasn't sure we would be able to do this in x-domain scenarios, but if we can then that is great.

cdata:
> I wonder if it'd be helpful to console.warn something when this engages.

Great idea. Adding this.

cdata:
> I imagine for future customization cases, there may be a render-scale value...

Filed https://github.com/GoogleWebComponents/model-viewer/issues/261 to track

cdata:
> Better idea: what are the downsides for always DPR 1 unless viewport defined?

Discussed elsewhere, this seems like a good idea for now, let's do that!

cdata:
Original comment now removed from change. Preserving here for posterity:

```
  // NOTE(cdata): if true, this suggests that the window contents have been
  // scaled such that the inner width is a multiple of the effective CSS pixel
  // dimensions of the actual screen. This is assumed to imply one of the
  // following conditions:
  //
  //  1. The window has been manually sized to extent beyond the bounds of
  // the user's display (probably only possible in desktop scenarios).
  //  2. There is no meta viewport on a high-DPR device, or else the meta
  // viewport has scaled the defuault width higher than the screen width of
  // the device.
  //
  // Condition #1 is not very common, but certainly possible. As long as there
  // is a meta viewport, we won't cap the DPR.
  // Condition #2 is usually not intentional but very easy to arrive at (a
  // user need only omit the meta viewport tag and view their site on a
  // contemporary mobile device).
  //
  // @see https://www.quirksmode.org/dom/w3c_cssom.html#screenview
```

cdata:
Filed https://github.com/GoogleWebComponents/model-viewer/issues/262 to track better testing capabilities relevant to this change

src/three-components/Renderer.js
cdata: It's not. I think I will remove this for now, as measuring DPR is no longer as expensive as it was at the beginning of this change. Instead. we can measure it in real time.

PR #260, opened 2018-12-05
Update three's extra components from r99

### LINKED ISSUE #240: Replace PatchedEquirectangularToCubeGenerator with r99 version
With the fix for the WebGLState memory leak landed upstream in r99 (https://github.com/mrdoob/three.js/pull/15330), we should swap out our workaround (`./src/three-components/PatchedEquirectangularToCubeGenerator.js`) with the r99 version once we update to r99.

jsantell:
Forgot to update three's examples components, which had my upstream patches I was looking for :laughing: no visual changes here, this updates the cube generator to remove the memory leak, and such our workaround. Fixes #240 

No changes to Sky or OrbitControls.

* GLTFLoader had some skin weight changes: [history](https://github.com/mrdoob/three.js/commits/dev/examples/js/loaders/GLTFLoader.js)
* CubeGenerator has been added to EquirectangularToCubeGenerator; a more desirable API, but will run into the same memory leak: [history](https://github.com/mrdoob/three.js/commits/dev/examples/js/loaders/EquirectangularToCubeGenerator.js)


cdata:
LGTM although I've been wondering: is there a reason why we continue to these if they don't contain modifications? They are distributed as part of the module: https://unpkg.com/three@0.99.0/examples/js/loaders/EquirectangularToCubeGenerator.js

jsantell:
> LGTM although I've been wondering: is there a reason why we continue to these if they don't contain modifications? They are distributed as part of the module: https://unpkg.com/three@0.99.0/examples/js/loaders/EquirectangularToCubeGenerator.js

ES module import/export statements with rollup config unfortunately, if there's a better way would love to switch but none that I know of unless we wanted to host an ES6-ified, otherwise unmodified version module, e.g. `@polymer/orbit-controls` or something, and join the long list of other ES module-ified packages in npm of three.js examples

cdata:
SG 👍

PR #264, opened 2018-12-07
Re-introduce the directional light to the scene

### LINKED ISSUE #179: some models are washed out
Some models appear very washed out. One example is: https://poly.google.com/view/4lZW20kkyMD?jsmode=du

Washed out version (current `<model-viewer>`):

washed-out

More detail (hosted at https://poly.google.com/view/4lZW20kkyMD?jsmode=du):

more-detail

We have several work items to tackle this (at a glance, #34, #6), so this may be fixed as we pick up those.

cdata:
This change is a candidate to address #179

 - Re-introduces a directional light by adding the "shadow" light to the scene
 - Shadow light hue is tinted to match background color if configured

### Before / after samples

astrocolordl

dhpbrdl 2018-12-06 15_54_17

shiskbgdl 2018-12-06 16_01_45

vf1adl 2018-12-06 15_34_12

Fixes #179

### DIFF
--- a/src/features/environment.js
+++ b/src/features/environment.js
@@ -19,6 +19,8 @@ import {$needsRender, $onModelLoad, $renderer, $scene, $tick} from '../model-vie
 const DEFAULT_BACKGROUND_COLOR = '#ffffff';
 const GAMMA_TO_LINEAR = 2.2;
 
+const WHITE = new Color('#ffffff');
+
 const $currentCubemap = Symbol('currentCubemap');
 const $setEnvironmentImage = Symbol('setEnvironmentImage');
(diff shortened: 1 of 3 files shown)


cdata:
> Does tint color change the overall brightness of the model? Like does a model appear darker/less direct light if it's on a darker background?

The tint is lerped strongly towards white. So, even on a black background it will appear to be illuminated (albeit less so than if it was on a white background). For example:

shishlerpdl 2018-12-06 17_44_44

Originally made the tint stronger and it was quite obvious. I decided a subtler tint was more pleasing than both pure white and the strong tint.

> I think the light should reset to ffffff when displaying a skysphere.

This seems naive, but maybe the best we can do for now. Will make this change.

PR #265, opened 2018-12-07
Auto center and ground the model so it appears correct in most cases in AR

jsantell:
This does not have any affect in 2D displays.


cdata:
A lot of warning signs for me here:

 - No issue referenced
 - Lacking PR details in description
 - No tests

PR #266, opened 2018-12-07
Fixing shadows in WebXR

jsantell:
WebXR requires rendering to a specific framebuffer, and internally in three, it made some assumptions about the default framebuffer (I forget the details) and `setFramebuffer()` (https://github.com/mrdoob/three.js/pull/14132) was added to WebGLRenderer for shadows, which is handled because WebGLShadowMap [sets a render target](https://github.com/mrdoob/three.js/blob/2db634d94955e119d5a519d091cb57ebf52f55e3/src/renderers/webgl/WebGLShadowMap.js#L225). Ultimately needed to add an onBeforeRender for any(?) RenderTarget when rendering to the AR framebuffer, AFAICT.

### DIFF
--- a/src/three-components/ModelScene.js
+++ b/src/three-components/ModelScene.js
@@ -13,20 +13,10 @@
  * limitations under the License.
  */
 
-import {AmbientLight, BackSide, Box3, Color, DirectionalLight, Mesh, MeshBasicMaterial, Object3D, PerspectiveCamera, Scene, SphereBufferGeometry, Vector3} from 'three';
+import {AmbientLight, BackSide, Box3, Color, DirectionalLight, Matrix4, Mesh, MeshBasicMaterial, Object3D, PerspectiveCamera, Scene, SphereBufferGeometry, Vector3} from 'three';
 
 import Model from './Model.js';
 import StaticShadow from './StaticShadow.js';
-
-// Valid types for `setScaleType` -- 'framed' scales the model
-// so that it fits within its 2D plane nicely. 'lifesize' is
-// unaltered scaling and uses whatever size the model provides,
-// such that 1 unit === 1 meter, used in AR.
-export const ScaleTypes = {
-  Framed: 'framed',
-  Lifesize: 'lifesize',
-};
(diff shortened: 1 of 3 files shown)


jsantell:
Filed upstream WebXR/WebGLRenderTarget issue: https://github.com/mrdoob/three.js/issues/15393

src/three-components/ModelScene.js
```diff
@@ -47,6 +37,7 @@ const FOV = 45;
 const DPR = window.devicePixelRatio;
 
 const $paused = Symbol('paused');
+const tempMat4 = new Matrix4();
```
cdata: let's just call this lowercase `matrix4`

src/three-components/ModelScene.js
```diff
@@ -282,8 +273,9 @@ export default class ModelScene extends Scene {
    * Called to update the shadow rendering when the room or model changes.
    */
   updateStaticShadow() {
+    this.model.remove(this.shadow);
```
cdata: FWIW `remove` can have unanticipated side-effects (see https://github.com/mrdoob/three.js/blob/master/src/core/Object3D.js#L423). It seems more hygienic only to remove it under the appropriate condition.

cdata:
Shadows actually seem to be completely gone for me, see below:

screenshot_20181213-122748
screenshot_20181213-122802

src/three-components/ModelScene.js
jsantell: :+1:

src/three-components/ModelScene.js
jsantell: :+1:

src/three-components/ModelScene.js
mrdoob: Yeah. It is better to do `this.shadow.visible = false;`.

jsantell: TIL, thanks all!

cdata:
I'm going to go ahead and close this out, as it's very out of date. We can revisit in the future when we get back on top of WebXR features. Thanks for the work here @jsantell !

Issue #216, opened 2018-11-15
Add N-context "naive" rendering mode


cdata:
This change will introduce the first additional rendering mode. The scope of this change is somewhat large because it includes implementing the ability to intelligently switch between rendering modes depending on the result os some heuristic.

Use this mode for some low threshold of model viewers on the page. Possibly the threshold should be different on mobile compared to not-mobile.

pushmatrix:
How is progress going with this?

cdata:
The rendering modes work is going to rear its head soon, but I'm not sure how soon. @pushmatrix Would you characterize your interest due to this issue's relevance to #268 ?

pushmatrix:
100%

DavidPeicho:
Right now one scene canno't be viewed in two different viewers, but if that changes in the future, having **N** WebGL contexts will be a memory killer.

Having one scene shared by two viewers isn't something stupid, for instance you could want to just have two viewers, one with a camera pointing down, and one with a camera pointing up. If the library starts at some point to implement this possibility, having N-viewer will make all the GPU duplicated **N** times unfortunately.

elalish:
I think we're going with the approach in #995 instead, which does this for n=1.

PR #268, opened 2018-12-07
WIP: transparent rendering

forresto:
### Reference Issue
Conversation in #182

TODO:
- [x] Fix trails
- [x] Wire up `background-color="transparent"`
- [ ] Decide if/how to get transparent attribute into shared renderer
- [ ] Add docs

### DIFF
--- a/src/three-components/ModelScene.js
+++ b/src/three-components/ModelScene.js
@@ -105,9 +105,6 @@ export default class ModelScene extends Scene {
     // depthTest=false
     this.skysphere.renderOrder = 1;
 
-    // TODO: hook to transparent attribute
-    this.skysphere.visible = false;
-
     this.add(this.pivot);
     this.add(this.light);
     this.add(this.skysphere);
--- a/src/three-components/ModelScene.js
+++ b/src/three-components/ModelScene.js
@@ -105,6 +105,9 @@ export default class ModelScene extends Scene {
     // depthTest=false
     this.skysphere.renderOrder = 1;
 
+    // TODO: hook to transparent attribute
+    this.skysphere.visible = false;
+
     this.add(this.pivot);
     this.add(this.light);
     this.add(this.skysphere);
--- a/src/three-components/Renderer.js
+++ b/src/three-components/Renderer.js
@@ -45,7 +45,7 @@ export default class Renderer extends EventDispatcher {
     const webGlOptions = {antialias: true};
 
     // TODO: hook to transparent attribute
-    cosnt isTransparent = true;
+    const isTransparent = true;
 
     // Only enable certain options when Web XR capabilities are detected:
     if (IS_AR_CANDIDATE || isTransparent) {
--- a/src/three-components/Renderer.js
+++ b/src/three-components/Renderer.js
@@ -45,9 +45,10 @@ export default class Renderer extends EventDispatcher {
     const webGlOptions = {antialias: true};
 
     // Only enable certain options when Web XR capabilities are detected:
-    if (IS_AR_CANDIDATE) {
+    // TODO: hook to transparent attribute
+    // if (IS_AR_CANDIDATE) {
       Object.assign(webGlOptions, {alpha: true, preserveDrawingBuffer: true});
-    }
+    // }
 
     this.canvas = document.createElement('canvas');
     // Need to support both 'webgl' and 'experimental-webgl' (IE11).
@@ -60,7 +61,8 @@ export default class Renderer extends EventDispatcher {
       canvas: this.canvas,
       context: this.context,
     });
-    this.renderer.autoClear = false;
+    // TODO: hook to transparent attribute
+    this.renderer.autoClear = true;
     this.renderer.gammaOutput = true;
     this.renderer.gammaFactor = GAMMA_FACTOR;
     this.renderer.setPixelRatio(resolveDpr());
--- a/src/three-components/Renderer.js
+++ b/src/three-components/Renderer.js
@@ -44,11 +44,13 @@ export default class Renderer extends EventDispatcher {
 
     const webGlOptions = {antialias: true};
 
-    // Only enable certain options when Web XR capabilities are detected:
     // TODO: hook to transparent attribute
-    // if (IS_AR_CANDIDATE) {
+    cosnt isTransparent = true;
+
+    // Only enable certain options when Web XR capabilities are detected:
+    if (IS_AR_CANDIDATE || isTransparent) {
       Object.assign(webGlOptions, {alpha: true, preserveDrawingBuffer: true});
-    // }
+    }
 
     this.canvas = document.createElement('canvas');
     // Need to support both 'webgl' and 'experimental-webgl' (IE11).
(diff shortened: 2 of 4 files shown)


forresto:
To test I set `this.skysphere.visible = false;` ... would there need to be more logic around that? There are a few place that expect the skysphere to exist.

With `this.renderer.autoClear = false;` other models on the page jump in, and trails:

image

With `this.renderer.autoClear = true;` gets rid of the other models, but still leaves these trails... any hints?

image

src/three-components/ModelScene.js
```diff
@@ -105,6 +105,9 @@ export default class ModelScene extends Scene {
     // depthTest=false
     this.skysphere.renderOrder = 1;
 
+    // TODO: hook to transparent attribute
+    this.skysphere.visible = false;
```
forresto: ... would there need to be more logic around that? There are a few place that expect the skysphere to exist.

cdata:
Thanks for the PR! We will take a look.

I wonder if there is a connection to #239

cdata:
@forresto can you confirm which browsers/platforms you are experiencing this in as we go to test for ourselves?

src/three-components/Renderer.js
```diff
@@ -45,9 +45,10 @@ export default class Renderer extends EventDispatcher {
     const webGlOptions = {antialias: true};
 
     // Only enable certain options when Web XR capabilities are detected:
-    if (IS_AR_CANDIDATE) {
+    // TODO: hook to transparent attribute
+    // if (IS_AR_CANDIDATE) {
```
cdata: Was this removed intentionally, or just a vestigial comment?

jsantell:
Quick comments:

* the folding you're seeing is from changing when the canvas is cleared due to using one WebGLRenderingContext for all elements on the page -- the skysphere does not render to the depth buffer -- this is different than #239 
* three r99 added the ability for `scene.background` to accept cube maps, which will make a lot of the skysphere logic throughout the codebase much nicer. #237 has some of that implemented that still needs a few other things to shake out.
* This will of course add overhead for all elements and possibly require recreating the WebGLRenderingContext
* What would happen if both `background-color` and `transparent` are defined? Should `transparent` be an option of `background-color`?
* What's the expected way to style this background with CSS? Something like `model-viewer { background-color: var(--some-gradient); }`?

src/three-components/Renderer.js
forresto: I was reusing the {alpha: true, preserveDrawingBuffer: true} in the next line there... This might make it more clear: f4f779c09f0b12d5bffaafe2ca283a6edbeb0e1e

forresto:
Safari, Chrome, and Firefox on Mac.

> Should transparent be an option of background-color?

I think that would make sense. I wasn't sure if there might be a rendering feature that could use both `background-color` and `transparent`.

> What's the expected way to style this background with CSS?

`style` on the `model-viewer` element seems to work. Also, the general case of a parent background showing through.

forresto:
> using one WebGLRenderingContext for all elements on the page

Would that prevent having transparent and non-transparent elements on the same page?

forresto:
Now I see how clearing works :)

image

src/three-components/Renderer.js
```diff
@@ -44,8 +44,11 @@ export default class Renderer extends EventDispatcher {
 
     const webGlOptions = {antialias: true};
 
+    // TODO: hook to transparent attribute
+    const isTransparent = true;
```
forresto: Since Renderer is a singleton we don't have a way to get an attribute in here. TBD how that should work. Ideas?

jsantell: This renderer is responsible for all elements/scenes, currently, so a renderer can be responsible for both a transparent and non-transparent scene -- `ModelScene` is unique per element, however

src/three-components/Renderer.js
```diff
@@ -60,7 +63,8 @@ export default class Renderer extends EventDispatcher {
       canvas: this.canvas,
       context: this.context,
     });
-    this.renderer.autoClear = false;
+    // TODO: hook to transparent attribute
+    this.renderer.autoClear = isTransparent;
```
jsantell: A handful of components use the renderer outside of elements -- composing environment maps or shadows, or any WebGLRenderTargets -- `autoClear` should not change as they affect all of those framebuffers

forresto: How can I keep it `false` and avoid other scenes bleeding in?

cdata: @forresto we do per-scene (in other words, per-element) renderer configuration in the render loop itself. See https://github.com/GoogleWebComponents/model-viewer/blob/master/src/three-components/Renderer.js#L135-L171.

However, this may not be suitable for all kinds of renderer configuration, and `autoClear` is probably not going to be suitable as it clears the whole context.

In the future, we intend to support additional rendering modes (Jordan already referenced this, but I'll ping it again as an example: https://github.com/GoogleWebComponents/model-viewer/issues/216). As we start to get into that, we should be able to layer on configurations like this.

forresto: Are you recommending that I hold more work on this PR for #216? I can use my branch for now, since it seems to work.

cdata: 👍 thanks @forresto yeah I think it would be best to pin this PR for now and come back to it soon as we start to upgrade our approaches to rendering.

pushmatrix:
Adding a big 👍 to this feature. Super looking forward to it being possible.

mrdoob:
@pushmatrix we were considering creating 2 internal webgl contexts (when needed) one for `alpha: false`, and another one for `alpha: true`.

tijmenvangurp:
Hey, I am really looking forward to use `background-color="transparent"`are there any workarounds in the meantime? Basically, I want my full page to have a certain background with gradient where my object is just floating on, which is not possible with the current implementation.

forresto:
Here's a build of my branch: https://uploads-ssl.webflow.com/5c013e2442f37ff004567ee6/5c17c0247910f0e6fe21dde9_model-viewer.min.js.txt (you can rename to .js)

Demo: https://3d-model-viewer-component.webflow.io/transparent-hack 

Note that the build is a few months behind the official version.

tijmenvangurp:
@forresto  Thanks for your reply :) will try to incorporate your build, hope that background transparent can be incorporated in the official build soon.

cdata:
I spent a little bit of time poking at this problem. The main roadblock that I ran into is that we need to set `alpha: true` when creating the GL context in order to support transparency, but this causes incorrect rendering in some alpha blending scenarios (see https://github.com/GoogleWebComponents/model-viewer/issues/381).

So, we should not _force_ `alpha: true` on the context. Perhaps we can add an attribute to let folks opt-in or out to background transparency? What would be the right default?

Relatedly, it seems desirable for us to generally move away from `background-color` and `background-image` as attributes, because they conflate `<model-viewer>`'s responsibilities with those of CSS (see #661 ).

elalish:
Closing in favor of #975. @forresto Thanks for working on this and sorry it took us so long to come to a solution. Hope the current state solves your needs!

PR #274, opened 2018-12-11
Place model via XRInputSource

### LINKED ISSUE #273: WebXR should use XRInputSource instead of canvas click handler
This would let us support future WebXR devices that may not have 2D input, e.g. Hololens

jsantell:
Change input from click handler on XRPresentationContext to XRInputSource for placing a model in WebXR. Currently only 'screen' inputs are supported (but as of now, the only input that exists in the wild), but will be able to use the same structure for those types of inputs. Will be easier to support 6DOF input in the future, as the input is more permanent (as opposed to the 'screen' input which only exists for the duration of the 'touch'), it currently works with `inputsourceschange` event, and we'll only need to listen to `selectend` on the input. Due to unstable nature of WebXR, we can tackle that once we have a device that implements 6DOF input in AR.

Also the ray now originates from where the touch occurs -- while this is expected for the 6DOF input, this might be weird on mobile AR, since the reticle is in the middle of the screen. In 6DOF inputs, the reticle would be tracking the input pose, instead of the viewer pose. Should we go back to firing it center of the screen?

The reticle has also been updated to later in the render loop so it uses the most recent pose to fire its ray, rather than being a frame behind.

Fixes #273 (and maybe need a follow up for 6DOF AR inputs? although no way to test yet...)


src/three-components/ARRenderer.js
```diff
@@ -246,28 +239,52 @@ height: 100%;`);
     return this[$outputContext];
   }
 
+  /**
+   * Calls the underlying WebXR `requestHitTest` and supports
+   * both the two- and three-argument forms, as well as
+   * implementation bugs.
+   */
+  async castRay(ray) {
```
jsantell: This handling will also need to occur in `Reticle.js` -- this should probably be moved to a util

cdata:
It would be nice to get links to relevant docs on low-level WebXR APIs. I have found that they tend to be scattered and sometimes there are multiple sources with different levels of freshness.

src/utils.js
```diff
@@ -17,6 +17,10 @@ import {Vector3} from 'three';
 
 import {HAS_FULLSCREEN_API, HAS_WEBXR_DEVICE_API, HAS_WEBXR_HIT_TEST_API, IS_AR_CANDIDATE} from './constants.js';
 
+export const domPointToArray = point => {
+  return new Float32Array([point.x, point.y, point.z]);
```
cdata: It seems marginally wasteful to create this object (sometimes more than once) on every frame (which will happen once reticle is updated to use this).

src/three-components/ARRenderer.js
```diff
@@ -299,19 +342,23 @@ height: 100%;`);
...
-      const viewMatrix = matrix4.fromArray(pose.getViewMatrix(view));
-
-      this.camera.matrix.getInverse(viewMatrix);
-      this.camera.updateMatrixWorld(true);
-      // NOTE: Clearing depth caused issues on Samsung devices
-      // @see https://github.com/googlecodelabs/ar-with-webxr/issues/8
-      // this.renderer.clearDepth();
-      this.renderer.render(this.scene, this.camera);
-    }
+    const view = frame.views[0];
```
cdata: What is the rationale behind this change? It seemed like it was important to handle all of the views. Is this no longer the case?

src/three-components/ARRenderer.js
cdata: https://github.com/GoogleWebComponents/model-viewer/issues/281

src/three-components/ARRenderer.js
```diff
@@ -279,6 +296,34 @@ height: 100%;`);
...
+    // before `selectstart` and removed immediately after `selectend` event.
+    // If we have a 'screen' source here, it means the output canvas was tapped.
+    const sources = session.getInputSources().filter(
+        input => input.targetRayMode === 'screen');
+
+    if (sources.length === 0) {
+      return;
+    }
+
+    const pose = frame.getInputPose(sources[0], this[$frameOfReference])
```
cdata: So if I understand this correctly, `screen` input sources are ephemeral and depend on actual input to show up in the list, so if one does it implies there was some touch input to select a placement?

cdata:
Here is a recording of placement now that we are no longer doing it based on casting directly from the center of the screen:

xrinputplacement

Overall this feels like a UX regression to me because the reticle position is not related to where the model is ultimately placed, and there is no other indication of what the placement rules should be. It also feels weird that the center of volume of the model seems to be aligned to where I touch (I guess I expected the bottom of the model to be aligned to where I touch).

Can we change this so that placement occurs where the reticle is placed?

src/three-components/ARRenderer.js
jsantell: Correct; although when held, the input will exist until release, so existence doesn't necessarily imply 'touchstart'

jsantell:
@cdata agreed on going back to using center of the screen for placement with reticle

src/utils.js
jsantell: Yeah -- actually this can be handled in `castRay` which can cache those arrays too

src/three-components/ARRenderer.js
```diff
@@ -246,28 +239,52 @@ height: 100%;`);
...
+   */
+  async castRay(ray) {
+    const hitArgs = this[$currentSession].requestHitTest.length;
 
-  async[$onOutputCanvasClick]() {
-    if (this[$currentSession] == null) {
-      return;
+    let hits;
+    try {
+      if (hitArgs === 3) {
```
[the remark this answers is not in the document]
jsantell: :+1:

src/three-components/ARRenderer.js
jsantell: The rational was for smartphone AR being monoscopic, but that falls apart for things like Hololens. Will revert, good catch!

jsantell:
Added tests, removed the patching of `requestHitTest`, since no implementation of the XRRay form exists, and XRRays are defined differently in the hit-test spec and the WebXR Device API spec, that is to say, it does not yet need patched -- no user visible changes

PR #296, opened 2019-01-16
exercise specific accessibility scenarios

smalls:
Leverage the existing demos where possible, otherwise create specific pages to exercise accessibility scenarios for basic usage of `<model-viewer>`.

This will allow further investigation of the accessibility of the component, to see if there are areas we can continue to improve.

Part of #295

### DIFF
--- a/examples/accessibility.html
+++ b/examples/accessibility.html
@@ -44,43 +44,22 @@
 </head>
 <body>
     <div id="container">
-        <h1>&lt;model-viewer&gt; - Examples</h1>
+        <h1>&lt;model-viewer&gt; - Accessibility Tests</h1>
         <div class="desc">
-            <p>There are several HTML pages containing usages of the <code>&lt;model-viewer&gt;</code> element with different configurations for demonstrative purposes and testing specific scenarios.</p>
+            <p>There are several different tests in order to investigate different aspects of accessibility and <code>&lt;model-viewer&gt;</code>. This page enumerates the test cases, and provides links to the individual scenarios.</p>
+            <p>Note that all these examples include <code>alt</code>.</p>
         </div>
-
-        <h2>Demos &amp; Scenarios</h2>
         <ul class="desc">
             <li><a href="default.html">default.html</a>: The barebones default case.</li>
-            <li><a href="background-color.html">background-color.html</a>: Using the <code>background-color</code> attribute to style the background color.</li>
-            <li><a href="background-image.html">background-image.html</a>: Using the <code>background-image</code> attribute to specify an equirectangular projection image.</li>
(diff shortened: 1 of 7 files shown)


examples/accessibility.html
```diff
@@ -0,0 +1,66 @@
...
+  <script src="../node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+
+  <!-- Intersection Observer polyfill is required for Safari and IE11 -->
+  <script src="../node_modules/intersection-observer/intersection-observer.js"></script>
+
+  <!-- Resize Observer polyfill is optional, and improves resize behavior in non-Chrome browsers: -->
+  <script src="../node_modules/resize-observer-polyfill/dist/ResizeObserver.js"></script>
+
+  <!-- Fullscreen polyfill is required for using experimental AR features in Canary: -->
+  <script src="../node_modules/fullscreen-polyfill/dist/fullscreen.polyfill.js"></script>
```
cdata: We no longer require this one. We haven't removed it from existing examples, but let's not add it to new ones.

examples/accessibility.html
```diff
@@ -0,0 +1,66 @@
...
+            <p>Note that all these examples include <code>alt</code>.</p>
+        </div>
+        <ul class="desc">
+            <li><a href="default.html">default.html</a>: The barebones default case.</li>
+            <li><a href="accessibility/poster.html">accessibility/poster.html</a>: A single sample exercising <code>poster</code>.</li>
+	          <li><a href="accessibility/reveal.html">accessibility/reveal.html</a>: A single sample exercising <code>poster</code>, <code>preload</code>, and <code>reveal-when-loaded</code>.</li>
+	          <li><a href="accessibility/controls.html">accessibility/controls.html</a>: A single sample exercising <code>controls</code>.</li>
+	          <li><a href="accessibility/auto-rotate.html">accessibility/auto-rotate.html</a>: A single sample exercising <code>auto-rotate</code>.</li>
+	          <li><a href="accessibility/multiple.html">accessibility/multiple.html</a>: Multiple models on the same page with a variety of attributes.</li>
```
cdata: Is there a version of this where all examples live on one page? I'm not sure what we get out of splitting the examples up into one test case per page - did you have anything in mind?

smalls: The `multiple.html` does cover most of the scenarios. I thought it might make tests and conversations about the pages simpler to exercise a single scenario on an individual page, especially since the "many models on a page" was a specific scenario that we wanted to test.

Happy to change it, though!

smalls:
> Did you check these out locally before you pushed up to Github to make sure the links all work?

Yup, they all work locally.

smalls:
@cdata any other thoughts?

I don't think there's any more work left so this will close #295 (once merged).

PR #311, opened 2019-01-28
Better scrolling on mobile devices

### LINKED ISSUE #75: natural controls and scrolling
Let's look at making the interaction between controls and scrolling better, and more natural.

The issue now is that the model element, with controls enabled (after #56 is fixed), captures all events, including those that would scroll the page. On mobile where the element takes up a significant portion of screen real estate we may end up preventing any scrolling past a certain point.

We discussed some solutions that we think have potential, including (I may be missing some):
- By default, having the model only move vertically a small amount; after that reaches the maximum, go back to scrolling the page
- Pass initial vertical movements up to the page; the user would have to interact with the model by rotating it horizontally before they could rotate it vertically

Other options:
- Using 2 fingers to move the model, a single finger scrolls the page
- A button to toggle model controls

cdata:
This change is a follow up to #288 . One of the primary goals of that change was to enable better interaction with pages containing `<model-viewer>` while scrolling. Due to a bug in the change, it turned out that mobile devices (or any that relied on touch input, really), still had a poor default interaction experience.

This change address scrolling on devices with touch input. Now all interaction (including via touch input) requires focus by default:

Fixes #75


src/three-components/SmoothControls.ts
```diff
@@ -400,6 +401,15 @@ export class SmoothControls extends EventDispatcher {
     }
   }
 
+  private get[$canInteract](): boolean {
+    if (this[$options].interactionPolicy === 'always-allow') {
+      return true;
+    }
+
+    const rootNode = this.element.getRootNode() as Document | ShadowRoot;
+    return rootNode.activeElement === this.element;
```
jsantell: Maybe should ensure that the interaction policy is 'allow-when-focused' (had to check SmoothControls to see if these were the only two valid tokens for interaction policy)

cdata: 👍 will fix

PR #312, opened 2019-01-28
Update three.js to r100

jsantell:
Only examples/ widget that changed is GLTFLoader -- in the future, work's being done so we can just pull these in via npm (https://github.com/mrdoob/three.js/pull/15518), so next three update may be able to use that.

[Migration guide](https://github.com/mrdoob/three.js/wiki/Migration-Guide) does not mention anything that would affect us.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1046,19 +1046,19 @@
     },
     "@types/clean-css": {
       "version": "3.4.30",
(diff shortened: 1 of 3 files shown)


cdata:
RIP `Octree`
pouroneout

cdata:
LGTM 👍 we are having Sauce Labs drama in CI this morning (tunnel sometimes doesn't connect), so we might have to reboot the build until it runs properly...

PR #313, opened 2019-01-28
Poster can be focused, hidden via keyboard

### LINKED ISSUE #297: poster attribute prevents keyboard interaction
There's no way to use the keyboard to interact with a `<model-viewer>` component with the `poster` attribute. Tabbing through the page just skips over the element.

After using the mouse to interact with the element, it can be tab-selected as normal.

Reproducing case:

- Use tab to navigate on a page containing `<model-viewer poster ...>`. A sample page is examples/accessibility/poster.html (if #296 is merged).

#### Live Demo
examples/accessibility/poster.html (if #296 is merged)
Or, the second element on https://googlewebcomponents.github.io/model-viewer/examples/loading.html

cdata:
This change improves keyboard interaction and general a11y when the `poster` attribute is used:

 - The poster element now has an `aria-label` that prompts the user to interact with it
 - The poster is now focusable, and receives focus when the `<model-viewer>` is focused if it is visible
 - When `space` or `enter` are pressed while the `<model-viewer>` is focused, the poster will be dismissed
 - When the poster is dismissed, focus is forwarded to the internal `<canvas>`

posterfocus

Fixes #297

### DIFF
--- a/src/features/loading.js
+++ b/src/features/loading.js
@@ -153,22 +153,28 @@ export const LoadingMixin = (ModelViewerElement) => {
 
       super.update(changedProperties);
 
+      const posterElement = this[$posterElement];
+
       if (changedProperties.has('alt')) {
-        this[$posterElement].setAttribute(
+        posterElement.setAttribute(
             'aria-label',
             `${this[$ariaLabel]}. ${this[$ariaLabelCallToAction]}`);
       }
 
       if (this[$shouldHidePoster]) {
-        this[$posterElement].classList.remove('show');
+        posterElement.classList.remove('show');
+        posterElement.setAttribute('aria-hidden', 'true');
+        posterElement.removeAttribute('tabindex');
       } else {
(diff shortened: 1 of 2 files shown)


cdata:
Just an advanced heads up: the test I crafted to test `poster` interaction when focused is flakey. I need to fix it before we proceed with this change.

cdata:
Flakey test deflaked. Should be good to review now!

examples/accessibility/poster.html
```diff
@@ -49,6 +49,7 @@ <h1>&lt;model-viewer&gt; - Accessibility - Poster</h1>
             <h2>Display poster until user interaction</h2>
             <template>
                 <model-viewer id="poster"
+                    controls
```
smalls: Is this required to transition from poster to a regular view?

cdata: No, but it was required to test that the controls were immediately usable after the transition

PR #314, opened 2019-01-29
Fix incorrect scale when zooming

### LINKED ISSUE #298: zooming in results in an off-center model
### Description
Zooming in far enough (175% on my screen/window size) results in the model getting smaller again and moving off-center, which persists even after zooming back out to 100%.

Reloading when zoomed in results in a properly centered, enlarged model.

Starting point (100%, never zoomed):
nozoom

Zooming in to 150%:
zoominto150

Zooming in to 175%:
zoominto175

Zooming back out to 100%:
zoomoutto100

#### Live Demo
https://googlewebcomponents.github.io/model-viewer/examples/default.html

### Browser Affected
- [X] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS
- [ ] Android
- [ ] iOS
- [X] Linux
- [ ] MacOS
- [ ] Windows

### Versions
- model-viewer: 5d084c0b5c68f4788e732ea9acaa078f5acef30e

cdata:
Zooming (e.g., with `⌘ +`/`⌘ -`) changes the effective `devicePixelRatio` and lays everything out accordingly. Previously, WebGLRenderer's DPR was not being updated under these circumstances, causing the model to be incorrectly staged under zoom events (and, incidentally, when turning on device emulation in dev tools).

Current behavior | New behavior
-----------------|---------------
resizedprbad|resizedprgood
resizedprmobilebad|resizedprmobilegood

Fixes #298

### DIFF
--- a/src/model-viewer-base.js
+++ b/src/model-viewer-base.js
@@ -125,6 +125,7 @@ export default class ModelViewerElementBase extends UpdatingElement {
           if (renderer.isPresenting) {
             return;
           }
+
           for (let entry of entries) {
             if (entry.target === this) {
               this[$updateSize](entry.contentRect);
--- a/src/test/three-components/Renderer-spec.js
+++ b/src/test/three-components/Renderer-spec.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import ModelViewerElementBase, {$canvas} from '../../model-viewer-base.js';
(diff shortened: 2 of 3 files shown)


cdata:
> Is there any cost to window.devicePixelRatio on every frame?

I doubt it, but I will check.

(IIUC we already read it on every frame)

cdata:
Seems like the cost is negligible:

image

image

image

image

PR #319, opened 2019-01-29
Improved a11y for <model-viewer> controls

### LINKED ISSUE #299: no prompt to inform the user that they can interact with the model
There isn't a prompt (either for screen readers or visually) that a model can be interacted with when using the `controls` attribute. There may be an opportunity for us to include a good default (which developers could choose to override).

There's also an opportunity to give screen reader feedback as the model is rotated by the user. If we assume that the model is oriented so the "front" faces the camera by default, we can give the user some prompts as the object is rotated ("Looking at the front of a 3D model", "Looking at the right side of a 3D model", "Looking at the back of a 3D model") but we need to be careful not to over-prompt the user.

This has been broken out into sub-issues:

 - [x] #315 
 - [x] #316
 - [x] #317

### LINKED ISSUE #315: Camera orientation is announced to screen readers
_As a screen reader user,
when I interact with a `<model-element>`,
I would like to hear about significant camera orientation changes,
so that I know what is being presented to me_

This is a sub-issue of #299

### LINKED ISSUE #316: Presence of camera controls is announced to screen readers
_As a screen reader user,
when I focus on a `<model-viewer>`,
I would like to be told about the camera controls,
so that I know that I can interact with it_

This is a sub-issue of #299

### LINKED ISSUE #317: Presence of camera controls announced by visual prompt
_As a web content viewer,
when I focus on a `<model-viewer>`,
I would like to be told about the camera controls,
so that I know that I can interact with it_

This is a sub-issue of #299

cdata:
# Improved a11y for `<model-viewer>` controls

This change is the rolled up PR including work from #322 and #325. Since this work all ties in to #299, I have copied over relevant graphics and descriptions so that everything can be tested / signed-off-on in its integrated form before it lands in master. Detailed info related to the change follows:

## Prompt screen reader users to interact w/ camera

This change proposes an ARIA-based interaction prompt for users who are focused on an activated `<model-viewer>` element with `controls` configured. The logic for this prompt is roughly:

 1. A user focuses the `<model-viewer>` (after any `poster` or equivalent is dismissed)
 2. After a period of time (currently ~3s), if the user has not interacted w/ the controls, they are prompted to interact ("Use mouse, touch or arrow keys to control the camera!")
 3. If the user blurs and then re-focuses the `<model-viewer>` without having interacted, repeat step 2
 4. If the user interacts with the `<model-viewer>` controls, we no longer prompt them to interact

interactionprompt

**Note:** the interaction prompt copy has been updated since the GIF was recorded, but the flow is the same.

## Visually prompt viewer to try controls

This change proposes a visual counterpart to the controls prompt suggested in #319 .

 - The timing and dismissal logic is the same as the ARIA prompt
 - An animated SVG is bundled to offer a reasonable default experience
 - Default prompt can be overridden by distributing "light tree" content to the `controls-prompt` slot

### Default visual prompt

controlsprompt

### Customized visual prompt

```html
<model-viewer src="something.gltf" controls>
  <!-- NOTE: Any custom content (not just images) can be used in place
       of the default prompt: -->
  <img src="something.gif" slot="controls-prompt">
</model-viewer>
```

controlspromptcustom

## Announce camera orientation to screen readers

This change is meant to complement a11y / interaction improvements brought by #319 and #322 .

 - When the orientation of the camera changes significantly, the new orientation is announced to screen readers
 - When the model loses focus and then regains focus, the original ARIA label is announced
 - Azimuth is separated into quadrants: front, left, back and right
 - The top third of the sphere is the "upper" pole, the bottom third is the "lower" pole

---
announceviewdirection

Fixes #316 
Fixes #317 via #322 
Fixes #315 via #325 
Fixes #299

### DIFF
--- a/examples/background-image.html
+++ b/examples/background-image.html
@@ -69,7 +69,6 @@
                     background-image="assets/whipple_creek_regional_park_04_1k.jpg"
                     alt="A 3D model of a damaged helmet with a forest in the background"
                     src="assets/DamagedHelmet/DamagedHelmet.gltf">
-                  <img src="https://upload.wikimedia.org/wikipedia/commons/c/c4/24-cell-orig.gif" slot="controls-prompt" width="200">
                 </model-viewer>
             </template>
         </example-snippet>
--- a/examples/background-image.html
+++ b/examples/background-image.html
@@ -68,7 +68,9 @@
                     controls
                     background-image="assets/whipple_creek_regional_park_04_1k.jpg"
                     alt="A 3D model of a damaged helmet with a forest in the background"
-                    src="assets/DamagedHelmet/DamagedHelmet.gltf"></model-viewer>
+                    src="assets/DamagedHelmet/DamagedHelmet.gltf">
+                  <img src="https://upload.wikimedia.org/wikipedia/commons/c/c4/24-cell-orig.gif" slot="controls-prompt" width="200">
+                </model-viewer>
             </template>
         </example-snippet>
       </div>
--- a/src/assets/controls-svg.js
+++ b/src/assets/controls-svg.js
@@ -1,0 +1,71 @@
+export default `
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="250px" height="200px" viewBox="0 0 250 200" fill="transparent">
+  <!-- Rotation arc -->
+  <path id="orbit"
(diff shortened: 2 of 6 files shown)


jsantell:
Looks great! I'd imagine this has value for non-keyboard navigating users as well (well, maybe more #322) for being in view and not touched after X seconds, or other triggers

Issue #259, opened 2018-12-04
add (and clean up existing) documentation


smalls:
Clean up our current documentation (mostly in README.md and examples/*) into a set of docs focused on helping new users with key use cases.

Note that there may be benefit to preserving what's currently in examples/* as it's initial purpose as a set of tests.

smalls:
As part of #320, we're pushing more documentation our of `README.md` and into `index.html`. Right now there's duplication, but as part of this let's take a consolidation pass to make sure the organization makes sense.

cdata:
I'm going to call this one closed as of #402 . Feel free to re-open if y'all disagree!

PR #320, opened 2019-01-29
Design update

yuinchien:
The index.html is now at the root level to serve as the landing page. Let's edit the content so we can publish the new design. A few suggestions/questions i have:

**sources.html**
Example 1 looks like a big rectangle surface instead of a cube, can we update the model and make it more cube looking?
Example 3 is empty, can we add some explanation? 
Example 4 what's the virtue of showing no model file here? 

**list.html & multiple-elements.html**
Should we consider adding code snippet so the left column is less empty. Or maybe not listing them in the landing page.


cdata:
Note that as of https://github.com/GoogleWebComponents/model-viewer/pull/320/commits/3b3c79fc72345623b703f7e7a84a5bbeedf5f647 this change breaks IE11 (my fault). We'll need to address that before this change can merge.

smalls:
> sources.html
> Example 1 looks like a big rectangle surface instead of a cube, can we update the model and make it more cube looking?
Yeah, this is no good. I'm not sure if there's a strong reason to prefer that model so I've changed to `reflective-sphere.gltf` in my upcoming PR.
> Example 3 is empty, can we add some explanation?
Yes will do!
> Example 4 what's the virtue of showing no model file here?
I've taken a stab but I hope for more thoughts from @cdata or @jsantell 
> 
> list.html & multiple-elements.html
> Should we consider adding code snippet so the left column is less empty. Or maybe not listing them in the landing page.
I think we could move these to a either a different performance page, like I've done for a11y. What do you think?

Some notes as I'm reviewing:

- `loading.html` is weird, but there have been recent changes. Let's check after we've updated to the latest. I'll revisit the text once I do that.

index.html
```diff
@@ -0,0 +1,208 @@
...
+          </li>
+          <li>
+            <a href="examples/pbr.html"><h4>Physically Based Rendering </h4></a>
+            <p>Showcase of high-resolution models with multiple textures and environment maps.</p>
+          </li>
+        </ul>
+
+        <h3 class="grouping-title">ATTRIBUTES</h3>
+
+        <ul class="list-attribute">
```
smalls: I like the inclusion of the attributes here. Should we remove the list from `README.md` to avoid duplication?

yuinchien: Sounds good to me. @cdata what do you think?

cdata: Sounds fine as long as we link folks to the docs from the README

smalls: Agreed. We need to take a large pass through the docs as part of #259, I've added a note there to check this as well.

jsantell:
> **sources.html**
> Example 1 looks like a big rectangle surface instead of a cube, can we update the model and make it more cube looking?
> Example 3 is empty, can we add some explanation?
> Example 4 what's the virtue of showing no model file here?
> 
> **list.html & multiple-elements.html**
> Should we consider adding code snippet so the left column is less empty. Or maybe not listing them in the landing page.

A lot of these examples were for testing different scenarios (e.g. example 1 should look like a flat square, testing boundary/resizing, and empty or broken models were for testing those, found a memory leak in one of the examples that was toggling through different background images quickly, using spheres that have a full range of metallic/roughness, multiple-elements showing what happens with 16 elements in view at once for stress test), and the new example pages are more geared towards devs learning how they work -- and we have more tests now too, so these "test" examples are probably less valuable now -- `fuzzer.html` probably makes the weirder cases redundant as well. SGTM to remove some of these "test cases"!

Except fuzzer :grin: which could live somewhere else outside of examples if needed. And the `multiple-elements.html` page is probably valuable still since we don't really have benchmarks -- could they still live in repo and hidden/not linked on the site? Low value in publicizing those test cases

yuinchien:
@jsantell i think having these tests in the repo is great. Since there will be more tests moving forward, should we consider create a *tests* folder so people know where to poke around if they want to?

yuinchien:
@cdata @jsantell per Jordan's feedback, he's up for keeping the examples featured in the site more "learning" focused. Can we remove list.html & multiple-elements.html from the index.html, and create a new folder called "tests" in the repo?

PR #323, opened 2019-01-30
Implement HDR image support (fixes #65) and transitions to using `scene.background`

### LINKED ISSUE #65: Support HDR environment maps, and HDR-ify default environment map

jsantell:
Broken up form of #237 with only HDR support and initial work for using `scene.background`, which greatly simplifies our skybox logic. The PR is big enough with enough test changes (and scene.background is necessary since we can't use two different mappings with the same texture), so PMREM can be in a follow up.

There was an issue with using `scene.background` with cube targets (https://github.com/mrdoob/three.js/pull/15662) so the work around here is until we update to r101 with the change -- tracking the removal of this hack via #196.

The HDR environments aren't terribly exciting without PMREM unfortunately, so this is pre-work for that, and to scope PMREM issues in its own branch.

Of note, below is an LDR env (top) and an HDR env; it appears the HDR map does not handle any roughness values at all without the PMREM, at least for the specular (the other lights appear to be more diffused as roughness increases). Not sure what's going on there, not what I'd expect three to do (and exactly the same area I was looking at for https://github.com/mrdoob/three.js/pull/15644, so a problem for another day if this is unintended. Either way, this should be thought of as part 1 of 2, i.e. not ready to promote/highlight.

screenshot from 2019-01-30 12-57-03


jsantell:
TIL IE doesn't have Array.prototype.find; repush test

jsantell:
`Object doesn't support property or method 'findIndex'` oh come on

src/test/features/environment-spec.js
```diff
@@ -70,18 +72,24 @@ const modelUsingEnvmap = (scene, meta) => {
...
 
 /**
  * Returns a promise that resolves when a given element is loaded
  * and has an environment map set that matches the passed in meta.
  * @see textureMatchesMeta
  */
-const waitForLoadAndEnvmap = (scene, element, meta) => Promise.all(
-    [waitForEvent(element, 'load'), waitForEnvmap(scene.model, meta)]);
+// const waitForLoadAndEnvmap = (scene, element, meta) => Promise.all(
+//    [waitForEvent(element, 'load'), waitForEnvmap(scene.model, meta)]);
```
cdata: Remove?

src/test/features/environment-spec.js
```diff
@@ -70,18 +72,24 @@ const modelUsingEnvmap = (scene, meta) => {
  * @param {Model} model
  * @param {Object} meta
  */
-const waitForEnvmap = (model, meta) => waitForEvent(
-    model,
-    'envmap-change',
-    e => textureMatchesMeta(e.value, {...meta, type: 'EnvironmentMap'}));
+const waitForEnvmap = (model, meta) =>
+    waitForEvent(model, 'envmap-change', e => {
```
cdata: `e => event`

src/test/features/environment-spec.js
```diff
@@ -70,18 +72,24 @@ const modelUsingEnvmap = (scene, meta) => {
  * @param {Model} model
  * @param {Object} meta
  */
-const waitForEnvmap = (model, meta) => waitForEvent(
-    model,
-    'envmap-change',
-    e => textureMatchesMeta(e.value, {...meta, type: 'EnvironmentMap'}));
+const waitForEnvmap = (model, meta) =>
```
cdata: `Envmap => EnvironmentMap` throughout

src/test/three-components/ModelScene-spec.js
```diff
@@ -26,13 +26,6 @@ function invertPad(vec3) {
   return vec3.clone().multiplyScalar(ROOM_PADDING_SCALE);
 }
 
-// Checks that the skysphere is within the camera's far plane
-function ensureSkysphereVisible(scene) {
```
cdata: Testing and assertions reduced here. What changed?

src/three-components/TextureUtils.js
```diff
@@ -70,42 +71,47 @@ export default class TextureManager extends EventDispatcher {
...
-    return cubemap;
+    return generator.renderTarget;
   }
 
   /**
    * @param {string} url
    * @return {Promise<THREE.Texture>}
    */
   async load(url) {
+    const isHDR = /\.hdr$/.test(url);
```
cdata: Let's name this `RegExp` e.g., `const HDR_FILE_RE = /\.hdr$/;` and hoist it to the module scope.

src/three-components/TextureUtils.js
```diff
@@ -115,39 +121,55 @@ export default class TextureManager extends EventDispatcher {
...
+      skybox = await this.equirectangularToCubemap(equirect);
+
+      // Use the same texture (directly instead of WebGLRenderTargetCube)
+      // for the environment map until #215
+      envmap = skybox.texture;
+
+      equirect.dispose();
+
+      return {envmap, skybox};
     } catch (e) {
```
cdata: `e => error`

src/three-components/TextureUtils.js
```diff
@@ -115,39 +121,55 @@ export default class TextureManager extends EventDispatcher {
...
-      if (cubemap) {
-        cubemap.dispose();
+      if (skybox) {
+        skybox.dispose();
+      }
+      if (envmap) {
+        envmap.dispose();
       }
+
+      console.error(e);
```
cdata: Remove and/or re-throw

cdata:
HDR looks so good 🤩thanks for all your work to make this happen @jsantell

src/test/three-components/ModelScene-spec.js
jsantell: The positioning of the skysphere is in the hands of three now, and the environment feature tests cover whether or not it's visible, these tests here ensured the skybox was positioned and sized correctly

cdata: 👍

examples/background-image.html
```diff
@@ -144,7 +143,7 @@ <h2>Cycling between <span class="attribute">background-image</span> and <span cl
             <template>
                 <model-viewer
                     id="toggle-image"
-                    backgroud-color="#ff0077"
```
cdata: Nice catch 👍

src/features/environment.js
```diff
@@ -134,18 +127,16 @@ export const EnvironmentMixin = (ModelViewerElement) => {
...
 
-      this[$scene].skysphere.material.map = null;
-      this[$scene].skysphere.material.needsUpdate = true;
+      this[$setShadowLightColor](parsedColor);
 
-      // TODO can cache this per renderer and color
-      const cubemap = textureUtils.generateDefaultEnvMap();
-      this[$currentCubemap] = cubemap;
-      this[$scene].model.applyEnvironmentMap(this[$currentCubemap]);
+      // TODO(#336): can cache this per renderer and color
```
cdata: 🍻 thanks for filing!

src/test/three-components/TextureUtils-spec.js
```diff
@@ -81,25 +90,53 @@ suite('TextureUtils', () => {
...
+      textures = await textureUtils.generateEnvironmentTextures(EQUI_URL);
+      expect(textures.skybox.texture.isTexture).to.be.ok;
+      expect(textures.environmentMap.isTexture).to.be.ok;
 
       expect(textureMatchesMeta(
-                 textures.equirect, {type: 'Equirectangular', url: EQUI_URL}))
+                 textures.skybox.texture, {mapping: 'Cube', url: EQUI_URL}))
           .to.be.ok;
+
+      // TODO (#215): Re-enable once PMREM is added
```
cdata: 👍

src/three-components/TextureUtils.js
```diff
@@ -36,32 +41,29 @@ const defaultConfig = {
 // describe the type of texture within the context of this application.
 const userData = {
   url: null,
-  // 'Equirectangular', 'EnvironmentMap'
-  type: null,
+  // 'Equirectangular', 'Cube'
```
cdata: 💯

src/three-components/TextureUtils.js
```diff
@@ -70,91 +72,110 @@ export default class TextureManager extends EventDispatcher {
...
-    return cubemap;
+    return generator.renderTarget;
   }
 
   /**
    * @param {string} url
    * @return {Promise<THREE.Texture>}
    */
   async load(url) {
+    const isHDR = HDR_FILE_RE.test(url);
```
cdata: 👍

src/three-components/TextureUtils.js
```diff
@@ -70,91 +72,110 @@ export default class TextureManager extends EventDispatcher {
...
+      }
+
       return null;
     }
   }
 
   dispose() {
-    this.envMapGenerator.camera.renderTarget.dispose();
-    this.envMapGenerator = null;
+    this.environmentMapGenerator.camera.renderTarget.dispose();
```
cdata: Should `EnvironmentMapGenerator` have its own `dispose` method so that we don't have to dig this deeply into its implementation details here?

jsantell: yeah, probably, this is a bit much

cdata:
Excellent work 👍 LGTM

PR #329, opened 2019-02-01
Make <model-viewer> resilient to null GL context

### LINKED ISSUE #327: <model-viewer> is incompatible with Google's crawler
`<model-viewer>` cannot be rendered by Chrome 41, which is [a strong indicator](https://developers.google.com/search/docs/guides/rendering) that the Google crawler cannot directly index `<model-viewer>` content right now.

### LINKED ISSUE #328: Documentation cannot be viewed in legacy browsers / crawlers
1) We serve our standard bundles on our doc site
2) We host our doc site on Github Pages

Our site cannot be viewed by crawlers and legacy browsers due to number 1.
We cannot take advantage of UA-detection to perform differential serving due to number 2.

In order for our docs to be properly indexed by crawlers, and (arguably less important) viewable in legacy browsers, we need address the above constraints.

This work has been broken out from #327

cdata:
This change proposes hardening `<model-viewer>` against lack of WebGL support.

Currently, when `<model-viewer>` is instantiated on a page that does not support WebGL, it throws an uncaught exception. This result of this might be benign if `<model-viewer>` is already present in the serialized HTML form of the document.

However, if `<model-viewer>` is being produced dynamically (e.g., being stamped by JS from a `<template>` or being created imperatively with `document.createElement`), the effect is much more catastrophic: it almost certainly results in `<model-viewer>` *not appearing at all* in the content document.

Regardless, it would always be better for `<model-viewer>` to recover gracefully from a scenario where WebGL is not supported. Even if `<model-viewer>` cannot render, it could still offer critical semantic context to vision impaired users or crawlers.

Fixes #327

Does **not** address #328


cdata:
>  I'd initially expect we'd be able to short circuit this earlier

Same, but unfortunately the way we pass inter-related objects around right now undermines this a bit. It definitely deserves some cleanup.

> what can we do to test the full-fledge element in that scenario?

We can copy over the `null` context test to all of the feature test suites. That's the best thing I can think of for now.

cdata:
@jsantell in response to your feedback in a more integrated strategy, I have "templatized" the basic test checks ("Can I construct the element?" "Can I append it?" "What if there is no WebGL?") and included this template in all of the feature suites.

src/test/templates.ts
```diff
@@ -0,0 +1,55 @@
+import ModelViewerElementBase from '../model-viewer-base.js';
```
jsantell: Very nice! :+1:

PR #345, opened 2019-02-08
Sparsely announce loading status to screen readers

### LINKED ISSUE #301: verbalize loading status of models (but not too often)
Models may be loading in the background (with `reveal-when-loaded preload`). This is useful information for the user, but we don't want to overprompt. A suggestion:

- When the page first loads, if models are loading in the background give a single global prompt "This page includes 3D models that are loading"
- Once *all* models have been loaded, prompt "All 3D models are loaded"

Each model will give appropriate information as normal

cdata:
This change proposes providing screenreader users with sparse updates on the current loading status of all `<model-viewer>` elements on the page. Some details of the change:

 - A singleton manages aggregation and observation of the loading state of all `<model-viewer>` instances that apply the `LoadingMixin`
 - `<model-viewer>` instances are observed when they connect to the document, and ignored after they disconnect
 - Two announcements are made: an initial one ("There are 3D models that are loading") and a final one ("All 3D models have loaded")
 - An initial announcement is made if (and only if) there is at least one `<model-viewer>` in the document
 - When all currently observed instances have either loaded or errored, the final announcement is made
 - Note: since the status announcer is bundled with the `<model-viewer>` implementation, at least one `<model-viewer>` must be upgraded and in the document for the initial to be made

Here is a demo of the behavior, using emulated network conditions ("Fast 3G" in DevTools):

announceloading

Fixes #301

### DIFF
--- a/src/features/loading/status-announcer.ts
+++ b/src/features/loading/status-announcer.ts
@@ -14,13 +14,14 @@
  */
 
 import ModelViewerElementBase from '../../model-viewer-base.js';
-import {debounce} from '../../utils.js';
+import {debounce, getFirstMapKey} from '../../utils.js';
 
 export const INITIAL_STATUS_ANNOUNCEMENT =
     'This page includes one or more 3D models that are loading';
 export const FINISHED_LOADING_ANNOUNCEMENT =
     'All 3D models in the page have loaded';
 export const UPDATE_STATUS_DEBOUNCE_MS = 100;
+
 
 const $modelViewerStatusInstance = Symbol('modelViewerStatusInstance');
 const $updateStatus = Symbol('updateStatus');
@@ -131,8 +132,10 @@ export class LoadingStatusAnnouncer {
     instanceStatus.instanceWasUnregistered();
 
     if (this.modelViewerStatusInstance === modelViewer) {
-      this.modelViewerStatusInstance =
-          statuses.size > 0 ? statuses.keys().next().value : null;
+      this.modelViewerStatusInstance = statuses.size > 0 ?
+          getFirstMapKey<ModelViewerElementBase, InstanceLoadingStatus>(
+              statuses) :
+          null;
     }
   }
 
@@ -163,7 +166,6 @@ export class LoadingStatusAnnouncer {
       return;
     }
 
-    console.log('Initial status');
     this.statusElement.textContent = INITIAL_STATUS_ANNOUNCEMENT;
     this.statusUpdateInProgress = true;
 
@@ -173,7 +175,6 @@ export class LoadingStatusAnnouncer {
       await Promise.all(loadingPromises);
     }
 
-    console.log('Finished status');
     this.statusElement.textContent = FINISHED_LOADING_ANNOUNCEMENT;
     this.statusUpdateInProgress = false;
   }
--- a/src/features/loading/status-announcer.ts
+++ b/src/features/loading/status-announcer.ts
@@ -13,6 +13,7 @@
  * limitations under the License.
  */
 
+import {EventDispatcher} from 'three';
 import ModelViewerElementBase from '../../model-viewer-base.js';
 import {debounce, getFirstMapKey} from '../../utils.js';
 
@@ -20,15 +21,14 @@ export const INITIAL_STATUS_ANNOUNCEMENT =
     'This page includes one or more 3D models that are loading';
 export const FINISHED_LOADING_ANNOUNCEMENT =
     'All 3D models in the page have loaded';
-export const UPDATE_STATUS_DEBOUNCE_MS = 100;
+const UPDATE_STATUS_DEBOUNCE_MS = 100;
 
 
 const $modelViewerStatusInstance = Symbol('modelViewerStatusInstance');
 const $updateStatus = Symbol('updateStatus');
 
 interface InstanceLoadingStatus {
-  instanceWasUnregistered: () => void;
-  loadAttemptCompletes: Promise<any>;
+  onUnregistered: () => void;
 }
 
 /**
@@ -42,7 +42,7 @@ interface InstanceLoadingStatus {
  *  1. There are <model-viewer> elements that have yet to finish loading
  *  2. All <model-viewer> elements in the page have finished attempting to load
  */
-export class LoadingStatusAnnouncer {
+export class LoadingStatusAnnouncer extends EventDispatcher {
   /**
    * The "status" instance is the <model-viewer> instance currently designated
    * to announce the loading status of all <model-viewer> elements in the
@@ -73,6 +73,7 @@ export class LoadingStatusAnnouncer {
       debounce(() => this.updateStatus(), UPDATE_STATUS_DEBOUNCE_MS);
 
   constructor() {
+    super();
     this.statusElement.setAttribute('role', 'status');
     this.statusElement.style.position = 'absolute';
     this.statusElement.style.color = 'transparent';
@@ -88,28 +89,29 @@ export class LoadingStatusAnnouncer {
       return;
     }
 
-    let instanceWasUnregistered = () => {};
+    let onUnregistered = () => {};
     const loadShouldBeMeasured =
         modelViewer.loaded === false && !!(modelViewer as any).src;
-    const loadAttemptCompletes = loadShouldBeMeasured ?
-        new Promise((resolve) => {
-          const resolveHandler = () => {
-            resolve();
+    const loadAttemptCompletes = new Promise((resolve) => {
+      if (!loadShouldBeMeasured) {
+        resolve();
+        return;
+      }
 
-            modelViewer.removeEventListener('load', resolveHandler);
-            modelViewer.removeEventListener('error', resolveHandler);
-          };
+      const resolveHandler = () => {
+        resolve();
 
-          modelViewer.addEventListener('load', resolveHandler);
-          modelViewer.addEventListener('error', resolveHandler);
+        modelViewer.removeEventListener('load', resolveHandler);
+        modelViewer.removeEventListener('error', resolveHandler);
+      };
 
-          instanceWasUnregistered = resolveHandler;
-        }) :
-        Promise.resolve();
+      modelViewer.addEventListener('load', resolveHandler);
+      modelViewer.addEventListener('error', resolveHandler);
 
-    this.registeredInstanceStatuses.set(
-        modelViewer, {instanceWasUnregistered, loadAttemptCompletes});
+      onUnregistered = resolveHandler;
+    });
 
+    this.registeredInstanceStatuses.set(modelViewer, {onUnregistered});
     this.loadingPromises.push(loadAttemptCompletes);
 
     if (this.modelViewerStatusInstance == null) {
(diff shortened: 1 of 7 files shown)


src/features/loading/status-announcer.ts
```diff
@@ -0,0 +1,181 @@
...
+ * with a LoadingStatusAnnouncer singleton. As they are disconnected, the are
+ * also unregistered. Announcements are made to indicate the following
+ * conditions:
+ *
+ *  1. There are <model-viewer> elements that have yet to finish loading
+ *  2. All <model-viewer> elements in the page have finished attempting to load
+ */
+export class LoadingStatusAnnouncer {
+  /**
+   * The "status" instance is the <model-viewer> instance currently designated
```
smalls: bikeshed: I initially assumed that status meant something else, although this comment cleared it up for me.

Would "announcer" be clearer?

smalls: Ooh, I see - https://www.w3.org/TR/wai-aria-1.1/#status (linked below) gives a bit more context. I'll leave this comment here but I feel even less strongly about this bikeshed.

An idea might be "statusAnnouncer" or something, but that's cumbersome.

src/test/helpers.js
```diff
@@ -93,4 +93,34 @@ export const dispatchSyntheticEvent = (element, type, properties = {
...
+ * decide to move files around.
+ *
+ * @param {string} name
+ * @return {string}
+ */
 export const assetPath = (name) => `${ASSETS_DIRECTORY}${name}`;
+
+
+/**
+ * Returns truu if the given element is in the tree of the document of the
```
smalls: Should be true, not truu.

smalls:
```suggestion
 * Returns true if the given element is in the tree of the document of the
```

src/test/features/loading/status-announcer-spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        test('sets finished status when all models are loaded', async () => {
+          const elementOne = new ModelViewerElement();
+          const elementTwo = new ModelViewerElement();
+
+          elementOne.src = elementTwo.src = assetPath('Astronaut.glb');
+
+          loadingStatusAnnouncer.registerInstance(elementOne);
+          loadingStatusAnnouncer.registerInstance(elementTwo);
+
+          await Promise.all(
```
jsantell: `await until(() => elementOne.loaded && elementTwo.loaded)`

src/utils.ts
```diff
@@ -158,3 +163,32 @@ export const resolveDpr = (() => {
...
+
+  let firstKey: T|null = null;
+
+  try {
+    map.forEach((_value: U, key: T, _map: Map<T, U>) => {
+      firstKey = key;
+      // Stop iterating the Map with forEach:
+      throw new Error();
+    });
+  } catch (_error) {
```
jsantell: :cold_sweat:  what a workaround!

src/features/loading/status-announcer.ts
```diff
@@ -0,0 +1,181 @@
...
+   * status for all <model-viewer> elements currently in the page. It has its
+   * role attribute set to "status", which causes screen readers to announce
+   * any changes to its text content.
+   *
+   * @see https://www.w3.org/TR/wai-aria-1.1/#status
+   */
+  readonly statusElement: HTMLParagraphElement = document.createElement('p');
+  protected statusUpdateInProgress: boolean = false;
+
+  protected[$updateStatus]: () => void =
```
jsantell: Is setting the output of `debounce` to `void` to ensure that it returns a `void 0` value? Shouldn't this be invalid? Is this a typescript thing?

src/features/loading/status-announcer.ts
```diff
@@ -0,0 +1,181 @@
...
+      value.shadowRoot.appendChild(statusElement);
+    } else if (statusElement.parentNode != null) {
+      statusElement.parentNode.removeChild(statusElement);
+    }
+
+    this[$modelViewerStatusInstance] = value;
+    this[$updateStatus]();
+  }
+
+  protected async updateStatus() {
```
jsantell: IIUC, setting `modelViewerStatusInstance` starts the debounced polling, which only happens when an element is registered, and it's the first element to do so (when `modelViewStatusInstance` has not yet been created), and when an element is disconnected.

What of a scenario where `n` models are loaded, and later, a new model has been added, calling `registerInstance`, and then not starting the polling because `modelViewerStatusInstance` already exists? This could already be covered and I'm missing a piece, it looks like the last test *could* cover this, but I suspect both elements are registered immediately, while the initial debounce cycle is still running

src/features/loading/status-announcer.ts
```diff
@@ -0,0 +1,181 @@
...
+    'All 3D models in the page have loaded';
+export const UPDATE_STATUS_DEBOUNCE_MS = 100;
+
+
+const $modelViewerStatusInstance = Symbol('modelViewerStatusInstance');
+const $updateStatus = Symbol('updateStatus');
+
+interface InstanceLoadingStatus {
+  instanceWasUnregistered: () => void;
+  loadAttemptCompletes: Promise<any>;
```
jsantell: `loadAttemptComplete` doesn't appear to be used as a member of `InstanceLoadingStatus`

src/features/loading/status-announcer.ts
```diff
@@ -0,0 +1,181 @@
...
+export const FINISHED_LOADING_ANNOUNCEMENT =
+    'All 3D models in the page have loaded';
+export const UPDATE_STATUS_DEBOUNCE_MS = 100;
+
+
+const $modelViewerStatusInstance = Symbol('modelViewerStatusInstance');
+const $updateStatus = Symbol('updateStatus');
+
+interface InstanceLoadingStatus {
+  instanceWasUnregistered: () => void;
```
jsantell: This name was a bit confusing... `onUnregister`? `cleanup`?

src/features/loading/status-announcer.ts
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * This element is a node that floats around the document as the status
+   * instance changes (see above). It is a singleton that represents the loading
+   * status for all <model-viewer> elements currently in the page. It has its
+   * role attribute set to "status", which causes screen readers to announce
+   * any changes to its text content.
+   *
+   * @see https://www.w3.org/TR/wai-aria-1.1/#status
+   */
+  readonly statusElement: HTMLParagraphElement = document.createElement('p');
```
jsantell: I wonder if it'd be worthwhile for a singleton element somewhere outside of a <m-v> element that could be used for any global interactions like this; reduce complexity, but of course, any changes to the DOM are unexpected, should be noted

src/features/loading/status-announcer.ts
```diff
@@ -0,0 +1,181 @@
...
+  protected statusUpdateInProgress: boolean = false;
+
+  protected[$updateStatus]: () => void =
+      debounce(() => this.updateStatus(), UPDATE_STATUS_DEBOUNCE_MS);
+
+  constructor() {
+    this.statusElement.setAttribute('role', 'status');
```
jsantell: Do any browsers require any sort of element attributes before being used for ARIA props? (e.g. requires a minimum width/height or something)

src/test/features/loading/status-announcer-spec.ts
```diff
@@ -0,0 +1,171 @@
...
+              [until(() => elementOne.loaded), until(() => elementTwo.loaded)]);
+          await timePasses(TIME_TO_UPDATE_MS);
```
jsantell: This seems a bit fragile; similar to some of the async environment work, I wonder if we can expose some internal events for the debounce starting/stopping, similar to `Model`'s `envmap-change` event, which is really only for tests.

src/features/loading/status-announcer.ts
cdata: Will fix

src/features/loading/status-announcer.ts
cdata: Will fix

src/features/loading/status-announcer.ts
cdata: We took this approach with Polymer Elements in the past (see [iron-a11y-announcer](https://github.com/PolymerElements/iron-a11y-announcer)), and my opinion of it is that it is basically not good. Any time a component modifies the document outside of its shadow root, it risks running afoul of user expectations of how their document changes over time.

There is definitely room for a status announcer that is generalized for more than just loading status updates. But, I'm skeptical that we can generalize it beyond the domain of status announcements related to a specific custom element.

src/features/loading/status-announcer.ts
cdata: This is just `TypeScript` speak for "doesn't return anything" (literally means returns `undefined`).

src/features/loading/status-announcer.ts
cdata: Visibility is important. The element will probably be ignored if you use `opacity: 0` or `visibility: hidden`, for example.

src/features/loading/status-announcer.ts
```diff
@@ -0,0 +1,181 @@
...
+   */
+  registerInstance(modelViewer: ModelViewerElementBase) {
+    if (this.registeredInstanceStatuses.has(modelViewer)) {
+      return;
+    }
+
+    let instanceWasUnregistered = () => {};
+    const loadShouldBeMeasured =
+        modelViewer.loaded === false && !!(modelViewer as any).src;
+    const loadAttemptCompletes = loadShouldBeMeasured ?
```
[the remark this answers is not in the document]
cdata: Will fix

src/features/loading/status-announcer.ts
cdata: BTW `void` is the implicit return signature for any function or method where it is not otherwise specified. You can omit it in many cases, but not if you are describing a return value type for a function that returns nothing (as we are doing here).

src/features/loading/status-announcer.ts
cdata: My thoughts were that there were two important scenarios to cover:

 1. First load of the page, where we will pick up N elements and then conclude when they have loaded (including lazily added ones in the mean time)
 2. The case where the site is a SPA, so a page navigation implies a new battery of elements might be added, which means we should do the announcement flow over again

Notably this leaves out a third case (as I think you are pointing out):

 3. A `<model-viewer>` is added after the final announcement, but before any of the previous `<model-viewer>` elements have been removed

This was done deliberately. My thinking here is that the risk of spammy loading announcements was a worse outcome than not saying anything about a sufficiently late-added `<model-viewer>`.

src/test/features/loading/status-announcer-spec.ts
cdata: 👍 will fix

src/test/features/loading/status-announcer-spec.ts
cdata: I'll investigate this

src/test/helpers.js
cdata: Will fix

src/utils.ts
cdata: bleeding-eyes.gif

src/features/loading/status-announcer.ts
cdata: Yeah, I was trying to stick to the ARIA terminology. I'll ponder changing it.

src/test/features/loading/status-announcer-spec.ts
cdata: `LoadingStatusAnnouncer` now inherits from `EventDispatcher`. The tests have been updated to listen to events rather than wait for a fixed amount of time.

cdata:
@googlebot I signed it

Issue #374, opened 2019-02-18
Bad render results for KHR_materials_unlit Khronos test model


cdata:
The [Khronos Unlit Test sample model](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/UnlitTest) README says that for the test model using this extension:

> The visible surface ... should appear as a uniform color, so much so that the 3D effect is intentionally lost.

And offers the following as examples of good and bad render results:

Good | Bad
------|-----
![image](https://user-images.githubusercontent.com/240083/52973400-d8513600-3372-11e9-97a5-67d7e3574643.png)|![image](https://user-images.githubusercontent.com/240083/52973413-e1420780-3372-11e9-889b-70a78dcbb5b7.png)

Here are examples of how `<model-viewer>` renders this model w/ various settings, all if which seem incorrect (possibly for different reasons):

model only|HDR `background-image`|`experimental-pmrem`
-----------|--------------------------|-----------------------
![image](https://user-images.githubusercontent.com/240083/52973636-8eb51b00-3373-11e9-9ca2-2c070ecac74c.png)|![image](https://user-images.githubusercontent.com/240083/52973650-9a084680-3373-11e9-957c-0813354df7fc.png)|![image](https://user-images.githubusercontent.com/240083/52973671-a9878f80-3373-11e9-8798-b797c2b2f5c8.png)

I also tested to see how Filament's `gltf_viewer` handles this case. It seems to handle the model correctly (presumably it supports the `KHT_materials_unlit` extension).

Filament `gltf_viewer`|
----------------------|
![image](https://user-images.githubusercontent.com/240083/52973721-d89e0100-3373-11e9-88cc-1ce3714c36a6.png)|

donmccurdy:
The viewer will need to avoid setting .envMap on MeshBasicMaterial - that material type is not subject to all the limits of the glTF unlit material, and adding the envMap will “break” it. Not sure what’s going on in the “model only” case though.

jsantell:
> The viewer will need to avoid setting .envMap on MeshBasicMaterial - that material type is not subject to all the limits of the glTF unlit material, and adding the envMap will “break” it. Not sure what’s going on in the “model only” case though.

A "default" env map is created in that case, leading to the "lit" appearance there. For PMREM, this is all black since an unlit material doesn't know what to do with a PMREM environment map

Issue #372, opened 2019-02-17
Support for models with multiple sets of UV coordinates


cdata:
Currently, when presenting the [Khronos Multi-UV sample glTF](), we get the wrong result. Below is a comparison between `<model-viewer>` and Filament `gltf_viewer` render results:

`<model-viewer>` | Filament `gltf_viewer`
-------------------|----------------------
![image](https://user-images.githubusercontent.com/240083/52920776-52c37c80-32c5-11e9-9b93-e59b970ba76c.png)|![image](https://user-images.githubusercontent.com/240083/52920781-6838a680-32c5-11e9-90a0-a98580a25ff2.png)

Issue #371, opened 2019-02-17
Failure to render glTFs with animations


cdata:
`<model-viewer>` cannot render the [Khronos Brain Stem sample glTF](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BrainStem). It fails with the following exception before anything is drawn:

![image](https://user-images.githubusercontent.com/240083/52918968-29e5bc00-32b2-11e9-8209-dd50de8d6d01.png)

The failure is on a call to update the model's skeleton: `object.skeleton.update();`.

Notably, this is a sample model with a built-in animation.

This bug was discovered while pursuing #360

cdata:
This same issue occurs when trying to load the [Khronos Cesium Man sample glTF](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/CesiumMan). It's probably animation related.

cdata:
Ditto: [Khronos Monster sample glTD](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Monster)

jsantell:
Looks like we have a SkinnedMesh without any `skeleton`

PR #361, opened 2019-02-14
Add fidelity test for HDR/PMREM

cdata:
This change proposes a final addition to #338 : a fidelity test to help us track changes and potential regressions related to HDR and PMREM. Some details:

 - A new HDR was added from HDRI Haven: airport workshop 01
 - The new test pairs the high-rez Damaged Helmet w/ airport workshop HDR and `experimental-pmrem`
 - A new event is now dispatched by `<model-viewer>` when the environment map changes
   - This is necessary for the fidelity test to know when it is okay to take a screenshot of `<model-viewer>`
 - A golden of `<model-viewer>` was added so that we can track changes compared to ourselves
 - This change led to an update to #351 such that the skybox in `gltf_renderer` now matches the orientation of the skybox in `<model-viewer>`

`<model-viewer>` <-> Filament|
--------------------------------|
image|

PR #375, opened 2019-02-18
Add Khronos glTF sample models to fidelity test suite

### LINKED ISSUE #360: Add Khronos glTF Sample Models to fidelity test suite

cdata:
<div align="center">
  
</div>

## Khronos glTF sample fidelity tests

This change proposes the addition of several tests to the fidelity suit, all based on [Khronos glTF sample models](https://github.com/KhronosGroup/glTF-Sample-Models). The highlights of this change include:

 - **27 new fidelity tests**, all based on Khronos glTF sample models
   - Some tests omitted, due to outright failure (see #371) or negligible testing value
   - See below for observations, analysis and remarks related to initial results
 - Screenshot update automation expanded to include support for updating `<model-viewer>` "master" screenshots
   - Used to test for regressions against our own accepted renderings
   - `npm run update-screenshots` will update Filament and `<model-viewer` "master" at the same time
 - Filament-based `gltf_renderer` updated to use the same model framing strategy as <model-viewer>
   - Puts us on a good path for keeping staging in line as we evolve `<model-viewer>` going forward
   - See below for additional details
 - A new command `npm run checkout-samples` will clone and/or update Khronos samples directly from Github
   - Runs automatically during `npm install` and `npm ci`

🛂**Note for reviewers:** these tests bring attention to a noticeable lighting / saturation difference between Filament and `<model-viewer>` for many of the cases. This is currently expected, and is assumed to be due to default lighting, tonemapping and exposure configurations across both renderers. For relevant background discussion, please refer to https://github.com/GoogleWebComponents/model-viewer/pull/338, where PMREM support was introduced.

#### 16bpc PNG textures

The first model to show a disparity is Antique Camera:

`<model-viewer>`|Filament `gltf_renderer`
-------------------|--------
image|image

I dug into this problem and discovered that Filament was failing to load some of the model's textures because [it was using a third-party library that did not support loading 16bpc PNGs](https://github.com/google/filament/issues/832). [A fix for this issue has already landed in Filament](https://github.com/google/filament/pull/835), so thankfully this particular disparity should be gone by the time we update the screenshots again.

#### Vertex colors

The next stand-out issue is that for an as-yet undiagnosed reason, Filament does not support vertex colors in glTFs yet. This issue is already documented by the Filament project (https://github.com/google/filament/issues/528). Here are the results for the two fidelity tests that check for vertex color support:

`<model-viewer>`|Filament `gltf_renderer`
-------------------|-------
image|image
image|image

#### Texture transforms

The Khronos body includes [a sample model that tests support for the `KHR_texture_transform` extension](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/TextureTransformTest). Filament shows an incorrect result for scale, transform and rotation ([subsequent discussion](https://github.com/google/filament/issues/528#issuecomment-465631129) on Filament project suggests this will be fixed by changing their glTF loader):

`<model-viewer>`|Filament `gltf_renderer`
-------------------|------------------------
image|image

#### Multiple UVs

One of the notable gaps in `<model-viewer>`'s glTF support that is uncovered by these tests is the lack of [multiple UV](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/MultiUVTest) support (#372 filed to track). Filament's render matches the expected result in this case:

`<model-viewer>`|Filament `gltf_renderer`
-------------------|-----------------------
image|image

#### Normal tangents

The glTF samples include two test cases ([1](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/NormalTangentTest), [2](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/NormalTangentMirrorTest)) related to normal tangents. The mirrored case tests whether the tangents supplied by the model are used (as opposed to the tangents being calculated by the renderer). The mirrored case is notably incorrect in Filament (already documented in https://github.com/google/filament/issues/528):

`<model-viewer>`|Filament `gltf_renderer`
-------------------|------------------------
image|image

#### Unlit materials

These tests uncovered that `<model-viewer>` completely fails to do the right thing with unlit materials (#374 filed to track). Thankfully, @jsantell has already put a fix (#377) up for consideration. Here is an example of the current rendering failure (Filament's render result is what we expect):

`<model-viewer>`|Filament `gltf_renderer`
-------------------|-----------------------
image|image

#### Alpha blending

The Khronos models include [a test case for alpha blending](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AlphaBlendModeTes). These render results surprised me: `<model-viewer>` seems to yield an incorrect result in all columns except for the second one (filed #381 to track). Filament renders the expected result:

`<model-viewer>`|Filament `gltf_renderer`
-------------------|------------------------
image|image

#### Other inconsistencies

There were a few render results where there were obvious inconsistencies or rendering artifacts, but the cause is less certain.

🛂**Reviewers: please consider the following cases and speculate about what might be causing these inconsistencies and whether we can/should take action to address them:**

##### [Environment Test](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/EnvironmentTest)

The bottom row seems very different. `<model-viewer>` seems closer to the result shown in the Khronos source repo:

`<model-viewer>`|Filament `gltf_renderer`
-------------------|------------------------
image|image

##### [Orientation Test](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/OrientationTest)

Particularly I want to call out the weird flecks of white on our render:

`<model-viewer>`|Filament `gltf_renderer`
-------------------|------------------------
image|image

### Staging models in `gltf_renderer`

In [our initial effort to compare `<model-viewer>` to other renderers](https://github.com/GoogleWebComponents/model-viewer/pull/305), we hand-crafted tests and manually captured screenshots. In the course of doing this, we were lucky that Filament and `<model-viewer>` happened to stage certain models with near-pixel-perfect camera projection and relative positioning. For example, consider this fidelity test added in #361 :

Example of coincidentally similar staging|
----------------------------------------|
image|

Unfortunately, this coincidence of staging strategies is just that: a coincidence. After adding the battery of Khronos sample model tests, it became immediately obvious that many types of models were not staged the same way by both renderers.

So, this change includes a modification of our Filament-based `gltf_renderer`. Model staging has been updated to closely match the strategy used by `<model-viewer>`'s own `ModelScene`:

https://github.com/GoogleWebComponents/model-viewer/blob/917251cb8fd1f65508078895862f22ac0504567a/src/three-components/ModelScene.js#L163-L204

https://github.com/GoogleWebComponents/model-viewer/blob/917251cb8fd1f65508078895862f22ac0504567a/src/three-components/ModelScene.js#L214-L240

Original, inconsistent staging|Updated staging strategy
-----------------------------|-------------------------
image|image

Our Filament screen captures will now be staged as expected. Hopefully this change will allow us to easily/coherently adapt our Filament staging over time as we change `<model-viewer>` as well.

### Follow-on work

The following issues have been filed during the course of work on this change, and represent follow-on work towards improved compatibility/conformance of `<model-viewer>` and consistency across renderers:

 - #371 `<model-viewer>` fails to render glTFs w/ animations
 - #372 Support for multiple UVs
   - Should be addressed in Three.js via https://github.com/mrdoob/three.js/issues/12608
 - #381 Fix rendering for alpha blend modes
 - #374 Fix rendering for unlit models
   - Change already available for consideration courtesy of @jsantell : #377 
 - #379 To address pixelation in environment maps
 - https://github.com/google/filament/issues/832 Support 16bpc PNGs in Filament's `gltf_viewer`
   - Fixed as of https://github.com/google/filament/pull/835
 - Shared relevant compatibility results with Filament project [here](https://github.com/google/filament/issues/528#issuecomment-465622950)

Fixes #360


donmccurdy:
Excited to see these tests set up, thanks! :)

**Texture transforms**

The description suggests Filament is correct on offset/scale and incorrect on rotation, but I think all three are wrong – looks like Filament doesn't implement that extension yet.

**Multiple UVs**

ThreeJS does support multiple UVs for the most common case – AO maps. But arbitrary textures can't be assigned to arbitrary UV maps, which is what this test shows, so that's a known issue.

**Alpha blending**

This is unexpected – I've commented on https://github.com/GoogleWebComponents/model-viewer/issues/381.

Overall – the staging looks similar on these, but the lighting does not. I can't tell from the PR if that's expected. In particular `<model-viewer/>` seems heavily front-lit, maybe? Are we matching lighting in `<model-viewer/>` or Filament, or is that a known caveat for now?

cdata:
> The description suggests Filament is correct on offset/scale and incorrect on rotation, but I think all three are wrong – looks like Filament doesn't implement that extension yet.

Good call, will fix the description.

> arbitrary textures can't be assigned to arbitrary UV maps, which is what this test shows, so that's a known issue.

Is there an issue we can reference from here?

> This is unexpected – I've commented on #381.

👍 

> Overall – the staging looks similar on these, but the lighting does not. I can't tell from the PR if that's expected.

Yeah, definitely. There was a bit of background discussion that I should have linked to that occurred during the PMREM change: https://github.com/GoogleWebComponents/model-viewer/pull/338

I'll include a brief summary in the PR description.

donmccurdy:
> > arbitrary textures can't be assigned to arbitrary UV maps, which is what this test shows, so that's a known issue.
> 
> Is there an issue we can reference from here?

See: https://github.com/mrdoob/three.js/issues/12608

scripts/model-viewer-screenshot.js
```diff
@@ -0,0 +1,56 @@
...
+  console.error(' Test slug not specified!');
+  process.exit(1);
+}
+
+if (outputFile == null) {
+  console.error(' Output file not specified!');
+  process.exit(1);
+}
+
+screenshotCreator
```
jsantell: For top level promise usage, I like to wrap an async iife `(async function(){ ... })()` to use async await.. which now I mention it, it's been awhile, I *think* node has native async awaits now.

scripts/install-third-party-renderers.sh
```diff
@@ -87,7 +87,7 @@ fi
 
 pushd $FILAMENT_DIR
 
-git fetch --depth=1 origin
+#git fetch --depth=1 origin
```
jsantell: Remove or add comment?

package.json
```diff
@@ -41,6 +42,7 @@
     "compare-fidelity": "./scripts/compare-fidelity-to-ref.sh",
     "install-renderers": "./scripts/install-third-party-renderers.sh",
     "update-screenshots": "node ./scripts/update-screenshots.js",
+    "checkout-samples": "./scripts/checkout-khronos-gltf-samples.sh",
```
jsantell: nit, maybe `get-samples` or `fetch-samples` is a bit more obvious at what this does

scripts/checkout-khronos-gltf-samples.sh
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+##
+
+# This script is responsible for cloning and / or updating the local repository
+# of Khronos glTF sample models
+
+REPO_URL=https://github.com/KhronosGroup/glTF-Sample-Models.git
+CLONE_PATH=./examples/assets/glTF-Sample-Models
+
+if [ ! -d "$CLONE_PATH/.git" ]; then
```
jsantell: Maybe a warning here, or in the README, that this will pull down a large amount of data

scripts/install-third-party-renderers.sh
cdata: Removed by accident (couldn't `git fetch` from the commuter ferry 🙃)

scripts/checkout-khronos-gltf-samples.sh
cdata: Good call, will fix 👍

package.json
cdata: Will fix 👍

scripts/model-viewer-screenshot.js
cdata: It does. Will fix 👍

README.md
```diff
@@ -275,6 +275,7 @@ The following npm scripts are available:
 * `npm test` - Runs tests.
 * `npm run check-fidelity` - Compare rendering to third-party renderers
 * `npm run install-renderers` - Install or update third-party renderers
+* `npm run fetch-samples` - Create or update a local copy of sample models. This is needed for running fidelity checks and updating screenshots. Warning: the samples are hundreds of megabytes large in aggregate!
```
smalls: If this is required, does it make sense to merge it in with (perhaps) `install-renderers` (another long, expensive prep step) into a single script named something like `prepare-fidelity`?

smalls: Or, if the thinking is that this is always run in `npm install`, maybe mention that (or leave it out entirely?)

cdata: I guess for factorization purposes it would be nice to keep these things separated. For usability purposes, we probably want a single command.

It's easy enough for us to make a combined command and just not document the sub-commands.

I'm also happy to leave it out entirely.

scripts/checkout-khronos-gltf-samples.sh
cdata: There is a warning in both places now.

scripts/install-third-party-renderers.sh
cdata: Fixed

scripts/model-viewer-screenshot.js
cdata: Fixed

README.md
smalls: Agreed re factorization, I think particularly the backend scripts (the `.sh` or `.js` files) having those split and encapsulated helps to understand the code.

I like both a combined commands and undocumented sub-commands (I've used a `-` prefix to connotate internal sub-commands) or leaving that sub-commands out entirely.

The thing I like most is having the combined command for usability.

smalls: Not a blocker to a merge, though.

cdata: Added a command called `npm run bootstrap-fidelity-dev` that checks out samples and builds filament in one go. Removed docs for the individual fetch / install tasks.

PR #377, opened 2019-02-19
Do not apply environment maps to unlit models. Fixes #374

### LINKED ISSUE #374: Bad render results for KHR_materials_unlit Khronos test model

jsantell:
Currently with a default env map, image env map, and PMREM'd env map currently:

localhost_5000_examples_fullscreen2 html 11

With this patch:

localhost_5000_examples_fullscreen2 html 10

Note that `UnlitTest.glb` is included (4kb) -- can reference the Khronos models if/when they land. The Astronaut is *not* unlit AFAICT (or, does not use the unlit extension), maybe should change the background-image demo describing it as such.

### DIFF
--- a/src/test/features/environment-spec.js
+++ b/src/test/features/environment-spec.js
@@ -22,6 +22,8 @@ const expect = chai.expect;
 const ALT_BG_IMAGE_URL = assetPath('quick_4k.png');
 const BG_IMAGE_URL = assetPath('spruit_sunrise_2k.jpg');
 const MODEL_URL = assetPath('reflective-sphere.gltf');
+const UNLIT_MODEL_URL = assetPath('glTF-Sample-Models/2.0/UnlitTest/glTF-Binary/UnlitTest.glb');
+const MULTI_MATERIAL_MODEL_URL = assetPath('Triangle.gltf');
 
 const backgroundHasMap =
     (scene, url) => {
@@ -48,16 +50,29 @@ const backgroundHasColor =
 const modelUsingEnvMap = (scene, meta) => {
   let found = false;
   scene.model.traverse(object => {
-    if (!object.material || !object.material.envMap) {
-      return;
-    }
-
-    if (textureMatchesMeta(object.material.envMap, meta)) {
-      found = true;
+    if (Array.isArray(object.material)) {
+      found = found || object.material.some(m => {
+        return textureMatchesMeta(m.envMap, meta);
+      });
+    } else if (object.material && object.material.envMap) {
+      found = found || textureMatchesMeta(object.material.envMap, meta);
     }
   });
   return found;
 };
+
+const modelHasEnvMap = (scene) => {
+  let found = false;
+  scene.model.traverse(object => {
+    if (Array.isArray(object.material)) {
+      found = found || object.material.some(m => m.envMap);
+    }
+    else if (object.material && object.material.envMap) {
+      found = true;
+    }
+  });
+  return found;
+}
 
 /**
  * Takes a model object and a meta object and returns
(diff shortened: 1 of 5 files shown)


src/three-components/Model.js
```diff
@@ -50,7 +50,10 @@ export default class Model extends Object3D {
 
   applyEnvironmentMap(map) {
     this.modelContainer.traverse(obj => {
-      if (obj && obj.isMesh && obj.material) {
+      // Unlit models (using MeshBasicMaterial) should not apply
+      // an environment map, even though `map` is the currently configured
+      // environment map.
+      if (obj && obj.isMesh && obj.material && !obj.material.isMeshBasicMaterial) {
```
donmccurdy: For some edge cases, `obj.material` could be an array of materials. We should probably avoid creating those in GLTFLoader at some point, but for now it's worth handling.

cdata:
@jsantell since we have merged #375 , would you be willing to update this change to use the unlit sample from the Khronos samples so that we don't have to check the model into our repo?

jsantell:
@cdata  
> since we have merged #375 , would you be willing to update this change to use the unlit sample from the Khronos samples so that we don't have to check the model into our repo?

wouldn't that require needing to fetch all the models before being able to run tests? UnlitTest.glb is 4kb as well.

@donmccurdy 
> For some edge cases, `obj.material` could be an array of materials. We should probably avoid creating those in GLTFLoader at some point, but for now it's worth handling.

Any models that you know that do this to test?

cdata:
>> since we have merged #375 , would you be willing to update this change to use the unlit sample from the Khronos samples so that we don't have to check the model into our repo?
> wouldn't that require needing to fetch all the models before being able to run tests? UnlitTest.glb is 4kb as well.

Yes, and we do this (models are fetched as part of `npm prepare`, so they happen after `npm install` and `npm ci`).

In practice this is probably fine. We have set up Travis to cache the directory they are checked out to, and it doesn't seem to take much time. We also update in place if we detect that the models are already fetched.

cdata:
FYI looks like `UnlitTest.glb` is still checked in

jsantell:
Using the Khronos sample models now, and handle when `material` is an array

src/three-components/Model.js
```diff
@@ -49,10 +49,24 @@ export default class Model extends Object3D {
...
-        obj.material.envMap = map;
-        obj.material.needsUpdate = true;
+      if (obj && obj.isMesh && Array.isArray(obj.material)) {
+        materialStack = materialStack.concat(obj.material);
+      }
+      else if (obj && obj.isMesh && obj.material) {
+        materialStack.push(obj.material);
+      }
+
+      for (let material of materialStack) {
```
cdata: Did you mean to do iteration of the material stack on every invocation of the `traverse` callback?

jsantell: Mm, yes, but missing clearing it out. I don't like any way that attempts to not duplicate the logic for the (untested) edge case as well. Will change

donmccurdy:
It's a bit contrived, but here's an example that will result in `mesh.material` being an array. The model contains a single mesh with two primitives, and the primitives are identical except for using different materials.

[Triangle.gltf.zip](https://github.com/GoogleWebComponents/model-viewer/files/2902385/Triangle.gltf.zip)

jsantell:
* Fixed the material array logic
* Added Triangle.gltf for test case (added PBR values to see envmap), shook out some issues in our GLTFLoader wrapper
* Remove UnlitTest.glb

src/test/features/environment-spec.js
```diff
@@ -160,6 +175,34 @@ suite('ModelViewerElementBase with EnvironmentMixin', () => {
...
+          });
+          element.src = UNLIT_MODEL_URL;
+          await onLoad;
+        });
+        test('applies no environment map on unlit model', async function() {
+          expect(modelHasEnvMap(scene)).to.be.false;
+        });
+      });
+
+      suite('on a model with multi-material meshes', () => {
```
cdata: 💯

PR #383, opened 2019-02-20
Add CXXFLAGS for Filament on Debian

smalls:
Without `CXXFLAGS=-stdlib=libc++`, Filament won't build on my Debian machine.

It doesn't impact compilation on MacOS, but I had to make other changes that I'll ask about below.

### DIFF
--- a/scripts/install-third-party-renderers.sh
+++ b/scripts/install-third-party-renderers.sh
@@ -96,6 +96,7 @@ cp $GLTF_RENDERER_CPP_PATH $FILAMENT_DIR/samples
 # Export critical environment variables for bulding Filament
 export CC=clang-7
 export CXX=clang++-7
+export CXXFLAGS="-stdlib=libc++"
 export FILAMENT_REQUIRES_CXXABI=true
 
 ./build.sh -j release
--- a/scripts/install-third-party-renderers.sh
+++ b/scripts/install-third-party-renderers.sh
@@ -93,7 +93,7 @@ git reset --hard origin/master
 git apply $FILAMENT_PATCH_PATH
 cp $GLTF_RENDERER_CPP_PATH $FILAMENT_DIR/samples
 
-# Export critical environment variables for bulding Filament
+# Export critical environment variables for building Filament
 export CC=clang-7
 export CXX=clang++-7
 export CXXFLAGS=-stdlib=libc++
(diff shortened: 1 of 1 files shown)


scripts/install-third-party-renderers.sh
```diff
@@ -93,9 +93,10 @@ git reset --hard origin/master
 git apply $FILAMENT_PATCH_PATH
 cp $GLTF_RENDERER_CPP_PATH $FILAMENT_DIR/samples
 
-# Export critical environment variables for bulding Filament
+# Export critical environment variables for building Filament
 export CC=clang-7
```
smalls: @cdata on my Mac, I had to commend out both the `export CC` and `export CXX` lines, as the default `clang` doesn't have a version. Were you working around it some other way? If not, I can add an OS check here.

cdata: Well on _my_ mac....

Seriously though, looks like I had not regenerated my cmake and it meant I was benefiting from an earlier set of environment variables.

Please add a platform check!

cdata: FWIW on Mac I need to remove the `FILAMENT_REQUIRES_CXXABI=true` envvar in addition to setting `CC=clang` and `CXX=clang++`.

smalls: 🤒 

I was able to build on Mac like this:

```
# Export critical environment variables for bulding Filament
#export CC=clang-7
#export CXX=clang++-7
export CXXFLAGS=-stdlib=libc++
export FILAMENT_REQUIRES_CXXABI=true
```

Maybe a difference in `cc` or `c++`? Mine are:

```
> echo $CC; which cc; cc --version

/usr/bin/cc
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
> echo $CXX; which c++; c++ --version

/usr/bin/c++
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

I'll add a check and try without `FILAMENT_REQUIRES_CXXABI=true`, to see if it makes a difference. If it improves compatibility I'm all for it.

cdata: Me ten years ago: don't care about no linker gonna be a web guy
Me today: 😱

PR #388, opened 2019-02-27
Add configurable shadow intensity, change default

### LINKED ISSUE #206: Shadows should be off by default and can be turned on with an attribute

cdata:
shadow-strength

This change proposes the addition of configurable shadow intensity, with a new default of `0.0`, effectively causing shadows to be hidden by default.

The image above demonstrates the shadow being changed on a timer between values `0.0` and `2.0`. The previous default intensity can be achieved by setting the value to `1.0`.

Fixes #206

### DIFF
--- a/src/features/environment.js
+++ b/src/features/environment.js
@@ -16,8 +16,10 @@
 import {BackSide, BoxBufferGeometry, Color, Mesh, ShaderLib, ShaderMaterial, UniformsUtils} from 'three';
 
 import {$needsRender, $onModelLoad, $renderer, $scene, $tick} from '../model-viewer-base.js';
+import {BASE_SHADOW_OPACITY} from '../three-components/StaticShadow.js'
 
 const DEFAULT_BACKGROUND_COLOR = '#ffffff';
+const DEFAULT_SHADOW_STRENGTH = 0.0;
 
 const WHITE = new Color('#ffffff');
 
@@ -30,6 +32,7 @@ const $hasBackgroundImage = Symbol('hasBackgroundImage');
 const $hasBackgroundColor = Symbol('hasBackgroundColor');
 const $deallocateTextures = Symbol('deallocateTextures');
 const $updateSceneLighting = Symbol('updateSceneLighting');
+const $updateShadow = Symbol('updateShadow');
 
 export const EnvironmentMixin = (ModelViewerElement) => {
   return class extends ModelViewerElement {
@@ -38,8 +41,14 @@ export const EnvironmentMixin = (ModelViewerElement) => {
         ...super.properties,
         backgroundImage: {type: String, attribute: 'background-image'},
         backgroundColor: {type: String, attribute: 'background-color'},
-        experimentalPmrem: {type: Boolean, attribute: 'experimental-pmrem'}
+        experimentalPmrem: {type: Boolean, attribute: 'experimental-pmrem'},
+        shadowStrength: {type: Number, attribute: 'shadow-strength'}
       };
+    }
+
+    constructor(...args) {
+      super(...args);
+      this.shadowStrength = DEFAULT_SHADOW_STRENGTH;
     }
 
     get[$hasBackgroundImage]() {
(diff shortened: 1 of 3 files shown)


src/features/environment.js
```diff
@@ -167,6 +176,16 @@ export const EnvironmentMixin = (ModelViewerElement) => {
       this[$needsRender]();
     }
 
+    [$updateShadow]() {
+      const {shadowStrength} = this;
+      const shadowStrengthIsNumber =
```
jsantell: Doesn't lit-element handle some of this?

src/features/environment.js
```diff
@@ -38,10 +41,16 @@ export const EnvironmentMixin = (ModelViewerElement) => {
         ...super.properties,
         backgroundImage: {type: String, attribute: 'background-image'},
         backgroundColor: {type: String, attribute: 'background-color'},
-        experimentalPmrem: {type: Boolean, attribute: 'experimental-pmrem'}
+        experimentalPmrem: {type: Boolean, attribute: 'experimental-pmrem'},
+        shadowStrength: {type: Number, attribute: 'shadow-strength'}
```
jsantell: I think `shadow-intensity` may be more in line with how shadow "strength" is normally defined

src/features/environment.js
```diff
@@ -16,8 +16,10 @@
 import {BackSide, BoxBufferGeometry, Color, Mesh, ShaderLib, ShaderMaterial, UniformsUtils} from 'three';
 
 import {$needsRender, $onModelLoad, $renderer, $scene, $tick} from '../model-viewer-base.js';
+import {BASE_SHADOW_OPACITY} from '../three-components/StaticShadow.js'
```
jsantell: I may be biased due to my history with `features/environment`, but could the shadow feature live in a different feature? This is the first attribute that's outside of the skybox -- while it handles IBL, I wonder if a `lights` feature would be appropriate for the shadow setting, as well as other future attributes? Just spitballing (more than OK for now)

src/features/environment.js
```diff
@@ -167,6 +176,16 @@ export const EnvironmentMixin = (ModelViewerElement) => {
       this[$needsRender]();
     }
 
+    [$updateShadow]() {
+      const {shadowStrength} = this;
+      const shadowStrengthIsNumber =
+          typeof shadowStrength === 'number' && !Number.isNaN(shadowStrength);
+
+      this[$scene].shadow.material.opacity = BASE_SHADOW_OPACITY *
+          (shadowStrengthIsNumber ? shadowStrength : DEFAULT_SHADOW_STRENGTH);
```
jsantell: I think I prefer moving some of the internal logic into the component, e.g. `this[$scene].shadow.setIntensity(...)` and let the component handle the base shadow opacity, similar to `model.applyEnvironmentMap(...)` in this module, and leave the feature focused on plumbing the different configurations/components, and let the components handle the details. This idea sort of fell a part a bit with the removal of the Skysphere, since `scene.background` is being manipulated directly here, and we don't have an abstraction/three-component for it otherwise.

jsantell: I don't think it'll matter much, but wonder if there's anything we need to do with the default case (e.g. rendering an opacity=0 mesh) -- I think it'll be skipped in the renderer somewhere, and the only potential overhead is the initial creation of the shadow. Even if we're rendering an opacity=0 shadow, it's a fake shadow anyway, so a negligible cost (outside of possibly the LUT on mobile??)

src/features/environment.js
cdata: Lit won't stop the value from being `NaN` when a non-number value is set in the attribute. And, any value may be assigned to the property, Lit does not do any filtering or validation there.

src/features/environment.js
cdata: Will fix

src/features/environment.js
cdata: I am planning to add some "intensity" configurations that control lighting. Since shadow is the absence of light, intensity seemed like a weird word to use.

src/features/environment.js
cdata: Yeah, will move per suggestion below

cdata: Also re: lights, the thought occurred to me as well. The main point of uncertainty for me is that technically the environment is also lighting (in graphics contexts, at least).

src/features/environment.js
cdata: Changed to intensity, as @johnpallett points out we may want a complimentary value in the future called "softness" which seems weird to pair with "strength." Then again, must one always be either soft or strong, or is this just a bias instilled in me by social prejudice?

PR #391, opened 2019-02-28
Add configurable stage light intensity

cdata:
This change proposes adding a `stage-light-intensity` attribute/property for configuring the intensity of Three.js ambient/directional lights in the scene.

 - Intensity in this context is a scalar that is applied to the base intensity of the ambient and directional lights separately
   - Rationale here is that the base intensity for each of these lights is different, and also based on the current role (see below)
 - Stage lights now also have a "primary" and "secondary" role, for non-PMREM and PMREM rendering respectively
   - The rationale here is that when PMREM is activated, the stage lights should be de-emphasized (they become "secondary" lighting)
   - This is an implementation detail, not directly configurable by the `<model-element>` user
 - `0.0` intensity turns off the stage lights; `1.0` is the current defaults as we know them; there is no upper bound

#### <div align="center">`stage-light-intensity` oscillating from 0.0 to 6.0</div>

| PMREM on                     | PMREM off                     |
| ----------------------- | ----------------------- |
lights-osc-pmrem 2019-02-27 18_26_45|lights-osc-no-pmrem 2019-02-27 18_31_44|

`stage-light-intensity="1.0"` Toggling PMREM on and off |
--------------------------------------------------------- |
lights-pmrem-no-pmrem|

### DIFF
--- a/src/test/three-components/TextureUtils-spec.js
+++ b/src/test/three-components/TextureUtils-spec.js
@@ -30,7 +30,9 @@ suite('TextureUtils', () => {
   let textureUtils;
   setup(() => {
     const renderer = new WebGLRenderer({canvas});
-    textureUtils = new TextureUtils(renderer);
+    // NOTE(cdata): We need to lower the samples here or else tests that use
+    // PMREM have a tendency to time out on iOS Simulator
+    textureUtils = new TextureUtils(renderer, {pmremSamples: 4});
   });
   teardown(() => textureUtils.dispose());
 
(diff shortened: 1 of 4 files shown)


src/test/three-components/TextureUtils-spec.js
```diff
@@ -30,7 +30,9 @@ suite('TextureUtils', () => {
   let textureUtils;
   setup(() => {
     const renderer = new WebGLRenderer({canvas});
-    textureUtils = new TextureUtils(renderer);
+    // NOTE(cdata): We need to lower the samples here or else tests that use
+    // PMREM have a tendency to time out on iOS Simulator
+    textureUtils = new TextureUtils(renderer, {pmremSamples: 4});
```
jsantell: unfortunate, but no benefit of higher samples here anyway

cdata:
> This is a whole can of worms of course that can be kicked down the road a bit. Everything else looks solid

Filed https://github.com/GoogleWebComponents/model-viewer/issues/396

PR #398, opened 2019-03-03
Make IntersectionObserver optional

### LINKED ISSUE #386: Make IntersectionObserver polyfill optional
Rationale: this will limit the number of polyfills required for basic, broad x-browser support to 1 (the Web Components polyfill).

Background:

Currently, support for `IntersectionObserver` via native implementation or polyfill is required for `<model-viewer>` to work. However, `IntersectionObserver` is used as part of a performance optimization strategy for rendering. So, we should make the code that uses it fall back to a naive path when `IntersectionObserver` is unavailable (similar to what we do with `ResizeObserver`), so that folks just getting started with `<model-viewer>` can benefit from a smaller list of things they have to do to see it working as advertised.

cdata:
This change proposes that `IntersectionObserver` should be an optional, performance-enhancing feature rather than a required feature to make `<model-viewer>` work (as it is now). Documentation updates related to this change are in progress, and will be included in a forthcoming PR.

Fixes #386

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -13,6 +13,8 @@ export const HAS_FULLSCREEN_API = document.documentElement != null &&
 
 export const HAS_RESIZE_OBSERVER = self.ResizeObserver != null;
 
+export const HAS_INTERSECTION_OBSERVER = self.IntersectionObserver != null;
+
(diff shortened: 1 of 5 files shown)


jsantell:
I don't see how the element would ever render without it -- the Renderer requires `scene.isVisible = true` to be set, which is only manipulated by the observer

cdata:
Good call, will fix that

PR #401, opened 2019-03-04
Use and recommend module/nomodule snippet

### LINKED ISSUE #399: Provide ESM bundle and improve compatibility recommendations
Currently, we ship 2 distinct bundles for easy consumption by users without a build pipeline:

 1. A "modern" bundle that includes all dependencies wrapped as a UMD module.
 2. A "legacy" bundle that includes all dependencies, transpiled to IE11-compatible ES5 and wrapped as a UMD module.

However, we don't offer a recommended path for users to best take advantage of both bundles and maximize browser compatibility. In fact, this was an oversight when we originally configured these bundles. There is no trivial way to use our distributed bundles (as currently configured) to correctly deliver `<model-viewer>` to all supported browsers.

To fix this, we should change our bundle configurations and suggest a code snippet leveraging the `nomodule` attribute. Background, recommendations and caveats related to this topic can be found in [this article by Jake Archibald](https://jakearchibald.com/2017/es-modules-in-browsers/).

cdata:
This change proposes a new format for including `<model-viewer>`-related scripts on a page, based on the `nomodule` strategy recommended in [Jake Archibald's article](https://jakearchibald.com/2017/es-modules-in-browsers/) on the topic.

 - The new format gives users a standard, tested way to achieve maximum compatibility without a specialty build pipeline or server-side application logic
 - All relevant examples have been updated to use the new format
 - Test suites have been consolidated so that we no longer need an alternative entrypoint to test legacy browsers
 - The standard bundle (`model-viewer.js`) has been changed to export an ECMAScript Module
 - A new `model-viewer-umd.js` bundle has been added, primarily to serve as input for generating so-called legacy bundles
 - The same treatment as above has been given to documentation and unit test script bundles

✨**Bonus**: I tested our documentation in Chrome 41 with this change, and it looks like the site should now fully render for the Googlebot.

The new format looks like this:

```html
<!-- 💁 Include both scripts below to support all browsers! -->

<!-- Loads <model-viewer> for modern browsers: -->
<script type="module"
    src="https://unpkg.com/@google/model-viewer/dist/model-viewer.js">
</script>

<!-- Loads <model-viewer> for old browsers like IE11: -->
<script nomodule
    src="https://unpkg.com/@google/model-viewer/dist/model-viewer-legacy.js">
</script>
```

One important caveat here: in the case of IE11 and older Safari, **both scripts will download** though only the `nomodule` script will execute. This is an unfortunate tradeoff of network cost for simplicity. In our case, it is an especially hard cost to bare as the script in question is rather large.

The rationale for accepting this tradeoff is that IE11 and old Safari users are, as a group, more likely to be on desktop and therefor less network constrained.

Fixes #399


rollup.config.js
```diff
@@ -68,6 +82,19 @@ export default [
     input: './lib/test/index.js',
     output: {
       file: './dist/unit-tests.js',
+      format: 'esm',
```
jsantell: On master, the build times are already pretty significant:

Cold:

```
real	0m19.718s
user	0m28.090s
sys	0m1.101s
```

Hot:

```
real	0m16.310s
user	0m24.035s
sys	0m0.846s
```

I suspect some breaking up the build in the future and revisiting https://github.com/GoogleWebComponents/model-viewer/pull/235 -- although we left off on the dependency relationship of tests

examples/augmented-reality.html
```diff
@@ -110,7 +110,23 @@ <h4>unstable-webxr</h4>
     <div class="copyright">©Copyright 2018 Google Inc. Licensed under the Apache License 2.0.</div>
   </div>
 
-  <script src="./built/dependencies.js"></script>
-  <script src="../dist/model-viewer.js"></script>
+  <!-- Documentation-specific dependencies: -->
```
jsantell: hyper nit: I'd prefer the script tags on one line, and removing new line between; makes it look like one computational unit; WDYT?

```html
   <!-- Loads <model-viewer> only on modern browsers: -->
  <script type="module" src="../dist/model-viewer.js"></script>
   <!-- Loads <model-viewer> only on old browsers like IE11: -->
  <script nomodule src="../dist/model-viewer-legacy.js"></script>
```

src/three-components/Renderer.js
```diff
@@ -92,12 +92,20 @@ export default class Renderer extends EventDispatcher {
   }
 
   set exposure(exposure) {
+    if (!this.canRender) {
```
jsantell: Is this related to this PR?

examples/augmented-reality.html
jsantell: I suspect this will fit on one line per 80 char limit on examples outside of the full CDN URL

rollup.config.js
cdata: Definitely welcome work to make this better in the future.

examples/augmented-reality.html
cdata: I personally find this easier to gloss over while scanning the code, and my intention was for the snippet to be eye-catching to a person viewing source.

That said, this is a super-subjective assessment and I'm open to the change depending on what others think is best.

src/three-components/Renderer.js
cdata: Yes, in Chrome 41 without WebGL this code path was being hit. We weren't seeing it before because without this snippet Chrome 41 wasn't even getting this far in our docs.

jsantell: Heh, good catch!

cdata:
> Is that the same issue as this: https://gist.github.com/samthor/64b114e4a4f539915a95b91ffd340acc? 

These issues are tangentially related. @samthor 's gist is a polyfill it make `nomodule` do what you expect in Safari < 11. Thankfully, we no longer support Safari < 11 (due to last 2 major versions policy).

> since the UMD build isn't publicized, is there any scenario where someone would want the UMD build without the legacy overhead

This is an open question that I haven't explored deeply. The main consideration I had in mind was existing users (if there even are any) who have become attached in some way to the current bundle format (which, prior to this change, is UMD).

jsantell:
> > Is that the same issue as this: https://gist.github.com/samthor/64b114e4a4f539915a95b91ffd340acc?
> 
> These issues are tangentially related. @samthor 's gist is a polyfill that makes `nomodule` do what you expect in Safari < 11. Thankfully, we no longer support Safari < 11 (due to last 2 major versions policy).
> 
> > since the UMD build isn't publicized, is there any scenario where someone would want the UMD build without the legacy overhead
> 
> This is an open question that I haven't explored deeply. The main consideration I had in mind was existing users (if there even are any) who have become attached in some way to the current bundle format (which, prior to this change, is UMD).

That all SGTM, ok with publishing it for now, and referencing if/when needed

smalls:
> ✨Bonus: I tested our documentation in Chrome 41 with this change, and it looks like the site should now fully render for the Googlebot.

That's a significant bonus!!

README.md
```diff
@@ -30,7 +30,15 @@ You can load a _bundled build_ via
 [unpkg.com](https://unpkg.com/@google/model-viewer/dist/model-viewer.js):
```
smalls:
```suggestion
unpkg.com by including the snippet below. This will automatically load the correct version for the user's browser.
```

smalls: My thought was this'd make it a little clearer for folks that they should just include the whole snippet, and the browser will choose the proper version?

README.md
```diff
@@ -51,11 +59,8 @@ dependencies.
 
 ## Usage
 
-If you are using a _bundled build_, first add a script tag to your page to load it
-
-```html
-<script src="path/to/bundled/model-viewer.js"></script>
-```
+If you are using a _bundled build_, first add a script tag to your page to load
+`<model-viewer>` as described in the Installing section.
```
smalls:
```suggestion
`<model-viewer>` as described in the [Installing](#installing) section.
```

smalls: I think that'll work.

examples/augmented-reality.html
smalls: Another option might be as follows. It might make it clearer that they're a logical block?

```
   <!-- Loads <model-viewer> (the first line on modern browsers, the second for legacy) -->
  <script type="module" src="../dist/model-viewer.js"></script>
  <script nomodule src="../dist/model-viewer-legacy.js"></script>
```

README.md
cdata: Agree, but it was a little confusing what `This` is referring to in `This will automatically load...`. Maybe use `The snippet will automatically load...`?

smalls: Yeah, that sounds good to me.

examples/augmented-reality.html
cdata: Thanks for the discussion everyone. After spending a lot of time thinking it over, I can see benefits and drawbacks to all proposed approaches and I am still not sure what the best answer is. I would like to suggest that we defer this decision and continue discussion in the context of https://github.com/GoogleWebComponents/model-viewer/issues/408

smalls: Nice dodge. Sounds goodl.

PR #410, opened 2019-03-06
Hotfix for PMREM environment map regression

cdata:
We experienced an untracked regression after the environment map refactor in #392 . The regression manifested as artifacts shown when using an environment map that we are currently excluding from our fidelity tests while we investigate PMREM generation quirks. Here is the rendering before and after this hotfix:

Without hotfix | With hotfix
---------------|-----------
image|image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google/model-viewer",
-  "version": "0.0.7",
+  "version": "0.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@google/model-viewer",
-  "version": "0.0.7",
+  "version": "0.1.0",
```
jsantell: Wonder why 5ad748cfd818ecfc4076809509fe5286b72b5eb7 didn't pick this up? Something to scope out for next release

Issue #434, opened 2019-03-15
Update gltf_renderer to use Filament's new gltfio library


cdata:
@prideout tells us that their `gltf_viewer` sample that `gltf_renderer` was based on has been updated to use the new library.

smalls:
Looks like the change is here: https://github.com/google/filament/commit/7bc6b3c818204d606192efa4c833ce520f5f4d07

Giorat:
What does this mean? It means that model-viewer will use Filament wasm to show models?

cdata:
No, we do comparisons to Filament as part of our fidelity test suite: https://googlewebcomponents.github.io/model-viewer/test/fidelity/results-viewer.html

elalish:
Also, filament's TangentNormal test from Khronos is broken, and the fix is in gltfio.

PR #436, opened 2019-03-18
update to latest filament

smalls:
Filament's `samples/CMakeLists.txt` changed slightly


cdata:
Do you think this change is related to #434 ? Is this tweak to the patch sufficient to build a working `gltf_renderer`?

smalls:
No, unfortunately. This just corrects a patch that failed to apply with the current assimp-based `gltf_renderer`.

cdata:
LGTM 👍

smalls:
To clarify - this still uses the old glTF library, but that's still supported (we can build & render with Filament).

PR #448, opened 2019-03-27
Begin work to support more renderers

smalls:
- Refactor renderer installation into sub-scripts
- Refactor screenshot utility to allow for easier extensions
- Update to new Filament

### DIFF
--- a/scripts/update-screenshots.js
+++ b/scripts/update-screenshots.js
@@ -67,16 +67,6 @@ const updateScreenshots = async (config) => {
   for (const scenario of scenarios) {
     const {goldens, slug} = scenario;
     const scenarioDirectory = path.join(fidelityTestDirectory, slug);
-    const testHtmlPath = path.join(scenarioDirectory, 'index.html');
-
-    const html = (await fs.readFile(testHtmlPath)).toString();
-
-    const backgroundImageMatch = html.match(backgroundImageRe);
-    const backgroundImage =
-        backgroundImageMatch != null ? backgroundImageMatch[1] : null;
-
-    const modelSourceMatch = html.match(modelSourceRe);
-    const modelSource = modelSourceMatch != null ? modelSourceMatch[1] : null;
 
     for (const golden of goldens) {
       const {name, file} = golden;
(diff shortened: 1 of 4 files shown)


scripts/update-screenshots.js
```diff
@@ -106,69 +96,85 @@ const updateScreenshots = async (config) => {
...
+      path.resolve(path.dirname(testHtmlPath), backgroundImage);
+
+  const modelSourcePath =
+      path.resolve(path.dirname(testHtmlPath), modelSource);
+
+  await new Promise((resolve, reject) => {
+    console.log(`🖌️  Rendering ${name} screenshot for ${scenario.slug}...`);
+
+    const childProcess = spawn(
+        filamentScreenshotScript,
```
cdata: Does this work? Shouldn't this be `script`?

scripts/update-screenshots.js
```diff
@@ -106,69 +96,85 @@ const updateScreenshots = async (config) => {
...
   }
 };
 
+const screenshotFromScript = async (scenario, scenarioDirectory, filePath, name, script) => {
+  const testHtmlPath = path.join(scenarioDirectory, 'index.html');
+
+  const html = (await fs.readFile(testHtmlPath)).toString();
+
+  const backgroundImageMatch = html.match(backgroundImageRe);
+  const backgroundImage =
```
cdata: The original implementation extracts these test configuration values in the least good way that actually works (RegExp parsing attributes from HTML strings).

It might make more sense to parse the HTML strings once outside of this function and pass the extracted values in here as arguments.

scripts/update-screenshots.js
smalls: OMG. Thanks.

scripts/update-screenshots.js
smalls: I'm going to give this a pass (in a separate CL). I agree, either this needs to be cleaned up, but I wonder if we should switch to a JSON storage format to clean up adding both new renderers and new models.

cdata: Yes, and yes.

cdata: Sorry, but also, yes, but maybe later to save time. Your call :)

Issue #450, opened 2019-03-29
Establish concrete definition of 1 meter as it relates to user-configurable API


cdata:
As we begin to land changes for features such as #98 , users will begin to express configuration with metric units that should correspond to reliable positions and distances in world space, regardless of how models are framed.

Currently in 3D inline rendering, we "frame" models by establishing a virtual "room" box and scaling the room and the model with the following steps (approximately):

 1. Determine the `aspectRatio` of the `<model-viewer>` element's bounding rect
 2. Set the height of `room` to `10`, and the width of `room` to `10 * aspectRatio` in world units
 3. Apply a scale to `model` so that its extents all fit inside of `room`
 4. If the depth of `room` is greater than the depth of `model`, shorten the depth of `room` by the difference

It is important for users to be able to trust that as a model scales based on responsive layout conditions, the camera and other configurable elements that rely on world space positioning update in a predictable, desirable way.

This issue covers the work needed to review what it means to specify positions and distances in metric units, and how those values are affected as the layout of `<model-viewer>` changes over time (thus causing the scaling of the model itself to change over time).

elalish:
The most obvious solution to this that I see, and for #197 as well, is to remove all concept of a "room", leaving model at model scale and adjusting the camera as necessary in model-based units (including near and far plane). However, that would be a significant change, so I'd like to know if I'm missing anything before I start designing that.

What was the original goal of the room? Should we try to keep the default framing identical to how it is now, or is there anything in the current version that someone isn't happy with? I'd be temped bound the model with a cylinder rather than a box, given our az-el camera.

cdata:
I would say that the floor is open for a good design proposal that addresses scaling in a way that works coherently across inline and unbounded (e.g., AR placement) use cases. Naive approaches have their limitations here.

The "room" dates back to the `<xr-model>` prototype and was meant to serve as a framing helper. It is useful to conceptualize a room when deciding how to adjust the camera frustum for some of the immersive use cases we were exploring, but I would not characterize it as a critical element of framing for today's `<model-viewer>`.

Issue #447, opened 2019-03-26
Support for panning the camera via keyboard, mouse and touch input


cdata:
This is a feature of `OrbitControls` that we left out of `SmoothControls` when we implemented it in #288 .

We might implement panning any number of ways, but we should strive to do it such that it helps the user maintain focus on the object in frame.

Here is an example of one possible way to "curve" the camera as it pans out farther and farther so that it never looks fully away from the object in focus:

<div align="center">
  <img width="75%" src="https://user-images.githubusercontent.com/240083/54964354-5df37300-4f29-11e9-8674-328ca0a5c577.png">
  <img width="75%" src="https://user-images.githubusercontent.com/240083/54964355-5df37300-4f29-11e9-924e-3427a42913b7.png">
  <img width="75%" src="https://user-images.githubusercontent.com/240083/54964356-5df37300-4f29-11e9-92fb-5ab7fb1d6fce.png">
</div>

elalish:
This is a major can of worms in terms of 3D UX, and probably one of the biggest things that makes CAD packages hard to navigate. I would advise against adding this to the default controls unless we have a clear idea of what we're trying to solve with it. #453 would give devs the ability to do it themselves if they need to. Do we have a known pain point we'd like to solve with panning?

cdata:
It's a good thing our job is to make hard things look easy 😅

At the surface of things there are a few facts to consider:

 - Three.js' `OrbitControls` supports panning, so we lost that behavior when we switched to a different control implementation
 - Some users have asked us for panning support (in other channels, unfortunately)
 - It is an option to enable this behavior as an opt-in via user configuration

You are right that #453 is highly relevant here. One could make the case that with #453 in place, panning is a matter of coordinating a change between `camera-orbit` and `camera-target` (or whatever we call it) over time.

The argument for this feature is that we should make it easy for users by implementing a reasonable, built-in strategy for coordinating this change over time, which includes support for multiple input modes (keyboard, mouse, touch).

elalish:
Okay, and here is my argument against: 

1) Why do you need to pan? Generally because you're highly zoomed in on a non-spherical model. Is this a use case we expect? Currently zoom is pretty limited by default.

2) Panning means moving an invisible target point in 3D space using 2D input. This is what makes panning so counter-intuitive: after a pan, the effect of subsequent viewer rotations are surprising. 

[...]

PR #452, opened 2019-03-29
Implement camera-orbit attribute

### LINKED ISSUE #98: Camera configuration: starting position (along the sphere of orbit) and radius of camera when model loads
When we introduced "smooth" controls to `<model-viewer>` in #288 , it became possible to easily configure the controls with constraints. However, the starting position is automatically determined based on the constraints configuration.

We should add a user-configurable attribute to enable configuration of the starting position of the camera, which includes azimuthal and polar angles, as well as orbit radius (effectively controlling "zoom").

### LINKED ISSUE #435: Control interaction prompt appears too quickly in most cases
We have a visual and audible (to screen readers) interaction prompt that appears after a fixed time (currently 3 seconds) of idleness starting when `<model-viewer>` is focused. In many cases, `<model-viewer>` will be focused as soon as it appears in the document. In this case, we begin measuring idleness even though the model has not finished loading. The result of this is that the visual prompt might appear before or near to when the model has appeared to the user.

We should fix this by only measuring idleness once the model has finished loading.

### LINKED ISSUE #437: Configuration to control time before displaying interaction prompt
In #322 we introduced a visual interaction prompt. In that change, the prompt appears _after_ the `<model-viewer>` is focused without interaction for three seconds.

We should add a user-configurable attribute so that the amount of time that passes before the prompt is displayed can be easily controlled.

### LINKED ISSUE #446: Controls SVG discoverable by VoiceOver on iOS
### Description

When using swipe gestures with VoiceOver & Safari on iOS, the controls SVG is discoverable and announced as, _"image"_. This announcement may lead to a confusing user experience as there's no further information on why the image exists. Further, it creates an unexpected and unnecessary "tab-stop."

Adding `focusable="false"` and `aria-hidden="true"` directly onto the `svg` element removes this extra tab-stop from occurring.

```html
<!-- controls-svg.js -->
<svg … focusable="false" aria-hidden="true">
  <!-- … -->
</svg>
```

IMG_0003

#### Live Demo
1. Go to [`<model-viewer>`](https://googlewebcomponents.github.io/model-viewer/index.html) with Safari for iOS
2. Load up [VoiceOver screen reader](https://support.apple.com/en-ca/guide/iphone/turn-on-and-practice-voiceover-iph3e2e415f/ios)
3. Swipe right with one finger until the _"image"_ content is found

### Browser Affected

- [ ] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [x] Safari

### OS

- [ ] Android
- [x] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### Versions

- model-viewer: v0.1.1

### LINKED ISSUE #456: Visual interaction prompt incurs compositing cost on every frame
The animated SVG that we use as a visual interaction prompt includes SVG SMIL animations. These animations have the effect of causing paint/composite on every frame, even when the container's opacity is 0.

cdata:
## Configurable `camera-orbit` attribute

This change proposes the addition of a `camera-orbit` attribute to enable users to control the starting and subsequent positions of the camera long a spherical orbit centered on the model.

<div align="center">

</div>

---

### Highlights

 - `camera-orbit` attribute allows CSS-like configuration of starting and subsequent camera positions
    - Accepts values in the form of `$theta $phi $radius`
    - Example accepted values include: `0deg 75deg 1m`, `1 2 3`, `1rad 2deg 100cm`
    - Default units are radians for angles, meters for radius
    - Special radius value `auto`: refers to the "ideal" camera distance as determined by scene framing
    - Camera interpolates from its current position to subsequent values of `camera-orbit` (see also #454)
 - `getCameraOrbit` method for real-time querying of current orbital position
    - Realtime position may not align to attribute on account of interpolation or user interaction
 - `interaction-prompt-threshold` attribute controls the time that must have passed before an interaction prompt is presented to the user
 - **🚨 BREAKING CHANGE:** The `controls` attribute has been renamed to `camera-controls`
    - Rationale: disambiguates from `controls` attribute in the `<video>` element
 - New modules for parsing and converting complex attribute values have been introduced
 - Lots of new tests for all of above have been added
 - Controls feature has been converted to TypeScript
 - A new documentation section for "Staging & Camera Control" has been added
 - A new model has been contributed to help document the `align-model` attribute introduced in #441

<div align="center">
  
</div>

### Name rationale

I waffled a bit when deciding what name to propose for the attribute. Initially, I thought that `camera-position` was a good name. But, `position` is already used by Three.js (and other libraries) to refer to the cartesian coordinates of things in 3D space. This attribute, by contrast, represents the spherical coordinates of the camera relative to a separately defined target origin.

As far as API shape is concerned, the spherical position can be queried by the user, and is represented by a plain-object version of Three.js's `Spherical` data structure e.g., `{ theta, phi, radius }`. However, it might make sense to name this object's properties more explicitly, e.g., `{ azimuth, inclination, radius }`. Feedback here is requested.

### Future work

 - Camera panning support #447 
 - Establish concrete definition of 1m in world space #450 
 - Attribute to configure the camera target #453
 - Opt-out of interpolation when changing camera orbit #454

Fixes #98 

Additional fixes have been rolled in from #457 :

Fixes #435 
Fixes #437 
Fixes #446 
Fixes #456


src/conversions.ts
```diff
@@ -0,0 +1,116 @@
...
+// TODO(#450): We are using FRAMED_HEIGHT as a proxy for 1 unit distance in
+// world space. In other words, 1 * FRAMED_HEIGHT = 1m. It's possible that this
+// is not ideal.
+import {FRAMED_HEIGHT} from './three-components/ModelScene.js';
+
+
+/**
+ * Convert a number expressing degrees to a number expressing radians.
+ */
+const degreesToRadians = (() => {
```
mrdoob: https://github.com/mrdoob/three.js/blob/master/src/math/Math.js#L8

cdata: 👍 good catch, will fix!

src/test/conversions-spec.ts
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+import {Math as ThreeMath} from 'three';
+import {deserializeSpherical} from '../conversions.js';
+
+const expect = chai.expect;
+
+suite('conversions', () => {
+  suite('deserializeSpherical', () => {
```
smalls: Nice! So glad to see the unit tests :)

src/parsers.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const parseAtomicValue = (valueString: string): ValueNode => {
+  const value = extractValue(valueString);
+  const unit = (value != null ? valueString.replace(value, '') : null) || null;
+
+  return {type: 'value', value, unit};
+};
+
+const extractValue = (() => {
+  const VALUE_RE = /^(([^-^.^0-9].*)|([-]?[0-9.]+))/;
```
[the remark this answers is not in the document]
smalls: This makes a bit more logical sense to me (ha, I know, regexes), although it still only extracts the values. I think it's a bit stricter (although still not strict) - specifically, "#a3#fd" now only matches "#a3".

`^((?:(?:#|[a-zA-Z])[a-zA-Z\d]*)|(?:-?[\d.]+))`

Oh! This will get the values too (in a second capture group) if my comment above makes sense, and then the whole string can be matched (taking advantage of the strictness).

`^((?:(?:#|[a-zA-Z])[a-zA-Z\d]*)|(?:-?[\d.]+))([a-zA-Z]*)$`

src/parsers.ts
```diff
@@ -0,0 +1,55 @@
+export interface ValueNode {
```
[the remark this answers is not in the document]
cdata: Will fix

src/conversions.ts
```diff
@@ -0,0 +1,104 @@
...
+ * only ValueNodes that represent a metric value (m, cm, mm) are supported.
+ *
+ * If no unit is specified, assumes meters. Returns 0 for a ValueNode that
+ * cannot be parsed.
+ */
+const lengthValueNodeToMeters = (lengthValueNode: ValueNode): number => {
+  const value = parseFloat(lengthValueNode.value as any);
+
+  if ((self as any).isNaN(value)) {
+    return 0;
```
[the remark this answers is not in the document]
cdata: The rationale here is that 0 is a reasonable default for any given meters length that cannot be parsed correctly. This is an homage to the CSS roots of this configuration value style. CSS doesn't throw or even warn you for invalid values. Instead, it substitutes a reasonable default.

src/parsers.ts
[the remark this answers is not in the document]
cdata: Doing everything in one JavaScript `RegExp` match is way faster than approaches that require an extra `String.prototype.replace` (in Chrome, anyway):

image

Good call on this. Will fix.

src/test/conversions-spec.ts
cdata: 🍻

cdata:
FYI @svinkle this change will pull in the fix for screen reader SVG discoverability.

src/features/controls.ts
```diff
@@ -0,0 +1,372 @@
...
+const HALF_PI = Math.PI / 2.0;
+const THIRD_PI = Math.PI / 3.0;
+const QUARTER_PI = HALF_PI / 2.0;
+const PHI = 2.0 * Math.PI;
+
+const AZIMUTHAL_QUADRANT_LABELS = ['front', 'right', 'back', 'left'];
+const POLAR_TRIENT_LABELS = ['upper-', '', 'lower-'];
+
+const ORBIT_CAMERA_NEAR_PLANE = FRAMED_HEIGHT / 10.0;
+const ORBIT_CAMERA_FAR_PLANE = FRAMED_HEIGHT * 10.0;
```
cdata: FYI @mrdoob I renamed these :)

PR #457, opened 2019-03-31
Fix various bugs related to interaction prompt

### LINKED ISSUE #435: Control interaction prompt appears too quickly in most cases

### LINKED ISSUE #437: Configuration to control time before displaying interaction prompt

### LINKED ISSUE #446: Controls SVG discoverable by VoiceOver on iOS

### LINKED ISSUE #456: Visual interaction prompt incurs compositing cost on every frame

cdata:
Fixes #435 
Fixes #437 
Fixes #446 
Fixes #456

### DIFF
--- a/index.html
+++ b/index.html
@@ -231,6 +231,15 @@
             for in-depth details related to this lighting mode.</p>
           </li>
           <li>
+            <h4>interaction-prompt-threshold</h4>
+            <p>When camera-controls are enabled, &lt;model-viewer&gt; will
+            prompt the user visually (and audibly, for screen readers) to
+            interact if they focus it but don't interact with it for some time.
+            This attribute allows you to set how long &lt;model-viewer&gt;
+            should wait (in milliseconds) before prompting to interact. Defaults
+            to 3000.</p>
+          </li>
+          <li>
             <h4>ios-src</h4>
             <p>The url to a <a href="https://graphics.pixar.com/usd/docs/Usdz-File-Format-Specification.html">USDZ</a> model which will be used on <a href="https://www.apple.com/ios/augmented-reality/">supported iOS 12+ devices</a> via <a href="https://developer.apple.com/videos/play/wwdc2018/603/">AR Quick Look</a> on Safari. See <a href="https://github.com/GoogleWebComponents/model-viewer#augmented-reality">Augmented Reality</a>.</p>
           </li>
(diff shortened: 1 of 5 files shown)


cdata:
cc @svinkle

cdata:
I'm going to roll this PR into #452 and ship all these as a single change.

PR #471, opened 2019-04-04
Auto-rotate doesn't rotate the model pivot if no model is loaded

### LINKED ISSUE #467: Don't start rotating the model before the poster is dismissed
### Description

If you have `auto-rotate` on with a poster, the model will start rotating before the poster is dismissed. The better behaviour would be for the model to only start rotating once it is visible. That way the initial pose of the model will be consistent.

This is especially important if you want the poster to be an image of the initial pose, so that it transitions smoothly from image to model.

#### Live Demo
Try it out here. Wait 5-10 seconds before clicking the poster, and you will see the astronaut has already rotated a fair bit
https://codepen.io/pushmatrix/pen/BEoExG

### Browser Affected

- [x] Chrome
- [x] Edge
- [x] Firefox
- [x] Helios
- [x] IE
- [x] Safari

### OS

- [x] Android
- [x] iOS
- [x] Linux
- [x] MacOS
- [x] Windows

### Versions
All versions

pushmatrix:
### Fixes #467

This fix checks to make sure that auto-rotation only starts once the poster is no longer visible and the model is interactable.

One thing is that this will not prevent the small stutter that happens when the poster transitions into 3D. That's because the rotation will start when the poster is dismissed, but the poster has a 0.3 second opacity transition. So during the transition you'll see the model rotation while the poster which is an image of the first frame will be fading away. It's not super noticeable, but it's also not perfectly smooth. 

It may be more trouble than it's worth solving that case since we'd have to listen for when the opacity transition ends before triggering the rotation. And since the loading mixin and the staging mixin don't know about each other, that could be tricky.

### DIFF
--- a/src/features/staging.ts
+++ b/src/features/staging.ts
@@ -84,7 +84,7 @@ export const StagingMixin = (ModelViewerElement:
         [$tick](time: number, delta: number) {
           super[$tick](time, delta);
 
-          if (this.autoRotate) {
+          if (this.autoRotate && (this as any)[$scene].model.hasModel()) {
             (this as any)[$scene].pivot.rotation.y +=
                 ROTATION_SPEED * delta * 0.001;
             this[$needsRender]();
--- a/src/features/staging.ts
+++ b/src/features/staging.ts
@@ -84,7 +84,7 @@ export const StagingMixin = (ModelViewerElement:
         [$tick](time: number, delta: number) {
           super[$tick](time, delta);
 
-          if (this.autoRotate && (this as any)[$scene].model.hasModel()) {
+          if (this.autoRotate && this.modelIsVisible) {
             (this as any)[$scene].pivot.rotation.y +=
                 ROTATION_SPEED * delta * 0.001;
             this[$needsRender]();
--- a/src/test/features/staging-spec.ts
+++ b/src/test/features/staging-spec.ts
@@ -116,4 +116,28 @@ suite('ModelViewerElementBase with StagingMixin', () => {
       });
     });
   });
+
+  suite('without a loaded model', () => {
+    setup(async () => {
+      element = new ModelViewerElement();
+      document.body.appendChild(element);
+    });
+
+    teardown(() => {
+      document.body.removeChild(element);
+    });
+
+    suite('auto-rotate', () => {
+      setup(() => {
+        element.autoRotate = true;
+      });
+
+      test('does not cause the model root to rotate over time', async () => {
+        const {turntableRotation} = element;
+        await timePasses(50);  // An arbitrary amount of time, greater than one
+                               // rAF though
+        expect(element.turntableRotation).to.be.equal(turntableRotation);
+      });
+    });
+  });
 });
--- a/src/test/features/staging-spec.ts
+++ b/src/test/features/staging-spec.ts
@@ -114,29 +114,19 @@ suite('ModelViewerElementBase with StagingMixin', () => {
                                // rAF though
         expect(element.turntableRotation).to.be.greaterThan(turntableRotation);
       });
-    });
-  });
 
-  suite('without a loaded model', () => {
-    setup(async () => {
-      element = new ModelViewerElement();
-      document.body.appendChild(element);
-    });
+      suite('when the model is not visible', () => {
+        setup(() => {
+          Object.defineProperty(
+            element, 'modelIsVisible', {value:false});
+        });
 
-    teardown(() => {
-      document.body.removeChild(element);
-    });
-
-    suite('auto-rotate', () => {
-      setup(() => {
-        element.autoRotate = true;
-      });
-
-      test('does not cause the model root to rotate over time', async () => {
-        const {turntableRotation} = element;
-        await timePasses(50);  // An arbitrary amount of time, greater than one
-                               // rAF though
-        expect(element.turntableRotation).to.be.equal(turntableRotation);
+        test('does not cause the model to rotate over time', async () => {
+          const {turntableRotation} = element;
+          await timePasses(50);  // An arbitrary amount of time, greater than one
+                                 // rAF though
+          expect(element.turntableRotation).to.be.equal(turntableRotation);
+        });
       });
     });
   });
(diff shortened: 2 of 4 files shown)


src/features/staging.ts
```diff
@@ -84,7 +84,7 @@ export const StagingMixin = (ModelViewerElement:
         [$tick](time: number, delta: number) {
           super[$tick](time, delta);
 
-          if (this.autoRotate) {
+          if (this.autoRotate && (this as any)[$scene].model.hasModel()) {
```
pushmatrix: The assumption here is that the base model only has children in it when the gltF is loaded and the poster is dismissed( or if there was no poster to begin with). That's how it seems to be working, with model.hasModel() only being true once `setSource` is called after the poster is dismissed. If this no longer becomes the case, this fix might not work.

I also don't know if `[$scene].model.hasModel()`will incur a performance cost being called on each tick. Right now hasModel calls `!!this.modelContainer.children.length;`

src/test/features/staging-spec.ts
```diff
@@ -116,4 +116,28 @@ suite('ModelViewerElementBase with StagingMixin', () => {
...
+      setup(() => {
+        element.autoRotate = true;
+      });
+
+      test('does not cause the model root to rotate over time', async () => {
+        const {turntableRotation} = element;
+        await timePasses(50);  // An arbitrary amount of time, greater than one
+                               // rAF though
+        expect(element.turntableRotation).to.be.equal(turntableRotation);
+      });
```
pushmatrix: It seems like mixins are intended to no be dependent on each other, so I couldn't specifically test behaviour with having a poster dismissed or not. That does mean that this test also operates under the assumption that if the poster is visible, then the model is not considered "loaded".

cdata: It's true that we haven't really done tests that integrate multiple mixins. That isn't to say we can't do it. But, I don't expect you to innovate in this regard.

That said, with `poster` and `preload` set together, it's possible for the model to be loaded but still hidden by the poster image.

src/features/staging.ts
cdata: If I understand the issue correctly, the psuedocode for the correct logic should be something along the lines of:

```javascript
if (autoRotate AND modelIsLoaded AND posterIsHidden) {
  rotateModel();
}
```

The clause `modelIsLoaded AND posterIsHidden` could be satisfied by a common notion of `modelIsVisible`, which would let us know that the model has been loaded, rendered and is actually being shown to the viewer.

What if we add an accessor `get modelIsVisible() { ... }` in the base class that can be overridden by mixins (like `LoadingMixin`) to specialize the definition? Then the conditional logic in `StagingMixin` can be:

```javascript
if (this.autoRotate && this.modelIsVisible) {
  // ...
}
```

And  this way `StagingMixin` doesn't need to worry about the implementation details of `LoadingMixin`. I think this would also excuse you from testing the explicit integration of `StagingMixin` and `LoadingMixin` (for now) :)

src/test/features/staging-spec.ts
pushmatrix: > That said, with `poster` and `preload` set together, it's possible for the model to be loaded but still hidden by the poster image.

That's what I thought, but it doesn't appear to be the case. 

The model is only considered "loaded" with all its children set when `setSource` is called: https://github.com/GoogleWebComponents/model-viewer/blob/master/src/three-components/Model.js#L186

If `poster` is set, then the `setSource` function is still only called once the poster is dismissed: https://github.com/GoogleWebComponents/model-viewer/blob/master/src/features/loading.js#L212. This is regardless of whether `preload` is on it or not.

You can test the behaviour by putting `poster` and `preload` attributes, and putting a breakpoint here https://github.com/GoogleWebComponents/model-viewer/blob/master/src/three-components/Model.js#L186. It will only trigger when you press the poster.

This seems really odd. Conceptually the flow seems weird that a function called `setSource` is called when a poster is dismissed. And I can see that with `preload` the glb is indeed being downloaded, but it is not loaded into the scene hierarchy until the poster is dismissed.

src/features/staging.ts
pushmatrix: That makes sense. I suppose if two mixins override `modelIsVisible`, then whichever mixin is included last in the list is the one that sticks?

cdata: @pushmatrix Actually, our mixins use inheritance to layer onto the class. So, as long as a mixin invokes the accessor on its super class, it can retain the implementation of other mixins without knowing what that implementation does. In this case, it would look like this in `LoadingMixin`:

```typescript
get modelIsVisible(): boolean {
  return super.modelIsVisible && this[$posterHidden];
}
```

src/test/features/staging-spec.ts
cdata: You're right, this is a bit confusing.

We have two parallel concepts:

 1. The notion of the model being loaded, which is to say that the model has been loaded over the network and cached by `<model-viewer>`'s internal glTF loader.
 2. The notion of setting the source on the `Model` instance, which causes that class to request the glTF from the loader (which might then load it over the network or pull it from cache) and then subsequently _render_ the model.

When I said it could be loaded, what I meant is that it could have been loaded by the internal glTF loader. We don't invoke `setSource` until interaction in order to avoid rendering before the poster is dismissed, but the model may already be "loaded".

I agree that `setSource` is an odd name given the context in this explanation. You can probably tell by the fact that it is still in JS that this part of the flow dates back to our prototyping days. So, bare with us 🙇

I think the main point is that regardless of whether `setSource` is invoked, the model could be considered "loaded".

pushmatrix: The model might be "loaded" but `hasModel()` still returns false until the poster is dismissed. Regardless, I prefer the `modelIsVisible` approach.

src/test/features/staging-spec.ts
```diff
@@ -114,6 +114,20 @@ suite('ModelViewerElementBase with StagingMixin', () => {
                                // rAF though
         expect(element.turntableRotation).to.be.greaterThan(turntableRotation);
       });
+
+      suite('when the model is not visible', () => {
+        setup(() => {
+          Object.defineProperty(
+            element, 'modelIsVisible', {value:false});
```
pushmatrix: Is this how we're mocking getters in this repo?

cdata: This is a totally reasonable way to do it, ya! 👌

pushmatrix:
@cdata This is ready for re-review :)

PR #476, opened 2019-04-05
Updated link text style & multiple sample page division(bg->border-bo…

yuinchien:
…ttom).

### DIFF
--- a/examples/styles/examples.css
+++ b/examples/styles/examples.css
@@ -367,7 +367,6 @@ code {
 paper-button {
   margin-top: 16px;
   font-weight: 500;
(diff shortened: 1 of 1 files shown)


examples/styles/examples.css
```diff
@@ -396,8 +397,8 @@ paper-button {
 }
 #list-example a {
   margin: 0;
-  color: #222;
-  text-decoration: underline;
+  /* color: #222; */
+  /* text-decoration: underline; */
```
cdata: Remove these?

yuinchien: one second

PR #478, opened 2019-04-05
[WIP] Elastic extent

### LINKED ISSUE #465: Elastic min/max extents of camera orbit
In AR QuickLook, there is a maximum and minimum to how much you can tilt the model (like model-viewer has). However if you keep dragging down, the model keeps tilting slightly with diminishing effect, allowing you to go beyond the limits.

quicklook

Then when you release, it snaps back as though there is some elastic force on it.

I wonder what people feel about this and if it makes for a more polished UX that we should port to the viewer.

mikkoh:
### Reference Issue

Fixes #465

elastic-astronaut

This PR makes it so that you can over rotate objects beyond maximum angles. If you go beyond the maximum angle there's a feeling like you're pulling on an elastic. Once you let go it will ease to the original clamped target rotation.

To make this PR not be *[WIP]*

- [x] Remove clamping azimuthal angles
- [x] Write tests 🤷‍♂️

### DIFF
--- a/src/test/three-components/SmoothControls-spec.js
+++ b/src/test/three-components/SmoothControls-spec.js
@@ -66,6 +66,8 @@ suite('SmoothControls', () => {
     camera = new PerspectiveCamera();
     controls = new SmoothControls(camera, element);
 
+    controls.applyOptions({overRotateAmount: 0});
+
     element.style.height = '100px';
     element.tabIndex = 0;
 
--- a/src/test/three-components/SmoothControls-spec.js
+++ b/src/test/three-components/SmoothControls-spec.js
@@ -324,5 +324,139 @@ suite('SmoothControls', () => {
         });
       });
     });
+
+    suite('elastic over rotate', () => {
+      setup(() => {
+        controls.applyOptions({
+          overRotateAmount: 1,
+          minimumPolarAngle: QUARTER_PI,
+          maximumPolarAngle: THREE_QUARTERS_PI
+        });
+        controls.setOrbit(0, 0, 0);
+        settleControls(controls);
+      });
+
+      test('can over rotate below minimumPolarAngle', () => {
+        const minPolar = controls.options.minimumPolarAngle;
+
+        controls.setOrbit(0, minPolar, 0);
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.equal(minPolar);
+
+        controls.adjustOrbit(0, QUARTER_PI * 0.25, 0);
+
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.lessThan(minPolar);
+      });
+
+      test('can over rotate above maximumPolarAngle', () => {
+        const maxPolar = controls.options.maximumPolarAngle;
+
+        controls.setOrbit(0, maxPolar, 0);
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.equal(maxPolar);
+
+        controls.adjustOrbit(0, -QUARTER_PI * 0.25, 0);
+
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.greaterThan(maxPolar);
+      });
+
+      test('can return to minimumPolarAngle on mouseup', () => {
+        const minPolar = controls.options.minimumPolarAngle;
+
+        controls.setOrbit(0, minPolar, 0);
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.equal(minPolar);
+
+        controls.adjustOrbit(0, QUARTER_PI * 0.25, 0);
+
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.lessThan(minPolar);
+
+        dispatchSyntheticEvent(element, 'mouseup');
+
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.equal(minPolar);
+      });
+
+      test('can return to maximumPolarAngle on mouseup', () => {
+        const maxPolar = controls.options.maximumPolarAngle;
+
+        controls.setOrbit(0, maxPolar, 0);
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.equal(maxPolar);
+
+        controls.adjustOrbit(0, -QUARTER_PI * 0.25, 0);
+
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.greaterThan(maxPolar);
+
+        dispatchSyntheticEvent(element, 'mouseup');
+
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.equal(maxPolar);
+      });
+
+      test('increment decreases when going beyond maximumPolarAngle', () => {
+        const maxPolar = controls.options.maximumPolarAngle;
+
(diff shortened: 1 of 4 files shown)


src/three-components/SmoothControls.ts
```diff
@@ -42,6 +42,8 @@ export interface SmoothControlsOptions {
   // A scalar in 0..1 that changes the dampening factor, corresponding to
   // factors from 0.05..0.3
   dampeningScale?: number;
+  // a percentage representing how much the model can be rotated beyond max min angles
+  overRotateAmount?: number;
```
mikkoh: I feel like this should be two values for polar angles and azimuth angles. But this might work for now. Thoughts?

src/three-components/SmoothControls.ts
```diff
@@ -339,9 +369,28 @@ export class SmoothControls extends EventDispatcher {
...
+
+    if (targetPhi < minPhi!) {
+      phiScale = 1 - Math.min((targetPhi - minPhi!) / (this[$minOverSpherical].phi - minPhi!), 1);
+    }
+
+    if (targetPhi > maxPhi!) {
+      phiScale = 1 - Math.min((targetPhi - maxPhi!) / (this[$maxOverSpherical].phi - maxPhi!), 1);
+    }
+
+    targetPhi = phi - deltaPhi * phiScale;
```
mikkoh: Basically all this is doing is slowing down `deltaPhi` based on how far we potentially would have gone over the max and min values.

src/three-components/SmoothControls.ts
```diff
@@ -541,6 +590,11 @@ export class SmoothControls extends EventDispatcher {
   private[$handlePointerUp](_event: MouseEvent|TouchEvent) {
     this.element.style.cursor = 'grab';
     this[$pointerIsDown] = false;
+    this.setOrbit(
+      this[$clampedTargetSpherical].theta,
+      this[$clampedTargetSpherical].phi,
+      this[$clampedTargetSpherical].radius
```
mikkoh: Basically when the user "lets go" we just want to orbit back to the clamped target.

src/three-components/SmoothControls.ts
```diff
@@ -597,4 +651,41 @@ export class SmoothControls extends EventDispatcher {
...
+    let minOverRotateAzimuthal = minimumAzimuthalAngle!;
+    let maxOverRotateAzimuthal = maximumAzimuthalAngle!; 
+
+    if (minimumAzimuthalAngle! > -Math.PI) {
+      minOverRotateAzimuthal = (minimumAzimuthalAngle! - Math.PI) * overRotateAmount! + minimumAzimuthalAngle!;
+    }
+
+    if (maximumAzimuthalAngle! < Math.PI) {
+      maxOverRotateAzimuthal = (Math.PI - maximumAzimuthalAngle!) * overRotateAmount! + maximumAzimuthalAngle!;
+    }
```
mikkoh: Azimuth angles are a bit weird with `overRotateAmount`. This might be easier if we just :fire: `overRotateAmount` and have a distinct over amount to rotate for azimuth. Not sure if it should be the same for polar angles.

With the above if `maximumAzimuthalAngle` is greater than PI no "elastic" effect is applied.

cdata:
Wow, awesome @mikkoh !

Regarding the clamped azimuth stuff.... I think we might need to _implement_ clamped azimuth as an API first, so maybe we can explore that as a separate change? BTW I have some WIP on that particular feature...

mikkoh:
Ok. Maybe Monday I’ll :fire: all things related to azimuth and in future PRs we can handle it.

src/test/three-components/SmoothControls-spec.js
```diff
@@ -322,5 +324,139 @@ suite('SmoothControls', () => {
...
+        controls.setOrbit(0, minPolar, 0);
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.equal(minPolar);
+
+        controls.adjustOrbit(0, QUARTER_PI * 0.25, 0);
+
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.lessThan(minPolar);
```
elalish: Shouldn't this be equal to minPolar after sufficient settling, because it springs back?

src/test/three-components/SmoothControls-spec.js
```diff
@@ -322,5 +324,139 @@ suite('SmoothControls', () => {
...
```
elalish: This is the same as above; is it necessary to duplicate?

src/test/three-components/SmoothControls-spec.js
```diff
@@ -322,5 +324,139 @@ suite('SmoothControls', () => {
...
+        expect(controls.getCameraSpherical().phi).to.be.equal(maxPolar);
+      });
+
+      test('increment decreases when going beyond maximumPolarAngle', () => {
+        const maxPolar = controls.options.maximumPolarAngle;
+
+        controls.setOrbit(0, maxPolar, 0);
+        settleControls(controls);
+
+        expect(controls.getCameraSpherical().phi).to.be.equal(maxPolar);
```
elalish: This is also a duplicate check.

src/three-components/SmoothControls.ts
```diff
@@ -42,6 +42,8 @@ export interface SmoothControlsOptions {
   // A scalar in 0..1 that changes the dampening factor, corresponding to
   // factors from 0.05..0.3
   dampeningScale?: number;
+  // a percentage representing how much the model can be rotated beyond max min angles
```
elalish: Is this really a percentage? We generally use ratios [0,1], instead of percentages [0,100].

src/three-components/SmoothControls.ts
```diff
@@ -597,4 +649,33 @@ export class SmoothControls extends EventDispatcher {
...
+      minimumAzimuthalAngle,
+      maximumAzimuthalAngle,
+      minimumPolarAngle,
+      maximumPolarAngle,
+      minimumRadius,
+      maximumRadius,
+      overRotateAmount
+    } = this[$options];
+
+    const minOverRotatePolar = (0 - minimumPolarAngle!) * overRotateAmount! + minimumPolarAngle!;
```
elalish: These calculations are simple enough that I don't think they warrant a whole extra method and two members. Plus I would not have intuited this function from the overRotateAmount name or description. Perhaps we should change overRotateAmount to specify the number of radians of extra rotation, which would both simplify the description and the calculations.

src/three-components/SmoothControls.ts
```diff
@@ -338,10 +364,31 @@ export class SmoothControls extends EventDispatcher {
...
+    } = this[$options];
+
+    let phiScale = 1;
+
+    if (targetPhi < minPhi!) {
+      phiScale = 1 - Math.min((targetPhi - minPhi!) / (this[$minOverSpherical].phi - minPhi!), 1);
+    }
+
+    if (targetPhi > maxPhi!) {
+      phiScale = 1 - Math.min((targetPhi - maxPhi!) / (this[$maxOverSpherical].phi - maxPhi!), 1);
```
elalish: The math here doesn't seem quite right. if phi = maxPhi, and I command a deltaPhi that would take me past maxOverSpherical, this would change that delta to zero, when I would instead expect it to be clamped. I think you want a mapping that takes the input delta: [0, inf] and maps it asymptotically to an output delta of [0, overRotateAmount] (see my comment below).

mikkoh:
This PR is so old I'm just gonna close this... I think it would be a nice addition but maybe it's better to start fresh again

Issue #486, opened 2019-04-15
Use dynamically colored linear gradient for progress mask


cdata:
In fulfilling #335 , we will have some basic, customizable progress bars built in to `<model-viewer>`. @yuinchien has described the default look and feel of these bars as including a linear gradient that is color-matched to the configured background-color of the element (or possibly the poster). See https://yuinchien.github.io/sandbox/progressbar/ for a reference of how the linear gradient should look and change over the course of the progress bar advancement.

We should implement this look and behavior as the default for our progress bars.

PR #487, opened 2019-04-16
Progress bars

### LINKED ISSUE #335: Add a visual progress bar / loading state to indicate that model is loading
It would be nice for the `<model-viewer>` to offer some kind of built-in progress bar to explain to users that the model is loading, and to help them understand the progress of the loading. For large models, it can be very confusing to stare at a blank space in the page for many seconds while the model loads.

### LINKED ISSUE #489: Pixel rounding of canvas size leads to unsightly margins
For example, the horizontal white line in this image should not be there: 

image

cdata:
# Progress Bars

This change proposes the addition of the notion of instance-wide loading progress, as well as corresponding customizable visualizations and API. In other words: progress bars.

Some highlights of this change include:

 - A new `progress` event is introduced, maps directly to our internal notion of progress
 - A built in progress bar UI has been introduced
   - Can be customized by adding child nodes to `slot="progress-bar"`
   - Can also be customized with CSS custom properties
 - The poster now has a default background color, and is always displayed before the model is loaded
   - Can now be customized by adding child nodes to `slot="poster"`
   - Can also be customized with CSS custom properties
 - A new `reveal` attribute has been introduced to control model visibility
   - Default configuration is equivalent to `reveal-when-loaded` semantics
   - `reveal="interaction"` defers loading / revealing model until user interaction
 - **🚨 BREAKING CHANGE:** `reveal-when-loaded` has been removed in favor of new `reveal` attribute and related semantics
 - A new `ProgressTracker` utility class has been introduced to assist in tracking progress across multiple, uncoordinated, asynchronous actions at once
 - Several modules have been converted to TypeScript, including: `model-viewer-base`, `features/loading`, `template` and `three-components/CachingGLTFLoader`
 - Asynchronous actors have been instrumented with progress callback support throughout
 - Type declarations for `GLTFLoader` have been copied over from the Three.js project
 - A bug was fixed relating to how DPR was applied when zooming
 - A bug was fixed that caused `<model-viewer>` to have 1px-wide empty margins sometimes

## New CSS custom properties

In keeping with our goal of enabling as much idiomatic customization as we can, the following CSS custom properties have been introduced by this change:

CSS property                         | Description
---------------------------- | ------------
`--poster-color`                    | Sets the `background-color` of the poster. Falls back to `--background-color` if set. Defaults to `#fff`.
`--poster-image`                  | Sets the `background-image` of the poster. **This is currently overridden by the `poster` attribute if it is set.** Defaults to `none`.
`--progress-bar-color`         | Sets the `background-color` of the progress bar. Defaults to `rgba(0, 0, 0, 0.4)`.
`--progress-mask` &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;      | Sets the `background` of the progress mask. Defaults to `#fff`.

## Tracking progress

In order to present the model at the correct time, we must take into account an uncertain number of possible network actions (environment map, glTF w/ multiple sub-resources etc) as well as an uncertain number of script/GPU processing steps (env map generation, skybox, PMREM etc) in order for the model to be considered presentable. The notion of "progress" introduced by this change encompasses all such actions that are implemented today.

Here is an example of how complex the process of presenting one model can be. Note that this is a fairly heavy configuration, resulting in almost 70MB of total assets transferred before we reveal the model:

Artboard 1@4x

## Examples

Basic progress bar w/ poster  | `reveal="interaction"`
----------------------------- | -----------------------------------
progressbar | progressbar-reveal-interaction

`src` attribute changing over time | Custom progress bar
---------------------------------- | ---------------------
progressbar-src-changing | progressbar-custom-bar

The following is an abbreviated form of the markup used to create the custom progress bar example above:

```html

<style>
  #progress {
    position: absolute;
    width: calc(100% / 3);
    height: 10px;
    color: rgba(255, 255, 255, 0.75);
    box-shadow: 0px 0px 8px 6px rgba(0, 0, 0, 0.25);
    border: 2px solid currentColor;
    border-radius: 10px;
    top: calc(50% - 5px);
    left: calc(50% - 100% / 6);
    opacity: 0;
    transition: opacity 0.3s 0.3s;
  }

  #progress.show {
    transition-delay: 0s;
    opacity: 1;
  }

  #progress>.bar {
    width: 100%;
    height: 100%;
    background-color: currentColor;
    transform-origin: top left;
    transform: scaleX(0);
  }
</style>

<!-- ... -->

<model-viewer src="./any.gltf">
  <div id="progress" slot="progress-bar">
    <div class="bar"></div>
  </div>
</model-viewer>

<script>
  const modelViewer = document.querySelector('model-viewer');
  const progress = document.querySelector('#progress');
  const bar = progress.querySelector('.bar');

  modelViewer.addEventListener('progress', (event) => {
    const {totalProgress} = event.detail;
    progress.classList.toggle('show', totalProgress < 1);
    bar.style.transform = `scaleX(${totalProgress})`;
  });
</script>
```

## Future work

 - Use a dynamically colored linear-gradient for progress mask #486 
 - Add CSS custom property details to docs site (@yuinchien is working on this) #488 
 - Make it possible to load environment map lazily #490
 - Make it possible to configure background color with CSS https://github.com/GoogleWebComponents/model-viewer/issues/492
 - A pretty bad Chrome bug related to `delegatesFocus` was discovered in the course of this change [crbug.com/950357](https://bugs.chromium.org/p/chromium/issues/detail?id=950357)

Fixes #335 
Fixes #489

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -57,106 +57,103 @@ const outputOptions = [{
   onwarn,
 }];
 
-if (NODE_ENV === 'production') {
-  plugins.unshift(
-    cleanup({
-      // Ideally we'd also clean third_party/three, which saves
-      // ~45kb in filesize alone... but takes 2 minutes to build
-      include: ['lib/**'],
-      comments: 'none',
-    }),
-  );
+if (NODE_ENV !== 'development') {
+  plugins.unshift(cleanup({
+    // Ideally we'd also clean third_party/three, which saves
+    // ~45kb in filesize alone... but takes 2 minutes to build
+    include: ['lib/**'],
+    comments: 'none',
+  }));
 
   outputOptions.push(
-    {
-      input: './lib/model-viewer.js',
-      output: {
-        file: './dist/model-viewer-umd.js',
-        sourcemap: true,
-        format: 'umd',
-        name: 'ModelViewerElement'
+      {
+        input: './lib/model-viewer.js',
+        output: {
+          file: './dist/model-viewer-umd.js',
+          sourcemap: true,
+          format: 'umd',
+          name: 'ModelViewerElement'
+        },
+        watch: {
+          include: 'lib/**',
+        },
+        plugins,
+        onwarn,
       },
-      watch: {
-        include: 'lib/**',
+      {
+        input: './lib/test/index.js',
+        output: {
+          file: './dist/unit-tests.js',
+          format: 'esm',
+          name: 'ModelViewerElementUnitTests'
+        },
+        watch: {
+          include: 'lib/**',
+        },
+        plugins,
+        onwarn,
       },
-      plugins,
-      onwarn,
-    },
-    {
-      input: './lib/test/index.js',
-      output: {
-        file: './dist/unit-tests.js',
-        format: 'esm',
-        name: 'ModelViewerElementUnitTests'
+      {
+        input: './lib/test/index.js',
+        output: {
+          file: './dist/unit-tests-umd.js',
+          format: 'umd',
+          name: 'ModelViewerElementUnitTests'
+        },
+        watch: {
+          include: 'lib/**',
+        },
+        plugins,
+        onwarn,
       },
-      watch: {
-        include: 'lib/**',
+      {
+        input: './lib/test/fidelity/components/image-comparison-app.js',
+        output: {
+          file: './dist/image-comparison-app.js',
+          sourcemap: true,
+          format: 'iife',
+          name: 'ImageComparisonApp'
+        },
+        watch: {
+          include: '{lib/test/fidelity/**,lib/third_party/**}',
+        },
+        plugins,
+        onwarn,
       },
-      plugins,
-      onwarn,
-    },
-    {
-      input: './lib/test/index.js',
-      output: {
-        file: './dist/unit-tests-umd.js',
-        format: 'umd',
-        name: 'ModelViewerElementUnitTests'
+      {
+        input: './lib/test/fidelity/image-comparison-worker.js',
+        output: {
+          file: './dist/image-comparison-worker.js',
+          sourcemap: true,
+          format: 'iife',
+          name: 'ImageComparisonWorker'
+        },
+        watch: {
+          include: '{lib/test/fidelity/**,lib/third_party/**}',
+        },
+        plugins,
+        onwarn,
       },
-      watch: {
-        include: 'lib/**',
+      {
+        input: './lib/documentation/components/example-snippet.js',
+        output: {
+          file: './examples/built/dependencies.js',
+          format: 'esm',
+          name: 'DocumentationDependencies'
+        },
+        plugins,
+        onwarn,
       },
-      plugins,
-      onwarn,
-    },
-    {
-      input: './lib/test/fidelity/components/image-comparison-app.js',
-      output: {
-        file: './dist/image-comparison-app.js',
-        sourcemap: true,
-        format: 'iife',
-        name: 'ImageComparisonApp'
-      },
-      watch: {
-        include: '{lib/test/fidelity/**,lib/third_party/**}',
-      },
-      plugins,
-      onwarn,
-    },
-    {
-      input: './lib/test/fidelity/image-comparison-worker.js',
-      output: {
-        file: './dist/image-comparison-worker.js',
-        sourcemap: true,
-        format: 'iife',
-        name: 'ImageComparisonWorker'
-      },
-      watch: {
-        include: '{lib/test/fidelity/**,lib/third_party/**}',
-      },
-      plugins,
-      onwarn,
-    },
-    {
-      input: './lib/documentation/components/example-snippet.js',
-      output: {
-        file: './examples/built/dependencies.js',
-        format: 'esm',
-        name: 'DocumentationDependencies'
-      },
-      plugins,
-      onwarn,
-    },
-    {
-      input: './lib/documentation/components/example-snippet.js',
-      output: {
-        file: './examples/built/dependencies-umd.js',
-        format: 'umd',
-        name: 'DocumentationDependencies'
-      },
-      plugins,
-      onwarn,
-    }
-  );
+      {
+        input: './lib/documentation/components/example-snippet.js',
+        output: {
+          file: './examples/built/dependencies-umd.js',
+          format: 'umd',
+          name: 'DocumentationDependencies'
+        },
+        plugins,
+        onwarn,
+      });
 }
 
 export default outputOptions;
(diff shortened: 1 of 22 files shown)


rollup.config.js
```diff
@@ -57,106 +57,103 @@ const outputOptions = [{
...
-if (NODE_ENV === 'production') {
-  plugins.unshift(
-    cleanup({
-      // Ideally we'd also clean third_party/three, which saves
-      // ~45kb in filesize alone... but takes 2 minutes to build
-      include: ['lib/**'],
-      comments: 'none',
-    }),
-  );
+if (NODE_ENV !== 'development') {
```
cdata: The default run mode for the build script was set as "development", but should probably be "production" so that naively running Rollup with the config builds all the things.

src/template.ts
```diff
@@ -21,46 +21,16 @@ template.innerHTML = `
...
-    .poster.show {
-      opacity: 1;
-      transition: none;
-    }
 
-    .poster:not(.show) {
-      pointer-events: none;
-    }
-
-    .click-to-view {
```
cdata: Removed this entirely, as it is vestigial from `<xr-model>` prototype days

yuinchien:
@cdata I wonder if we should simplify the default css properties for progress bar & future features. My concern is that the list of properties can be endless if we aren't extremely selective. I believe developers can further customize CSS if they want to, we should keep defaults to a minimum.

To me, the most crucial properties for progress bar are: 
- --background-color
- --poster-image
- --progress-bar-color

If we set default poster image & progress-mask to *background-size: cover*, we can still have a pretty ok baseline results. The progress-mask color should be directly linked to background-color, without transparency. We should recommend using poster image with the same background color for consistency. WDYT?

src/features/loading.ts
```diff
@@ -0,0 +1,402 @@
+/*
```
smalls: Just a nit, this would have been easier to read broken out into separate PRs. But, it was broken out into separate commits, so it was pretty easy to just go back there to take a look:

- https://github.com/GoogleWebComponents/model-viewer/pull/487/commits/3ad1274c0e784cd1e957769ef0f981079f359fe9
- https://github.com/GoogleWebComponents/model-viewer/pull/487/commits/38c9e2bf9dbe32061925a90d4b9e2dfc8d8636e5#diff-5fa0cbec670bc4c1c710be11a06e2376

src/features/loading.ts
```diff
@@ -0,0 +1,402 @@
...
+    get[$modelIsReadyForReveal](): boolean {
+      const {src} = this;
+
+      return !!src && CachingGLTFLoader.hasFinishedLoading(src) &&
+          this[$lastReportedProgress] === 1.0 && this[$shouldRevealModel]
+    }
+
+    get[$shouldRevealModel](): boolean {
+      return this.reveal === RevealStrategy.AUTO ||
+          this.reveal === this[$posterDismissalSource];
```
smalls: Would the second clause be clearer as `PosterDismissalSource.INTERACTION === this[$posterDismissalSource]` or `!!this[$posterDismissalSource]`?

smalls: I suppose my thought (particularly with the second) is that we probably don't care what the dismissal source is, only if we have one?

src/model-viewer-base.ts
```diff
@@ -208,7 +223,12 @@ export default class ModelViewerElementBase extends UpdatingElement {
     // sure that the value has actually changed before changing the loaded flag.
     if (changedProperties.has('src') && this.src !== this[$scene].model.url) {
       this[$loaded] = false;
-      this[$updateSource]();
+      (async () => {
+        const updateSourceProgress = this[$progressTracker].beginActivity();
+        await this[$updateSource](
+            (progress: number) => updateSourceProgress(progress * 0.9));
```
smalls: Why `* 0.9`? So it doesn't go to 1.0 until we're actually done (below)?

src/utilities/progress-tracker.ts
```diff
@@ -0,0 +1,141 @@
+import {clamp} from '../utils.js';
```
smalls: File header?

src/model-viewer-base.ts
cdata: Yeah

src/utilities/progress-tracker.ts
cdata: Will fix

src/features/loading.ts
cdata: Yeah, agreed. Lot's of regret.

cdata:
@yuinchien thanks for the feedback. The CSS properties have been pared down to four.

src/features/loading.ts
cdata: You're right, at this stage we only care that a dismissal source is set. As written, it's only helpful if we had N+1 possible sources.

src/features/loading.ts
smalls: :)

AYTSistemas13:
const modelViewer = document.querySelector('model-viewer');
const progress = document.querySelector('#progress');
const bar = progress.querySelector('.bar');

modelViewer.addEventListener('progress', (event) => {
    const {totalProgress} = event.detail;
    progress.classList.toggle('show', totalProgress < 1);
    bar.style.transform = `scaleX(${totalProgress})`;
  });

when I put this code into an Angular Project, typescript show a few of error messages, I already achieved most of them but one it is still going, and the error says: Property 'detail' does not exist on type 'ProgressEvent<EventTarget>'.

This is my code in my component from Angular project:

const modelViewerload = document.querySelector('model-viewer') as ModelViewerElement;
    const progress = document.querySelector('#progress') as HTMLElement;
    const bar = progress.querySelector('.bar') as HTMLElement;
    
    modelViewerload.addEventListener('progress', (event) => {
      const totalProgress:any = event.detail;
      progress.classList.toggle('show', totalProgress! < 1);
      bar.style.transform = `scaleX(${totalProgress})`;
    } );

I installed @google/model-viewer via npm, and this are my imports:
import { ModelViewerElement, RGBA } from "@google/model-viewer/lib/model-viewer";

I hope you can read me and find a solution. Thanks.

elalish:
maybe `const {totalProgress} = event.detail as any;`? We may have an error in our types.

AYTSistemas13:
> 

Yes, I did this and worked:
const totalProgress = (event as any).detail.totalProgress;

Issue #466, opened 2019-04-02
Interaction prompt does not animate in Shadow DOM in some browsers


pushmatrix:
### Description

The animation prompt does not render properly in Safari or Firefox.

What it's supposed to look like:

<img width="371" alt="Screen Shot 2019-04-02 at 5 12 28 PM" src="https://user-images.githubusercontent.com/190692/55436507-83583080-556a-11e9-81ed-b5e6b14ebdea.png">

What it looks like:

<img width="222" alt="Screen Shot 2019-04-02 at 5 12 57 PM" src="https://user-images.githubusercontent.com/190692/55437070-0037da00-556c-11e9-99c5-2f5a6c876e40.png">
(it's also not animating in firefox)

I believe the issue is related to Shadow DOMs. SVG transforms seem to not have consistent behaviour when inside of a Shadow DOM. I isolated the issue to show that it is not related to model viewer:

#### Live Demo
https://codepen.io/pushmatrix/pen/XQmrvx

On Firefox the animation doesn't even happen. On Safari the animation happens, but some elements are off.

Here's a more minimal one with just a rotating triangle: 
https://codepen.io/pushmatrix/pen/ZZbzPo.
The animation doesn't work on Firefox.

You can also see this behaviour in model viewer by loading it in Safari or Firefox, and clicking the model viewer to get focus, and waiting for the prompt to appear.

### Browser Affected

- [ ] Chrome
- [ ] Edge (not tested)
- [x] Firefox 66.0.2 
- [ ] Helios (not tested)
- [ ] IE (not tested)
- [x] Safari 12.0.3

### OS

- [ ] Android (not tested)
- [x] iOS 12.2
- [ ] Linux (not tested)
- [x] MacOS 10.14.3
- [ ] Windows (not tested)

### Versions
This is version agnostic. The svg animation issue is happening regardless of whether it is in the model viewer or not.

cdata:
Great catch. Due to weirdness in compositing, we are talking about potentially switching this prompt to an animated GIF or something like that. Either way, we will get this fixed.

cdata:
FYI I was able to adjust the SVG for the near term so that it looks less broken when the animations don't play:

![slightlybettersvg](https://user-images.githubusercontent.com/240083/56689345-b8e6ca00-668f-11e9-89bb-8c699ef9b58b.gif)

elalish:
This type of interaction prompt was removed quite awhile ago.

PR #503, opened 2019-04-24
Partially fix interaction prompt SVG

cdata:
This change partially fixes the issue reported in #466 . I'm going to leave #466 open until we have a full fix.

With this change, the prompt still doesn't animate, but doesn't look totally broken (as it did before) with the animations disabled:

slightlybettersvg


cdata:
Also note: I tried reproducing the problem in desktop Safari and Firefox, but was only able to reproduce the broken animations in Firefox.

PR #516, opened 2019-05-01
fixed flex syntax

### LINKED ISSUE #511: Fidelity test results viewer broken layout in Chrome 74
Reported by multiple users, and also reproducible in my local environment.

It looks like either Chrome 74 introduced a Flex bug, or we previously relied on a Flex bug that was fixed. Either way, this is what the results viewer looks like in Chrome 74 on Mac OS:

image

We also confirmed that the problem does not occur in Chrome 73.

elalish:
### Reference Issue
Fixes #511 

Looking at https://www.w3schools.com/cssref/css3_pr_flex.asp, it seems like their example code doesn't really match with their CSS syntax. I changed it from the example way to the CSS way, and it seems to be happy now. If someone has some other browsers handy, it'd be good to double check they all like it too.

### DIFF
--- a/src/test/fidelity/components/rendering-scenario.ts
+++ b/src/test/fidelity/components/rendering-scenario.ts
@@ -120,7 +120,7 @@ h2 {
 
 .screenshot > img {
   max-width: 256px;
-  flex: 1;
+  flex: 1 1 auto;
 }
 
 .check {


cdata:
Looking solid for me in Safari and Firefox, which is good enough for this test tool. If there are issues in other browsers, we can fix them as they are reported.

PR #518, opened 2019-05-01
Design update

yuinchien:


examples/styles/examples.css
```diff
@@ -13,25 +13,24 @@
...
   --heading-h3-max-width: 480px;
+  --heading-h1-margin-top: 40px;
   --browser-support-desc-margin-top: 40px;
   --browser-support-icons-margin-top: 16px;
-  --button-size: 56px;
-  --icon-size: 32px;
+  --button-size: 36px;
+  --icon-size: 28px;
   --header-height: 72px;
-  --poster-color: #455A64;
```
cdata: Is `white` okay as the poster color for everything?

examples/styles/examples.css
```diff
@@ -55,6 +54,34 @@ a {
...
+  background: white;
+  height: 56px;
+  align-items: center;
+  z-index: 1000;
+  overflow: hidden;
+  border-bottom: var(--border-style);
+}
+#sticky-shortcut > div {
+  margin-right: 12px;
+  /* font-weight: 500; */
```
cdata: Remove commented code or leave a note explaining why it is there!

examples/styles/examples.css
yuinchien: That's a mistake. white it is. will fix.

examples/styles/examples.css
```diff
@@ -314,32 +350,71 @@ code {
...
+
+#button-home {
+  font-weight: 500;
+  font-size: 20px;
+  margin: 0px 0 12px 0;
+  display: block;
+  display: flex;
+  align-items: center;
+}
+  #button-home .icon-modelviewer {
```
cdata: Funny indentation here

cdata:
Overall this looks really cool 😎

I wonder with the posters if we should do something to make it more obvious that they are posters. Right now I have hard time knowing if the poster has been dismissed or not.

One thing I did with a poster for another demo was add a gaussian blur:

progressbar

Or another thing we could do is add some text to the poster, or an icon or something... WDYT?

yuinchien:
The blur animation is nice, maybe start with less blurriness, what's the blur px in this camera demo?

cdata:
Haha uhm... I don't remember, but the last time I used the filter in Photoshop it was `21.0` 🤷‍♂

mrdoob:
I was very confused by that blur at first: "If the model is ready so we can blur it, why not show it non-blurred instead?" I did not understand it was a poster.

But now I understand we were trying to do like platforms like medium where a blurred version of the image shows while loading. Not sure if "poster" is conceptually a good name for this though. Placeholder maybe?

mrdoob:
I think it would be more clear if we showed a screenshot of the first frame as poster. Then, once it loads, the model replaces the image and starts rotating.

cdata:
We use `poster` because that is the name of a similar attribute for [the `<video>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-poster).

I don't feel strongly about the name. It's worth filing an issue to discuss it if you have the time @mrdoob 

The problem with showing the first frame is that the model can be (and often is) configured not to dismiss the poster until interaction. To me it's confusing when it looks like the model is there but you have to click on the model, wait for it to finish loading, and then click on it again to truly interact.

yuinchien:
@cdata Let me know if more change is needed. Thanks!

Issue #537, opened 2019-05-07
Fidelity tests show zero matching pixels for identical images with threshold 0


elalish:
### Description

Sample output for identical image comparison from npm run check-fidelity:

Comparing <model-viewer> to <model-viewer> (master)

  📏 Using threshold 0.0
  📊 Matching pixels: 0.00%
  📊 Mean color distance: 0.00%
  📊 Mean color distance (mismatching pixels only): 0.00%

  📏 Using threshold 1.0
  📊 Matching pixels: 100.00%
  📊 Mean color distance: 0.00%
  📊 Mean color distance (mismatching pixels only): 0.00%

  📏 Using threshold 10.0
  📊 Matching pixels: 100.00%
  📊 Mean color distance: 0.00%
  📊 Mean color distance (mismatching pixels only): 0.00%

The problem is that identical pixels are defined as < threshold apart, and they can't be less than zero. We could just remove the threshold 0 version, since that is actually what threshold 1 means, but I think it would probably be clearer to make identical mean <= threshold.

cdata:
Thanks for pointing this out!

PR #544, opened 2019-05-10
Remove ROOM_PADDING_SCALE

### LINKED ISSUE #527: ROOM_PADDING_SCALE is unnecessary
### Description

This is just a simple code cleanup thing, but there is no need to pad the room, as the camera near plane is not set by the room's dimensions. Also, once #526 is merged, there won't even be a room anymore. Removing this will simplify things, but will also slightly change default framing, thus changing the fidelity tests.

elalish:
### Reference Issue
Fixes #527, #537 

As noted in #526, the room padding has not been necessary for awhile, so this removes it for code clarity, especially since there is no longer a room. It has the effect of slightly zooming in all of our fidelity tests and removing the thin boarders around the square ones, as framing is now tight. Thus, all the golden images have been updated.


cdata:
Our margins were already a bit tight for my liking, if I'm honest. Would it make sense to compensate by adjusting the computed "ideal distance" in order to preserve a bit of the margin?

elalish:
I'm happy to do that if you'd like. One possibility is to change the ideal
distance to the safe distance where any rotation is still fully visible,
though that might be more margin than you want. Also, the margin you see is
highly dependent on how much the model fills its bounding box. Are there
particular models in our fidelity tests that you don't like the framing of?
Really the change is quite subtle.

cdata:
@elalish I would say that the antique camera is an example where the margin makes a big difference in the framing. Check out the bottom of the model, where the front leg of the tripod approaches the edge of the frame:

image

elalish:
Closing as out of date, since some of the cleanup here has landed elsewhere. May revisit later.

Issue #495, opened 2019-04-19
WebXR is broken in Android Chrome Canary


stars66:
Hi Developer Does it support android AR?

cdata:
Hi, thanks for the question! Currently, we have features that intend to integrate with the experimental and volatile WebXR APIs. These APIs only exist in Chrome Canary behind a flag today. So, the short answer is that you can do AR on Android with `<model-viewer>`, but only on an experimental basis (our integration tends to break a lot without us noticing). For production purposes, it is currently not possible to do AR on Android.

Esteban26:
Hi @cdata I'm trying to complete a POC for future purposes but I'm facing some issues to display the AR button and see the astronaut model with AR...

I'm following these docs to integrate the AR functionality on Android.

Device: Huawei Mate 20 Lite
Android version: 8.0.1
Chrome Canary version:  I tried 70.0.3538.2  and 72.0.3585.0

My code is this, and currently I'm getting the 3d model displayed, but AR button is not getting displayed.

I checked my browser support with this tool and everything seems to be ok, 
https://googlecodelabs.github.io/ar-with-webxr/work/

I also opened your AR example and it seems to be working good, but my local page doesn't work.
https://googlewebcomponents.github.io/model-viewer/examples/augmented-reality.html

Could you please help me to find a solution?

`<script type="module" src="https://unpkg.com/@google/model-viewer/dist/model-viewer.js"></script>
<script nomodule="" src="https://unpkg.com/@google/model-viewer/dist/model-viewer-legacy.js"></script>
<script src="https://unpkg.com/@magicleap/prismatic"></script>

<model-viewer alt="3D model" src="/dist/test/astronaut.glb" ios-usdz="/dist/test/astronaut.usdz" background-color="#222" magic-leap unstable-webxr></model-viewer>`

cdata:
@Esteban26 a lot of what you are describing could be affected by your project setup. Can you share a live demo that exhibits the problem, perhaps?

Esteban26:
@cdata Thanks for your quick response, I made a test page example and I uploaded to a demo server, this is the URL.
It may take some time to display the gbl model due to the low bandwith of the demo server, sorry for that.
http://graycomm.b.cctv.com.mx:82/viewer/dist/test/ar.html

cdata:
[...]

PR #558, opened 2019-05-19
Finish typescript conversion

### LINKED ISSUE #16: Convert to TypeScript

cdata:
The title says it all. I tried to keep each commit limited to the changes needed for just the named module that was being converted, but the changes per commit are in some ways order-dependent. For example, I began making `ModelScene` class fields mostly private, but had to change most of them to `public` as I uncovered access of them from other modules (as a side-note, this was an interesting exercise demonstrating the importance of explicit API access).

Some notes for this change:

 * The feature mixins have had counterpart type interfaces layered on. This is annoying because it expresses the API redundantly. However, it is also necessary to properly express the type signature of the return value of a mixin (which is always the type of the input class plus some interface).
 * Three.js types are mostly correct, but are incomplete in areas. Boolean fields like `isMesh` or `isTexture` seem to be missing. I had to add [a custom type](https://github.com/GoogleWebComponents/model-viewer/pull/558/files#diff-c89bf09b7a1f2994e20b6bd378d74f61R4) because `ACESFilmicToneMapping` wasn't represented in the constants.
 * I added a bunch of up-to-date types to our WebXR type definitions. I did not remove any existing types. This means there is likely a mix of up-to-date (as far as the spec is concerned) and deprecated types in this file. I would like to defer "fixing" this until we address #495 , when we can assess how close the latest APIs in Canary are to what is written in the spec.
 * Instead of converting our checked-in, ESM-ified "third party" Three.js add-ons, I created temporary type definition files for them. I hope that this is good enough for our purposes, but we may have to fully TypeScript-ify them down the road.

 

Fixes #16


src/three-components/ModelScene.ts
```diff
@@ -49,27 +61,47 @@ const $modelAlignmentMask = Symbol('modelAlignmentMask');
...
+  public activeCamera: Camera;
+
+
   /**
    * @param {ModelViewerElementBase} options.element
    * @param {CanvasHTMLElement} options.canvas
    * @param {number} options.width
    * @param {number} options.height
-   * @param {THREE.WebGLRenderer} options.renderer
+   * @param {Renderer} options.renderer
```
elalish: Are these type comments still necessary now that they're defined above?

src/three-components/Model.ts
```diff
@@ -108,18 +117,19 @@ export default class Model extends Object3D {
...
+  setObject(model: Object3D) {
     this.clear();
     this.modelContainer.add(model);
     this.updateBoundingBox();
     this.dispatchEvent({type: 'model-load'});
   }
 
   /**
-   * @param {String} url
+   * @param {String?} url
```
elalish: Again, should we have these decorations anymore?

src/test/features/controls-spec.ts
```diff
@@ -80,7 +87,7 @@ suite('ModelViewerElementBase with ControlsMixin', () => {
         // element:
         await timePasses();
 
-        settleControls(element[$controls]);
+        settleControls((element as any)[$controls]);
```
elalish:
```suggestion
        settleControls(controls);
```

src/test/three-components/SmoothControls-spec.ts
```diff
@@ -363,16 +366,16 @@ suite('SmoothControls', () => {
...
 
             controls.addEventListener('change', ({source}) => {
               changeSource.push(source);
             });
 
             dispatchSyntheticEvent(element, 'keydown', {keyCode: KeyCode.UP});
-            controls.update(performance.now, ONE_FRAME_DELTA);
-            controls.update(performance.now, ONE_FRAME_DELTA);
-            controls.update(performance.now, ONE_FRAME_DELTA);
+            controls.update(performance.now(), ONE_FRAME_DELTA);
```
elalish: Considering the first argument is unused, might we just pass 0?

src/test/three-components/TextureUtils-spec.ts
```diff
@@ -45,22 +47,22 @@ suite('TextureUtils', () => {
...
+      expect((texture as any).isTexture).to.be.ok;
       expect(textureMatchesMeta(
                  texture, {mapping: 'Equirectangular', url: EQUI_URL}))
           .to.be.ok;
     });
     test('loads a valid HDR texture from URL', async () => {
       let texture = await textureUtils.load(HDR_EQUI_URL);
       texture.dispose();
-      expect(texture.isTexture).to.be.ok;
+      expect((texture as any).isTexture).to.be.ok;
```
elalish: why is "as any" so often needed? Does that mean TS can't deduce its type?

src/types/webxr.d.ts
```diff
@@ -42,14 +42,104 @@ interface XR extends EventTarget {
   requestDevice(): Promise<XRDevice>;
 }
 
-interface XRSession {
+interface XRRigidTransform {
```
jsantell: This will be great once the spec is more stable and could find a home in some master typescript def list!

src/three-components/Model.ts
jsantell: These are a bit redundant with TS

src/three-components/ModelScene.ts
cdata: They can be removed. I filed https://github.com/GoogleWebComponents/model-viewer/issues/581

src/test/three-components/SmoothControls-spec.ts
cdata: We know the method's implementation, so technically we might. We go through the motions here and pretend not to know them, because the contract is the method signature (not its implementation).

src/test/three-components/TextureUtils-spec.ts
cdata: Yes. Unfortunately, Three.js type definitions aren't always complete. `as any` is an escape hatch that basically tells `tsc` it could be anything, so don't assert anything about the type. The less we use it the better. With Three.js we have to use it a fair amount (for now).

src/three-components/TextureUtils.ts
```diff
@@ -223,7 +245,7 @@ export default class TextureUtils extends EventDispatcher {
           nonPmremEnvironmentMap.dispose();
         }
       } else if (environmentMapWasGenerated) {
-        environmentMap.userData = {
+        (environmentMap as any).userData = {
```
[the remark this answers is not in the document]
cdata: Sadly, `userData` is one of the things that is not properly annotated on Three.js classes

Issue #498, opened 2019-04-22
[feature] Auto Zoom


devcline:
Hello i already saw a camera-orbit that can zoom a model, but what about  slow auto zoom after loading. Thanks

cdata:
@devcline are you referring to the sort of zoom-out effect that Poly has?

You can do something similar today with `camera-orbit` and the `model-visibility` event. Consider the following example:

```html
<model-viewer src="any.glb" camera-orbit="75deg 40deg 0.75m">
</model-viewer>

<!-- ... later ... -->
<script>
(() => {
  const modelViewer = document.querySelector('model-viewer');
  modelViewer.addEventListener('model-visibility', (event) => {
    if (event.detail.visible) {
      modelViewer.cameraOrbit = '0deg 75deg auto';
    }
  }, { once: true });
})();
</script>
```

I have made a live example for you to check out here: http://lopsided-motion.glitch.me/zoom.html

devcline:
This is very cool! Thanks you for this , i appreciate !

cdata:
@devcline eventually there will be. Kind of inline with #458 we would like to make other properties like maximum velocity configurable.

mikkoh:
I almost wonder if the `camera-orbit` attribute and `cameraOrbit` should animate. Like in @devcline's case he could write/use code that's external to animate.

One interesting thing that comes up with #458 that it might be nice in JS to have something like:
```javascript
const couldOrbitToPosition = modelViewer.setCameraOrbit('0deg 75deg auto');
```

cdata:
Yeah, great point. I think we have #454 to allow users to disable interpolation for exactly that case @mikkoh 🙌

cdata:
I'm going to close this in favor of #598 , which is asking for the same thing but has pretty GIFs :)

PR #563, opened 2019-05-21
Adjustable FOV

### LINKED ISSUE #232: Adjustable camera FOV
Currently, this is set to 45 degrees, with a robust rationale of "this looked good for some test models awhile back" -- of course there's not a one-size-fits-all, so this will need to be configured. Here's an example of zooming in on a long model, where the spheres on the sides really start to bend due to the small FOV:

screenshot from 2018-11-20 17-25-45

elalish:
### Reference Issue
Fixes #232, #537 

This adds the camera-FOV attribute to the <model-viewer> element, allowing control of the camera's field of view. In the process of making SmoothControls properly interpolate FOV, I also rewrote the interpolation scheme to be simpler, with fewer parameters and extensible to any number of interpolated states, which should help when we want to add configurable targets. I also removed the camera from controls.ts, such that SmoothControls is now the only place where the camera is updated.

One thing I have not done, but I mentioned in #232 was to change the definition of FOV from vertical FOV to widest FOV. I think this would be a good change to make, but the rest of this PR should not affect framing at all, so that might be best in a separate one, assuming others agree to the change. Also, it looks like the fidelity goldens are out of date again.


src/features/controls.ts
```diff
@@ -170,6 +170,17 @@ export const ControlsMixin = (ModelViewerElement:
...
+            this[$updateCameraFov]();
+          }
+        }
+
+        [$updateCameraFov]() {
+          let fov = deserializeAngleToDeg(this.cameraFOV);
+          if (fov == null) {
+            fov = deserializeAngleToDeg(DEFAULT_CAMERA_FOV);
+          }
+          this[$controls].setFov(fov as number);
```
cdata: Nit: since you `null`-checked this value, it's sufficient to assert it is not null to the type checker with `!`: `this[$controls].setFov(fov!)`

src/three-components/SmoothControls.ts
```diff
@@ -34,65 +34,53 @@ export interface SmoothControlsOptions {
...
   eventHandlingBehavior?: EventHandlingBehavior;
   // Controls when interaction is allowed (always, or only when focused)
   interactionPolicy?: InteractionPolicy;
 }
 
 export const DEFAULT_OPTIONS = Object.freeze<SmoothControlsOptions>({
-  minimumRadius: 8,
-  maximumRadius: 32,
+  minimumRadius: 1,
+  maximumRadius: 2,
```
cdata: 💯

src/three-components/SmoothControls.ts
```diff
@@ -34,65 +34,53 @@ export interface SmoothControlsOptions {
...
   minimumPolarAngle: Math.PI / 8,
   maximumPolarAngle: Math.PI - Math.PI / 8,
   minimumAzimuthalAngle: -Infinity,
   maximumAzimuthalAngle: Infinity,
-  decelerationMargin: 0.25,
-  acceleration: 0.15,
-  dampeningScale: 0.5,
```
cdata: The `Damper` helper seems reasonable, but it would be nice to make some of these qualities configurable in time. It was missing an issue, so I filed https://github.com/GoogleWebComponents/model-viewer/issues/580

See also this earlier discussion on the topic: https://github.com/GoogleWebComponents/model-viewer/issues/498#issuecomment-486709830

For now, YAGNI and thanks for cleaning all of this stuff up 🙇

src/three-components/SmoothControls.ts
```diff
@@ -154,6 +140,35 @@ export interface ChangeEvent extends Event {
...
+      // This clamps at the point where a large time step would cause the
+      // discrete step to overshoot.
+      this[$velocity] = 0;
+      return xDest;
+    }
+    if (timeStepMilliseconds < 0)
+      return x;
+    // Critically damped
+    const acceleration = NATURAL_FREQ * NATURAL_FREQ * (xDest - x) -
+        2 * NATURAL_FREQ * this[$velocity];
```
cdata: Avoid abbreviations (`NATURAL_FREQ` and `xDest` in this case)

src/three-components/SmoothControls.ts
```diff
@@ -154,6 +140,35 @@ export interface ChangeEvent extends Event {
...
+    this[$velocity] += acceleration * timeStepMilliseconds;
+    if (Math.abs(this[$velocity]) < NIL_SPEED * xNormalization &&
+        acceleration * (xDest - x) <= 0) {
+      // This ensures the controls settle and stop calling this function instead
+      // of asymptotically approaching their destination.
```
cdata: 🙌

src/three-components/SmoothControls.ts
```diff
@@ -154,6 +140,35 @@ export interface ChangeEvent extends Event {
   source: string,
 }
 
+class Damper {
```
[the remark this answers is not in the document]
cdata: Also, a descriptive comment would be a welcome sight for future readers 🙏

src/three-components/SmoothControls.ts
```diff
@@ -173,16 +188,25 @@ export interface ChangeEvent extends Event {
...
-  private[$spherical]: Spherical = new Spherical();
-  private[$targetSpherical]: Spherical = new Spherical();
-  private[$previousPosition]: Vector3 = new Vector3();
   private[$isUserChange]: boolean = false;
 
+  private[$spherical]: Spherical = new Spherical();
+  private[$destSpherical]: Spherical = new Spherical();
+  private[$thetaDamper]: Damper = new Damper();
+  private[$phiDamper]: Damper = new Damper();
+  private[$radiusDamper]: Damper = new Damper();
```
cdata: We are completely inconsistent with this, but I recently learned that TypeScript will infer a class field or variable's type based on initial assignment. Feel free to take advantage of this for brevity whenever you feel like it 💁‍♂

src/three-components/SmoothControls.ts
```diff
@@ -226,13 +247,11 @@ export class SmoothControls extends EventDispatcher {
...
         this[$handlePointerMove](event as TouchEvent);
 
-    this[$previousPosition].copy(this.camera.position);
-    this[$positionToSpherical](this.camera.position, this[$spherical]);
-    this[$targetSpherical].copy(this[$spherical]);
-
     this[$options] = Object.assign({}, DEFAULT_OPTIONS);
 
-    this.setOrbit();
+    this.setOrbit(0, Math.PI / 2, 1);
```
cdata: Previously, initial orbital position was inferred from the configured camera's position. One reason why this seems preferable is that it gives the user the ability to change camera control schemes without forcing the camera to reset its position to some fixed default.

OTOH this is of marginal benefit to `<model-viewer>`. But, the fixed defaults used here seem separable from `SmoothControls` implementation...

src/three-components/SmoothControls.ts
```diff
@@ -173,16 +188,25 @@ export interface ChangeEvent extends Event {
...
 export class SmoothControls extends EventDispatcher {
+  protected[$idealCameraDistance]: number = -1;
+
   private[$interactionEnabled]: boolean = false;
 
   private[$options]: SmoothControlsOptions;
   private[$upQuaternion]: Quaternion = new Quaternion();
   private[$upQuaternionInverse]: Quaternion = new Quaternion();
-  private[$spherical]: Spherical = new Spherical();
-  private[$targetSpherical]: Spherical = new Spherical();
```
cdata: Thanks for fixing our terminology usage here 🍻

src/three-components/SmoothControls.ts
```diff
@@ -306,17 +332,41 @@ export class SmoothControls extends EventDispatcher {
...
   }
 
   /**
-   * Resets the zoom sensitivity based on model scale, here represented by
-   * framedHeight.
-   * @param framedHeight
+   *
    */
-  updateFramedHeight(framedHeight: number) {
+  updateFraming(framedHeight: number, modelDepth: number, aspect: number) {
```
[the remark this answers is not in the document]
elalish: Framing seems pretty central to camera controls to me. What was strange was that both controls.ts and SmoothControls.ts were manipulating the camera, and when I was working on them I was running into a lot of bugs because it was unclear when one was stomping on the other. Also they were passing a lot of values back and forth, which is an indicator to me that the concepts are not well isolated. If you want to go down the bike shed route, I might rename SmoothControls to CameraControls and controls to CameraAttributes to make the current separation more clear.

src/three-components/SmoothControls.ts
```diff
@@ -340,21 +390,30 @@ export class SmoothControls extends EventDispatcher {
       maximumRadius
     } = this[$options];
 
-    const {theta, phi, radius} = this[$targetSpherical];
+    const {theta, phi, radius} = this[$destSpherical];
+
+    if (typeof destRadius === 'string') {
+      switch (destRadius) {
+        default:
+        case 'auto':
```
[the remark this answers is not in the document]
elalish: The "auto" keyword means apply the ideal camera distance, which I've moved into SmoothControls because I think it is more generically useful than to just <model-viewer>. However, we could tweak the API shape to clean it up a bit; maybe skip the string and use null to represent auto?

src/three-components/SmoothControls.ts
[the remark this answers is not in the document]
cdata: These are the questions that are running through my mind as I consider this:

 - If I used `SmoothControls` in a Three.js application that had nothing to do with `<model-viewer>`, would "framing" have meaning to me?
 - If I replaced Three.js with another renderer (thus requiring a new version of `SmoothControls` that worked in terms of that renderer's API), would I have to re-implement the notion of framing entirely?
 - Can the concept of framing be implemented without direct access to the private implementation details of `SmoothControls`?

Framing as a concept is currently part of the `ModelScene`. This makes sense after a fashion. After all, what purpose does the scene have if not to set the stage for the viewer?

OTOH it may make sense to extract it from `ModelScene` and implement it in a new construct that integrates a `ModelScene` and `SmoothControls`. But, adding it directly into `SmoothControls` is conflating two concepts without clear justification. Yes, it's obvious that there is overlap in terms of values that are used in both cases. But it's not obvious that there is conceptual overlap (or API even access requirements) to justify moving the implementation here.

src/conversions.ts
```diff
@@ -102,3 +106,15 @@ export const deserializeSpherical =
       return null;
     };
 
+export const deserializeAngleToDeg = (angleString: string): number|null => {
+  try {
+    const angleValueNode = parseValues(angleString);
+
+    if (angleValueNode.length === 1) {
```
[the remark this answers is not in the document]
elalish: No, that's the desired (output) units I'm specifying there; the input units are inferred.

src/conversions.ts
```diff
@@ -102,3 +106,15 @@ export const deserializeSpherical =
       return null;
     };
 
+export const deserializeAngleToDeg = (angleString: string): number|null => {
```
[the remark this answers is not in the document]
elalish: Will do

src/three-components/SmoothControls.ts
elalish: Yeah, I figured that as well; at least now we only need one parameter. For a future PR...

src/three-components/SmoothControls.ts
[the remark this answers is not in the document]
elalish: I put this back.

src/three-components/SmoothControls.ts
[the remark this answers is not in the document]
elalish: Put this back too, but keeping camera ownership in one place.

src/three-components/SmoothControls.ts
elalish: I removed that because I thought it was confusing. Sure, it uses the camera's existing position, but then ignores its orientation and instead points at the target, which you were not allowed to configure yet. Plus the most common case is that the camera is at the origin, which leads to a divide by zero, which was only solved by some surprising side effects of the default minimum radius. I think this will be more clear for developers.

src/three-components/SmoothControls.ts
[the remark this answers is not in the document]
elalish: Will do.

src/three-components/SmoothControls.ts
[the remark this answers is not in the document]
cdata: Thanks!

We spoke f2f about this topic. `SmoothControls` is responsible for manipulating the position (and now FOV) of the camera, while another entity (currently `ModelScene`) is responsible for determining the "framing" of the model by adjusting the frustum and repositioning the camera as needed.

Position / FOV manipulation is the domain of `SmoothControls`, so `ModelScene` should never perform those changes directly on the camera instance.

FOV sits at an intersection between these responsibilities. It seems like there is a lot of meat left on the bone for improvement in how the camera reference is manipulated by these different constructs.

src/conversions.ts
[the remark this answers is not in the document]
cdata: My mistake, I misread it

src/features/controls.ts
```diff
@@ -90,23 +90,24 @@ export const ControlsMixin = (ModelViewerElement:
             {type: String, attribute: 'camera-orbit', hasChanged: () => true})
         cameraOrbit: string = DEFAULT_CAMERA_ORBIT;
 
+        @property(
+            {type: String, attribute: 'camera-fov', hasChanged: () => true})
+        fieldOfView: string = DEFAULT_CAMERA_FOV;
```
cdata: 🍻

src/features/controls.ts
```diff
@@ -90,23 +90,24 @@ export const ControlsMixin = (ModelViewerElement:
             {type: String, attribute: 'camera-orbit', hasChanged: () => true})
         cameraOrbit: string = DEFAULT_CAMERA_ORBIT;
 
+        @property(
+            {type: String, attribute: 'camera-fov', hasChanged: () => true})
```
cdata: Make sure to update the attribute name here to `field-of-view`

src/features/controls.ts
```diff
@@ -90,23 +90,24 @@ export const ControlsMixin = (ModelViewerElement:
...
-        protected[$idealCameraDistance] = 1.0;
-
         protected[$idleTime]: number = 0;
         protected[$userPromptedOnce]: boolean = false;
         protected[$waitingToPromptUser]: boolean = false;
         protected[$shouldPromptUserToInteract]: boolean = true;
 
         protected[$controls]: SmoothControls;
 
+        protected[$idealCameraDistance] = -1.0;
```
cdata: Is `-1.0` a stand-in for "no value" here? If so, I would recommend using `null` instead, as it is more conventionally consistent with the rest of the library.

src/features/controls.ts
```diff
@@ -125,17 +126,20 @@ export const ControlsMixin = (ModelViewerElement:
...
+          this[$updateCameraOrbit]();
+          this[$updateCameraFov]();
         }
 
         getCameraOrbit(): SphericalPosition {
           const {theta, phi, radius} = this[$lastSpherical];
           return {theta, phi, radius};
         }
 
+        getCameraFov(): number {
```
cdata: Let's change this to `getFieldOfView()` to be inline with the property/attribute

src/features/controls.ts
```diff
@@ -177,6 +181,17 @@ export const ControlsMixin = (ModelViewerElement:
           if (changedProperties.has('cameraOrbit')) {
             this[$updateCameraOrbit]();
           }
+          if (changedProperties.has('fieldOfView')) {
+            this[$updateCameraFov]();
+          }
+        }
+
+        [$updateCameraFov]() {
```
cdata: Ditto, update naming here

src/features/controls.ts
```diff
@@ -246,39 +261,38 @@ export const ControlsMixin = (ModelViewerElement:
          * state.
          */
         [$updateCamera]() {
```
cdata: Thanks for maintaining a clear separation of concerns here for the time being!

src/features/controls.ts
```diff
@@ -246,39 +261,38 @@ export const ControlsMixin = (ModelViewerElement:
...
+              controls.getCameraSpherical().radius /
                   this[$idealCameraDistance] :
-                                  1;
-          this[$idealCameraDistance] = near + scene.modelDepth / 2;
+              1;
+          this[$idealCameraDistance] = framedDistance + scene.modelDepth / 2;
 
-          camera.aspect = scene.aspect;
-          camera.updateProjectionMatrix();
+          controls.updateIntrinsics(near, far, scene.aspect, zoomSensitivity);
```
cdata: 💯

src/features/controls.ts
```diff
@@ -246,39 +261,38 @@ export const ControlsMixin = (ModelViewerElement:
...
 
           controls.applyOptions({minimumRadius, maximumRadius});
-          controls.updateFramedHeight(scene.framedHeight);
-
-          controls.target.copy(scene.target);
 
           controls.setRadius(zoom * this[$idealCameraDistance]);
-          controls.jumpToDestination();
+          controls.setTarget(scene.target);
+          controls.jumpToGoal();
```
cdata: We previously agreed to use `Destination` as the domain term here. What changed?

src/features/controls.ts
```diff
@@ -378,11 +392,11 @@ export const ControlsMixin = (ModelViewerElement:
         }
 
         [$onChange]({source}: ChangeEvent) {
-          this[$deferInteractionPrompt]();
           this[$updateAria]();
           this[$needsRender]();
 
           if (source === 'user-interaction') {
+            this[$deferInteractionPrompt]();
```
cdata: Is this intentional? The implications are meaningful...

src/test/conversions-spec.ts
```diff
@@ -48,4 +49,19 @@ suite('conversions', () => {
       expect(deserializeSpherical('  0 0\n   0 ')).to.be.eql([0, 0, 0]);
     });
   });
+
+  suite('deserializeAngleToDeg', () => {
```
cdata: 💯

src/test/three-components/SmoothControls-spec.ts
```diff
@@ -60,6 +60,26 @@ const cameraIsLookingAt = (camera: Camera, position: Vector3) => {
 export const settleControls = (controls: SmoothControls) =>
     controls.update(performance.now(), FIFTY_FRAME_DELTA);
 
+suite('Damper', () => {
```

src/three-components/SmoothControls.ts
```diff
@@ -154,6 +139,44 @@ export interface ChangeEvent extends Event {
...
+ * xNormalization is a number to provide the rough scale of x, such that
+ * NIL_SPEED clamping also happens at roughly the same convergence for all
+ * states.
+ */
+export class Damper {
+  private[$velocity]: number = 0;
+
+  update(
+      x: number, xGoal: number, timeStepMilliseconds: number,
+      xNormalization: number): number {
```
cdata: thx `clang-format` 🙄 very readable

src/three-components/SmoothControls.ts
```diff
@@ -154,6 +139,44 @@ export interface ChangeEvent extends Event {
...
+        2 * NATURAL_FREQUENCY * this[$velocity];
+    this[$velocity] += acceleration * timeStepMilliseconds;
+    if (Math.abs(this[$velocity]) < NIL_SPEED * xNormalization &&
+        acceleration * (xGoal - x) <= 0) {
+      // This ensures the controls settle and stop calling this function instead
+      // of asymptotically approaching their goal.
+      this[$velocity] = 0;
+      return xGoal;
+    } else
+      return x + this[$velocity] * timeStepMilliseconds;
```
cdata: This should be enforced by linting in some way, but conventionally we always express `if/else` clauses as blocks. Please wrap this clause and the others above in braces `{ }`

Filed https://github.com/GoogleWebComponents/model-viewer/issues/608 to make improvements here

src/three-components/SmoothControls.ts
```diff
@@ -306,31 +338,37 @@ export class SmoothControls extends EventDispatcher {
...
-   * framedHeight.
-   * @param framedHeight
+   * Sets the non-interpolated camera parameters
    */
-  updateFramedHeight(framedHeight: number) {
-    // Make zoom sensitivity scale with model size:
-    this[$zoomMeters] = framedHeight / 10;
+  updateIntrinsics(
+      nearPlane: number, farPlane: number, aspect: number,
+      zoomSensitivity: number) {
```
cdata: 👍

src/features/controls.ts
elalish: I had a pile of variables called "destX" and you told me not to use shorthand, so instead of making every variable super long, I switched from destination to goal. I switched this function for consistency (along with the comments).

cdata: The new name works fine. The justification for making the change is one that I would prefer to avoid in the future. My general opinion on names is that code clarity trumps the cost of typing, and also the perceived byte cost (tools will fix this for us). In this case, both words are perfectly clear, but previously we had agreed to use the former word, and consensus should also be weighed in the balance 👍 

No need to revert, so long live `Goal`!

elalish: Understood, will keep in mind in future.

src/features/controls.ts
elalish: Good catch; I must have changed this when I was trying to diagnose a test failure. Reverted.

src/three-components/SmoothControls.ts
elalish: Not sure why this doesn't show as outdated, but I did fix this, along with any other unbracketed if statements I could find. I'm used to the Google C++ style guide where the brackets are omitted.

PR #566, opened 2019-05-23
switched HDR env map from half float to RGBM16

### LINKED ISSUE #565: Astronaut is too shiny on Android devices
### Description

On Android, the Astronaut model (and presumably others) is way too shiny compared to its matte look on other platforms. 

Correct: 
Screenshot from 2019-05-23 15-22-23
Incorrect: 
Screenshot from 2019-05-23 15-23-45

### Browser Affected
- [x] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS
- [x] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### Versions
- model-viewer: v0.3.1
- three.js: r104

elalish:
### Reference Issue
Fixes #565 

It seems some android devices don't support linear interpolation of float or half float textures, which also means they can't generate mipmaps for those. In order to make our generated HDR environment map work on those devices, we convert it to RGBM16, which looks like 8-bit RGBA, but gets 16x dynamic range with only moderate nonlinearity so that linear interpolation still does a reasonably decent job. 

We may be able to extend this approach to RGBELoader to convert that highly nonlinear format into RGBM16 to enable linear interpolation of background images.

### DIFF
--- a/src/three-components/EnvironmentMapGenerator.ts
+++ b/src/three-components/EnvironmentMapGenerator.ts
@@ -152,9 +152,7 @@ export default class EnvironmentMapGenerator extends EventDispatcher {
         uniform samplerCube tCube;
         varying vec3 vWorldDirection;
         void main() {
-          vec4 texColor = textureCube( tCube, vec3( - vWorldDirection.x, vWorldDirection.yz ), 2.0 );
-          vec4 texColorLinear = mapTexelToLinear( texColor );
-          gl_FragColor = linearToOutputTexel( texColorLinear );
+          gl_FragColor = textureCube( tCube, vec3( - vWorldDirection.x, vWorldDirection.yz ), 2.0 );
         }
       `,
       side: BackSide,
--- a/src/three-components/EnvironmentMapGenerator.ts
+++ b/src/three-components/EnvironmentMapGenerator.ts
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


mrdoob:
You found the magic combination!! I'll test this in a bit.

mrdoob:
Bad news... [This example](https://googlewebcomponents.github.io/model-viewer/examples/model-formats.html) should look like this:

Screen Shot 2019-05-23 at 8 48 22 PM

After this PR, it looks like this:

Screen Shot 2019-05-23 at 8 47 55 PM

elalish:
Shoot! The one page I must have forgotten to look at... yes indeed. At least it's the same on desktop and mobile now... I'll take a look.

elalish:
Okay, so I fixed the bug, but this has revealed a deeper problem, namely lack of precision in RGBM16:
RGBM16env
I was a bit concerned about this anyway; by using the max function we effectively only get the 8 bits of the M channel for all of our 16x dynamic range. A different encoding could better distribute the bits, but seems likely to degrade its linearity and cause interpolation artifacts. I'll poke around it some more. 

Another option is to make RGBE work by building custom interpolation, but that would require plumbing texture sizes into the fragment shaders, as well as making custom mipmap generators, and I think that would all have to happen in three, which sounds like a lot. Thoughts?

mrdoob:
Yeah, I bumped into that result too. I did not know how to solve it.
I thought RGBM16 had a way of adjusting the range but I guess not?

elalish:
@cdata Before I close this PR, can you check if this fixes the Samsung black model bug that I can't repro? This isn't the right fix, but it would tell us that it's a float texture problem.

Issue #519, opened 2019-05-02
Exit fullscreen button doesn't always exit fullscreen


cdata:
We received a report from a user that when they manually caused `<model-viewer>` to go fullscreen (by invoking the `requestFullscreen()` method on it directly), the exit fullscreen button introduced in #501 appeared but did not seem to do anything.

PR #579, opened 2019-05-29
Remove fullscreen request needed only for experimental WebXR

### LINKED ISSUE #543: Samsung Internet has prefixed Fullscreen API and requires a polyfill
~We have received some reports that the AR button does not work on the Samsung Galaxy S9 (that possibly extends to other Samsung devices). We should investigate what is happening to ensure compatibility with these devices.~

We should update our compatibility docs to include Samsung Internet, since it is pegged at a version of Chrome that is ~10 versions behind stable. And, we should see if we can mitigate the issue by using prefixed versions of the API if we detect them.

mkeblx:
### Reference Issue
Fixes #543 

Unless I'm not understanding the use of the Fullscreen API call purpose here, doing an Intent to the native ARCore viewer does not require the Fullscreen API and this matches the description https://github.com/GoogleWebComponents/model-viewer/blob/master/POLYFILLS.md#regarding-fullscreen-api of only needing for the 'experimental Web XR Device API-based AR mode'.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11432,6 +11432,12 @@
         "triple-beam": "^1.2.0"
       }
     },
+    "wkfs-polyfill": {
+      "version": "0.2.1",
+      "resolved": "https://registry.npmjs.org/wkfs-polyfill/-/wkfs-polyfill-0.2.1.tgz",
+      "integrity": "sha512-BEpaZJrcoSv8j+vIJfrBBUAYyFymlx156he8WaCs603qbnujx9vxtVociFPTEcIWQY9/y68rIAHh1Rl+8MzBqw==",
+      "dev": true
+    },
     "wordwrap": {
       "version": "0.0.3",
       "resolved": "https://registry.npmjs.org/wordwrap/-/wordwrap-0.0.3.tgz",
--- a/package.json
+++ b/package.json
@@ -107,7 +107,8 @@
     "rollup-plugin-node-resolve": "^3.0.0",
     "typescript": "^3.1.3",
     "wct-browser-legacy": "^1.0.1",
-    "web-component-tester": "^6.9.2"
+    "web-component-tester": "^6.9.2",
+    "wkfs-polyfill": "^0.2.1"
   },
   "dependencies": {
     "lit-element": "^2.0.0",
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -13,6 +13,7 @@
  * limitations under the License.
  */
 
+import 'wkfs-polyfill';
 import {property} from 'lit-element';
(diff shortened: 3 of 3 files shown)


Esteban26:
I think it is needed, it opens the fullscreen as a fallback behavior, btw fullscreen exit event is triggered by the browser when the android intent is executed.

cdata:
@mkeblx this is indeed intended, because there is no consistent way to detect that the intent failed in all browsers.

There is a documentation bug here: we aren't properly describing all of the cases when the Fullscreen API is required.

And there are also these related bugs: https://github.com/GoogleWebComponents/model-viewer/issues/543 https://github.com/GoogleWebComponents/model-viewer/issues/519

mkeblx:
I see, too bad can't detect failed intent. Samsung Internet will eventually support un-prefixed but trying to find a good solution for out-of-the-box usage prior to that.

From #543:
> It might be trivial for us to look for a prefixed version of the Fullscreen API and use that internally, so a future version may no longer require the polyfill if we fix this bug.

Do you think doing internally would be your preferred approach? If so I can upload a patch that does so, but maybe cleaner to do in a polyfill-like way in one spot (any maybe only need for `webkit` prefix as only care about Android for code path?)

One option that does solve, in case that didn't want to polyfill by default, while preserving fallback option when ARCore not installed/updated + `requestFullsceen` not available.
```js
case ARMode.AR_VIEWER:
  if (this.requestFullscreen)
    this.requestFullscreen();
  openARViewer(this.src, this.alt || '');
  break;
```

cdata:
@mkeblx Yeah, I think we will get a lot of mileage out of aliasing the API in the case that a `webkit` prefix is detected. [caniuse.com](https://caniuse.com/#search=fullscreen) suggests that the `webkit` alias alone cover every major browser except IE11.

The major challenge here is that it doesn't take much before we have written something bigger than [the actual polyfill](https://github.com/nguyenj/fullscreen-polyfill), which is quite small as it is.

If you would like to propose a change, that would be great. The criteria for acceptance will be that it is smaller in total byte size than the actual polyfill and doesn't require us to repeat ourselves twice everywhere we use the API :)

If it's helpful, I think the parts of the API that we currently use include:

 - `element.requestFullscreen()`
 - `element.exitFullscreen()`
 - `fullscreenchange` event
 - `document.fullscreenElement`

mkeblx:
Here is a smaller polyfill just for webkit, about as small as can get: <500 bytes. Basically the same as previous used polyfill on example pages implementation wise. Tests in the referenced package test successfully on m75 and m67 (version prior to removing prefix) and on device.
 
I see you've updated docs with Samsung Internet, thanks! Can update as well.

mkeblx:
No longer needed.

Issue #584, opened 2019-05-30
Fix "ModelViewerElementBase with ControlsMixin » when registered » camera-controls » a11y » does not prompt if user already interacted"


cdata:
Flakes seen on macOS 10.13 safari 12.0 and iPhone XR Simulator

Issue #583, opened 2019-05-30
Fix "ModelViewerElementBase with ControlsMixin » when registered » camera-orbit » can independently adjust azimuth"


cdata:
Flakes seen on macOS 10.13 safari 11.1

Issue #582, opened 2019-05-30
Fix "ModelViewerElementBase with StagingMixin » with a loaded model » auto-rotate » pauses rotate after user interaction"


cdata:
Flakes seen on iPhone XR Simulator

PR #586, opened 2019-05-30
Quell flakey tests

cdata:
This change removes three known flakey tests from the suite. The following issues have been filed to address fixing these tests: #582 #583 #584 

This change also removes some fidelity tests from rotation because of unreproducible script errors causing them to bail. These issues were filed to cover re-activating them: #609 #610 #611 #612


cdata:
Looks like we may have introduced a catastrophic failure that is particular to IE11 in #552 . It may be a side-effect of legitimate test failure, but "IndexSizeError" is apparently thrown when a call to `drawImage` on 2D context invokes dimensions that are larger than the canvas size.

cdata:
I was unable to reproduce the "IndexSizeError" in a live Sauce Labs session. However, I did run into IE11 hitting its memory ceiling during test runs, so we're going to try more aggressively clearing the cache.

cdata:
I'm incrementally removing the fidelity tests that are failing to run. I will file issues for each.

cdata:
Related to `IndexSizeError`: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/16192680/

cdata:
The IE `IndexSizeError` was a race condition related to async polyfill upgrade timing. Essentially, the scene was sized to 0 x 0 because the constructor was invoked before the element had dimensions, and render was then being invoked before the next resize. I think I have resolved it by setting the standard unsized media dimensions (300 x 150) in the case when the constructor is invoked before the element is connected.

cdata:
I have also split our fidelity tests out into a separate Travis build. This should speed up unit test results substantially.

PR #591, opened 2019-05-31
Remove experimental-pmrem flag

### LINKED ISSUE #336: Cache generated environment map

### LINKED ISSUE #396: Deprecate stage-light-intensity / change default to 0
In #391 we added a configuration that introduced a `stage-light-intensity` attribute and formalized a de-facto detail of lighting that arose from our fidelity explorations: there are different baseline levels of stage lighting for PMREM and non-PMREM lighting scenarios.

@jsantell left this feedback on the topic:

> Quick comment on the primary/secondary 'baseline' for the lights: while I think these settings make sense to balance the overall illumination with PMREM, it seems a bit strange in the scenario where devs have control over the intensity of specifically non-environmental lighting (e.g. a dimly lit environment map, stage lights change drastically when toggling on PMREM).

We should review how the lighting configuration can result in dramatic changes depending whether PMREM is enabled or not. Certainly it would be bad for devs to be surprised when their lighting configuration changes drastically when flipping a single attribute. And, it requires folks to have multiple understandings of the effect various lighting attributes might have on the scene.

### LINKED ISSUE #565: Astronaut is too shiny on Android devices

### LINKED ISSUE #574: Environment map generation fails on some devices
### Description

Model looks very different on different devices. 

On an Android, Honor 7X, it looks like this:
signal-2019-05-28-141018

On iPhone XR:
IMG_5890

On Android, Xiaomi Mi 9:
Screenshot_2019-05-28-17-41-26-090_com android chrome

The last two are pretty similar but reflections still look different (not very visible in the screenshots).

#### Live Demo
https://bramble-geranium.glitch.me/

### Browser Affected

- [x] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [x] Safari

### OS

- [x] Android
- [x] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

Devices where this is broken:

Honor 7X (BND-L21)
Android 8.0
Chrome 74.0.3729.157

Samsung Galaxy S6 (SM-G920F)
Android 7.0
Chrome 74.0.3729.157

### LINKED ISSUE #587: Rendering inconsistencies with fine detailed normals + roughness
### Description

I'm having an issue with the lighting on certain PBR models that have a lot of small details in the normal map. The way the details in the normal map are reflecting light seems to be off and the model comes across a bit too shiny and wet.

Below are images in various viewers. They use the same environment map from here: https://github.com/GoogleWebComponents/model-viewer/issues/521#issuecomment-488694339. 

## Model Viewer
Screen Shot 2019-06-10 at 3 34 52 PM
Screen Shot 2019-06-10 at 3 34 58 PM

## Sketchfab
Screen Shot 2019-06-10 at 3 29 51 PM
Screen Shot 2019-06-10 at 3 29 55 PM

## Substance
Screen Shot 2019-06-10 at 3 31 37 PM

## AR Quick Look
(Doesn't use the same environment map)
arql

We created the material in substance painter, and Sketchfab & AR Quick Look render it as designed. Model-viewer makes it look like the normal map roughness is really low and reflective.

#### Live Demo
https://ninth-paste.glitch.me/

Download the model [here](https://github.com/GoogleWebComponents/model-viewer/files/3273539/fabricball.zip)

### Browser Affected

- [x] Chrome (74.0.3729.157)
- [x] Firefox (66.0.5)
- [x] Safari (12.0.3)

### OS

- [ ] Android (not tested)
- [x] iOS (12.3.1)
- [x] MacOS (10.14.3)

### Versions
- model-viewer: v0.3.1 and latest master #f5b3b55  
- three.js: 105

### LINKED ISSUE #601: LDR environment maps without PMREM do not light nonmetallic, rough objects at all.
### Description

All of our lighting for objects like this (metal = 0, roughness = 1 is the default for unspecified materials, like the Astronaut) comes from our stage light; the environment is completely ignored for jpgs. It has very little effect on shiny objects (they are still much too dark), but with metal = 1, everything is fine.

#### Live Demo
The duck, when the stage light turns off: https://googlewebcomponents.github.io/model-viewer/examples/lighting-and-environment.html
Screenshot from 2019-06-04 11-11-30

### Browser Affected

- [x] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [x] Android
- [ ] iOS
- [x] Linux
- [ ] MacOS
- [ ] Windows

### Versions
- model-viewer: v0.3.1
- three.js: r104

### LINKED ISSUE #648: Select / author fidelity tests to judge effects of PMREM overhaul
We are about to embark on work to simplify our rendering pipeline (#396, #647). As part of this work we will be enabling PMREM for rendering in every configuration. We will also be deprecating and subsequently removing the `stage-light-intensity` attribute.

In order to confirm rendering parity and track regressions over the course of this work, we should select and/or implement a battery of fidelity tests that represent model details that stand to be affected or tell us the most about what changed (if anything).

elalish:
Fixes #587 
Fixes #565
Fixes #601
Fixes #396
Fixes #648
Fixes #574
Fixes #336

The only real trouble with bug #587 is that our HDR environment maps really don't work at all without pmrem. Additionally, #565 can be solved by turning on PMREM for the generated enviroment map, and #601 can be solved by turning on PMREM for LDR environment maps. Given all of that, this PR removes the experimental-pmrem flag and instead enables PMREM in all cases, which has the nice side effect of simplifying both the API and the code.

The only downside is a performance hit. However, we plan to look into improving our rendering in the near future, and I believe we can attain what PMREM has been doing with a simpler and faster algorithm.


elalish:
Actually, upon further inspection, it turns out #565 can be fixed by turning on PMREM for the default generated environment map, so that answers my question above. Two birds with one stone.

src/three-components/TextureUtils.ts
```diff
@@ -233,7 +232,10 @@ export default class TextureUtils extends EventDispatcher {
         }
       }
 
-      if (options.pmrem) {
+      // Apply PMREM to HDR textures, which if loaded from file have
+      // RGBEEncoding, and if generated are HalfFloatType
+      if (environmentMap.encoding === RGBEEncoding ||
+          environmentMap.type === HalfFloatType) {
```
cdata: Woah, we weren't even applying PMREM to the generated map? That... doesn't seem right...

cdata:
Marked as breaking because as of this change we turn on `experimental-pmrem` for everything.


@elalish what do you think about adding an escape hatch for folks who might need to migrate? I'm thinking something along the lines of a `disable-pmrem` flag...

elalish:
Well, dusting off my PM hat, I'd ask if we know of any situations where this behavior is worse (less physically accurate) than the current behavior? Also, for the sake of our own testing and maintenance, I think it's important that we reduce the number of rendering paths we expose. Better to do that now so that we can get the bug reports and fix them before our stable release.

cdata:
The proposal would be to give folks an off-ramp and then remove the flag entirely in a subsequent version. The risk of not doing it is that you lose those users entirely when they feel left behind/blind-sided without a transition path into the future.

elalish:
I think there are too many serious rendering bugs that this fixes to warrant keeping the old path open. I'm also concerned about the dev burden it will put on us to try and keep up both paths as we improve the rendering in 0.5.

cdata:
I think we should wait for v0.5.0 as a matter of timeliness. v0.4.0 has been lingering too long, and our partners need some incremental improvements to become unblocked. I'm also worried that without low-barrier migration path accompanying this change, we could inadvertently hurt the uptake of v0.4.0. v0.5.0 will give us the time to live with this change for a bit before shipping it to users.

Keep in mind that we can start landing v0.5.0 changes as soon as v0.4.0 goes out.

cdata:
Just had a thought as I filed #649 : if we fix this first, we'll cut the number of PMREM generation passes by about half in the tests. Maybe we should go ahead and fix this early in case it wins us some wiggle room in Safari.

cdata:
Let's give this another shot now that #649 is fixed.

src/features/environment.ts
```diff
@@ -213,16 +208,11 @@ export const EnvironmentMixin = (ModelViewerElement:
...
-          const environmentIntensity = this.experimentalPmrem ?
-              this.environmentIntensity * 0.65 :
-              this.environmentIntensity;
-
+          const illuminationRole = IlluminationRole.Secondary;
           scene.configureStageLighting(
               this.stageLightIntensity, illuminationRole);
-          scene.model.setEnvironmentMapIntensity(environmentIntensity);
+          scene.model.setEnvironmentMapIntensity(
+              this.environmentIntensity * 0.65);
```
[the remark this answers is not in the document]
elalish: I've removed this in a follow-on PR.

src/three-components/TextureUtils.ts
```diff
@@ -168,118 +171,140 @@ export default class TextureUtils extends EventDispatcher {
...
+  /**
+   * Loads a skybox from a given URL, then PMREM is applied to the
+   * skybox texture and the resulting WebGLRenderTarget is returned,
+   * with the assumption that it will be used as an environment map.
+   */
+  private[$loadEnvironmentMapFromSkyboxUrl](
+      url: string,
+      progressTracker?: ProgressTracker): Promise<WebGLRenderTarget> {
+    if (!this[$environmentMapCache].has(url)) {
+      const skyboxLoads = this[$loadSkyboxFromUrl](url, progressTracker);
```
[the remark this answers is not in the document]
cdata: The cache will be populated synchronously (relative to this line). The trick is that we populate it with a promise for the eventual value, and that promise can be looked up and `await`d in parallel by other routines trying to load from the same URL.

elalish:
Ah, very cool. Thanks for the explanation!

src/three-components/TextureUtils.ts
```diff
@@ -168,118 +171,140 @@ export default class TextureUtils extends EventDispatcher {
...
+      const environmentMapGenerator =
+          new EnvironmentMapGenerator(this.renderer);
+      const interstitialEnvironmentMap = environmentMapGenerator.generate();
+
+      this[$generatedEnvironmentMap] =
+          this.pmremPass(interstitialEnvironmentMap.texture);
+
+      // We should only ever generate this map once, and we will not be using
+      // the environment map as a skybox, so go ahead and dispose of all
+      // interstitial artifacts:
```
elalish: Is it true that we don't use the generated environment map as a skybox? I'm not sure why that hadn't occurred to me before. Maybe I should update our fidelity tests accordingly.

src/three-components/TextureUtils.ts
```diff
@@ -299,22 +324,43 @@ export default class TextureUtils extends EventDispatcher {
...
+    }
+
+    this[$environmentMapCache].clear();
+
+    for (const skyboxLoads of this[$skyboxCache].values()) {
+      const skybox = await skyboxLoads;
+      if (skybox != null) {
+        skybox.dispose();
+      }
+    }
```
elalish: This for loop could be a method to skip the repetition.

test/wct-sauce.conf.json
```diff
@@ -7,7 +7,7 @@
   ],
   "clientOptions": {
     "mochaOptions": {
-      "timeout": 90000
+      "timeout": 180000
```
elalish: This actually didn't help as I recall, so we could probably revert it. Up to you.

index.html
```diff
@@ -166,13 +166,6 @@ <h3 class="grouping-title">Attributes</h3>
               is for models with unlit materials. Set this value to 0 to hide the
               environment entirely. Defaults to 1.</p>
             </li>
-            <li>
-              <div>experimental-pmrem</div>
```
cdata: 🎉

src/features/environment.ts
```diff
@@ -69,9 +67,6 @@ export const EnvironmentMixin = (ModelViewerElement:
         @property({type: String, attribute: 'background-color'})
         backgroundColor: string = DEFAULT_BACKGROUND_COLOR;
 
-        @property({type: Boolean, attribute: 'experimental-pmrem'})
```
cdata: 🎉

src/features/environment.ts
```diff
@@ -213,16 +205,8 @@ export const EnvironmentMixin = (ModelViewerElement:
...
-          const illuminationRole = this.experimentalPmrem ?
-              IlluminationRole.Secondary :
-              IlluminationRole.Primary;
-          const environmentIntensity = this.experimentalPmrem ?
-              this.environmentIntensity * 0.65 :
-              this.environmentIntensity;
-
-          scene.configureStageLighting(
-              this.stageLightIntensity, illuminationRole);
-          scene.model.setEnvironmentMapIntensity(environmentIntensity);
```
cdata: 💯

src/test/three-components/TextureUtils-spec.ts
```diff
@@ -30,16 +30,19 @@ const HDR_EQUI_URL = assetPath('spruit_sunrise_2k.hdr');
 
 suite('TextureUtils', () => {
   let textureUtils: TextureUtils;
-  let renderer: WebGLRenderer;
+  // The renderer can retain state, so these tests have the possibility of
+  // getting different results in different orders. However, our use of the
+  // renderer *should* always return its state to what it was before to avoid
+  // this kind of problem (and many other headaches).
```
cdata: 💯

src/test/three-components/TextureUtils-spec.ts
```diff
@@ -100,7 +103,8 @@ suite('TextureUtils', () => {
   });
 
   suite('generating an environment map and skybox', () => {
-    let textures: {skybox: WebGLRenderTargetCube|null, environmentMap: Texture}|
+    let textures:
+        {environmentMap: WebGLRenderTarget, skybox: WebGLRenderTargetCube|null}|
```
cdata: Let's import the type we created for this and use it here.

src/test/three-components/TextureUtils-spec.ts
```diff
@@ -130,57 +135,34 @@ suite('TextureUtils', () => {
...
-            mapping: 'Cube',
+          expect(textureMatchesMeta(textures.environmentMap.texture, {
+            mapping: 'PMREM',
             url: HDR_EQUI_URL
           })).to.be.ok;
         });
 
     test(
-        'returns an environmentMap and skybox texture from url with PMREM',
+        'returns an environmentMap and skybox texture from two urls',
```
cdata: Did we lose a test condition in this cleanup? I think there were two cases tested previously:

 1. A skybox is generated, and an environment map is generated for me from the skybox
 2. A skybox is generated, and an environment map is generated for me from the given URL

src/third_party/three/EquirectangularToCubeGenerator.d.ts
```diff
@@ -10,7 +10,8 @@ export interface CubemapGeneratorOptions {
 export class CubemapGenerator {
   constructor(renderer: WebGLRenderer);
   public fromEquirectangular(
-      texture: Texture, options?: CubemapGeneratorOptions): WebGLRenderTarget;
+      texture: Texture,
+      options?: CubemapGeneratorOptions): WebGLRenderTargetCube;
```
cdata: 👍

src/three-components/ModelScene.ts
```diff
@@ -97,16 +90,8 @@ export default class ModelScene extends Scene {
 
     this.model = new Model();
     this.shadow = new StaticShadow();
-    this.light =
-        new HemisphereLight(0xBBBBBB, 0x444444, AMBIENT_LIGHT_HIGH_INTENSITY);
-    this.light.name = 'HemisphereLight';
-    this.light.position.set(2, 4, 2);
```
cdata: 👍

src/three-components/ModelScene.ts
```diff
@@ -237,16 +221,8 @@ export default class ModelScene extends Scene {
...
-  configureStageLighting(
-      intensityScale: number, illuminationRole: IlluminationRole) {
-    this.light.intensity = intensityScale *
-        (illuminationRole === IlluminationRole.Primary ?
-             AMBIENT_LIGHT_HIGH_INTENSITY :
-             AMBIENT_LIGHT_LOW_INTENSITY);
-    this.shadowLight.intensity = intensityScale *
-        (illuminationRole === IlluminationRole.Primary ?
-             DIRECTIONAL_LIGHT_HIGH_INTENSITY :
-             DIRECTIONAL_LIGHT_LOW_INTENSITY);
```
cdata: 💯

src/three-components/TextureUtils.ts
```diff
@@ -36,7 +40,16 @@ const HDR_FILE_RE = /\.hdr$/;
 const ldrLoader = new TextureLoader();
 const hdrLoader = new RGBELoader();
 
+const $environmentMapCache = Symbol('environmentMapCache');
+const $skyboxCache = Symbol('skyboxCache');
```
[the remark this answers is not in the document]
cdata: 🙌

src/three-components/TextureUtils.ts
cdata: I think that it's fine to let the fidelity tests have a skybox. This is the existing behavior when specifying a `background-image`, so it is expected. We can change this behavior if we want 🤷‍♂ but let's file an issue to discuss it?

test/fidelity/config.json
```diff
@@ -213,6 +213,34 @@
...
+        "file": "model-viewer-golden.png"
+      }, {
+        "name": "Filament",
+        "file": "filament-golden.png"
+      }],
+      "dimensions": {
+        "width": 1536,
+        "height": 1536
+      }
+    },
```
cdata: 🍻

test/fidelity/khronos-MetalRoughSpheres-HDR/index.html
```diff
@@ -0,0 +1,38 @@
+<!--
+/*
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
cdata: `2019`

test/wct-sauce.conf.json
cdata: It's arbitrarily high either way. No harm in keeping it this high.

cdata:
This PR includes several threads of work by @elalish , the descriptions for which I am quoting here for posterity:

> ## [Cache environments and skyboxes](https://github.com/GoogleWebComponents/model-viewer/pull/667)
> 
> Really #336 was already fixed, but not in a complete way as only the generated environment was cached, but not any loaded environments, skyboxes, or PMREM results. Now all of the above are cached in a single system. The near-term goal of this PR is to get our automated tests to pass on Safari with PMREM, as we're guessing the current problem is they use too much memory from duplicate environments.

> ## [Reuse renderer in tests](https://github.com/GoogleWebComponents/model-viewer/pull/668)
>
> Still working on getting our tests to pass. All those Safari warnings about too many WebGL contexts made me dig in a little. I found this thread pixijs/pixi.js#2233 (comment) which says that Three's renderer.dispose() method only simulates a context being destroyed and there's no guarantee the browsers actually clean up at that point. Everyone seems to think the best answer to create a renderer (and therefore a context) once, and always reuse it. We were doing that everywhere except our tests, so I've updated them, and even on Chrome they run markedly faster. We still create and (try to) destroy a few in tests that don't involve an element. I'm not sure if it's worth it to try and make those share a renderer too.

> ## [Switch environment generation from HalfFloat to RGBE](https://github.com/GoogleWebComponents/model-viewer/pull/670)
>
> Several major Samsung devices don't support halfFloat textures, causing our environment generation to fail. Switching to RGBE fixes this, but kills our current shader for blurring that environment, so I've created a new gaussianBlur filter for cubeMaps to replace it.

> ## [Add new fidelity tests](https://github.com/GoogleWebComponents/model-viewer/pull/672)
>
> I added two new fidelity tests as copies of the Khronos-MetalRoughSpheres, one with an HDR environment and one with an LDR environment (both spruit_sunrise, which has very high dynamic range, so good for catching bugs). Combined with the original that uses lightroom (similar to our generated environment), we now cover all of our lighting cases with a single, canonical model.
>
> Interestingly, this exposed a bug in filament, or possibly our usage thereof: Our Filament renders for the HDR and LDR environments are identical, and both appear to be LDR. This explains why we thought our renders looked "too yellow" when using spruit_sunrise_2k.hdr; in actuality Filament was not yellow enough, as it was lacking HDR.

> ## [Default stage light to zero](https://github.com/GoogleWebComponents/model-viewer/pull/674)
>
> In addition to changing the default, I removed the hemisphere light as I believe its only purpose was to hide an old LDR environment lighting bug that has since been fixed. I also removed all the arbitrary constants I could find relating to lighting.
>
> I also added a top-light to our environment generation to take the place of our default stage light, but we'll need to update our fidelity tests to properly compare.

src/three-components/TextureUtils.ts
```diff
@@ -168,118 +165,253 @@ export default class TextureUtils extends EventDispatcher {
...
+    tempTarget.texture.magFilter = NearestFilter;
+    tempTarget.texture.minFilter = NearestFilter;
+    tempTarget.texture.generateMipmaps = false;
 
-      if (environmentMap != null) {
-        (environmentMap as any).dispose();
-      }
+    var gammaOutput = this.renderer.gammaOutput;
+    var toneMapping = this.renderer.toneMapping;
+    var toneMappingExposure = this.renderer.toneMappingExposure;
```
cdata: In modern JS, there are few good reasons to use `var` ever. Use `const` or `let` instead (in this case probably `const`).

src/test/three-components/TextureUtils-spec.ts
elalish: No, we only lost a case for doing it with/without PMREM. We still have the test above that makes them both from a skybox URL, and we have this one that makes them from separate URLs.

elalish:
Regarding illumination, I did play with this a bit, and chose to match the lightroom renders instead, since that's our default. Honestly, it bothers me that we still need exposure = 0.8 just to get the skyboxes to match. We should be able to find an equation to match that exactly. Then we can continue with trying to more accurately match their model illumination, but given the Filament HDR bug I found, I don't want to work on that until it's fixed, because I'd guess their fix may affect LDR lighting as well.

Issue #514, opened 2019-04-29
Add CSS property for customizing progress bar height


yuinchien:
Allow customization for progress bar height to better accommodate wide range of model sizes displayed in the browser.

pushmatrix:
+1 to this

cdata:
Fixed by #602

PR #602, opened 2019-06-04
Added css property for progress bar height.

pushmatrix:
Implements #514 

Now you can set a pixel value for the height of the progress bar.


pushmatrix:
Yep, 5% works. 50% works too if you're into that sort of thing.

Screen Shot 2019-06-04 at 5 28 32 PM

cdata:
Looks great! Thanks @pushmatrix

PR #613, opened 2019-06-07
Basic screenshot API via toDataURL

### LINKED ISSUE #596: Screenshot API: Ability to export render as a data URL
There's many cases where it'd be useful to export an image of the model-viewer. Either to download it, or to access and manipulate it with javascript.

Normally I'd use `canvas.toDataURL()`, but since we don't have access to the canvas element in the Shadow DOM, it'd be great to have something like `modelviewer.toDataURL()`

cdata:
This change proposes that we forward invocations of `toDataURL` to our internal `<canvas>` element so that users can take advantage of one of `<canvas>`'s built-in screenshot capabilities without reaching through `<model-viewer>`'s shadow root.

The semantics of `toDataURL` are the same as the native `<canvas>` one. Screenshots will only display what is rendered to the canvas.

~Depends on #586~

Fixes #596 

cc @pushmatrix

### DIFF
--- a/index.html
+++ b/index.html
@@ -307,6 +307,15 @@
               current animation to the beginning, you should also set the
               currentTime property to 0.</p>
             </li>
+            <li>
+              <div>toDataURL(type, encoderOptions)</div>
+              <p>Returns a screenshot of the current model render in the format
+              specified by <i>type</i> (defaults to image/png). The screenshot is
+              encoded as a data URL string. In formats that support a sliding scale
+              of quality (such as image/jpeg and image/webp) you can also specify a
+              value for <i>encoderOptions</i> between 0 and 1 (<i>encoderOptions</i>
+              defaults to 0.92 otherwise).</p>
+            </li>
           </ul>
 
         </div>
--- a/src/test/model-viewer-base-spec.ts
+++ b/src/test/model-viewer-base-spec.ts
@@ -146,7 +146,7 @@ suite('ModelViewerElementBase', () => {
       });
     });
 
-    suite.only('capturing screenshots', () => {
+    suite('capturing screenshots', () => {
       let element: ModelViewerElementBase;
       setup(async () => {
         element = new ModelViewerElement();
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -303,6 +303,15 @@ <h3 class="grouping-title">Methods</h3>
               current animation to the beginning, you should also set the
               currentTime property to 0.</p>
             </li>
+            <li>
+              <div>toDataURL(type, encoderOptions)</div>
+              <p>Returns a screenshot of the current model render in the format
+              specified by <i>type</i> (defaults to image/png). The screenshot is
+              encoded as a data URL string. In formats that support a sliding scale
+              of quality (such as image/jpeg and image/webp) you can also specify a
+              value for <i>encoderOptions</i> between 0 and 1 (<i>encoderOptions</i>
```
elalish: If the only encoder option is quality, should we just call this quality?

src/test/model-viewer-base-spec.ts
```diff
@@ -146,6 +146,36 @@ suite('ModelViewerElementBase', () => {
...
+
+      teardown(() => {
+        if (element.parentNode != null) {
+          element.parentNode.removeChild(element);
+        }
+      });
+
+      suite('toDataURL', () => {
+        test('produces a URL-compatible string', () => {
+          expect(/^data\:image\//.test(element.toDataURL())).to.be.true;
```
elalish: This is a new magical incantation I've never seen; what does it do?

index.html
cdata: Yeah, it's kind of a bad name. It was chosen for us, though: https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL

src/test/model-viewer-base-spec.ts
cdata: JavaScript has regular expression literal syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions

The usage here can also be written as: `new RegExp('data\:image\/')`.

I'll assign it to a variable so that it is easier to parse.

elalish:
Thanks, LGTM!

pushmatrix:
Woohoo!  Thanks guys <3

PR #614, opened 2019-06-07
Allow skipping interpolation when changing camera

### LINKED ISSUE #454: Ability to disable camera interpolation when changing camera orbital position
As of #452 users can configure the current camera orbital position with an attribute. By default (and partially as a consequence of how camera control works generally), when the position is configured to a new value, the camera interpolates to the new position (only on subsequent changes, not for the initial value).

This is a sound default, but we should also offer a way for users to opt out of this interpolation if they wish.

elalish:
Fixes #454 

This is a pretty simple change to skip interpolation by calling the jumpCameraToGoal method, here exposed on the element.


index.html
```diff
@@ -294,6 +294,17 @@ <h3 class="grouping-title">Methods</h3>
...
               of user interaction or camera interpolation.</p>
             </li>
+            <li>
+              <div>getFieldOfView()</div>
+              <p>Returns the realtime camera vertical field of view in degrees. 
+              This value may differ from the value configured with the camera-orbit 
+              attribute on account of user interaction or camera interpolation.</p>
+            </li>
+            <li>
+              <div>jumpToGoal()</div>
```
cdata: I can make the referential leap with `getFieldOfView` to understand that it is referring to the camera (what else could FOV be referring to?). `jumpToGoal` is relatively less clear.

WDYT about `jumpCameraToGoal()`?

elalish:
Thanks for the explanation. However, this fix still doesn't solve the problem. In fact, it looks for all the world like jumpToGoal is not having any effect, but this is odd since it works properly in internal methods. I tried putting time passes before it and after it, but it made no difference. Internally, this function is equivalent to settleControls, so the only difference seems to be whether it is called from SmoothControls or from the element. Any more ideas?

index.html
```diff
@@ -294,6 +294,17 @@ <h3 class="grouping-title">Methods</h3>
...
+            <li>
+              <div>getFieldOfView()</div>
+              <p>Returns the realtime camera vertical field of view in degrees. 
+              This value may differ from the value configured with the field-of-view 
+              attribute on account of user interaction or camera interpolation.</p>
+            </li>
+            <li>
+              <div>jumpCameraToGoal()</div>
+              <p>Instantaneously updates the camera to the specified goal state 
+              instead of interpolating the motion over time.</p>
```
cdata: It might be worth padding the language here to set the expectations about what "instantaneous" means. Maybe rephrase to read along the lines of, "the camera will immediately move to its last configured goals _on the next update_."

PR #615, opened 2019-06-07
Document Samsung Browser requirements

### LINKED ISSUE #543: Samsung Internet has prefixed Fullscreen API and requires a polyfill

cdata:
- Samsung Internet has been added to our compatibility tables
 - Language related to Fullscreen API requirements has been updated

Fixes #543

### DIFF
--- a/POLYFILLS.md
+++ b/POLYFILLS.md
@@ -15,27 +15,27 @@ gaps for some of the newest features.
 
 These browser features are **required** for `<model-viewer>` to work correctly:
 
-Feature                    | Chrome | Canary | Safari 12 | Firefox 65 | Edge | IE 11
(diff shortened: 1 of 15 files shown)


cdata:
That would be nice. Unfortunately, Sauce Labs does not currently support testing Android browsers other than Chrome.

PR #619, opened 2019-06-10
Fix for procedural resizing

### LINKED ISSUE #618: Render is blank during procedural resizing
### Description

@mrdoob found this issue and made a patch to fix it, but the issue was never filed, so here it is. When style.width or height are changed, the viewer waits too long to redraw the frame, so if they change continuously nothing is ever drawn.

#### Live Demo
https://model-viewer-scratch.glitch.me/resize.html

elalish:
Fixes #618 

This is an adjustment to #555, which ensures that reframing also occurs before render. Proof it works: https://model-viewer-scratch.glitch.me/resize-fixed.html

### DIFF
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -388,6 +388,9 @@ export const ControlsMixin = (ModelViewerElement:
         [$onResize](event: any) {
           super[$onResize](event);
           this[$updateCamera]();
+          // Render a frame right away
+          this[$scene].isDirty = true;
+          this[$scene].renderer.render(performance.now());
         }
 
         [$onModelLoad](event: any) {
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -389,7 +389,6 @@ export const ControlsMixin = (ModelViewerElement:
           super[$onResize](event);
           this[$updateCamera]();
           // Render a frame right away
-          this[$scene].isDirty = true;
           this[$scene].renderer.render(performance.now());
         }
 
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -388,8 +388,6 @@ export const ControlsMixin = (ModelViewerElement:
         [$onResize](event: any) {
           super[$onResize](event);
           this[$updateCamera]();
-          // Render a frame right away
-          this[$scene].renderer.render(performance.now());
         }
 
         [$onModelLoad](event: any) {
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -320,7 +320,6 @@ export default class ModelViewerElementBase extends UpdatingElement {
 
   [$onResize](e: {width: number, height: number}) {
     this[$scene].setSize(e.width, e.height);
-    this[$needsRender]();
   }
 
   /**
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -320,6 +320,7 @@ export default class ModelViewerElementBase extends UpdatingElement {
 
   [$onResize](e: {width: number, height: number}) {
     this[$scene].setSize(e.width, e.height);
+    this[$needsRender]();
   }
 
   /**
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -278,9 +278,6 @@ export default class ModelViewerElementBase extends UpdatingElement {
     const intWidth = parseInt(width, 10);
     const intHeight = parseInt(height, 10);
 
-    this[$container].style.width = `${width}px`;
-    this[$container].style.height = `${height}px`;
-
     if (forceApply || (prevWidth !== intWidth || prevHeight !== intHeight)) {
       this[$onResize]({width: intWidth, height: intHeight});
     }
--- a/src/template.ts
+++ b/src/template.ts
@@ -39,6 +39,8 @@ template.innerHTML = `
 
     .container {
       position: relative;
+      width: 100%;
+      height: 100%;
     }
 
     canvas {
--- a/src/template.ts
+++ b/src/template.ts
@@ -38,9 +38,13 @@ template.innerHTML = `
     }
 
     .container {
+      display: flex;
       position: relative;
+      align-items: center;
+      justify-content: center;
       width: 100%;
       height: 100%;
+      overflow: hidden;
     }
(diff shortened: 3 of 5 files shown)


src/features/controls.ts
```diff
@@ -353,6 +353,9 @@ export const ControlsMixin = (ModelViewerElement:
         [$onResize](event: any) {
           super[$onResize](event);
           this[$updateCamera]();
+          // Render a frame right away
+          this[$scene].isDirty = true;
+          this[$scene].renderer.render(performance.now());
```
cdata: If I didn't apply the `ControlsMixin` to the base element, what would happen when we resize?

src/model-viewer-base.ts
```diff
@@ -320,7 +320,6 @@ export default class ModelViewerElementBase extends UpdatingElement {
 
   [$onResize](e: {width: number, height: number}) {
     this[$scene].setSize(e.width, e.height);
-    this[$needsRender]();
```
cdata: Was this taken out because some other part of the base implementation invokes the re-render?

src/features/controls.ts
elalish: Good point, I wasn't thinking about using them independently.

src/model-viewer-base.ts
elalish: I was following the prior PR's method here, but you're right, there's a better way.

src/features/controls.ts
cdata: Is it possible that we don't even need the call to `renderer.render` here? It seems like the parent implementation of `[$onResize]` will cause us to re-render eventually..

elalish: The eventually part was exactly the problem; @mrdoob had a better explanation in #555.

cdata: Yes, I see. It seems like we should be invoking this in the base class (not in the mixin) because the problem occurs regardless of whether `ControlsMixin` is applied. Is that correct?

elalish: That's how it was in #555, but the problem there is that it gets redrawn without being reframed, since that happens in ControlsMixin.

cdata: It seems like we might be trading one timing problem for another (unintentionally). `[$needsRender]()` causes `isDirty` to be set. Then, you call `render` directly in the mixin implementation causing an immediate render (and `isDirty` gets reset). So, on the next frame, the scene may not be re-rendered as intended.

Perhaps this timing issue could be solved by invoking `[$udpateCamera]()` before calling the super implementation of this method? Then we only need to render synchronously in one place (the base implementation).

elalish: Hmm, I just tried this, only to realize that [$updateCamera] is dependent on the base class setting the width and height first, so now we're trying to call the base class both before and after. Should I just setSize twice?

cdata:
This was probably well known by some folks, but TIL that `<canvas>` [clears itself when you change it's (non-CSS) `width` or `height`](https://aquatic-hourglass.glitch.me/canvas-resize.html). Relatedly:

This change does appear to fix the symptoms of the problem, but I tweaked the test case above, and a profile demonstrates that we now will render _twice_ in a single frame for a typical resize scenario:

image
[Live version used for profiling](https://aquatic-hourglass.glitch.me/double-render.html)

It seems that the fundamental problem here relates to the timing of `ResizeObserver` callbacks compared to `requestAnimationFrame`. `ResizeObserver` apparently invokes its callback _after_ the `requestAnimationFrame` where we render the frame. This means that when the canvas `width`/`height` changes due to `ResizeObserver` callback, it has the effect of clearing the canvas immediately after we have rendered during the concurrent `requestAnimationFrame` callback.

All that to say:

I would like us to try an approach where the `<canvas>` element is not resized until we are about to invoke `drawImage` on it during render.

This will have the unfortunate trade-off of causing the `<canvas>` resize to happen a frame late, but that seems preferable to both the current scenario (`<canvas>` is cleared for a frame after resize) and the scenario where we render _twice_ during resize (double work during a performance sensitive time).

It might end up being really jarring to see the resize happen a frame late, but let's try it and see.

cdata:
I've been adapting the change to only set the non-CSS `width`/`height` on the `<canvas>` at render time. This has the predicted effect of a 1-frame-late resize.

Here are some examples of resize behavior **on `master`** today:

Orientation on docs | Re-layout on docs | Resize on rAF | Resize on interval |
--------------------| ------------------- | ------------- | -----------------  |
master-docs-swap | master-responsive | master-raf-resize | master-interval-resize |

And here are the same examples with **this change**:

Orientation on docs | Re-layout on docs | Resize on rAF | Resize on interval |
------------------- | ------------------ | -------------- | ------------------ |
pr-docs-swap | pr-responsive | pr-raf-resize | pr-interval-resize |

As you can see from the "Resize on interval" example, there is a one-frame lag between setting the CSS `width`/`height` and re-rendering the model after changing the non-CSS `width`/`height`. Overall, I would say that the results are acceptable, because 1-frame lag is not especially obvious most of the time. The lag is the trade-off of not rendering twice in a performance-sensitive frame.

I tested this out in a few browsers (Safari and Firefox). Safari was comparable to Chrome, but Firefox yielded obviously poor results for the "Orientation on docs" test:

Orientation on docs (Firefox) |
----------------------------- |
 |

So, at this point, I'm a little torn. It's a bad idea to render twice on a resize. It also seems lame that Firefox users would have to have a possibly degraded visual experience when resizing.

mrdoob:
> So, at this point, I'm a little torn. It's a bad idea to render twice on a resize. It also seems lame that Firefox users would have to have a possibly degraded visual experience when resizing.

After years of dealing with 1-frame lag when resizing windows in various OSes, I think having a double render in some frames is worth it 🤓

cdata:
👍 thanks for the feedback @mrdoob

I have experimented with two different approaches that skip the extra
render, but both have trade-offs for the presentation of the model during
the frame where resize happens. The current approach in this branch avoids
skewing the model, but would cause the model to be letter-boxed during the
resize frame in many configurations.

Let's take a double-render approach and see if it works okay in practice!
I'll update this branch to reflect that.

cdata:
I'm going to close this out in favor of #705 because the branch has potentially useful work for the future.

Issue #500, opened 2019-04-22
User configurable interaction policy


srussking:
The new interaction "allow-when-focused" policy which allows for scrolling over the model is cool, however, for my purposes it is a difficult to convey to the user, also it seems to get reset every time the user "fullscreens" the app or contracts (possibly this is due to our putting it into an iframe, unknown...).

I am wondering if it is possible to have that be an attribute that is passed in. i.e. <model-viewer ... interaction-policy="always-allow" />

Not familiar enough with the code to submit a pr for that but seems like it should be easy enough to add and be inline with the current strategy.

Just a suggestion though.

cdata:
Thanks for the feedback!

Yes, we want to make most of the controls constraints configurable as public API (a similar issue in this area is #458 ). I would say that the main thing we need to decide is the shape of the API.

`interaction-policy` is a reasonable name for an attribute that controls this, although I would love to hear ideas from other folks in the community if they have any.

For your specific use case @srussking it might help mitigate the problem if you manually re-focus the `<model-viewer>` when fullscreen state changes e.g., by calling `yourModelViewer.focus()`.

But in general I agree that we need to support changing the interaction policy because the default is best for non-fullscreen use cases.

srussking:
Cool, wanted to update this in case anyone else has this issue. Moving the fullscreen div inside the model-viewer made this issue less problematic, such that now fullscreen and back to normal keeps the focus on the model viewer...

Cheers

cdata:
I also want to call out that this is related to https://github.com/GoogleWebComponents/model-viewer/issues/479 which would enable a less strict heuristic for enabling interaction that might serve as a middle-ground between "always" and "only when focused."

cdata:
This has been fixed in #621

Issue #479, opened 2019-04-08
Implement better heuristic to enable user interactions than basic focus


cdata:
As of #288 , we began blocking user interaction unless a `<model-viewer>` has focus. This enables users to interact with the page in more natural ways (such as scrolling while passing over `<model-viewer>`) at the cost of requiring explicit interaction to focus `<model-viewer>` before the controls would work.

@mrdoob immediately pointed out that this was not very intuitive when interacting, especially on mobile. @yuinchien also pointed out the same, and has prototyped some potential alternative heuristics to decide if `<model-viewer>` should be interactive.

This issue covers discussion and work necessary to enable improved detection of explicit user interactions with `<model-viewer>` so that it feels intuitive without disrupting other activities such as scrolling the page.

yuinchien:
I played with horizontal gesture detection on <model-viewer>, as @cdata pointed out, most of the mobile page scroll is vertical, allowing auto-focus to happen when detecting user interaction feels a lot more accessible. 

```
const modelViewer = $('#demo-1');
let startX;
let startY;
let isFocused = false;
const THRESHOLD_X = 30;

function onTouchStart(e) {
  let touchobj = e.changedTouches[0];
  startX = touchobj.pageX;
  startY = touchobj.pageY;
}

function onTouchMove(e) {
  let touchobj = e.changedTouches[0];
  if( Math.abs(touchobj.pageX-startX)>THRESHOLD_X && !isFocused) {
      modelViewer.focus();
  }
}

function onFocus() {
  isFocused = true;
}
function onBlur() {
  isFocused = false;
}

modelViewer.addEventListener("focus", onFocus);
modelViewer.addEventListener("blur", onBlur);
modelViewer.addEventListener("touchstart", onTouchStart, false);
modelViewer.addEventListener("touchmove", onTouchMove, false);
```

elalish:
On a related note, we also have the inverse problem: when the <model-viewer> element covers the whole screen on mobile (either because it's big, or because the user zoomed the page on it before it got focus), then once it gets focus it is impossible to scroll the page, because all touch events are routed to rotating and zooming the model and there is no way to defocus.

elalish:
[...]

PR #621, opened 2019-06-11
Change default interaction policy to always-allow

pushmatrix:
From all of our user testing it seems clear that requiring focus on mobile to interact is not intuitive. @mrdoob said the same thing over at https://github.com/GoogleWebComponents/model-viewer/issues/479.

I do think we need a better heuristic for mobile (https://github.com/GoogleWebComponents/model-viewer/issues/479), and we should have the interaction policy be configurable (https://github.com/GoogleWebComponents/model-viewer/issues/500), but a sensible default should favor making it easy to interact with.

Thoughts?

### DIFF
--- a/index.html
+++ b/index.html
@@ -183,6 +183,14 @@
               any value between 20 and 45 degrees. Defaults to "45deg".</p>
             </li>
             <li>
+              <div>interaction-policy</div>
+              <p>Allows you to change whether the viewer requires focus before interacting with it.
+              If set to "allow-when-focused", the user must focus on the viewer (click / tap)
+              before being able to control it. If set to "always-allow", the user can 
+              control it even if the viewer is not the focused element on the page. 
+              Defaults to "always-allow".</p>
+            </li>
+            <li>
               <div>interaction-prompt</div>
               <p>Allows you to change the conditions under which the visual and
               audible interaction prompt will display. Allowed values are "auto"
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -32,11 +32,18 @@ export interface SphericalPosition {
 }
 
 export type InteractionPromptStrategy = 'auto'|'when-focused';
+export type InteractionPolicy = 'always-allow'|'allow-when-focused';
 
 const InteractionPromptStrategy:
     {[index: string]: InteractionPromptStrategy} = {
       AUTO: 'auto',
       WHEN_FOCUSED: 'when-focused'
+    };
+
+const InteractionPolicy:
+    {[index: string]: InteractionPolicy} = {
+      ALWAYS: 'always-allow',
+      WHEN_FOCUSED: 'allow-when-focused'
     };
 
 export const DEFAULT_CAMERA_ORBIT = '0deg 75deg auto';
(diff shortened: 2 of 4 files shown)


cdata:
This looks good. I think I would like to see this paired with a way for folks to opt-in to requiring focus until such time as we implement #479 . LMK if you have time to add an attribute for that, otherwise I can layer it on!

pushmatrix:
I can likely handle it, but it will be next week.

pushmatrix:
@cdata I added in the attribute **interaction-policy** so users can opt-in to requiring focus. This name came from the internal attribute that is currently being used.

- Is interaction-policy a good name for an attribute? Or should it be something like **require-focus** and have that just be a true/false? 

- SmoothControls.ts and Controls.ts now both define the same InteractionPolicy. Doesn't seem DRY to me. Thoughts?

cdata:
`interaction-policy` is a good name. I think attributes with explicit values are better than boolean attributes without values in a lot of cases (for the same reason that boolean method parameters are usually not ideal).

Agreed, there is a bit of repetition here. I think in the long-run, `interaction-policy` will be evolved or deprecated as we land on some solid built-in heuristics for interaction detection e.g. #479

src/features/controls.ts
```diff
@@ -32,13 +32,20 @@ export interface SphericalPosition {
...
 
 const InteractionPromptStrategy:
     {[index: string]: InteractionPromptStrategy} = {
       AUTO: 'auto',
       WHEN_FOCUSED: 'when-focused'
     };
 
+const InteractionPolicy:
+    {[index: string]: InteractionPolicy} = {
+      ALWAYS: 'always-allow',
```
cdata: Change `ALWAYS` to `ALWAYS_ALLOW`

src/features/controls.ts
```diff
@@ -207,6 +219,11 @@ export const ControlsMixin = (ModelViewerElement:
             }
           }
 
+          if (changedProperties.has('interactionPolicy')) {
+            const interactionPolicy = this.interactionPolicy;
+            controls.applyOptions({interactionPolicy});
```
cdata: 👍

src/three-components/SmoothControls.ts
```diff
@@ -54,7 +54,7 @@ export const DEFAULT_OPTIONS = Object.freeze<SmoothControlsOptions>({
   minimumFov: 20,
   maximumFov: 45,
   eventHandlingBehavior: 'prevent-all',
-  interactionPolicy: 'allow-when-focused'
+  interactionPolicy: 'always-allow'
```
cdata: 👍

index.html
```diff
@@ -182,6 +182,14 @@ <h3 class="grouping-title">Attributes</h3>
               units in both degrees and radians (e.g., "30deg" or "0.5rad"). Accepts
               any value between 20 and 45 degrees. Defaults to "45deg".</p>
             </li>
+            <li>
+              <div>interaction-policy</div>
```
cdata: 💯

pushmatrix:
Changes made :)

Issue #524, opened 2019-05-02
Sometimes modelIsVisible is wrong, causes autoRotate to break


cdata:
We have observed under some circumstances that `modelIsVisible` is incorrect, which in turn can cause `autoRotate` to break (because rotation does not occur until the model is visible as of #471 ).

cdata:
Moving this to v0.6.0. After discussion with @elalish there is some awkwardness and subtlety to the concept of `modelIsVisible` and we probably need to re-evaluate it on a more fundamental level.

pushmatrix:
I've seen behaviour where the model is not visible if the viewer was initially hidden and then made visible. This is a common design for online stores where you have one big product photo that you can switch by pressing thumbnails in a gallery.

However sometimes the model viewer shows blank when I change thumbnails (hide/unhide viewers). 

![gif](https://user-images.githubusercontent.com/190692/60614955-040e0f00-9d9c-11e9-8de9-c3c71feacbeb.gif)

@cdata suggested this may be related to `modelIsVisible`.  

Steps to reproduce:
- Go here https://third-dimensions.myshopify.com/products/model
- Press the second thumbnail
- The model viewer doesnt display anything.

Sometimes you can force refresh and click the second thumbnail during page load, and the viewer will show.

pushmatrix:
Is this still slated for 0.6?

cdata:
Yes!

pushmatrix:
I tested out my above example some more, and found that in Safari it throws this error:
<img width="1295" alt="Screen Shot 2019-09-18 at 2 11 25 PM" src="https://user-images.githubusercontent.com/190692/65174100-37b51880-da1e-11e9-8374-1924fb9579be.png">

This is not thrown in Chrome. However in Chrome it still doesnt show the second viewer when you try to unhid it.

Here's another barebones example of the issue: https://bedecked-profit.glitch.me
Press the *Show viewer 2* button at the bottom and you'll notice that nothing loads.

View it in Safari to see the error.

PR #622, opened 2019-06-11
Make interaction prompt independent of focus

### LINKED ISSUE #529: Configuration to raise interaction prompt automatically without requiring focus
There are some scenarios where it is desirable to raise the interaction prompt automatically, without requiring that the user focus on the element first. Unfortunately, it is not possible to enable this use case by invoking `.focus()` on `<model-viewer>` because not all browsers allow focusing elements prior to user interaction.

Let's add a configuration that causes the interaction prompt to be raised automatically (after the configured `interaction-prompt-threshold`), regardless of focus state.

elalish:
Fixes #529 

Now the prompt is activated by the interactionPromptThreshold seconds after modelIsVisible, and canceled by a user-interaction, never to be shown again. This also allowed the logic to be simplified quite a bit. I think I may be hitting a small issue with #524, as on a page with several models, like lighting-and-environment.html, once the prompt is up for the first model, it's also up for all the other models, even though they haven't scrolled into view yet.

### DIFF
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -53,19 +53,15 @@ const $updateCamera = Symbol('updateCamera');
 const $updateCameraOrbit = Symbol('updateCameraOrbit');
 const $updateFieldOfView = Symbol('updateFieldOfView');
 
-const $blurHandler = Symbol('blurHandler');
 const $focusHandler = Symbol('focusHandler');
 const $changeHandler = Symbol('changeHandler');
 const $promptTransitionendHandler = Symbol('promptTransitionendHandler');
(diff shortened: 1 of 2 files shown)


cdata:
This mostly looks good. I'm going to defer a full review until it comes time to land as part of #623 .

cdata:
While we have the `interaction-prompt` attribute, it seems like there is a low-hanging first step here to change it's default to `"auto"`. Let's make a change to do that first and move #623 into v0.7.0.

elalish:
Closing in favor of #633.

Issue #625, opened 2019-06-11
Implement tests for quick-look-browsers


cdata:
These tests are currently stubbed out and skipped. Implementation is blocked by #624 .

Issue #624, opened 2019-06-11
Ability to mock constants in tests


cdata:
We have a module that exports a bunch of constants that test various qualities of the browser one time so that consumer modules don't have to repeat such checks. Unfortunately, it is currently impossible change the values of these constants at runtime for the purposes of mocking in tests. We need to overhaul how we share these constants to make this possible.

PR #626, opened 2019-06-11
Opt-in to Quick Look for iOS Chrome

### LINKED ISSUE #534: Enable users to opt-in to Quick Look integration in iOS Chrome
Unfortunately, iOS Chrome is more strict than iOS Safari when it comes to launching Quick Look for previewing .usdz files. iOS Safari will launch Quick Look for any .usdz file regardless of `Content-Type`. However, iOS Chrome strictly requires that the appropriate `Content-Type` be set (`model/vnd.usdz+zip`).

Worse: iOS Chrome has all of the detectable features that iOS Safari has with regards to support for launching Quick Look, so it is very difficult to infer Chrome's additional strictness.

Many users cannot change the `Content-Type` response headers sent by their servers. So, we should ensure that folks can opt-out (or opt-in) to the Quick Look integration in iOS Chrome.

https://webkit.org/blog/8421/viewing-augmented-reality-assets-in-safari-for-ios/

cdata:
- It is now possible to opt-in or -out of Quick Look integration for a set of known browsers
 - Currently only "chrome" and "safari" are supported values
 - 🚨**BREAKING CHANGE**: iOS Chrome is now opt-in by default (previously it was enabled by default, with no way to opt-out)
 - To achieve the behavior before this change, use the following configuration: `<model-viewer quick-look-browsers="safari chrome"></model-viewer>`
 - Other browsers are not included because we don't test for them and don't currently know if they even integrate with Quick Look (#627)
 - Currently tests targeting specific browsers are prohibitively difficult to write due to how we set constants (#624, #625)

Fixes #534

### DIFF
--- a/src/features/ar.ts
+++ b/src/features/ar.ts
@@ -15,7 +15,8 @@
 
 import {property} from 'lit-element';
 
-import {IS_ANDROID, IS_AR_QUICKLOOK_CANDIDATE, IS_IOS, IS_WEBXR_AR_CANDIDATE} from '../constants.js';
+import {IS_ANDROID, IS_AR_QUICKLOOK_CANDIDATE, IS_IOS, IS_IOS_CHROME, IS_IOS_SAFARI, IS_WEBXR_AR_CANDIDATE} from '../constants.js';
+import {enumerationDeserializer} from '../conversions.js';
 import ModelViewerElementBase, {$container, $renderer, $scene} from '../model-viewer-base.js';
 import {Constructor, deserializeUrl} from '../utilities.js';
 
@@ -80,7 +81,12 @@ export const openARViewer = (() => {
   };
 })();
 
-type ARMode = 'quick-look'|'ar-viewer'|'unstable-webxr'|'none';
+export type QuickLookBrowser = 'safari'|'chrome';
+
+const deserializeQuickLookBrowsers =
+    enumerationDeserializer<QuickLookBrowser>(['safari', 'chrome']);
+
+export type ARMode = 'quick-look'|'ar-viewer'|'unstable-webxr'|'none';
 
 const ARMode: {[index: string]: ARMode} = {
   QUICK_LOOK: 'quick-look',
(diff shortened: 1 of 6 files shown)


src/features/ar.ts
```diff
@@ -290,6 +317,16 @@ configuration or device capabilities');
           event.preventDefault();
           this.activateAR();
         }
+
+        get[$isAllowedQuickLookBrowser](): boolean {
```
elalish: Nit, but I might call this $enableQuickLook, or $canLaunchQuickLook.

src/test/features/ar-spec.ts
```diff
@@ -41,14 +41,21 @@ suite('ModelViewerElementBase with ARMixin', () => {
...
+
     suite('with unstable-webxr', () => {
-      let element: ModelViewerElementBase;
+      let element: ModelViewerElementBase&ARInterface;
 
       setup(async () => {
         element = new ModelViewerElement();
         document.body.appendChild(element);
 
-        (element as any).unstableWebxr = true;
```
elalish: :+1:

src/features/ar.ts
cdata: Good suggestions, will fix 👍

cdata:
TIL IE11 does not support constructing a `Set` directly from an iterable 😢 `new Set([1, 2, 3])` will return a `Set` with zero items...

cdata:
Looks like the full test suite passed, including IE11

SokWonJeremyKim:
Hello.

I tried two tweaks found in this repo:
- adding compatibility enhancing libraries
- adding "quick-look-browsers" attribute as suggested here

But I still can't seem to start iOS quick look from chrome.

I can see the ar icon in the lower right corner but clicking it just downloads the usdz file.

Here's my model-viewer part
```
<model-viewer
    ar camera-controls
    quick-look-browsers="safari chrome"
    background-color="#BBB"
    alt={product.alt}
    magic-leap unstable-webxr
    src={product.src_gltf}
    ios-src={product.src_usdz}
    poster={product.thumbnail}
    reveal="interaction">
</model-viewer>
```

Thank you

cdata:
@SokWonJeremyKim Per the original issue (https://github.com/GoogleWebComponents/model-viewer/issues/534), the reason we added the requirement to opt-in to iOS Chrome via the `quick-look-browsers` attribute is because Chrome currently requires that you set the appropriate `Content-Type` header for the `.usdz` file in order to launch Quick Look.

If you do not set the appropriate `Content-Type` header, Chrome will just download the file instead.

The correct `Content-Type` header to set for `.usdz` is: `model/vnd.usdz+zip`

SokWonJeremyKim:
Thank you so much!!!!! That was super helpful!!

PR #628, opened 2019-06-11
Ensure model visibility works when element not renderable

cdata:
Previously, we keyed off of CSS transition events to know when to switch model visibility state internally. However, when an element is in a non-renderable state (such as when its layout is collapsed or when its computed CSS `display` property is `none`), transitions don't occur and so the associated events do not fire.

This change adds a check for "renderable" element state. The element is considered renderable if its layout dimensions are greater than 0. If it is renderable, we proceed as before. If it isn't, we perform the necessary side-effects immediately instead of waiting for a transition to complete.

Fixes https://github.com/GoogleWebComponents/model-viewer/issues/524

### DIFF
--- a/src/features/loading.ts
+++ b/src/features/loading.ts
@@ -15,7 +15,7 @@
 
 import {property} from 'lit-element';
 
-import ModelViewerElementBase, {$ariaLabel, $canvas, $progressTracker, $updateSource} from '../model-viewer-base.js';
+import ModelViewerElementBase, {$ariaLabel, $canvas, $onResize, $progressTracker, $updateSource} from '../model-viewer-base.js';
 import {CachingGLTFLoader} from '../three-components/CachingGLTFLoader.js';
 import {Constructor, debounce, deserializeUrl, throttle} from '../utilities.js';
 
@@ -68,8 +68,10 @@ const $shouldRevealModel = Symbol('shouldRevealModel');
 const $showPoster = Symbol('showPoster');
 const $hidePoster = Symbol('hidePoster');
 const $modelIsVisible = Symbol('modelIsVisible');
+const $changeModelVisibilityState = Symbol('changeModelVisibilityState');
 const $preloadAttempted = Symbol('preloadAttempted');
 const $sourceUpdated = Symbol('sourceUpdated');
+const $elementIsRenderable = Symbol('elementIsRenderable');
 
 const $updateLoadingAndVisibility = Symbol('updateLoadingAndVisibility');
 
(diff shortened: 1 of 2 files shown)


src/test/features/loading-spec.ts
```diff
@@ -178,6 +178,34 @@ suite('ModelViewerElementBase with LoadingMixin', () => {
...
+        });
+
+        suite('modelIsVisible', () => {
+          test(
+              'is true when the element eventually becomes renderable',
+              async () => {
+                const modelBecomesVisible =
+                    waitForEvent(element, 'model-visibility');
+                element.src = ASTRONAUT_GLB_PATH;
+                await modelBecomesVisible;
```
elalish: Doesn't this test include the previous one?

src/features/loading.ts
```diff
@@ -382,30 +407,39 @@ export const LoadingMixin = (ModelViewerElement:
...
+                    this[$canvas].focus();
+                  }
+
+                  finishHidingPoster();
+                });
+              }, {once: true});
+            } else {
+              // For non-renderable cases, we don't need to bother focusing the
+              // canvas because it cannot be focused anyway
+              this[$changeModelVisibilityState](true);
```
elalish: What does it mean for an element to be not renderable, but for its model to be visible?

cdata:
It looks like every browser that relies on the `ResizeObserver` polyfill has problems with this approach. Specifically, the polyfill appears to be unable to detect changes due to style property manipulation, so we don't get a resize update.

This has implications beyond the scope of this bugfix, but for now I will try to address the polyfill case for just this bug in isolation.

cdata:
(Side-note / silver lining: Edge, Firefox and Safari are all adding `ResizeObserver` in their next major releases https://caniuse.com/#search=resize).

src/test/features/loading-spec.ts
cdata: It depends on the same circumstances as the previous one. I tend to err on the side of overlap if it seems like there are two distinct behaviors that should be ensured. For example, I wouldn't remove a test that checks for the `'load'` event just because 50 other tests depend on observing that event as part of their setup.

src/features/loading.ts
cdata: That's a good question. I don't think we have been very crisp on this. Perhaps the element being in a non-renderable state should always imply that the model is not visible.

OTOH it seems like a valid case for someone to want to wait until the model would actually be rendered to a canvas before changing `display: none => block`.

elalish: Personally, as I alluded to elsewhere, I would not expect the model to be visible unless it's being displayed to the user. We don't check that at all here.

cdata: Yeah, for sure. I think the intention and the name don't really line up well here. I think the intention is something along the lines of, "if the element was rendered, the model would be displayed." The boolean is intended to capture the difference between the canvas been obscured by the poster vs. the poster being hidden.

cdata:
Closing this as it has become quite dated. Part of what this change endeavored to fix is resolved by #755 , and the remainder of the design work should be addressed with #733 .

PR #630, opened 2019-06-11
Update fidelity tests

### LINKED ISSUE #521: Change fidelity tests to use new default environment lighting
When we added Khronos models to our fidelity tests in #360 , we made a decision to use an environment map when staging most of them because our default environment lighting was pretty bad and Filament had no default IBL. As of #472 we have much improved default lighting conditions in `<model-viewer>`. We should change the environment lighting used in our tests to be based on the new default.

elalish:
Fixes #521, #527 

As I was comparing the fidelity results, I realized there was a subtle bug about the framing between model-viewer and filament, and removing room_padding fixed it, so that's bundled here. I also tweaked the exposure to get the backgrounds to look more similar and highlight the actual rendering differences. I also tweaked one of our scripts to make it easier to load one test at a time instead of having to wait for all of them.


elalish:
I fixed the patch and built filament, but their tangent rendering bug is still manifest. It appears this is because our gltf_renderer.cpp uses MeshAssimp to import GLTF models, but the fix is in their gltfio loader instead. Updating that seems like it probably deserves its own PR, and it's probably not a priority for 1.0?

PR #633, opened 2019-06-12
Configuration to raise prompt automatically

### LINKED ISSUE #529: Configuration to raise interaction prompt automatically without requiring focus

cdata:
- A new `interaction-prompt` attribute is introduced, with two supported values:
   1. **`when-focused`** (default): this is the current behavior, where the threshold is not measured until the element is first focused
   2. **`auto`**:  This enables the behavior described in #529 , and causes the threshold time to be measured as soon as the model is revealed
 - In both cases, the prompt is dismissed by user interaction the same way
 - This change precedes #622 , but in that change `auto` will become the default behavior (and we will possibly also remove `when-focused`)

Fixes #529

### DIFF
--- a/index.html
+++ b/index.html
@@ -174,6 +174,16 @@
               for in-depth details related to this lighting mode.</p>
             </li>
             <li>
+              <div>interaction-prompt</div>
+              <p>Allows you to change the conditions under which the visual and
+              audible interaction prompt will display. Allowed values are "auto"
+              and "when-focused". If set to "auto", the interaction prompt will
+              be displayed as soon as the interaction-prompt-threshold (see below)
+              time has elapsed (after the model is revealed). If set to "when-focused",
+              the interaction prompt will only be activated if the element has
+              first received focus. Defaults to "when-focused".</p>
+            </li>
+            <li>
               <div>interaction-prompt-threshold</div>
               <p>When camera-controls are enabled, &lt;model-viewer&gt; will
               prompt the user visually (and audibly, for screen readers) to
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -27,8 +27,15 @@ export interface SphericalPosition {
   radius: number;
 }
 
+export type InteractionPromptStrategy = 'auto'|'when-focused';
 
-const DEFAULT_CAMERA_ORBIT = '0deg 75deg auto';
+const InteractionPromptStrategy:
+    {[index: string]: InteractionPromptStrategy} = {
+      AUTO: 'auto',
+      WHEN_FOCUSED: 'when-focused'
+    };
+
+export const DEFAULT_CAMERA_ORBIT = '0deg 75deg auto';
 const DEFAULT_FIELD_OF_VIEW = '45deg';
 
 const HALF_PI = Math.PI / 2.0;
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -173,6 +173,16 @@ <h3 class="grouping-title">Attributes</h3>
               the built-in scene lights. Refer to <a href="https://wiki.jmonkeyengine.org/jme3/advanced/pbr_part3.html" target="_blank">this article</a>
               for in-depth details related to this lighting mode.</p>
             </li>
+            <li>
+              <div>interaction-prompt</div>
+              <p>Allows you to change the conditions under which the visual and
+              audible interaction prompt will display. Allowed values are "auto"
+              and "when-focused". If set to "auto", the interaction prompt will
+              be displayed as soon as the interaction-prompt-threshold (see below)
+              time has elapsed (after the model is revealed). If set to "when-focused",
```
elalish: I don't think the prompt should come up until the model is visible on the page, hence my comment about modelIsVisible not meaning much. Try this with auto on our lighting examples page. You'll notice as you scroll that the prompt has no delay except on the first one.

cdata: Yeah, acknowledged. I think we should work on this as a separate improvement when we square away what we mean by `modelIsVisible`.

cdata: (as part of #524 , that is)

cdata:
Thanks for the feedback @elalish . The original feedback that provoked this change was from Visible Body. They typically have one model per page, and will most likely set `interaction-prompt-threshold` to 0 (see [this typical usage](https://www.visiblebody.com/learn/muscular/muscle-movements?hs_preview=XsY0_-Nv-2410128992) as an example).

The change you are suggesting depends on us sorting out `modelIsVisible`, so I'm going to land this as-is for the sake of the v0.4.0 release and we will sort everything out in v0.6.0.

pinkp:
I've set mine up as follows:

   interaction-prompt="auto"
   interaction-prompt-threshold="0"

But the hand prompt doesn't appear until you focus. How can I get it to appear as soon as the page loads like your typical example usage. ? Thanks

cdata:
@pinkp I just tried it out locally and it worked. Could you file an issue with a reproducible test case?

pinkp:
[This site](http://www.artistvillage.co.uk) it only works when you focus on it.

cdata:
@pinkp you need to update to at least version v0.4.0 to use this feature. How are you incorporating `model-viewer.js` into your site today?

pinkp:
Yes! I was using a downloaded version linked to on the server, now I've updated to the one in the docs and it works perfectly! Thanks I knew it would be something simple.



<script type="module"
    src="https://unpkg.com/@google/model-viewer/dist/model-viewer.js">
</script>


<script nomodule
    src="https://unpkg.com/@google/model-viewer/dist/model-viewer-legacy.js">
</script>

PR #636, opened 2019-06-13
exact integration of SmoothControls

### LINKED ISSUE #634: Extreme lagginess / jerky movement in desktop Safari
### Description
On my 15" retina mbp in Safari, when I view a model-viewer that takes up a lot of the screen, the movement is suddenly very laggy and the model jerks around.

janky

It gets smoother the smaller the viewport is.

Repro:
- Safari 12.1.1 (14607.2.6.11)
- Visit https://googlewebcomponents.github.io/model-viewer/
- Make the window as big as possible
- Try rotating the astronaut

It is even more noticeable if you are in the scaled display: 

Screen Shot 2019-06-12 at 9 55 43 PM

I have noticed performance on iOS is not as good as on Android. My XS Max is somewhat choppy when I rotate models, but on a pixel 2 it's buttery smooth.

### Browser Affected

- [ ] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [x] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [x] MacOS (10.14.5)
- [ ] Windows

Screen Shot 2019-06-12 at 11 07 40 PM

### Versions
- model-viewer: v0.3.1
- three.js: r104

elalish:
Fixes #634 

Replaced our linear discrete integration of our second order system of SmoothControls with an exact solution, allowing large time steps to be calculated accurately.

### DIFF
--- a/src/three-components/SmoothControls.ts
+++ b/src/three-components/SmoothControls.ts
@@ -154,7 +154,7 @@ export class Damper {
   update(
       x: number, xGoal: number, timeStepMilliseconds: number,
       xNormalization: number): number {
-    if (!x) {
+    if (!isFinite(x)) {
       return xGoal;
     }
     if (timeStepMilliseconds < 0) {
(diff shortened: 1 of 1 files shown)


src/three-components/SmoothControls.ts
```diff
@@ -154,27 +154,30 @@ export class Damper {
...
-        2 * NATURAL_FREQUENCY * this[$velocity];
-    this[$velocity] += acceleration * timeStepMilliseconds;
-    if (Math.abs(this[$velocity]) < NIL_SPEED * xNormalization &&
-        acceleration * (xGoal - x) <= 0) {
+    // Exact solution to a critically damped second-order system, where:
+    // acceleration = NATURAL_FREQUENCY * NATURAL_FREQUENCY * (xGoal - x) -
+    // 2 * NATURAL_FREQUENCY * this[$velocity];
+    const c1 = (x - xGoal);
+    const c2 = this[$velocity] + NATURAL_FREQUENCY * c1;
+    const c3 = c1 + timeStepMilliseconds * c2;
```
cdata: [Please use descriptive names for these terms](https://github.com/GoogleWebComponents/model-viewer/wiki/Code-conventions#naming)

src/three-components/SmoothControls.ts
```diff
@@ -154,27 +154,30 @@ export class Damper {
   update(
       x: number, xGoal: number, timeStepMilliseconds: number,
       xNormalization: number): number {
-    if (timeStepMilliseconds >= DECAY_MILLISECONDS) {
-      // This clamps at the point where a large time step would cause the
-      // discrete step to overshoot.
-      this[$velocity] = 0;
+    if (!x) {
```
cdata: `!x` will be true if `x` is `0`. In most cases, you should use `x == null` for cases where you expect `x` to be unset. If `x` is ever `NaN`, it's usually a bug / code hygiene issue.

elalish: I see, the error was that x was undefined, which is different than null, right? I'll see if I can track that down. Is there a good way to check that x is just a valid number? Do we have an isFinite() function?

src/three-components/SmoothControls.ts
elalish: These don't have names, they are only saved here so we don't have to do the calculations multiple times. I can call c1 deltaX, but c2 and c3 are physically meaningless as far as I can tell.

elalish: I renamed them so that at least you can tell their units.

src/three-components/SmoothControls.ts
cdata: `==` and `===` in JavaScript are different, and the nature of `==` means `x == null` will catch both `null` and `undefined`

this-is-fine.jpg

PR #644, opened 2019-06-16
removed link to default.html which was a non-existent file

Frazer:


Frazer:
I signed it!

cdata:
Thanks for the cleanup!

PR #645, opened 2019-06-16
Add 2 more demos

Frazer:
Add 2 more demos to index.html


cdata:
These demos were actually removed from our main docs in a previous change. The argument was that they are interesting test cases but not representative demos.

Frazer:
Oh, sorry.  I find the pbr.html really impressive - and showing the high quality reflections is inspiring.

cdata:
Yeah, it would be nice to come up with a good way to re-incorporate these demos. However, for now we would like to keep them separate as they aren't being updated with the rest of our docs. Thanks!

PR #651, opened 2019-06-19
Remove duplicate environment creation

### LINKED ISSUE #649: PMREM appears to be applied to environment maps twice
@elalish pointed out that we appear to be generating PMREM for environment maps twice, possibly due to a redundant invocation in `firstUpdated`. We should quash this ASAP, as PMREM generation is quite costly.

elalish:
Fixes #649 

Our environments were getting loaded twice, once in firstUpdate, and again in update. I think generation only happened once because of three's caching, but reading in the HDR file and running PMREM were both definitely happening twice, and are both quite expensive. I can't see anything going wrong from removing the firstUpdate function, but I'm also not sure why it was there, so please let me know if I've missed some side effect.

### DIFF
--- a/src/features/environment.ts
+++ b/src/features/environment.ts
@@ -112,15 +112,6 @@ export const EnvironmentMixin = (ModelViewerElement:
           }
         }
 
-        firstUpdated(changedProperties: Map<string, any>) {
-          super.firstUpdated && super.firstUpdated(changedProperties);
-          // In case no environment-related properties were confiured, we should
-          // make sure that the environment is updated at least once:
-          if (this[$cancelEnvironmentUpdate] == null) {
-            this[$updateEnvironment]();
-          }
-        }
-
         [$onModelLoad](event: any) {
           super[$onModelLoad](event);
 
--- a/src/features/environment.ts
+++ b/src/features/environment.ts
@@ -167,6 +167,7 @@ export const EnvironmentMixin = (ModelViewerElement:
             }
 
             this[$applyEnvironmentMap](environmentMap);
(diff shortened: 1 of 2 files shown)


cdata:
👍 How can we catch regressions?

elalish:
I'm not sure; I'm hoping our current tests and fidelity tests will tell us. Can you think of new cases we should add?

cdata:
Our current test suite is not comprehensive, and we rely on folks layering on regression tests and feature tests as changes are made.

Off the top of my head, I see tests here for checking the background color when nothing is configured, but not the environment lighting:

https://github.com/GoogleWebComponents/model-viewer/blob/42553986abe851fbe966b838bad5c5eafc9704c6/src/test/features/environment-spec.ts#L148-L160

Since the comment that you removed claims that `firstUpdated` was needed in the case that no environment is configured, it sounds like this is an important test case to cover.

elalish:
Good idea, I've added the unconfigured tests. As for checking how many times the environment is generated, I would spend significantly more time making that test than on the whole PR and I still wouldn't be convinced it would reliably catch those perf bugs. I think this is a relatively unlikely regression, so I would skip this particular test, but you're welcome to overrule me if this is more important than our other work.

cdata:
> I would spend significantly more time making that test than on the whole PR

This is the reason why easy / small PRs are dangerous :)

cdata:
I think for the near term let's just add a test that waits for a reasonable window of time (call it one second) after the model loads and asserts that the environment did not re-generate in that window of time.

In some failure scenarios, this test will fail flakily. But, I think flakey failure in this case is better than no test at all.

PR #658, opened 2019-06-25
Camera change event

### LINKED ISSUE #506: Observable event indicating that the camera position changed

cdata:
This change proposes a new, public event `camera-change`, that can be used to observe and respond to changes in the `camera-orbit` and `field-of-view`.

The new event also replaces our internal, ad-hoc coordination with a moving/changing camera.

Fixes #506

### DIFF
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -323,9 +323,6 @@ export default class ModelViewerElementBase extends UpdatingElement {
     this[$needsRender]();
   }
 
-  [$onUserModelOrbit]() {
-  }
-
   /**
    * Parses the element for an appropriate source URL and
    * sets the views to use the new model based off of the `preload`
--- a/src/test/features/controls-spec.ts
+++ b/src/test/features/controls-spec.ts
@@ -13,9 +13,9 @@
  * limitations under the License.
  */
 
-import {$controls, $idealCameraDistance, $promptElement, ControlsInterface, ControlsMixin, DEFAULT_INTERACTION_PROMPT_THRESHOLD, INTERACTION_PROMPT, SphericalPosition} from '../../features/controls.js';
-import ModelViewerElementBase, {$scene} from '../../model-viewer-base.js';
-import {SmoothControls} from '../../three-components/SmoothControls.js';
+import {$controls, $idealCameraDistance, $promptElement, CameraChangeDetails, ControlsInterface, ControlsMixin, DEFAULT_INTERACTION_PROMPT_THRESHOLD, INTERACTION_PROMPT, SphericalPosition} from '../../features/controls.js';
+import ModelViewerElementBase, {$canvas, $scene} from '../../model-viewer-base.js';
+import {ChangeSource, SmoothControls} from '../../three-components/SmoothControls.js';
 import {Constructor} from '../../utilities.js';
 import {assetPath, dispatchSyntheticEvent, rafPasses, timePasses, until, waitForEvent} from '../helpers.js';
 import {BasicSpecTemplate} from '../templates.js';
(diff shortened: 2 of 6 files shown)


src/model-viewer-base.ts
```diff
@@ -323,9 +323,6 @@ export default class ModelViewerElementBase extends UpdatingElement {
     this[$needsRender]();
   }
 
-  [$onUserModelOrbit]() {
```
elalish: :+1:

src/test/features/controls-spec.ts
```diff
@@ -165,6 +165,29 @@ suite('ModelViewerElementBase with ControlsMixin', () => {
...
+      test('sets an appropriate event source', async () => {
+        const cameraChangeDispatches =
+            waitForEvent<CustomEvent<CameraChangeDetails>>(
+                element, 'camera-change');
+
+        const cameraOrbit = element.getCameraOrbit();
+        element.cameraOrbit = `${cameraOrbit.theta + 1}rad ${
+            cameraOrbit.phi}rad ${cameraOrbit.radius}m`;
+
+        const event = await cameraChangeDispatches;
```
elalish: I'm still a little dubious on the value of all this duplicated code, when if the first test fails, it'll just fail both of these anyway.

cdata: If the code that assigns the camera change source breaks in some way, one test will pass and the other will not.

elalish: True; is this related to how we seem to get zero test logging feedback in Safari, as compared to Chrome? Because I feel like with the output Chrome gives, that would be easy to disambiguate even if they were only one test.

cdata:
> How will the website know what about the camera changed? 

Code that cares about this will (for now) have to be measuring values on each frame and comparing them to the values from last frame.

My rationale for doing this is that it is non-zero cost for us to provide deltas (a cost we would pay on every camera change, observed or not), not everyone will need the deltas and those that will need them have enough information to compute them for themselves.

> I could imagine they might want to run some callback only if the FOV changed or something

Great feedback! Filed https://github.com/GoogleWebComponents/model-viewer/issues/664

elalish:
Makes sense, thanks for the explanation.

src/test/features/controls-spec.ts
cdata: It's more about not conflating behavioral tests. It is advantageous for tests to be based around discrete behaviors or features.

For example, if someday we decided to remove the `event.detail.source` property, we might come along here and delete this test at the same time. If it was the only test that covered the `camera-change` event dispatching in the first place, we might accidentally remove coverage for something because the test was pulling double duty.

frontendjorge:
> This change proposes a new, public event `camera-change`, that can be used to observe and respond to changes in the `camera-orbit` and `field-of-view`.
> 
> The new event also replaces our internal, ad-hoc coordination with a moving/changing camera.
> 
> Fixes #506

Hello, but how to get the update of canera-orbit I'm using: 

`const modelViewer = document.querySelector('model-viewer');
modelViewer.addEventListener('camera-change', (event) => {
    //const {totalProgress} = event;
    console.log(event)
  });`

but in camera-orbit and field-of-view is always the same value.

cdata:
@frontendjorge please consider this example:

```javascript
const modelViewer = document.querySelector('model-viewer');
modelViewer.addEventListener('camera-change', () => {
  const orbit = modelViewer.getCameraOrbit();
  const fov = modelViewer.getFieldOfView();

  const orbitString = `${orbit.theta}rad ${orbit.phi}rad ${orbit.radius}m`;
  const fovString = `${fov}deg`;

  // etc...
});
```

PR #665, opened 2019-06-27
Add missing files in src tree and gh-pages

cdata:
The `:focus-visible` change omitted some changes that caused our docs to break. Re-introducing them here.


cdata:
Related build that caused the problem: https://travis-ci.org/GoogleWebComponents/model-viewer/jobs/551409236

There is a bigger problem here: the build clearly failed but gh-pages was deployed anyway.

Issue #491, opened 2019-04-18
Enable better cache control and/or cache purging


cdata:
Today we aggressively cache models and artifacts related to environment maps, and we do not make any attempt to purge this cache. So, any scenario that might lead to models tens or hundreds of models and/or distinct environment maps being loaded is likely to hit memory ceilings. We should offer API that enables users to control how we cache models and other artifacts, and possibly also some way to purge the cache completely.

PR #667, opened 2019-06-28
Cache environments and skyboxes

### LINKED ISSUE #336: Cache generated environment map

elalish:
Fixes #336 

Really #336 was already fixed, but not in a complete way as only the generated environment was cached, but not any loaded environments, skyboxes, or PMREM results. Now all of the above are cached in a single system. The near-term goal of this PR is to get our automated tests to pass on Safari with PMREM, as we're guessing the current problem is they use too much memory from duplicate environments.

The next thing to consider as a follow-on to this work is probably #491.


src/three-components/TextureUtils.ts
```diff
@@ -215,32 +201,47 @@ export default class TextureUtils extends EventDispatcher {
...
       // texture from the source:
-      const nonPmremEnvironmentMap = environmentMap;
-
-      environmentMap = this.pmremPass(nonPmremEnvironmentMap);
+      if (targetCache.has(environmentMapUrl)) {
+        environmentMap = targetCache.get(environmentMapUrl)!;
+      } else {
+        const nonPmremEnvironmentMap = environmentMap!.texture;
+        environmentMap = this.pmremPass(nonPmremEnvironmentMap);
+        targetCache.set(environmentMapUrl, environmentMap);
```
cdata: This dualistic quality where the environment map may be PMREM'd or may not be adds an interesting complication. You would need to amend the skybox example I commented about above to do something like this when populating the cache in the environment map case:

```typescript
environmentCache.set(environmentUrl, this.loadEquirectAsCubeMap(
        environmentUrl,
        progressTracker ? progressTracker.beginActivity() : () => {})
    // Chain the promise so that the cached async work includes the PMREM pass:
    .then((environmentMap: WebGLRenderTarget) => this.pmremPass(environmentMap.texture)));
```

That would actually end up looking a lot cleaner than what we had before this PR was proposed...

PR #668, opened 2019-06-28
Reuse renderer in tests

elalish:
Still working on getting our tests to pass. All those Safari warnings about too many WebGL contexts made me dig in a little. I found this thread https://github.com/pixijs/pixi.js/issues/2233#issuecomment-271924403 which says that Three's renderer.dispose() method only simulates a context being destroyed and there's no guarantee the browsers actually clean up at that point. Everyone seems to think the best answer to create a renderer (and therefore a context) once, and always reuse it. We were doing that everywhere except our tests, so I've updated them, and even on Chrome they run markedly faster. We still create and (try to) destroy a few in tests that don't involve an element. I'm not sure if it's worth it to try and make those share a renderer too.

### DIFF
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -50,7 +50,6 @@ export const $onModelLoad = Symbol('onModelLoad');
 export const $onResize = Symbol('onResize');
 export const $onUserModelOrbit = Symbol('onUserModelOrbit');
 export const $renderer = Symbol('renderer');
-export const $resetRenderer = Symbol('resetRenderer');
 export const $progressTracker = Symbol('progressTracker');
 
 /**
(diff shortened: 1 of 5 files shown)


cdata:
Generally seems fine. It's worth noting that this effectively undoes one of the memory optimizations we put in place earlier in #438. Let's move forward and keep an eye on the Safari results for warning signs that this created a regression.

cdata:
It's also worth mentioning that calling `renderer.dispose()` is not about cleaning out the WebGL context. It's mainly for cleaning up Three.js internal object references.

https://github.com/mrdoob/three.js/blob/78b74d23c48493370d77e53b38137db8974ac835/src/renderers/WebGLRenderer.js#L571-L576

Issue #449, opened 2019-03-28
Improved, generalized staging for fidelity test scenarios


cdata:
We should look into improving the conditions of our fidelity test scenarios so that we can get the most actionable, consistent feedback from them.

Recording resources worth looking into here:

 - https://masteringcgi.com.au/mastering-cgi-shader-ball/
 - https://twitter.com/jcanadagarcia/status/1009152211157442560
 - https://twitter.com/jcanadagarcia/status/1009151582292840453
 - http://casual-effects.com/data/index.html

elalish:
@cdata Is this a dupe of #309?

cdata:
No, this issue was intended to cover how we present the models in fidelity test cases to maximize the useful details of the renders. In other words, picking useful camera angles and physical environment conditions that give us a higher chance of getting useful information when reviewing rendered pixels.

cdata:
This is probably worthy of inclusion as we do #616

elalish:
I'm declaring this fixed by #672.

PR #672, opened 2019-07-02
Add new fidelity tests

### LINKED ISSUE #648: Select / author fidelity tests to judge effects of PMREM overhaul

elalish:
### Reference Issue
Fixes #648, possibly #449?

I added two new fidelity tests as copies of the Khronos-MetalRoughSpheres, one with an HDR environment and one with an LDR environment (both spruit_sunrise, which has very high dynamic range, so good for catching bugs). Combined with the original that uses lightroom (similar to our generated environment), we now cover all of our lighting cases with a single, canonical model. 

Interestingly, this exposed a bug in filament, or possibly our usage thereof: Our Filament renders for the HDR and LDR environments are identical, and both appear to be LDR. This explains why we thought our renders looked "too yellow" when using spruit_sunrise_2k.hdr; in actuality Filament was not yellow enough, as it was lacking HDR.


elalish:
Indeed, our usage of Filament is wrong, filed #673

cdata:
> in actuality Filament was not yellow enough, as it was lacking HDR

You have got to be kidding me.

Great find @elalish !

I think we can land this PR if it's all the same to you. We'll update the goldens whenever Filament fixes the bug.

PR #674, opened 2019-07-03
Default stage light to zero

### LINKED ISSUE #396: Deprecate stage-light-intensity / change default to 0

elalish:
### Reference Issue
Fixes #396 

In addition to changing the default, I removed the hemisphere light as I believe its only purpose was to hide an old LDR environment lighting bug that has since been fixed. I also removed all the arbitrary constants I could find relating to lighting. 

I also added a top-light to our environment generation to take the place of our default stage light, but we'll need to update our fidelity tests to properly compare.

### DIFF
--- a/src/three-components/EnvironmentMapGenerator.ts
+++ b/src/three-components/EnvironmentMapGenerator.ts
@@ -93,35 +93,41 @@ export default class EnvironmentMapGenerator extends EventDispatcher {
     scene.add(box6);
 
 
-    // -z right
+    // -x right
     const light1 = new Mesh(geometry, this.createAreaLightMaterial(50));
     light1.position.set(-16.116, 14.37, 8.208);
     light1.scale.set(0.1, 2.428, 2.739);
     scene.add(light1);
 
-    // -z left
+    // -x left
     const light2 = new Mesh(geometry, this.createAreaLightMaterial(50));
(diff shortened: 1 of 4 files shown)


src/three-components/EnvironmentMapGenerator.ts
```diff
@@ -93,36 +93,42 @@ export default class EnvironmentMapGenerator extends EventDispatcher {
     scene.add(box6);
 
 
-    // -z right
+    // -x right
```
cdata: image

Issue #515, opened 2019-05-01
npm run dev doesn't pick up changes to fidelity tests


elalish:
### Description

When npm run dev is running, it seems to only pick up changes to model-viewer, but not to the other targets like image-comparison-app that the fidelity tests use. This is confusing, as all of these targets are served to 8000, but they simply don't update when changed. Also, it starts throwing Middlewear errors on the command line.

### OS

- [ ] Android
- [ ] iOS
- [x] Linux
- [ ] MacOS
- [ ] Windows

### Versions
master, as of May 1st, 2019.

cdata:
Yeah, I agree this is confusing (or at least unexpected).

In general, it seems as though the `npm run dev` workflow loses the ability to recompile based on discrete bundle changes (it invokes some form of `npm run build` when files change, which rebuilds everything).

If I run `tsc -w` in one terminal and `rollup -c -w` in another terminal, a rebuild only occurs for the bundle (or bundles) that are affected by discrete file changes. It would be nice if we could have this property in `npm run dev` as well, that way we would not have to pick between _just_ building `model-viewer.js` and rebuilding the whole world when going that route.

elalish:
If that's not hard, I'd agree that would be ideal. However, it's also quite easy to work around, so we could just take the tack of just updating the README to say that npm run dev only updates model-viewer and you have to do npm run build first and after any change outside of model-viewer (which is probably rare anyway).

PR #676, opened 2019-07-09
Switched Filament rendering from native to JS

### LINKED ISSUE #434: Update gltf_renderer to use Filament's new gltfio library

### LINKED ISSUE #673: Filament fidelity tests do not use HDR
### Description

We are using Filament incorrectly, so when we pass in an HDR environment map it is getting converted to LDR.

#### Live Demo
Note that the Filament renders of spruit_surise_2k.hdr and .jpg are identical:
Screenshot from 2019-07-02 15-10-43

elalish:
### Reference Issue
Fixes #434 

Really I'm hoping to also fix #673, but that hasn't happened yet, though at least this has opened a discussion with Filament about their renderer ([bug](https://github.com/google/filament/issues/1387)). Here I've switched from filament's native renderer to their JS version, which is a bit simpler and already has gltfio. If we go forward with this, I believe we can remove all of the filament building stuff from our repo, so long as we're willing to pull down a binary for cmgen. I checked in the linux executable for now, but we could also consider other means (like pulling it from the github release). 

I also updated our fidelity tests, switching from gltf to glb and removing some less useful ones.


elalish:
I think this one is good to go now. I'm going to say it also fixes #515, as I updated the Readme. We're blocked on a few of the khronos tests because they don't have GLB versions and FilamentJS has a bug with opening GLTFs in another directory. Once either of those is fixed, we should be able to add them back in (I only removed the config entries, so it should be quick to put them back).

elalish:
Closing in favor of #682, where I've rebased onto master.

PR #677, opened 2019-07-10
Magic Leap / Prismatic example updates

chrisdubya:
General fixes for Prismatic compatibility with model viewer.

I noticed that the Prismatic library was commented out of all of the examples, so I re-added it back to augmented-reality.html example and made some updates to improve the display of the model when viewed on Magic Leap.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -16,7 +16,7 @@ https://glitch.com/edit/#!/model-viewer
 - [ ] Chrome
 - [ ] Edge
 - [ ] Firefox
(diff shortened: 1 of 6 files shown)


cdata:
Thanks for the contribution! Would you be willing to post before / after image captures of the change? It looks like it's just the z-offset, but it would be nice to have a visual comparison.

chrisdubya:
Thanks, I'll close this and open an issue and elaborate the problem with screenshots attached.  Among some cosmetic fixes, the main issue is Prismatic not loading and the `ml-model` tag not receiving the `magic-leap` attribute.

PR #686, opened 2019-07-30
Adding actual unlit glTF

### LINKED ISSUE #653: Our claimed unlit model is not unlit
### Description

The third model here:
Claims to be unlit, but Astronaut.glb has no material defined, which means it gets the GLTF default material, which is PBR, roughness = 1, metalness = 1. 

Also the alt text wrong on most of the models on that page.

elalish:
Fixes #653 

Our model that claimed to be unlit was actually just a default material, which means PBR with roughness = 1 and metal = 0. Instead of replacing it, I added an identical scene with the same model converted to unlit, since I figured this would show the distinction better.


elalish:
Closing in favor of #694; needed a new branch to fix the merge.

PR #689, opened 2019-08-03
Improved cache control, fewer memory leaks

### LINKED ISSUE #491: Enable better cache control and/or cache purging

cdata:
This change proposes a first step towards better cache management and introduces limited cache control as a public API. Highlights include:

 - `CacheEvictionPolicy` has been introduced, to offer a consistent model for purging items from caches
 - A few memory leaks have been fixed:
    - We were leaking all `<model-viewer>` elements via our integration with `:focus-visible` polyfill
    - We have been leaking vertex buffers in a few places due to not disposing of them appropriately
 - `CachingGLTFLoader` has been integrated with `CacheEvictionPolicy` so that we no longer leak all of our models by design
 - A minor bug in `cloneGltf` has been fixed, which was leading us to incorrectly clone materials
 - A public, static, writable property `modelCacheSize` has been added, to enable imperative user control of GLTF caching behavior
    - The value is global to all `<model-viewer>` instances and sets the "eviction threshold" for `CachingGLTFLoader`
    - It can be used this way: `customElements.get('model-viewer').modelCacheSize = 10;`.
 - The Android test target has been removed, due to an undiagnosed critical failure that causes it to crash _only on Sauce Labs_

Future work:

 - Re-enable Android tests via Browserstack https://github.com/GoogleWebComponents/model-viewer/issues/691

Fixes #491


src/test/features/environment-spec.ts
```diff
@@ -122,6 +122,12 @@ const waitForLoadAndEnvMap =
     };
 
 suite('ModelViewerElementBase with EnvironmentMixin', () => {
+  suiteTeardown(() => {
+    // Reset the renderer once at the end of this spec, to clear out all
+    // of the heavy cached image buffers:
+    ModelViewerElementBase[$resetRenderer]();
```
elalish: Do we really want to destroy the gl context here? Would it be adequate to dispose of and recreate TextureUtils?

src/test/three-components/TextureUtils-spec.ts
```diff
@@ -29,23 +28,36 @@ const EQUI_URL = assetPath('spruit_sunrise_2k.jpg');
...
+    // renderer *should* always return its state to what it was before to avoid
+    // this kind of problem (and many other headaches).
+    renderer = new WebGLRenderer({canvas});
+    renderer.debug.checkShaderErrors = true;
+  });
+
+  suiteTeardown(() => {
+    // Ensure we free up memory from loading large environment maps:
+    Cache.clear();
+    renderer.dispose();
```
elalish: What does disposing of the renderer do that Cache.clear() doesn't do already? Every time we recreate a renderer we get a lot of console spew that makes reading the test output harder.

src/three-components/CachingGLTFLoader.ts
```diff
@@ -94,28 +141,42 @@ export class CachingGLTFLoader {
...
-          (node.material as Array<Material>).forEach(material => {
-            material.onBeforeCompile = updateShader;
-          });
-        } else if (node.material != null) {
-          node.material.onBeforeCompile = updateShader;
-        }
+      this[$evictionPolicy].retain(url);
+
+      Object.defineProperty(model, $releaseFromCache, {
+        value: (() => {
```
elalish: This doesn't seem like our usual style; what's the purpose of Object.defineProperty here?

src/three-components/ModelUtils.ts
```diff
@@ -80,14 +90,30 @@ export const cloneGltf = (gltf: Gltf): Gltf => {
   const cloneBones: BoneMap = {};
   const cloneSkinnedMeshes: SkinnedMeshMap = {};
 
-  if (hasScene && hasSkinnedMeshes) {
+  if (hasScene) {
```
elalish: Can we combine this with the same if statement above?

src/three-components/PMREMGenerator.ts
```diff
@@ -56,6 +56,10 @@ export const generatePMREM =
           cubeLods.forEach((target) => {
             target.dispose();
           });
+          meshes.forEach((mesh) => {
+            (mesh.material as Material).dispose();
+            mesh.geometry.dispose();
```
elalish: Disposing of the PMREM materials will be a significant hit on performance (it takes longer to compile them than to run them). If we'd like to control it, perhaps we should make it a class again that just holds the materials and is a member of TextureUtils, so that it's only disposed of when TextureUtils is disposed?

src/three-components/TextureUtils.ts
```diff
@@ -424,18 +424,26 @@ void main() {
...
-      const environmentMap = await environmentMapLoads;
-      if (environmentMap != null) {
-        environmentMap.dispose();
+      try {
+        const environmentMap = await environmentMapLoads;
+        if (environmentMap != null) {
+          environmentMap.dispose();
+        }
+      } catch (e) {
+        // Suppress errors
```
elalish: What errors are we getting?

src/utilities.ts
```diff
@@ -22,7 +22,7 @@ export {
 
 export type Constructor<T = object> = {
   new (...args: any[]): T,
-  prototype: T
+  prototype: T,
```
elalish: intentional?

src/utilities/cache-eviction-policy.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    for (let i = this[$recentlyUsed].length - 1; i >= this[$evictionThreshold];
+         --i) {
+      const key = this[$recentlyUsed][i];
+      const retainerCount = this[$retainerCount].get(key);
+
+      if (retainerCount === 0) {
+        this[$cache].delete(key);
+        this[$recentlyUsed].splice(i, 1);
```
elalish: I don't quite follow this splice. Why are we putting a new entry into $recentlyUsed when we just deleted that entry from the cache?

src/utilities/cache-eviction-policy.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+    this[$retainerCount].set(key, this[$retainerCount].get(key)! + 1);
+
+    const recentlyUsedIndex = this[$recentlyUsed].indexOf(key);
+
+    if (recentlyUsedIndex !== -1) {
+      this[$recentlyUsed].splice(recentlyUsedIndex, 1);
+    }
+
+    this[$recentlyUsed].unshift(key);
```
elalish: These two lines seem to be vaguely duplicate? Also, does the $recentlyUsed array grow without bound? Could we just make $recentlyUsed the same length as $retainerCount and just swap elements to the beginning? If only JS had a priority queue...

src/utilities/focus-visible.ts
```diff
@@ -34,39 +37,67 @@ export type MixableBaseClass = HTMLElement&OptionalConnnectedCallback;
...
-          coordinateWithPolyfill(this);
+          if (this[$endPolyfillCoordination] == null) {
+            this[$endPolyfillCoordination] = coordinateWithPolyfill(this);
+          }
+        }
+
+        disconnectedCallback() {
+          super.disconnectedCallback && super.disconnectedCallback();
+          // It's important to remove the polyfill event listener when we
+          // disconnect, otherwise we will leak the whole element via window:
```
elalish: Interesting!

index.html
```diff
@@ -305,6 +305,14 @@ <h3 class="grouping-title">Properties</h3>
               always start paused, and remain so unless the autoplay attribute
               is set or the .play() method is invoked.</p>
             </li>
+            <li>
+              <div>ModelViewer.modelCacheSize</div>
+              <p>This static, writable property sets &lt;model-viewer&gt;'s internal
+              cache size for GLTF models. Note that any number of models may be cached
+              at any given time. This number just sets the maximum number of models
+              that can be cached even if they are not being used by a &lt;model-viewer&gt;
```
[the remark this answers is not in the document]
cdata: 👍 will clarify

src/test/features/environment-spec.ts
cdata: As we have discussed before, the GL context isn't especially important. What this does do is clear out [a few internal Three.js caches](https://github.com/mrdoob/three.js/blob/dev/src/renderers/WebGLRenderer.js#L569-L572). It's possible that we could dispose of the artifacts held in these caches more discretely, but I'm not sure how to go about doing that comprehensively.

This is only ever done once in the entire test suite (after `environmennt-spec.ts` is run to completion), so it shouldn't be a huge cost 🤷‍♂

src/test/three-components/TextureUtils-spec.ts
cdata: See my comment above, I linked to the relevant lines. It's a hatchet, not a scalpel, for sure, but profiles bare out that there is a fair amount of memory (megabytes worth) being held by references in the renderer's internal cache.

src/three-components/CachingGLTFLoader.ts
cdata: It's the only way I could get `tsc` to stop complaining about me decorating this object.

It might be possible to change the type of the identifier and get a cleaner result. I'll try that out.

src/three-components/ModelUtils.ts
cdata: Yeah, good call, I think we can merge the conditions.

src/three-components/PMREMGenerator.ts
cdata: We need to dispose of them _or_ we need to ensure they are re-used. The profile suggested that we were holding very accumulating very large buffers in memory.

src/three-components/TextureUtils.ts
cdata: If a promise is rejected (for example. say a file fails to load), the promise is rejected for all time. If you later `await` that promise, it will throw an error.

src/utilities.ts
cdata: Will remove

src/utilities/cache-eviction-policy.ts
cdata: The JavaScript `Array` is unbounded and can grow indefinitely, runtime permitting. One cannot limit the size of an `Array`. JavaScript `Array` is more like a map than an actual array, where the indices are the keys. You can overload its `push` / `pop` etc. methods, but you cannot overload the index operator, and that allows you to insert to arbitrary indices (e.g., `this[$recentlyUsed][49283] = true` will totally do a thing here, and grow the reported `length` accordingly).

`[$retainerCount]` being a `Map`, it is more expensive to count the keys than to just do the book keeping directly as we retain new keys. So, surgically splicing elements out of the `Array` is probably the best we can do.

src/utilities/cache-eviction-policy.ts
cdata: [The signature for JavaScript splice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) is:

```
var arrDeletedItems = array.splice(start[, deleteCount[, item1[, item2[, ...]]]])
```

Added items start at the third argument. So, we are only removing things here, not adding anything.

src/three-components/PMREMGenerator.ts
elalish: I'm fine with disposing of the geometry, but the shaders are reused. Of course you could ditch them if you never wanted to load another environment map, but we don't really have a way to reason about that right now.

src/utilities/cache-eviction-policy.ts
elalish: Ah, that's what I missed when I read that definition; I thought these splices were adding elements, but now I see they're removing them. I thought elements weren't being removed at all, hence my other comment. Thanks!

src/utilities/cache-eviction-policy.ts
elalish: Got it, see above.

src/three-components/PMREMGenerator.ts
cdata: I will take a stab at caching them so that they can be re-used and properly disposed of. [As we know for our earlier spelunking](https://github.com/mrdoob/three.js/blob/master/src/renderers/webgl/WebGLPrograms.js#L293), `Material` instances need to be disposed _eventually_, otherwise their backing programs will leak.

cdata:
This change has been rebased onto master

PR #695, opened 2019-08-06
Blur skybox

### LINKED ISSUE #181: Ability to Blur out environment
A common feature in model viewers is the ability to apply a blur on the environment. This deemphasizes the environment, makes the model pop, more interesting than a flat color, and ability to provide a nicer environment map than the default one generated. I don't think this should change the PMREM generation, such that a very reflective model would reflect the unblurred environment, but curious how other platforms handle this with regard to the corresponding environment map on the model

### LINKED ISSUE #528: HDR environment maps are not interpolated

elalish:
Fixes #181 
Fixes #528 

This applies our cubemap Gaussian blur function to the skybox. In the process, it also converts an RGBE skybox to RGB so that it can be mipmapped and interpolated like an LDR skybox. I'm not 100% happy with the Gaussian blur function yet, as there is still a bit of aliasing, but I have an idea to improve it that we can look into as a follow-on.


examples/assets/attributions.md
```diff
@@ -49,21 +49,21 @@ licensed under <a href="https://hdrihaven.com/p/license.php">CC0</a>
...
 
-## aircraft_workshop_01_4k.hdr
+## aircraft_workshop_01_1k.hdr
 
-aircraft_workshop_01_4k.hdr by <a href="https://hdrihaven.com">HDRI Haven</a>
+aircraft_workshop_01_1k.hdr by <a href="https://hdrihaven.com">HDRI Haven</a>
 licensed under <a href="https://hdrihaven.com/p/license.php">CC0</a>
 (<a href="https://hdrihaven.com/hdri/?h=aircraft_workshop_01">source</a>)
 
-## music_hall_01_4k.hdr
```
cdata: 💯

src/three-components/TextureUtils.ts
```diff
@@ -49,15 +52,6 @@ const $loadEnvironmentMapFromSkyboxUrl =
     Symbol('loadEnvironmentMapFromSkyboxUrl');
 const $loadGeneratedEnvironmentMap = Symbol('loadGeneratedEnvironmentMap');
 
-export interface TextureUtilsConfig {
```
cdata: 👍

src/three-components/TextureUtils.ts
```diff
@@ -38,6 +40,7 @@ Cache.enabled = true;
 const HDR_FILE_RE = /\.hdr$/;
 const ldrLoader = new TextureLoader();
 const hdrLoader = new RGBELoader();
+const cubemapSize = 256;
```
cdata: Let's capitalize this e.g, `CUBEMAP_SIZE` (and yeah, `ldrLoader` / `hdrLoader` break our convention here a little; they should probably be static class members or something but we can fix that later).

src/three-components/TextureUtils.ts
```diff
@@ -192,9 +184,13 @@ export default class TextureUtils extends EventDispatcher {
         environmentMapLoads = this[$loadGeneratedEnvironmentMap]();
       }
 
-      const [environmentMap, skybox] =
+      let [environmentMap, skybox] =
           await Promise.all([environmentMapLoads, skyboxLoads]);
 
+      if (skybox != null) {
+        skybox = this.gaussianBlur(skybox, 0.004, GammaEncoding);
```
cdata: Doesn't this return a reference to a new, distinct render target? If so, who manages the lifecycle of this instance? When does it get disposed?

src/three-components/TextureUtils.ts
```diff
@@ -290,7 +287,8 @@ export default class TextureUtils extends EventDispatcher {
 
   gaussianBlur(
       cubeTarget: WebGLRenderTargetCube,
-      standardDeviationRadians: number = 0.04) {
+      standardDeviationRadians: number = 0.04,
+      outputEncoding: TextureEncoding|null = null): WebGLRenderTargetCube {
```
cdata: If you just want this to be an optional third argument, you can express it as `outputEncoding?: TextureEncoding`. The "loose" equality conditionals are useful for this particular scenario only: checking if a value is `undefined` _or_ `null`. So, the check further down the page will work as written.

src/three-components/TextureUtils.ts
```diff
@@ -366,39 +373,61 @@ void main() {
...
+    blurUniforms.latitudinal.value = false;
+    blurUniforms.tCube.value = cubeTexture;
+    blurUniforms.inputEncoding.value = encodings[cubeTexture.encoding];
+    blurUniforms.outputEncoding.value = encodings[tempTexture.encoding];
     blurCamera.update(this.renderer, blurScene);
 
-    this.renderer.toneMapping = toneMapping;
-    this.renderer.toneMappingExposure = toneMappingExposure;
-    this.renderer.gammaOutput = gammaOutput;
+    if (outputEncoding != null && outputEncoding === GammaEncoding &&
```
cdata: Strictly speaking, you probably don't need the first clause of this condition because `outputEncoding === GammaEncoding` will always be false for every non-number value.

src/three-components/shader-chunk/common.glsl.js
```diff
@@ -70,38 +72,46 @@ export const encodings = {
...
-  [RGBDEncoding]: 5
+  [RGBDEncoding]: 5,
+  [GammaEncoding]: 6
 };
 
 export const texelIO = /* glsl */ `
-vec4 inputTexelToLinear(vec4 value, int encoding){
-    if(encoding == 0){
+uniform int inputEncoding;
+uniform int outputEncoding;
```
cdata: 💯

src/three-components/TextureUtils.ts
```diff
@@ -366,39 +373,61 @@ void main() {
...
+        cubeTexture.encoding !== GammaEncoding) {
+      cubeTarget.dispose();
+      const outParams = {
```
cdata: Let's call this `blurTargetOptions` or something like that

src/three-components/TextureUtils.ts
elalish: Sometimes it is the same renderTarget (if the input is jpg); if it is hdr then gaussianBlur disposes of the one you gave it and returns this new one in its place.

cdata: Doesn't that mean we're possibly disposing of the cached render target?

elalish: Let's fix this in a following PR to extend our usage of your cachingPolicy.

elalish: To be clear, we're no longer disposing of the cached render target, but we are leaking backgrounds.

mrdoob:
Love the blurred skyboxes! 👌

I tried the feature in http://model-viewer-tester.glitch.me/ and I'm not sure the exposure works correctly for the background now.

Screen Shot 2019-08-13 at 4 35 40 PM

Is there something I should change in my test?

cdata:
I played around with it and couldn't tell if anything was wrong. Do you
have a screenshot of what you expected it to look like at that exposure
level?

mrdoob:
Yes, sorry, I did not know how to use a previous version of the element, just figured it out.

This is how it's supposed to look like:

Screen Shot 2019-08-13 at 5 03 16 PM

Basically, it looks like the blurred background loses the HDR information.

cdata:
👍 great catch @mrdoob I filed #707 to investigate

elalish:
Ah yes, good catch. This is because of the way I made the skybox
interpolate, by converting it to RGB. Works great except when the exposure
changes, which I neglected to test. I guess I should tweak the shader
instead to interpolate RGBE.

PR #698, opened 2019-08-07
Hotfix: address fullscreen fallback in Chrome m76

### LINKED ISSUE #693: "View in AR" button sometimes only opens fullscreen instead of SceneViewer
### Description

The "View in AR" button is correctly shown. However, for some models, pressing it just makes everything glitch around (some onsceen elements appear, some disappear, some just blank out), pressing it again makes the model go fullscreen, but SceneViewer isn't launched.

#### Live Demo
https://lemon-town.glitch.me/
but also happens with the default astronaut guy sometimes https://model-viewer.glitch.me/
Video of the behaviour:
https://youtu.be/t6uQEcQDnBU

### Browser Affected

- [X] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [X] Android, latest Chrome
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### Versions

happens with both 0.3x and 0.4x
Device is Xiaomi Mi 9, but have also seen on other Android phones

cdata:
This hotfix addresses a regression in Chrome m76, that causes our "fullscreen fallback" for Scene Viewer to not work as intended. The symptoms of this problem have been documented by several users in #693 . 

The nature of the fix is to split our click handling out into multiple event listeners. We now use two events to attempt fullscreen and trigger the intent respectively.

This "fix" is actually just a workaround for the problem. It preserves our behavior through Chrome m76, but **this workaround doesn't save us from potential future regressions**. The kernel of the problem is that we expect a handled `intent://` URL on Android to cancel a `requestFullscreen()` invocation. There is no spec or other explicit guarantee that this will always be true.

Special thanks go to @azakus for suggesting the workaround.

A related crbug is forthcoming.

Fixes #693

### DIFF
--- a/src/template.ts
+++ b/src/template.ts
@@ -243,7 +243,7 @@ template.innerHTML = `
 
     <div class="slot ar-button">
       <slot name="ar-button">
-        <a id="default-ar-button" class="fab" href="#"
+        <a id="default-ar-button" class="fab"
             tabindex="2"
             aria-label="View this 3D model up close">
           ${ARGlyph}
--- a/src/features/ar.ts
+++ b/src/features/ar.ts
@@ -313,6 +313,11 @@ configuration or device capabilities');
 
           if (showArButton) {
             this[$arButtonContainer].classList.add('enabled');
+            // NOTE(cdata): The order of the two click handlers on the "ar
+            // button container" is important, vital to the workaround described
+            // earlier in this file. Reversing their order will cause our Scene
+            // Viewer integration to break.
+            // @see https://github.com/GoogleWebComponents/model-viewer/issues/693
             this[$arButtonContainer].addEventListener(
                 'click', this[$arButtonContainerClickHandler]);
             this[$arButtonContainer].addEventListener(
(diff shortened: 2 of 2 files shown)


cdata:
Still need to confirm that our iOS Quick Look integration has not regressed.

src/template.ts
```diff
@@ -243,7 +243,7 @@ template.innerHTML = `
 
     <div class="slot ar-button">
       <slot name="ar-button">
-        <a id="default-ar-button" class="fab" href="#"
+        <a id="default-ar-button" class="fab"
```
elalish: What was href="#" for?

cdata:
iOS compatibility confirmed on a local iPad.

src/template.ts
cdata: The reason to put it on a valid `href` attribute on an `<a>` is because the browser gives anchors different _default_ styles depending on whether or not they have an `href` attribute, and `href="#"` is a commonly used value that won't cause a navigation. We no longer depend on the default styles for `<a>` and we also `preventDefault()` on the event, so the attribute is vestigial.

cdata:
> Is the order of the click event listeners important for this to work?

I think so, yes. I can add a note about that.

PR #710, opened 2019-08-15
Improved PMREM

### LINKED ISSUE #587: Rendering inconsistencies with fine detailed normals + roughness

### LINKED ISSUE #632: Moiré effect on certain materials
### Description

On one of the models we've tested, we've seen a pretty noticeable moire effect.

Screen Shot 2019-06-12 at 10 14 11 AM

Screen Shot 2019-06-12 at 10 14 15 AM

I have a live demo of it here:
https://seed-oatmeal.glitch.me/

I know this isn't a model-viewer issue, and more something in Three, but I'm wondering if there's anything we can do to soften out this effect.

I feel like some of the reflectiveness there is due to https://github.com/GoogleWebComponents/model-viewer/issues/587, but even with that fix the effect would likely persist.

For reference, this is what we're expecting it to look like out of Substance:

Screen Shot 2019-06-12 at 10 19 43 AM

Screen Shot 2019-06-12 at 10 19 49 AM

elalish:
Fixes #587 
Fixes #632

This is an improvement to the very simple PMREM approximation we were using before. It also includes several fixes to the three.js MeshStandardMaterial that we use. These changes mostly made HDR environment maps look more realistic, but also included is a change which cleans up the aliasing we were getting from high frequency normal maps. There is still room for improvement, but it's significantly better than what it was. I also updated our lighting & environment page to better showcase our rendering with HDR.


src/constants.ts
```diff
@@ -66,4 +67,6 @@ export const IS_AR_QUICKLOOK_CANDIDATE = (() => {
 
 // @see https://developer.chrome.com/multidevice/user-agent
 export const IS_IOS_CHROME = IS_IOS && /CriOS\//.test(navigator.userAgent);
-export const IS_IOS_SAFARI = IS_IOS && /Safari\//.test(navigator.userAgent);
\ No newline at end of file
+export const IS_IOS_SAFARI = IS_IOS && /Safari\//.test(navigator.userAgent);
+export const IS_IE =
+    !((window as any).ActiveXObject) && 'ActiveXObject' in window;
```
cdata: 👍 

This specifically detects IE11. Let's either note that in a comment or change the name to `IS_IE11`.

src/three-components/PMREMGenerator.ts
```diff
@@ -27,7 +29,9 @@ import {encodings, getDirectionChunk, getFaceChunk, texelIO} from './shader-chun
 export const generatePMREM =
     (cubeTarget: WebGLRenderTargetCube, renderer: WebGLRenderer):
         WebGLRenderTarget => {
-          const {cubeUVRenderTarget, cubeLods, meshes} = setup(cubeTarget);
+          const roughnessExtra = [0.5, 0.7, 1.0];
```
cdata: I eventually determined that these were referring to lods, but from this line the purpose of these numbers seems kind of obscure.

src/three-components/PMREMGenerator.ts
```diff
@@ -41,61 +45,81 @@ export const generatePMREM =
             target.dispose();
           });
           meshes.forEach((mesh) => {
-            (mesh.material as Material).dispose();
+            // (mesh.material as Material).dispose();
```
cdata: Did we decide that we aren't going to dispose of this stuff? If so, maybe the meshes should just be a static array that gets updated as needed?

src/three-components/PMREMGenerator.ts
```diff
@@ -41,61 +45,81 @@ export const generatePMREM =
...
+        anisotropy: cubeTarget.texture.anisotropy,
+        encoding: cubeTarget.texture.encoding
+      };
+
+      // Hard-coded to lodMax = 8 until we can add a uniform.
+      const lodMin = 3;
+      const lodMax = 8;
+      // lodBase is the mip Level that is integrated to form all of the extra
+      // levels, but is not output directly into the PMREM texture. DO NOT
+      // CHANGE, as the Blur shader below is hard coded for this size.
```
cdata: 💯

src/three-components/PMREMGenerator.ts
```diff
@@ -27,7 +29,9 @@ import {encodings, getDirectionChunk, getFaceChunk, texelIO} from './shader-chun
 export const generatePMREM =
```
cdata: This approach is becoming sophisticated enough that some additional detail would be really helpful here for future readers.

src/three-components/CachingGLTFLoader.ts
```diff
@@ -13,15 +13,13 @@
  * limitations under the License.
  */
 
-import {Scene} from 'three';
-import {Mesh} from 'three';
```
cdata: 👍

src/three-components/shader-chunk/common.glsl.js
```diff
@@ -9,19 +9,24 @@ import {texelConversions} from './encodings_pars_framgment.glsl.js'
 export const getDirectionChunk = /* glsl */ `
 vec3 getDirection(vec2 uv, int face) {
     uv = 2.0 * uv - 1.0;
-    vec3 direction;
+    vec3 direction = vec3(clamp(uv, -1.0, 1.0), 1.0);
+    uv = abs(uv);
+    float over = max(uv.x, uv.y) - 1.0;
+    if(over > 0.0) direction.z -= over;
```
cdata: This is an interesting tweak. What case is being handled here?

src/three-components/shader-chunk/cube_uv_reflection_fragment.glsl.js
```diff
@@ -3,26 +3,23 @@ import {getFaceChunk, getUVChunk} from './common.glsl.js';
 export const cubeUVChunk = /* glsl */ `
 #ifdef ENVMAP_TYPE_CUBE_UV
 
-#define cubeUV_faceSize (256.0)
-#define cubeUV_maxMipLevel (8.0)
+const float cubeUV_maxMipLevel = 8.0;
+const float cubeUV_minMipLevel = 3.0;
```
cdata: I almost wish these could be shared constants across this shader chunk and the PMREM generator (also `mipBase`). Or maybe uniforms? But that would be unnecessary since we can just "preprocess" them into the program source here..

src/three-components/shader-chunk/normalmap_pars_fragment.glsl.js
```diff
@@ -0,0 +1,44 @@
...
+		uniform mat3 normalMatrix;
+
+	#else
+
+		// Per-Pixel Tangent Space Normal Mapping
+		// http://hacksoflife.blogspot.ch/2009/11/per-pixel-tangent-space-normal-mapping.html
+
+		vec3 perturbNormal2Arb( vec3 eye_pos, vec3 surf_norm ) {
+
+			// Workaround for Adreno 3XX dFd*( vec3 ) bug. See #9988
```
cdata: Is this file forked from Three.js? If so, can we document the nature of the changes? (ditto for `lights_fragment_maps.glsl.js`)

src/three-components/shader-chunk/cube_uv_reflection_fragment.glsl.js
```diff
@@ -33,34 +30,50 @@ vec3 bilinearCubeUV(sampler2D envMap, vec3 direction, float mipInt) {
...
   return mix(tm, bm, f.y);
 }
 
-vec4 textureCubeUV(sampler2D envMap, vec3 reflectedDirection, float mipLevel) {
-  float mip = max(mipLevel, defaultMipmap(reflectedDirection));
-  mip = clamp(cubeUV_maxMipLevel - mip, 0.0, cubeUV_maxMipLevel);
-  float f = fract(mip);
+vec4 textureCubeUV(sampler2D envMap, vec3 sampleDir, float roughness) {	
+  float filterMip = 0.0;
+  if(roughness >= 0.7){
```
cdata: Nit:  kinda funky formatting in this conditional, maybe match the one below?

src/three-components/PMREMGenerator.ts
```diff
@@ -234,32 +265,114 @@ void main() {
   }
 }
 
-class PackingShader extends RawShaderMaterial {
+// This is a hack because IE claims that a shader which unrolls a loop to access
+// 96 texture lookups has "complexity which exceeds allowed limits", however 48
+// seems to be fine, so we subsample. This could be improved if someone cares a
+// lot about IE.
```
cdata: I wonder if there are low-end devices that might also have issues with this.

Separately, it might be cleaner to read if this was written as a [preprocessor directive](https://threejs.org/docs/#api/en/materials/ShaderMaterial.defines). Although as I write that, I'm not 100% sure if that will prevent IE from barfing when linking the shader.

src/three-components/shader-chunk/cube_uv_reflection_fragment.glsl.js
elalish: Yeah, really they should be uniforms, but since this is internal to three.js, we'll have to wait until it's upstreamed for that.

src/three-components/shader-chunk/normalmap_pars_fragment.glsl.js
elalish: Yes, I can do that.

src/three-components/PMREMGenerator.ts
elalish: The shaders take longer to compile than they take to run, so it's very important that we don't dispose of them, since three does such a nice job caching them. Are we worried about a leak here? The programs are always identical, so three ensures they don't get duplicated. We shouldn't delete them until <model-viewer> will not be loading any new environments anymore, and basing that on page navigation sounds reasonable?

src/three-components/PMREMGenerator.ts
elalish: That's a good idea; I'll give that a try.

src/three-components/shader-chunk/common.glsl.js
elalish: Added a comment above.

src/constants.ts
cdata: 🍻

src/three-components/PMREMGenerator.ts
cdata: 💯

src/three-components/PMREMGenerator.ts
cdata: Was it too cumbersome to use Three.js' `defines` property for `ShaderMaterial` to do this?

elalish: Yeah, those are really for defining constants, not just a switch. I noticed this is the way three tends to do it internally.

src/three-components/PMREMGenerator.ts
```diff
@@ -15,19 +15,28 @@
...
  * (PMREM) from a cubeMap environment texture. This allows different levels of
  * blur to be quickly accessed based on material roughness. It is packed into a
  * special CubeUV format that allows us to perform custom interpolation so that
- * we can support nonlinear formats such as RGBE.
+ * we can support nonlinear formats such as RGBE. Unlike a traditional mipmap
+ * chain, it only goes down to the lodMin level (below), and then creates extra
+ * even more filtered mips at the same lodMin resolution, associated with higher
+ * roughness levels. These extra mips are stored at X rather than Y offsets to
+ * help distiguish them. In this way we maintain resolution to smoothly
+ * interpolate diffuse lighting while limiting sampling computation.
```
cdata: 💯

src/three-components/shader-chunk/common.glsl.js
cdata: 💯

src/three-components/shader-chunk/envmap_physical_pars_fragment.glsl.js
```diff
@@ -40,21 +40,26 @@ export const envmapChunk = /* glsl */ `
 
 	}
 
-	// taken from here: http://casual-effects.blogspot.ca/2011/08/plausible-environment-lighting-in-two.html
-	float getSpecularMIPLevel( const in float blinnShininessExponent, const in int maxMIPLevel ) {
-
-		//float envMapWidth = pow( 2.0, maxMIPLevelScalar );
-		//float desiredMIPLevel = log2( envMapWidth * sqrt( 3.0 ) ) - 0.5 * log2( pow2( blinnShininessExponent ) + 1.0 );
+	// (elalish) Changed from Blinn-Phong to Trowbridge-Reitz distribution and added anti-aliasing.
```
cdata: 💯

src/three-components/shader-chunk/envmap_physical_pars_fragment.glsl.js
```diff
@@ -40,21 +40,26 @@ export const envmapChunk = /* glsl */ `
...
-		// clamp to allowable LOD ranges.
-		return clamp( desiredMIPLevel, 0.0, maxMIPLevelScalar );
+		desiredMIPLevel = clamp( maxMIPLevelScalar - desiredMIPLevel, 0.0, maxMIPLevelScalar );
+
+		return desiredMIPLevel;
 
 	}
 
-	vec3 getLightProbeIndirectRadiance( /*const in SpecularLightProbe specularLightProbe,*/ const in GeometricContext geometry, const in float blinnShininessExponent, const in int maxMIPLevel ) {
+	// (elalish) Changed the input from blinnShininessExponent to roughness, so that we can use the Trowbridge-Reitz distribution instead.
```
cdata: 👍

src/three-components/shader-chunk/envmap_physical_pars_fragment.glsl.js
```diff
@@ -66,9 +71,16 @@ export const envmapChunk = /* glsl */ `
 
 		#endif
 
+		// (elalish) Mixing the reflection with the normal is more accurate and keeps rough objects from gathering light from behind their tangent plane.
```
cdata: 👍

src/three-components/shader-chunk/normalmap_pars_fragment.glsl.js
```diff
@@ -0,0 +1,45 @@
...
+			vec2 st1 = dFdy( vUv.st );
+
+			float scale = sign( st1.t * st0.s - st0.t * st1.s ); // we do not care about the magnitude
+
+			vec3 S = normalize( ( q0 * st1.t - q1 * st0.t ) * scale );
+			vec3 T = normalize( ( - q0 * st1.s + q1 * st0.s ) * scale );
+			vec3 N = normalize( surf_norm );
+			mat3 tsn = mat3( S, T, N );
+
+            // (elalish) Biased the normal mip to anti-alias the normal's screen-space derivatives
```
cdata: 👍

test/index.html
```diff
@@ -21,6 +21,16 @@
   <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
   <title>&lt;model-viewer&gt; Unit Tests</title>
 
+  <!-- Uncomment the following script in order to debug shader errors in Sauce Labs. When 
+       a shader fails to compile, it logs a console.error ("Shader failed to compile!") before it logs
+       the debug output from WebGL. Without this script, WCT bails out on the first error, so you 
+       never see the useful log, which tends to be especially critical for IE11. -->
```
cdata: 👍

PR #713, opened 2019-08-20
Updated fidelity testing

### LINKED ISSUE #434: Update gltf_renderer to use Filament's new gltfio library

### LINKED ISSUE #515: npm run dev doesn't pick up changes to fidelity tests

### LINKED ISSUE #655: Quality Tests link is broken or down
Visiting https://googlewebcomponents.github.io/model-viewer/test/fidelity/results-viewer.html, I see the message `Loading...` and the console shows a 404 for `config.json`.

### LINKED ISSUE #673: Filament fidelity tests do not use HDR

cdata:
This change proposes a revised test authoring scheme and unified screenshot creation for our fidelity test suite. It builds upon @elalish 's initial work to migrate us towards an exclusively browser-based screenshot creation scheme in #676 / #682 .

Critical details include:

 - Canonical test configuration is now entirely contained in the config (`test/fidelity/config.json`)
 - A linkable "scenario selection" UI has been added w/ support for arbitrary browser-based renderers
 - A `<filament-viewer>` element has been introduced to streamline Filament's usage in fidelity tests
 - Specific renderers can now be excluded from individual tests via the config
 - Individual renderers can be configured with "setup" script (via the config) that will be run prior to generating a screenshot with that renderer
   - This pattern can be extended to other lifecycle stages as needed as we move to include additional renderers
 - A platform-specific `cmgen` binary is downloaded from Github when running on Linux and OSX
 - `.jpg` IBLs are automatically converted to `.png` before being processed by `cmgen`
 - Our Travis config should be fixed to allow for proper test result deployment to Github Pages, but we won't know for sure until we land this change

Unified fidelity test viewers |
----------------------------|
newfidelityviewer |

Latest results with this change |
-------------------------------|
latestresults |

Fixes #434
Fixes #515
Fixes #655
Fixes #673


examples/assets/attributions.md
```diff
@@ -51,7 +51,7 @@ Originally in HDR format
 
 ## spruit_sunrise_1k
 
-spruit_sunrise_1k.hdr, spruit_sunrise_1k.jpg by <a href="https://hdrihaven.com">HDRI Haven</a>
+spruit_sunrise_1k.hdr, spruit_sunrise_1k.jpg, spruit_sunrise_2k_HDR.hdr, spruit_sunrise_2k_LDR.jpg by <a href="https://hdrihaven.com">HDRI Haven</a>
```
elalish: I switched us from 2k environments to 1k, since that matches our 256 cubemaps. I think adding these was just a merge conflict, since I was trying to rename them.

cdata:
There is a dropdown that lets you review incremental changes by commit e.g., https://github.com/GoogleWebComponents/model-viewer/pull/713/commits/1e33320ae47f01f4fe12d7cd7d715ec158b02599

elalish:
Oh wow, almost all your changes are just in that commit, aren't they? Cool, thanks.

elalish:
I see what you mean about the framing being off randomly. Very strange, and definitely wasn't there in my version of this PR (I spent a fair bit of time checking that it was single-pixel aligned). However, I also noticed a couple of other bugs you can see in this screenshot:
image
First, both filament and <model-viewer> get a scrollbar, but only sometimes, again at random. When they are rendering I notice that they render first at one size and then they expand to roughly double, and that's when the scrollbar appears. Sadly, doesn't seem to correlate with the issue above.

Second, Filament is using the HDR sunrise for both of these. I originally fixed this by appending _LDR, _HDR to their names because otherwise they end up in the same cache entry; I'm guessing that's what happened here.

examples/assets/attributions.md
cdata: Will fix

cdata:
Good catch, I'll fix the filename issue.

Strangely, I have never seen that scrollbar. Regarding the offset, I was actually talking about this:

image

You can see from the outline in the bottom right that the two models are offset vertically from each other. Back to back comparison also shows the difference:

b2

elalish:
Yes, I saw it too; my point is it wasn't there in my PR. I'll see if I can find the culprit as I code review.

elalish:
Have you seen this? 
image
I went to http://127.0.0.1:8000/test/fidelity/renderers/filament/index.html and noticed that at first nothing loaded but a black square. Then I changed the drop down and it started working fine, but after changing models a few times, they now all render in grayscale. Then after a new more changes, only the skybox renders. Seems to be fine for the model-viewer site though.

cdata:
Yeah, I have talked to @prideout about some of this stuff. All of the models will load and render if they are the first model, and there are various bugs that occur for subsequent models. I decided that the best path forward would be for us to land these changes and then file issues against Filament with links to broken cases.

donmccurdy:
Screen Shot 2019-08-22 at 11 30 16 AM


Those diff stats. 👏

Do you think it's strongly preferable that any future third-party renderers added here should also be rendering in the browser? Just wondering about the feasibility of diffing offline renders from something like Mitsuba or Arnold, eventually.

cdata:
For path tracers, I doesn't seem like we'll get a lot of value from maintaining our own "runners" to launch renders with them the way we did for Filament. I suspect we will get the most bang for our buck by maintaining golden renders for those.

That said, to the extent that any path tracer can scripts / worked into some kind of automation, I think we could definitely incorporate them into our golden update flow. I see this is parallel to the actual fidelity testing flow.

For everything else, it would probably be helpful to do a survey of realtime renderers that we are likely to be interested in. The ones we have talked about as low hanging fruit include:

 - Babylon
 - Cesium
 - Khronos Reference Renderer

These are all able to be run on the web. Separately, we have found ourselves compared to realtime rendering in Adobe Dimension and Substance Painter. I'm sure we will eventually be compared to similar tools in the future. These will be a bit more tricky to incorporate, but I don't want to rule out the idea of incorporating them somehow.

donmccurdy:
Agreed on just maintaining golden renders for path tracers, I don't think we need to launch renders online either. 👍

I'd be very curious whether game engines like Unity or Godot that support WebGL+WASM packaging could be used to create a viewer that's sufficient for loading models at runtime, but that's probably a big project. PlayCanvas is another to think about.

The three web libraries you mention are the better places to start though, for sure.

cdata:
Yeah, that's a very interesting idea. We are already incorporating Filament via WASM. It might be mostly a licensing question in Unity's case. We already have some experience working with the Unity glTF loader plugin.

I don't know much about Godot, but says it is open source. Do they already support runtime loading of glTFs?

donmccurdy:
Looks like Godot's glTF support is good, but not available at runtime currently: https://github.com/godotengine/godot/issues/24768

cdata:
I suppose we could just compile all the models we are testing into a single Godot binary?

More realistically, we'll probably have to wait until that issue sees some movement.

PR #714, opened 2019-08-20
update offscreencanvas type script file

yiyix:
update the type script to version 3.5, so it support offscreencanvas


yiyix:
@cdata: could you please review this change? thank you

PR #715, opened 2019-08-20
parent 2d4aaf35735397a1d89f7f5df28f8a0a080d569c

yiyix:
compiled version with unmovable models


cdata:
@yiyix I think this PR is able to be closed, but let me know if I am mistaken and we can revisit.

PR #718, opened 2019-08-23
Add some docs, CSS to hide UI

### LINKED ISSUE #554: Docs for model-visibility event are missing
I just had a peek at the docs and noticed that the `model-visibilty` event was never properly documented. It replaces the `poster-visibility` event and was added in #487 .

cdata:
This change proposes two new CSS custom properties that will enable users to completely hide the UI via CSS. It also adds docs for these properties and missing docs for the `model-visibility` event.

Fixes #554

### DIFF
--- a/index.html
+++ b/index.html
@@ -263,6 +263,14 @@
 
           <ul class="list-attribute">
             <li>
+              <div>--ar-button-display</div>
+              <p>Sets the display property of the AR button. Intended to be used to force the button to be hidden. Defaults to block.</p>
+            </li>
+            <li>
+              <div>--interaction-prompt-display</div>
+              <p>Sets the display property of the interaction prompt. Intended to be used to force the prompt to be hidden. Defaults to flex.</p>
+            </li>
+            <li>
               <div>--poster-color</div>
               <p>Sets the background-color of the poster. Falls back to --background-color if set. Defaults to #fff.</p>
             </li>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -390,6 +398,13 @@ <h3 class="grouping-title">Events</h3>
               <div>load</div>
               <p>Fired when a model is loaded. Can fire multiple times per <span class="attribute">&lt;model-viewer&gt;</span> if the src attribute is changed.</p>
             </li>
+            <li>
+              <div>model-visibility</div>
+              <p>This event is fired when the visibility of the model changes. When the
+              model is loaded, rendered and fully visible (e.g., the poster is no longer
+              visible), this event will fire and <span class="attribute">event.detail.visible</span>
+              will be "true".</p>
```
elalish: This event will still fire if the model is not in the viewport, right? We might want to mention that.

index.html
```diff
@@ -262,6 +262,14 @@ <h3 class="grouping-title">Attributes</h3>
           <h3 class="grouping-title">CSS Custom Properties</h3>
 
           <ul class="list-attribute">
+            <li>
+              <div>--ar-button-display</div>
+              <p>Sets the display property of the AR button. Intended to be used to force the button to be hidden. Defaults to block.</p>
+            </li>
```
elalish: Should we list the values of the property that are intended to be used? Will any values do something really weird?

cdata: `none` is implied by the sentence explaining that this is only meant to hide the UI. But, realistically we cannot force folks not to use other values.

PR #719, opened 2019-08-23
added furnace test

elalish:
This adds a furnace test using pbr_spheres.glb, though I had to update it to make the material white instead of grey. I thought Three.js passed the furnace test, but it turns out they only test the top row. You can see we are perfect for the top and bottom rows (conductor and non), but lose a fair bit of energy in between. Filament is only right for conductors, but is off by less at maximum. 

I updated all the goldens, but the Filament ones all have the weird scrollbar I get on Linux. @cdata Can you commit corrected ones here? I also changed update-screenshots to accept a whitelist of scenarios rather than renderers, and changed the names of a few tests to avoid spaces, in order to make these easily whitelistable from the command line.


cdata:
FYI @jsantell

elalish:
So you don't have to search through the files below, here is our furnace test result:
image
And Filament's:
image

PR #729, opened 2019-08-28
Zoom using FoV

### LINKED ISSUE #542: Camera can enter object
### Description
Really there are two problems here. The first is the near and far plane are set arbitrarily to 0.1 and 10 times the framed height, respectively, which is not enough in all cases. More importantly, zooming the camera moves it physically closer to the center of model, and thus sometimes inside of the model, which can't be fixed by the near plane. I would propose solving this by zooming in by narrowing the FOV and keeping the camera at the ideal radius. 

#### Live Demo
https://maps3d.io/blog/
Just zoom in and you'll soon see something like this:
Screenshot from 2019-05-09 09-03-50

### Browser Affected

- [x] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [x] Linux
- [ ] MacOS
- [ ] Windows

elalish:
Fixes #542 

In addition to zooming using FoV instead of radius, I also interpolate using log(FoV) in order to make it perpetually linear (same movement is the same fractional change from any starting point). I also narrowed the minimum so you can get closer to the object, since you can never enter it now.

### DIFF
--- a/src/test/features/controls-spec.ts
+++ b/src/test/features/controls-spec.ts
@@ -204,7 +204,7 @@ suite('ModelViewerElementBase with ControlsMixin', () => {
         });
 
         test('updates with current orbit after interaction', async () => {
-          controls.adjustOrbit(0, 0.5, 0);
+          controls.adjustOrbit(0, 0.5, 0, 0);
           settleControls(controls);
 
           const orbit = element.getCameraOrbit();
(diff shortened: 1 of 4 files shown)


src/test/features/controls-spec.ts
```diff
@@ -149,7 +149,7 @@ suite('ModelViewerElementBase with ControlsMixin', () => {
       });
 
       test('defaults FOV correctly', async () => {
-        expect(element.getFieldOfView()).to.be.equal(DEFAULT_FOV);
+        expect(element.getFieldOfView()).to.be.closeTo(DEFAULT_FOV, 0.00001);
```
cdata: 👯‍♂

src/three-components/SmoothControls.ts
```diff
@@ -35,9 +35,9 @@ export interface SmoothControlsOptions {
   // The maximum angle between model-forward and the camera azimuthal position
   maximumAzimuthalAngle?: number;
   // The minimum camera field of view in degrees
-  minimumFov?: number;
+  minimumLogFov?: number;
```
cdata: `fieldOfView`

cdata: Separately, its incongruous for us to ask the user to think about "log field of view" here when configuring the controls, but separately set field of view directly in a method like `setFov`. It seems like we could get away with asking the user to set a desired range in degrees and we could `log(fov)` as an internal detail.

src/test/three-components/SmoothControls-spec.ts
```diff
@@ -280,13 +280,17 @@ suite('SmoothControls', () => {
           test('does not zoom when scrolling while blurred', () => {
```
cdata: Let's add test coverage for configuring min/max field of view.

src/three-components/SmoothControls.ts
elalish: Good call.

src/test/three-components/SmoothControls-spec.ts
```diff
@@ -237,6 +237,25 @@ suite('SmoothControls', () => {
         });
       });
 
+      suite('field of view', () => {
```

pushmatrix:
Amazing!!

PR #733, opened 2019-08-29
Camera target control

### LINKED ISSUE #453: Configuration to control camera target
As of #452 , we will support user configurable camera orbit positions. However, this is limited because the camera will always face the same target (currently fixed to the scene origin).

We should support an attribute (perhaps called `camera-target`) that allows this "target" to be changed.

### LINKED ISSUE #603: Deprecate / remove align-model attribute
In #441 , we introduced the `align-model` attribute as an escape-hatch for users who want to stage their models in a special way and did not want to use the default staging.

Once we implement #453 , `align-model` won't be as useful, and the escape hatch use case will mostly be covered by `camera-target`. At that point, we should deprecate and remove `align-model`.

### LINKED ISSUE #679: Object is black on some Android devices
### Description

The relevant THREE github issue on this case: https://github.com/mrdoob/three.js/issues/14137. It was opened more than year ago and there is still no solution. Model-view player based on the same THREE.js library, but it does not work on a lot of popular devices. For example, many Xiaomi phones (even flagman ones) can't work with it. Tested on Redmi 4A (Adreno 308), Mi A1 (Adreno 506), Mi6 (Adreno 540). More devices with black textures issue can be found by the link above. 

Errors in the console: 
- WEBGL_depth-texture extension not supported
- OES_texture_float_linear extension not supported
- EXT_shader_texture_lod extension not supported.

Screenshot of our own [test](https://image.faradise.ru/arcore.html) , but can also be produced on the main page with astronaut. 
ShooterScreenshot-22-18-07-19

### Browser Affected

- [x] Chrome
- [x] Firefox

I guess all available for Android platform browsers are affected. 

### OS

- [x] Android

### Additional information
1. The problem is in MeshStandartMaterial (https://threejs.org/docs/#api/en/materials/MeshStandardMaterial). We use difuse, normal, metallic and roughness maps with this shader and get black texture problem in our own project with THREE.js too. I think the astronaut in model-viewer inherits the same issue. 
2. From the documenation about MeshStandartMaterial we can get that it's about "physically based rendering (PBR) has recently become the standard in many 3D applications, such as Unity, Unreal and 3D Studio Max.". Ok, I'm Unity developer, so let's do the test. 
  - Open Unity 2019.1
  - Create project with LightWeightPipeline
  - Choose shader for material object similiar to THREE's MeshStandartMaterial. I'll attach the screenshots from Unity and THREE editor. Shaders parametres are very similiar. 
  - Make a build
  - That's it, rendering is fine! 

Unity-PBR-LWP
THREE-Editor-PBR

So, I supppose the phones are OK with PBR and can handle diffuse/normal/metallic/rougness shader. THREE's developers think that there is nothing they can do. 

I hope google engineers can handle this issue together with THREE developers, because Web AR really depends on it.

elalish:
Fixes #453 
Fixes #603 
Fixes #679 

In the process of removing align-model I was able to simplify our logic considerably, including cleaning up how the static shadow is placed, since I was having trouble following its logic as I was debugging. The target is interpolated along with the other camera parameters, so this should allow us to do a smooth transition that includes zooming, position and rotation, and once combined with #729, without going through the model. 

I only removed a function from staging.ts, but it seems it wasn't formatted before? Because everything got reflowed. I'm assuming it's correct now...

#679 turned out to be mostly not a <model-viewer> bug, but we noticed that it did show an incorrect shadow for large models. The refactor of StaticShadow here should fix this.

### DIFF
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -280,10 +280,15 @@ export const ControlsMixin = (ModelViewerElement:
         }
 
         [$updateCameraTarget]() {
-          let target = deserializeVector3(this.cameraTarget);
+          const targetValues = deserializeVector3(this.cameraTarget);
+          let target = this[$scene].model.boundingBox.getCenter(new Vector3);
 
-          if (target == null) {  // this happens in the default 'auto' case
-            target = this[$scene].model.boundingBox.getCenter(new Vector3);
+          if (targetValues != null) {
+            for (let i = 0; i < 3; i++) {
+              if (targetValues[i] !== 'auto') {
+                target.setComponent(i, targetValues[i] as number);
+              }
+            }
           }
 
           this[$controls].setTarget(target);
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -14,9 +14,9 @@
  */
 
 import {property} from 'lit-element';
-import {Event, Spherical} from 'three';
+import {Event, Spherical, Vector3} from 'three';
 
-import {deserializeAngleToDeg, deserializeSpherical} from '../conversions.js';
+import {deserializeAngleToDeg, deserializeSpherical, deserializeVector3} from '../conversions.js';
 import ModelViewerElementBase, {$ariaLabel, $needsRender, $onModelLoad, $onResize, $scene, $tick} from '../model-viewer-base.js';
 import {ChangeEvent, ChangeSource, SmoothControls} from '../three-components/SmoothControls.js';
 import {Constructor} from '../utilities.js';
@@ -276,20 +276,12 @@ export const ControlsMixin = (ModelViewerElement:
         }
 
         [$updateCameraTarget]() {
-          let target = deserializeSpherical(this.cameraOrbit);
+          let target = deserializeVector3(this.cameraTarget);
 
-          if (target == null) {
-            target = deserializeSpherical(DEFAULT_CAMERA_ORBIT)!;
+          if (target == null) {  // this happens in the default 'auto' case
+            target = this[$scene].model.boundingBox.getCenter(new Vector3);
           }
 
-          if (typeof radius === 'string') {
-            switch (radius) {
-              default:
-              case 'auto':
-                radius = this[$idealCameraDistance]!;
-                break;
-            }
-          }
           this[$controls].setTarget(target);
         }
 
@@ -370,7 +362,6 @@ export const ControlsMixin = (ModelViewerElement:
           controls.applyOptions({minimumRadius, maximumRadius});
 
           controls.setRadius(zoom * this[$idealCameraDistance]!);
-          controls.setTarget(scene.target);
           controls.jumpToGoal();
         }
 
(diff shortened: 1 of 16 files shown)


src/features/controls.ts
```diff
@@ -166,6 +172,10 @@ export const ControlsMixin = (ModelViewerElement:
           return {theta, phi, radius};
         }
 
+        getCameraTarget(): Vector3 {
+          return this[$controls].getTarget();
```
cdata: 👍

src/features/controls.ts
```diff
@@ -265,16 +279,23 @@ export const ControlsMixin = (ModelViewerElement:
           this[$controls].setOrbit(theta, phi, radius as number);
         }
 
+        [$updateCameraTarget]() {
+          let target = deserializeVector3(this.cameraTarget);
+
+          if (target == null) {  // this happens in the default 'auto' case
```
cdata: Let's deserialize to `'auto'` in this case. The code will be more self-explanatory, and it will line up with what we do for deserializing sphericals.

src/model-viewer-base.ts
```diff
@@ -114,6 +116,10 @@ export default class ModelViewerElementBase extends UpdatingElement {
     return this[$loaded];
   }
 
+  get loadedTime() {
```
cdata: No need to make this public, is there?

src/three-components/SmoothControls.ts
```diff
@@ -42,6 +42,8 @@ export interface SmoothControlsOptions {
   eventHandlingBehavior?: EventHandlingBehavior;
   // Controls when interaction is allowed (always, or only when focused)
   interactionPolicy?: InteractionPolicy;
+  // The approximate size of the model (m)
+  modelSize?: number;
```
cdata: Can you offer more context for this new configuration parameter?

src/features/controls.ts
cdata: Relatedly, it seems likely that I would want to be able to set `'auto'` per-axis of the vector. For example, maybe I want the target to be automatically selected for me along the `y` axis, while panning the camera along the `x` axis.

src/model-viewer-base.ts
elalish: Agreed, will fix.

src/features/controls.ts
elalish: Yeah, I was on the fence about that already. Sounds good.

src/three-components/SmoothControls.ts
elalish: Nevermind, that's not needed. I can just reuse maximumRadius.

src/test/features/controls-spec.ts
```diff
@@ -407,11 +422,9 @@ suite('ModelViewerElementBase with ControlsMixin', () => {
 
             canvas.focus();
 
-            await timePasses(DEFAULT_INTERACTION_PROMPT_THRESHOLD / 2.0);
-
             interactWith(canvas);
 
-            await timePasses(DEFAULT_INTERACTION_PROMPT_THRESHOLD + 100);
+            await timePasses(element.interactionPromptThreshold + 100);
```
cdata: 💯

PR #739, opened 2019-09-04
Scroll-correlated camera configuration properties

### LINKED ISSUE #600: Tilt effect on scroll
Another subtle effect for showing that what is in the viewer is realtime 3D is to have the model tilt slightly on scroll.  See demo here: https://panic.com/transmit/

tilt

One consideration is not too tilt it too much because the underside of models are not always the things you want to focus on. For example for products like a table, seeing the bottom of the table might not be where you want to draw user's eyes too since it might be a weird angle to show the product in.

In the truck example above it's nice that the truck already starts with the camera looking down on it, so you never get to a point where you're staring at the bottom of the truck.

Complimentary feature to: #595

### LINKED ISSUE #828: Interaction prompt displays even if src is null
Steps to reproduce:

1. Open a page with a properly configured `<model-viewer>` with a valid `src`
2. Before the interaction prompt displays, set the `src` to `null`
3. The interaction prompt displays anyway

cdata:
calc2

This change proposes the introduction of a declarative, CSS-like configuration syntax for `<model-viewer>` attributes and properties. The initial goal of this change is to enable users to correlate the camera's orbital position with the top-level scroll position on the page. The long-term goal of this change is to establish a common, intuitive, web-native syntax for describing complex attribute/property configuration schemas in `<model-viewer>` as it evolves.

**The main user-facing change is that CSS-like [`calc()`](https://developer.mozilla.org/en-US/docs/Web/CSS/calc) and [`env()`](https://developer.mozilla.org/en-US/docs/Web/CSS/env) functions can now be used when configuring `camera-orbit`.** Additionally, a single "environment variable" called `window-scroll-y` is available to be used by the `env()` function, which yields a value between `0` and `1` derived from the current global scroll position.

Previously, we used a very basic string parsing mechanism to parse `camera-orbit` values in order to support CSS-like numbers with units (e.g., `10deg`) and keywords like `auto`. This change expands the parsing facilities and layers on systems for evaluating the parsed result and observing signals that tell when the parsed result should optimally be evaluated.

A trivial usage of this new syntax looks like this:

```html
<model-viewer camera-orbit="calc(-1.5rad + env(window-scroll-y) * 4rad) 90deg auto"
              src="assets/Astronaut.glb"
              background-color="#622dcf"
              alt="A 3D model of an astronaut">
</model-viewer>
```

Some details of the new syntax include:

 - **All `calc()` algebra is supported**, including nested `calc()` terms
    - We currently do not enforce any maximum nesting depth
 - **Only supports angles and lengths**; no new unit types introduced at this time
 - **Adheres closely to CSS**, only deviating where needed
    - Everything is standard CSS with the notable exception of `window-scroll-y`
 - **No breaking changes required** to inherit the new capabilities; the new syntax is additive

## New `@style` decorator

A new `@style` decorator has been introduced in order to make it relatively easy to integrate sophisticated style parsing to any property already using the `@property` decorator from the `lit-element` module.

Use of the `@style` decorator looks like this:

```typescript
// Intrinsics describe the metrics required in order to parse foo,
// including basis values for percentages, fallback values and
// substitutes for keywords. Intrinsics can optionally be specified as
// a factory function that takes an element instance and returns an
// intrinsics object: 
const fooIntrinsics = {
  basis: [numberNode(1, 'm'), numberNode(Math.PI, 'rad')],
  keywords: {auto: [null, numberNode(200, '%')]}
};

class StyleableElement extends UpdatingElement {
  @style({intrinsics: fooIntrinsics, updateHandler: 'updateFoo'})
  @property({type: String})
  foo: string = '200cm 1rad';

  updateFoo(style: [number, number]) {
    // do something with parsed value of foo
  }
}
```

## Performance analysis

Some steps have been taken to mitigate the performance cost of parsing attribute values as an expressive CSS-like syntax. That said, the performance cost of this approach is not zero and is worth analyzing.

For the purposes of this analysis, I will be using a configuration of `<model-viewer>` that stresses the style subsystem proposed here. A test page has been crafted the produces N model viewer instances in this fashion (psuedocoded):

```javascript
for (i = 0; i < N; ++i) {
  const modelViewer = document.createElement('model-viewer');
  modelViewer.src = './examples/assets/Astronaut.glb';
  modelViewer.cameraOrbit = `calc(-${i}rad + env(window-scroll-y) * 8rad) calc(180deg - env(window-scroll-y) * 180deg + 5deg * ${i}) calc(3m - env(window-scroll-y))`;
  document.body.appendChild(modelViewer);
}
```

Note that the proposed parsing/evaluation strategy caches parsed ASTs across all elements, as well as the evaluated results of static sub-trees of the AST on a per-element basis. In this test case, every element will have a unique value for `camera-orbit`, so the parser cache will be subverted.

The test case has been run with various values of N on a Google Pixel 3 phone and a 2013 MacBook Pro in Chrome. The majority of the meaningful work occurs when updating the `cameraOrbit` property, so the `updateCameraOrbit` and `syncCameraOrbit` methods were instrumented to measure the work.

### Startup costs

The upfront parsing costs of a single element on the page is quite small in the grand scheme things, but still very significant. On a MacBook Pro, initial cost of `updateCameraOrbit` grows from `0.47ms` to `1.86ms`, and on a Pixel 3 it grows from `0.8ms` to `3.6ms`.

Branch | MacBook Pro (N=1) | Pixel 3 (N=1)
--------|--------------------|-------------
`master` | image|image
`scroll-correlated-camera-orbit`|image|image

While this time is clearly dwarfed by the time we spend preparing and rendering a 3D scene, we should consider these times in the context of when they will matter most. In the future, we will probably lazy load most of our rendering implementation so that 3D content can defer to other, more critical content on the page. In this future, we will have compartmentalized most of the cost of this element, but will still ask users to pay the cost of attribute/property parsing up front. If we apply this parsing scheme to other properties as well, it seems likely that the parsing / evaluation time will start to add up.

I also checked the scaling properties of this boot-up cost when multiple elements are on the page. The total cost scales linearly with the number of elements on the page, which makes sense given the stated conditions of this test.

MacBook Pro | Pixel 3
--------------|-------
image|image

Notably, the cost of `updateCamerOrbit` for N=1 is significantly higher than subsequent instances, which suggests that some significant amount of cost is paid upfront that I haven't quite put my finger on yet. It could be the cost of adding a listener for the scroll event (which only happens once for all elements), but that seems unlikely to explain all of the extra time. In this related data for the MacBook Pro, note the average time for N=1 compared to other sample counts (similar results were seen on the Pixel 3):

Element count | Sample size | Average time (ms) | Total time (ms)
-- | -- | -- | --
1 | 3 | 1.12167 | 3.365
5 | 15 | 0.38867 | 5.83
10 | 30 | 0.298 | 8.94
30 | 90 | 0.27767 | 24.99
100 | 300 | 0.37107 | 111.32

### Scroll synchronization costs

Additionally, I did some testing of the cost to synchronize scroll in real time. The cost to synchronize to scroll proved to be very small, so I only ran some light tests on my MacBook Pro. In these tests, I created a bunch of elements and then scrolled down the page to the bottom while collecting samples:

Method | Element count | Sample size | Average time (ms) | Total time (ms)
-- | -- | -- | -- | --
updateCameraOrbit | 10 | 30 | 0.32183 | 9.655
syncCameraOrbit | 10 | 2370 | 0.04226 | 100.15
updateCameraOrbit | 100 | 300 | 0.40555 | 121.665
syncCameraOrbit | 100 | 49600 | 0.31537 | 15642.15

We can see that the cost of `syncCameraOrbit`, which represents the cost of evaluating the AST (including incorporating dynamic scroll position values on each frame) is quite small compared to `updateCameraOrbit` (which must also parse configuration values). But, the `updateCameraOrbit` cost is a one-time startup cost that only needs to be paid when `camera-orbit` is assigned directly.

## Additional thoughts

There are a few reasons why we probably want to introduce a syntax like this to our attribute configurations:

**It gives us the flexibility to keep our API surface concise.** As time goes on, our configuration API must strike a balance between allowing complex attribute values and proliferating attribute names.

**It leverages existing user know-how.** CSS is a lingua franca of the web. Since this syntax carries with it most of the semantics of CSS, our users don't require special knowledge in order to use it.

**It scales well to different use cases.** CSS expressions can be as simple as a word (e.g., `blue`), but they also allow you to express lists, typed numbers, functions and bindings.

**It supports our goal to explore and experiment in this domain.** Over time, a syntax like this will inform our conception of a CSS-based style API as it might apply to a `<model>` tag that is built-in to the browser.

I hope that over time we will begin to express other attributes such as `field-of-view`, `camera-target` and perhaps many others in terms of the syntax afforded by this change.

Fixes #600 
Fixes #828


src/features/controls.ts
```diff
@@ -239,21 +249,15 @@ export const ControlsMixin = (ModelViewerElement:
         }
 
         [$updateFieldOfView]() {
-          let fov = deserializeAngleToDeg(this.fieldOfView);
+          let fov = deserializeAngle(this.fieldOfView);
```
elalish: Might be best to include the second argument here to make the units more clear.

src/styles/deserializers.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Parses an angle CSS number string and deserializes it as a number with the
+ * given unit type. For example:
+ *
+ * deserializeAngle('1rad', 'rad'); // Returns 1
+ * deserializeAngle('1rad', 'deg'); // Returns 57.29577951308232
+ * deserializeAngle('180deg', 'rad'); // Returns 3.141592653589793
+ */
+export const deserializeAngle =
+    (angleString: string, unit: 'deg'|'rad' = 'deg'): number|null => {
```
elalish: Perhaps we should remove the default to enforce descriptive units where it is used.

src/styles/style-effector.ts
```diff
@@ -0,0 +1,191 @@
...
+const $computeStyleCallback = Symbol('computeStyleCallback');
+const $astWalker = Symbol('astWalker');
+const $dependencies = Symbol('dependencies');
+
+const $scrollHandler = Symbol('scrollHandler');
+const $onScroll = Symbol('onScroll');
+
+/**
+ * The StyleEffector is configured with a callback that will be invoked at the
+ * optimal time that some array of CSS expression ASTs ought to be evaluated.
```
elalish: Speaking of avoiding contractions, what is AST?

src/styles/evaluators.ts
```diff
@@ -0,0 +1,417 @@
...
+ * external state that changes as the user scrolls, so it should always be
+ * re-evaluated to ensure we get the most recent value.
+ *
+ * Some important notes about this feature include:
+ *
+ *  - There is no such thing as a "window-scroll-y" CSS environment variable in
+ *    any stable browser at the time that this comment is being written.
+ *  - The actual CSS env() function accepts a second argument as a fallback for
+ *    the case that the specified first argument isn't set; our syntax does not
+ *    support this second argument.
```
elalish: A little more description here of what window-scroll-y does and how it relates to similar constructs would be nice here.

src/styles/evaluators.ts
```diff
@@ -0,0 +1,417 @@
...
+
+  get isConstant(): boolean {
+    return false;
+  };
+
+  [$evaluate](): NumberNode {
+    if (this[$identNode] != null) {
+      switch (this[$identNode]!.value) {
+        case 'window-scroll-y':
+          const verticalScrollPosition = window.pageYOffset;
```
elalish: Why do we want to have the calculation based on the absolute scroll position? It seems like it would be better UX to have it get updated in a relative way based on scrolling. It also seems like it would be simpler to accomplish this by intercepting a scroll event, but I assume there's a reason that doesn't work well?

cdata: We do accomplish this by intercepting a `scroll` event. The `scroll` event is the signal that causes this code path to run (after the first time, that is). But, [the event object dispatched during scroll is a basic `Event` and doesn't give you any special information.](https://developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event).

src/features/controls.ts
cdata: Good call, will fix.

src/styles/deserializers.ts
cdata: SGTM

src/styles/style-effector.ts
cdata: ["abstract syntax tree"](https://en.wikipedia.org/wiki/Abstract_syntax_tree)

For us CS/engineer types, it is a very common acronym (and "abstract syntax tree" is a bit of a mouthful).

src/styles/evaluators.ts
cdata: 👍

src/styles/evaluators.ts
elalish: This absolute behavior is growing on me. It is kind of nice to have this automatic viewport reset built in if someone gets lost zooming in on a model.

src/features/controls.ts
```diff
@@ -238,45 +283,26 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-      let fov = deserializeAngleToDeg(this.fieldOfView, DEFAULT_FOV_DEG);
-      this[$controls].setFieldOfView(fov!);
+    [$syncFieldOfView](style: EvaluatedStyle<Intrinsics<['rad']>>) {
+      this[$controls].setFieldOfView(style[0] * 180 / Math.PI);
     }
 
-    [$updateCameraOrbit]() {
-      let sphericalValues = deserializeSpherical(
-          this.cameraOrbit,
-          sphericalDefaults(this[$scene].model.idealCameraDistance));
```
elalish: :+1:

src/features/controls.ts
```diff
@@ -202,8 +247,8 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
       this[$controls].removeEventListener('change', this[$changeHandler]);
     }
 
-    updated(changedProperties: Map<string, any>) {
-      super.updated(changedProperties);
+    updated(changedProperties: Map<string|number|symbol, unknown>) {
+      super.updated(changedProperties as Map<string, any>);
```
elalish: Where did number, symbol and unknown come from? What's the difference between unknown and any?

src/features/controls.ts
```diff
@@ -238,45 +283,26 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-        this[$updateCameraOrbit]();
-      }
-
-      if (changedProperties.has('cameraTarget')) {
-        this[$updateCameraTarget]();
-      }
-
-      if (changedProperties.has('fieldOfView')) {
-        this[$updateFieldOfView]();
-      }
```
elalish: :+1:

src/styles/conversions.ts
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Ensures that a given number is expressed in radians. If the number is already
+ * in radians, does nothing. If the value is in degrees, converts it to radians.
+ * If the value has no specified unit, the unit is assumed to be radians. If the
+ * value is not in radians or degrees, the value is resolved as 0 radians.
+ */
+export const degreesToRadians =
+    (numberNode: NumberNode, defaultRadians: number = 0): NumberNode => {
+      let {number, unit} = numberNode;
+      if (isNaN(number) || !isFinite(number)) {
```
elalish: I believe NaN is already not finite.

src/styles/conversions.ts
```diff
@@ -0,0 +1,141 @@
...
+import {NumberNode, ZERO} from './parsers';
+
+/**
+ * Ensures that a given number is expressed in radians. If the number is already
+ * in radians, does nothing. If the value is in degrees, converts it to radians.
+ * If the value has no specified unit, the unit is assumed to be radians. If the
+ * value is not in radians or degrees, the value is resolved as 0 radians.
+ */
+export const degreesToRadians =
+    (numberNode: NumberNode, defaultRadians: number = 0): NumberNode => {
```
elalish: Perhaps change defaultRadians to fallbackRadiansValue and put a comment matching the lengthToBaseMeters comment below?

src/styles/conversions.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    if (unit == null) {
+      return node;
+    }
+
+    const normalize = unitNormalizers[unit];
+
+    if (normalize == null) {
+      return fallback;
```
elalish: It seems like we have duplicated logic here. This function handles the fallback as well as the unit conversion switches, yet the functions it calls into also handle both of these things. Could one or the other be simplified?

src/styles/deserializers.ts
```diff
@@ -0,0 +1,60 @@
...
+ *   'Monday',
+ *   'Tuesday',
+ *   'Wednesday',
+ *   'Thursday',
+ *   'Friday',
+ *   'Saturday',
+ *   'Sunday'
+ * ]);
+ */
+export const enumerationDeserializer = <T extends string>(allowedNames: T[]) =>
```
elalish: Topic for another day perhaps, but I notice this whole file is only used in one place, for distinguishing which quickLookBrowser we're willing to use. I'm not entirely sure why we do that? Curious if we have an opportunity to remove some code at some point...

src/styles/evaluators.ts
```diff
@@ -0,0 +1,561 @@
...
+ * one-time AST preparation work from work that necessarily has to happen upon
+ * each evaluation.
+ */
+export abstract class Evaluator<T> {
+  /**
+   * An Evaluatable is a NumberNode or an Evaluator that evaluates a NumberNode
+   * as the result of invoking its evaluate method. This is mainly used to
+   * ensure that CSS function nodes are cast to the corresponding Evaluators
+   * that will resolve the result of the function, but is also used to ensure
+   * that an percentage nested at arbitrary depth in the expression will always
```
elalish: nit: an -> a

src/test/decorators-spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  basis: [numberNode(1, 'm'), numberNode(Math.PI, 'rad')],
+  keywords: {auto: [null, numberNode(200, '%')]}
+};
+
+class StyleableElement extends UpdatingElement {
+  @style({intrinsics: fooIntrinsics, updateHandler: 'updateFoo'})
+  @property({type: String})
+  foo: string = '200cm 1rad';
+
+  fooUpdates: Array<[number, number]> = [];
```
elalish: Any reason for this to be an array and not just a couple of numbers that update?

src/test/features/controls-spec.ts
```diff
@@ -243,9 +244,10 @@ suite('ModelViewerElementBase with ControlsMixin', () => {
           element.cameraOrbit = cameraOrbit;
           const fieldOfView = 30;
           element.fieldOfView = `${fieldOfView}deg`;
+          console.log('JUMP!');
```
elalish: Is this still needed?

src/test/styles/evaluators-spec.ts
```diff
@@ -0,0 +1,294 @@
...
+          'calc', [expressionNode([
+            numberNode(1, null),
+            operatorNode('+'),
+            functionNode('calc', [expressionNode([numberNode(1, null)])])
+          ])]));
+
+      expect(evaluator.isConstant).to.be.true;
+    });
+
+    test('is not constant if any nested function is not constant', () => {
```
elalish: What if you have a percentage of window-scroll-y? I saw a comment earlier that makes me think this would return const. Also, does percentage of window-scroll-y actually make any sense, since it seems redundant?

src/three-components/SmoothControls.ts
```diff
@@ -443,8 +443,8 @@ export class SmoothControls extends EventDispatcher {
   /**
    * Sets the target the camera is pointing toward
    */
-  setTarget(target: Vector3) {
-    this[$goalTarget].copy(target);
+  setTarget(x: number, y: number, z: number) {
+    this[$goalTarget].set(x, y, z);
```
elalish: :+1:

src/features/controls.ts
cdata: From [this Stack Overflow answer](https://stackoverflow.com/questions/51439843/unknown-vs-any)

> unknown which is the type-safe counterpart of any. Anything is assignable to unknown, but unknown isn't assignable to anything but itself and any without a type assertion or a control flow based narrowing. Likewise, no operations are permitted on an unknown without first asserting or narrowing to a more specific type.

In other words, `unknown` is an actual type (technically the type that includes all possible types) whereas `any` is just a dark void in which all type information is discarded.

This is probably preferable to `any` in many cases. I guess we should use it where we can.

The reason why we changed the type signature here was to satisfy a TypeScript compiler check. The actual signature for this method is the new one, and our decorator operates on any derivative of `UpdatingElement`, so it was causing problems that our signatures were not  lining up.

src/test/decorators-spec.ts
cdata: Just to assert that we don't invoke it more times than expected.

src/test/features/controls-spec.ts
cdata: Nope, good catch.

src/test/styles/evaluators-spec.ts
cdata: Mathematically it makes sense. Say I have the following expression: `calc(50% + 50% * env(window-scroll-y))`.

For a basis of `1m` this would work out to `0.5m` at the minimum scroll position and `1m` at the maximum scroll position.

src/styles/conversions.ts
cdata: Yes, but JavaScript also has `Infinity` and `isNaN(Infinity) == false` 🙃

src/styles/conversions.ts
cdata: I'll take a stab at it

src/styles/deserializers.ts
cdata: Yah, this seems ripe for removal or refactor, but it's not obvious what to do and seemed orthogonal to this change.

src/decorators.ts
```diff
@@ -0,0 +1,157 @@
...
+
+          [$evaluateAndSync]: {
+            value: function() {
+              if (this[$styleEvaluator] == null) {
+                return;
+              }
+
+              const result = this[$styleEvaluator].evaluate();
+
+              (this as unknown as
```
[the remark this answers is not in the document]
cdata: don't ask me hard questions I just work here

Seriously though, this is the only way I could figure out to do this kind of dynamic dispatch "correctly." I collaborated with our local TypeScript gurus on this one and the recommendation was that there may be a more sound alternative but it will be extremely difficult to write correctly.

src/styles/conversions.ts
elalish: Sure, but isFinite(NaN) is false, so my point was that !isFinite(number) should be enough on its own.

cdata: Will fix

PR #747, opened 2019-09-12
Light cleanup for various small code issues

cdata:
This change consists mainly of cleanup, a lot of which is necessary or prudent for us to bring `<model-viewer>` into Google's internal code repository. Changes include:

 - Adding previously omitted type annotations
 - Ensuring some stricter TypeScript rules are satisfied
 - Awaiting or using promises in some cases where they were discarded
 - Re-arranging code to avoid unnecessary dependency cycles

Perhaps most notably, the combined class that makes up the standard `<model-viewer>` element is now correctly typed, meaning that TypeScript-aware code editors can provide intellisense-style contextual documentation inline with usages of `<model-viewer>`.

I recommend that reviewers view the files with [Github's whitespace filtering enabled](https://github.com/GoogleWebComponents/model-viewer/pull/747/files?w=1), as most of the PR's diff consists of indentation / formatting changes. When whitespace deltas are eliminated, the change is significantly smaller:
image


src/model-viewer.ts
```diff
@@ -21,27 +21,18 @@ import {LoadingMixin} from './features/loading.js';
...
-  FocusVisiblePolyfillMixin,
-  AnimationMixin,
-  LoadingMixin,
-  ARMixin,
-  ControlsMixin,
-  EnvironmentMixin,
-  StagingMixin,
-  MagicLeapMixin
-];
+export type ModelViewerElement = InstanceType<typeof ModelViewerElement>;
```
cdata: @rictic @aomarks could you comment on whether or not this is a reasonable approach to ensuring this element implementation is properly typed, given our usage of class mixins?

src/features/controls.ts
```diff
@@ -104,380 +104,375 @@ export interface ControlsInterface {
...
-        protected[$promptTransitionendHandler] = () =>
-            this[$onPromptTransitionend]();
+    protected[$promptTransitionendHandler] = () =>
+        this[$onPromptTransitionend]();
 
-        constructor() {
-          super();
-          const scene = (this as any)[$scene];
+    constructor(...args: any[]) {
+      super(...args);
```
elalish: What are these new args?

src/features/loading.ts
```diff
@@ -85,337 +76,333 @@ const $onClick = Symbol('onClick');
...
+    }
+
+    [$getModelIsVisible]() {
+      return super[$getModelIsVisible]() || this[$modelIsVisible];
+    }
+
+    [$getLoaded]() {
+      const src = this.src;
+      return super[$getLoaded]() ||
+          !!(src && CachingGLTFLoader.hasFinishedLoading(src));
```
elalish: I know the logic of these functions is unchanged, but it's still odd to me. Why super or this?

src/model-viewer-base.ts
```diff
@@ -295,6 +298,19 @@ export default class ModelViewerElementBase extends UpdatingElement {
...
+  // order to support overloading when TypeScript is compiled to ES5
+  // @see https://github.com/Polymer/lit-element/pull/745
+  // @see https://github.com/microsoft/TypeScript/issues/338
+  [$getLoaded](): boolean {
+    return this[$loaded];
+  }
+
+  // @see [$getLoaded]
+  [$getModelIsVisible](): boolean {
+    return true;
```
elalish: :disappointed:

src/three-components/ModelScene.ts
```diff
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import {BackSide, BoxBufferGeometry, Camera, Color, Object3D, PerspectiveCamera, Scene, Shader, ShaderLib, ShaderMaterial} from 'three';
+import {BackSide, BoxBufferGeometry, Camera, Color, Event as ThreeEvent, Object3D, PerspectiveCamera, Scene, Shader, ShaderLib, ShaderMaterial} from 'three';
```
elalish: :+1:

src/model-viewer.ts
aomarks: Looks good to me! I hadn't seen the `InstanceType<typeof MixinApplication>` pattern before -- that's really great! It solves the problem where the result of a mixin application call is a value but not a type in a really simple way.

src/features/animation.ts
```diff
@@ -23,121 +23,131 @@ const MILLISECONDS_PER_SECOND = 1000.0
...
-              if (this[$paused]) {
-                model.updateAnimation(0);
-                this[$needsRender]();
-              }
-            }
-          }
-
-          return AnimationModelViewerElement;
-        };
+export interface AnimationInterface {
```
aomarks: You might be able to avoid explicitly defining the mixin interfaces. TypeScript should infer the correct return type of the mixin based on the class returned from the function if you just leave off the return type:

```ts
const MixinA = <T extends Constructor<HTMLElement>>(Base: T) => {
  return class InnerClass extends Base {
    static aStatic = '';
    aInstance = '';
  }
}
```

If you do need an explicit reference to the interface of that inner mixin class, then it looks like a pattern like this could work:

```ts
// Extract the class defined by the mixin:
type MixinAConstructor = ReturnType<typeof MixinA>;
type MixinAInstance = InstanceType<ReturnType<typeof MixinA>>;
declare const ac: MixinAConstructor;
declare const ai: MixinAInstance;
ac.aStatic;
ai.aInstance;
```

Here's a playground where I've been trying out some mixin patterns:

http://www.typescriptlang.org/play/#code/C4TwDgpgBAwg9gOwM7AE4FcDGw6oDwAqAfFALxQDeUCEA7lABQB0LAhqgOZIBcUrCIANoBdAJS8CUAL4BuAFBzMiFFACyASwAe6hAEEyUQlAibgEBABMksZWiw58ACQKqAMgFEANhAC254EREDABCrEgQEqJkJBRyUFCoEMDoqAhQmJ5h1gCSCDSoMJlI1iZmltah4ZRx8VAorMDqmHwAysANTQYA5F3ytXy59QiY0OQ9fbVgqHBm2BAWfAAK07NmC2O9NVJy23IA9HtQ7qaorNhQwAAW0Dr5ALQ6ZqgAZmfQcM8X11A+WjpQFggzx06kaiDkoEgaj+engyDs2FwBgASkkUggCOAIHhIRAPtDtHoiPJcQSdLpBu1hqMoJT+CNMZA8KjkqlGdjcfiNITdIF5ICMuxoEp4XxMLxueS4SgMIjUPyIILEulbHx1BKYRT4fSIPIzkxWG0Opg9eoDXTqfJFKrJQhggYjKVzFYbPDZQ48M43F5fP5AiEwhEoAQoqQYjVEqy0oLirS8hACkUSqZnRVA9V+vVGs0AEZG7PdTb9HMWkaFibbXYx6yqECVaBO8pQL0ebx+BDADN1doFnOB-OdDYTPvhUs08Y7BQilS23RgMCedTzAy24IMWfruuB0SiK0HKAAdVBlzg6E7V3USAANGS9PPF8vMPwoIhPCAoDnoOhwgswnwLliN4IDM-4AG6sJ46AQEwEJYrec4Lku6xxkMDJYjiWJcpq95IcSCizjh8xMCOEADiacgEYhRGGj2TTyJRD4WMRZHyNOnaaBqPKEchND0AxSEMLuciaMRgZjvIImsOJwnEdJihJkcKaWLaxhKS6-HLrE8RZp0jYWGR5Y1HpY6GfEHBJKpZQWMcYBwOEywzIqayCZQCRoqkXyXgaDmrPMMjSPIuzHFZtqieELFyMFzqhTRxryFFykwsxtHkQlFihXpEVsVA77kLxikhTCgnyCAYUQNJpVSdqlpyKVJbVSMJVMMZDW6rVzVqTZdkQD5Tl+UAA

src/features/controls.ts
cdata: The stricter checks of the Google TypeScript compiler mark this as a case where we are using a mixin pattern. The rule for mixin class extension like this is that constructor args _must_ be blanket forwarded to the super constructor.

src/features/loading.ts
cdata: I agree, the logic here is somewhat inscrutable. The intention is to express the logic, "true if the model is already known to be loaded by the super class, or is known to be loaded by the loader." And, the reason why those states might not line up is kind of dubious and deserves critical review.

src/model-viewer-base.ts
cdata: 🙁

src/features/animation.ts
cdata: Thanks, that's an awesome insight. I would love it if we could omit the extra interfaces.

src/features/controls.ts
elalish: I see, maybe my question would be better expressed as why does only this Mixin seem to have a constructor? Is it even necessary?

src/features/animation.ts
cdata: Amazing. This totally works. Thanks @aomarks 🎉

src/features/controls.ts
cdata: No, it probably isn't. I think some of this stuff could be re-written as class field initializers, and the rest ought to be moved to `connectedCallback`.

src/features/animation.ts
cdata: I ran into trouble trying to use this approach while also extending `HTMLElementTagNameMap`. Here is a repro of my error with your playground implementation:

image

I don't understand where along the way the type is considered to be self-referential, though...

aomarks: Interesting, not sure what's going on there. Can you reproduce that error in the playground?

cdata: Hopefully / annoyingly, no! http://www.typescriptlang.org/play/#code/C4TwDgpgBAwg9gOwM7AE4FcDGw6oDwAqAfFALxQDeUCEA7lABQB0LAhqgOZIBcUrCIANoBdAJS8CUAL4BuAFByIADzC5gUTIhRQAsgEslehAEEyUQlGXAICACZJYWtFhz4AEgR0AZAKIAbCABbG2AiIgYAIVYkCAlRMhIKOSgoVAhgdFQEDT9ohwBJBBpUGFykBysbeygomMpklKgUVmA9TD4AZWAWtrMAcj75Rr5C5oRMaHIBocawVDhrbAhbPgAFecXrFanBhqk5fbkAeiOoHyU0VmwoYAALaCNigFoja1QAMyvoOHeb+6hAgYjFBbBB3kY9K1EIoVGobuBoPpDCZ4MhnNhcGYAErpTIIAgIvCgSA-XRAkxEeTKVSodTExHk4yjbrjSZQZn8CYEyB4HEZLLciBEhGkpFGYxheSgzC5NIaJx8TC8MUopwYDGoKUQGXsaCaNF8PTKxkc1nyK5MVhdHqYc16S2mibyGE09T67QqiJmCyVOwOVEodWuPAebz+IIhMKRaKxKAEeKkRINNL87I68rsooQEplCoXKoOWrQJLDZqtdoAI2t5f6u2GFcdbOmewOCmpcPTDh0ICLlnzfqgod8AWCCHUJaa3RrFZj1d6OxmM5ijdr8kO7dp8oNKuMYDAfj0yzMnoYO9PPZjolEzpOUAA6pDbnB0HTbnokAAaMnI3f7w8rTB+CgRA-BAKAK2gdAYhWaI+HhSAvwQBY4IAN1YPx0AgJgXThelv3FPcDyPchG0FYUSV+HdCP-SkFCov9liYJcIDnW05HoojbEtVj5A4-8mJ4uR3XUJRjR-ajiOoOh8JMCTbAYa85CUJiY0beRlNYNSlKYrScM3Tszn7WwVT7awBz4o8JzLXpfVsVjVwaWyVwXBoOHSUyqnOVQYnWBZtS2BTKFSXEsj+d9LV8zZlhkaQ1wUc4zOM8kVJiQSEqqFVuKnNp5HSuxMqrbK2LypLkSYWzBOEqAwPIGh6BKlUFPkEAUogLSWs0tFOQgZqdK6s05Bapz+qdQbyqMry4B8jZ-OihRpVlaAOD8OAZz8eoUlebNPgmQdPGHCMxwIVgOAAOVYYIdFYMANsaPpAWRJ5roY2w+jEgiXpmfYpCAA

I wonder what the meaningful delta could be...

cdata: One potentially significant difference is the compiler version. We are using 3.6.3, while the playground is apparently on 3.5.1.

cdata: Sure enough, 3.5.1 compiles fine. 3.6.3 shows the error.

cdata: I pinned us to 3.5.3, as this is the "stable" version used internally. We'll save this issue for another day.

aomarks: SGTM!

src/features/controls.ts
cdata: I factored out the explicit constructor, so now this chunk of implementation is a bit cleaner.

PR #756, opened 2019-09-16
Wrap theta

### LINKED ISSUE #590: Camera-orbit theta value rotates beyond 360 degrees
### Description

When you rotate the model a lot in one direction, the theta goes over 360 degrees instead of wrapping back around. When setting the camera orbit theta back to 0deg, then the model overspins.

See here: https://nickel-longship.glitch.me/

This makes it hard to reset the camera back to the original position.

### Browser Affected
All browsers

### OS
All OS
### Versions
- model-viewer: 0.3.1
- three.js: r101

elalish:
Fixes #590 

I added one additional check, so if min/max azimuth are set, then we don't attempt to go the short way, as this might result in interpolating outside of the allowed range.


src/three-components/SmoothControls.ts
```diff
@@ -494,6 +495,13 @@ export class SmoothControls extends EventDispatcher {
     const {maximumPolarAngle, maximumRadius, maximumFieldOfView} =
         this[$options];
 
+    const dTheta = this[$spherical].theta - this[$goalSpherical].theta;
+    if (Math.abs(dTheta) > Math.PI &&
+        !isFinite(this[$options].minimumAzimuthalAngle!) &&
+        !isFinite(this[$options].maximumAzimuthalAngle!)) {
+      this[$spherical].theta -= Math.sign(dTheta) * 2 * Math.PI;
```
cdata: Substitute constant `TAU` above?

src/three-components/SmoothControls.ts
```diff
@@ -567,6 +575,11 @@ export class SmoothControls extends EventDispatcher {
     return handled;
   }
 
+  // Wraps to bewteen -pi and pi
+  private[$wrapAngle](radians: number): number {
+    return (radians + Math.PI) % (2 * Math.PI) - Math.PI;
```
cdata: `TAU`?

src/three-components/SmoothControls.ts
elalish: I have a small vendetta against TAU, possibly just because I didn't go to MIT. I didn't even notice it was in the file. Personally I'd remove it, but I'll use it if you feel strongly.

cdata: I don't feel strongly, but it was added at the request of another reviewer so I feel weird taking it out.

Use your discretion 👍

pushmatrix:
Yay!

PR #757, opened 2019-09-16
Update WebXR mode to match current immersive-ar

### LINKED ISSUE #495: WebXR is broken in Android Chrome Canary

klausw:
Get WebXR mode working again using the current API as implemented in Chrome Canary using immersive-ar mode. 

This commit is a snapshot of an airplane hacking session, it's in rough shape due to having a barely-functioning dev environment. Please ignore superficial issues such as indentation, those should obviously be fixed before merging. There's also extra logging, and I was taking shortcuts to get it working such as overriding feature checks. I'm turning it into a PR at @elalish 's suggestion, at least it should be useful as a roadmap of areas that need changing.

FYI, the hit test API is likely to undergo more changes in the near future again, so that will need further tweaks at that time.

Fixes #495

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -17,21 +17,16 @@
 // the appropriate flags. However, just because we have the API does not
 // guarantee that AR will work.
 export const HAS_WEBXR_DEVICE_API = navigator.xr != null &&
-    self.XRSession != null && self.XRDevice != null &&
-    self.XRDevice.prototype.supportsSession != null;
+    self.XRSession != null && navigator.xr.supportsSession != null;
 
 export const HAS_WEBXR_HIT_TEST_API =
     HAS_WEBXR_DEVICE_API && self.XRSession!.prototype.requestHitTest;
-
-export const HAS_FULLSCREEN_API = document.documentElement != null &&
-    document.documentElement.requestFullscreen != null;
 
 export const HAS_RESIZE_OBSERVER = self.ResizeObserver != null;
 
 export const HAS_INTERSECTION_OBSERVER = self.IntersectionObserver != null;
 
-export const IS_WEBXR_AR_CANDIDATE =
-    HAS_WEBXR_HIT_TEST_API && HAS_FULLSCREEN_API;
+export const IS_WEBXR_AR_CANDIDATE = HAS_WEBXR_HIT_TEST_API;
 
 export const IS_MOBILE = (() => {
   const userAgent =
--- a/src/features/ar.ts
+++ b/src/features/ar.ts
@@ -17,7 +17,7 @@ import {property} from 'lit-element';
 
 import {IS_ANDROID, IS_AR_QUICKLOOK_CANDIDATE, IS_IOS, IS_IOS_CHROME, IS_IOS_SAFARI, IS_WEBXR_AR_CANDIDATE} from '../constants.js';
 import {enumerationDeserializer} from '../conversions.js';
-import ModelViewerElementBase, {$container, $renderer, $scene} from '../model-viewer-base.js';
+import ModelViewerElementBase, {$renderer, $scene} from '../model-viewer-base.js';
 import {Constructor, deserializeUrl} from '../utilities.js';
 
 /**
@@ -95,28 +95,15 @@ const ARMode: {[index: string]: ARMode} = {
   NONE: 'none'
 };
 
-const $exitFullscreenButtonContainer = Symbol('exitFullscreenButtonContainer');
 const $arButtonContainer = Symbol('arButtonContainer');
-const $defaultExitFullscreenButton = Symbol('defaultExitFullscreenButton');
 const $enterARWithWebXR = Symbol('enterARWithWebXR');
 const $canActivateAR = Symbol('canActivateAR');
 const $arMode = Symbol('arMode');
 const $canLaunchQuickLook = Symbol('canLaunchQuickLook');
 const $quickLookBrowsers = Symbol('quickLookBrowsers');
 
-const $arButtonContainerFallbackClickHandler =
-    Symbol('arButtonContainerFallbackClickHandler');
-const $onARButtonContainerFallbackClick =
-    Symbol('onARButtonContainerFallbackClick');
 const $arButtonContainerClickHandler = Symbol('arButtonContainerClickHandler');
 const $onARButtonContainerClick = Symbol('onARButtonContainerClick');
-
-const $exitFullscreenButtonContainerClickHandler =
-    Symbol('exitFullscreenButtonContainerClickHandler');
-const $onExitFullscreenButtonClick = Symbol('onExitFullscreenButtonClick');
-
-const $fullscreenchangeHandler = Symbol('fullscreenHandler');
-const $onFullscreenchange = Symbol('onFullscreen');
 
 export interface ARInterface {
   ar: boolean;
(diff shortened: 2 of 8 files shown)


klausw:
Deployed version is at https://klausw.github.io/model-viewer/www/examples/augmented-reality.html

cdata:
This is exciting 👍 nice to see some help in this area.  Thanks @klausw !

elalish:
@klausw Do you want to keep working on this PR, or should I just pull it and finish it on my side?

klausw:
Please review, I've rebased it and cleaned up the changes, including removing the no-longer-needed fullscreen mode workarounds and reusing the existing threejs WebGLRenderer instead of creating a new one. As a side effect, that made the regular Astronaut work, so I no longer need to switch to the unlit variant.

klausw:
FYI, updated test URL is https://klausw.github.io/model-viewer/examples/augmented-reality.html

src/three-components/ARRenderer.ts
```diff
@@ -232,18 +186,20 @@ export class ARRenderer extends EventDispatcher {
   }
 
   [$postSessionCleanup]() {
+    (this.renderer as any).setFramebuffer(null);
+    this.parentRenderer.setRendererSize(1, 1);
```
elalish: What is this for?

src/three-components/ARRenderer.ts
```diff
@@ -377,12 +314,13 @@ height: 100%;`);
...
 
-    for (const view of (frame as any).views) {
-      const viewport = (session.baseLayer as XRWebGLLayer).getViewport(view);
-      this.renderer!.setViewport(0, 0, viewport.width, viewport.height);
-      this.renderer!.setSize(viewport.width, viewport.height, false);
+    for (const view of (frame as any).getViewerPose(this[$refSpace]).views) {
+      const viewport =
+          (session.renderState.baseLayer as XRWebGLLayer).getViewport(view);
+      this.renderer.setViewport(0, 0, viewport.width, viewport.height);
+      this.renderer.setSize(viewport.width, viewport.height, false);
```
elalish: setSize seems weird here. Are there ever multiple views? If there were I don't think the size would be the same as the viewport.

src/types/webxr.d.ts
```diff
@@ -109,25 +106,30 @@ type XRTargetRayMode = 'gaze'|'tracked-pointer'|'screen';
...
-      Promise<XRReferenceSpace>;
-  requestHitTest(
-      origin: Float32Array, direction: Float32Array,
-      frameOfReference: XRFrameOfReference): Promise<XRHitResult[]>;
-  getInputSources(): Array<XRInputSource>;
+}
+
+interface XRSession extends EventTarget {
+  renderState: XRRenderState;
+  updateRenderState(obj: any): any;
```
elalish: Is this really all the typing we have for this method?

src/three-components/ARRenderer.ts
```diff
@@ -52,93 +47,66 @@ export class ARRenderer extends EventDispatcher {
...
 
   initializeRenderer() {
-    if (this.renderer != null) {
-      return;
-    }
-
-    this.renderer = new WebGLRenderer(
-        {canvas: this.inputCanvas, context: this.inputContext});
-    this.renderer.setSize(window.innerWidth, window.innerHeight);
     this.renderer.setPixelRatio(1);
```
elalish: If we're going to reuse our renderer, we'll need to reset this to the device DPR when we're done.

src/constants.ts
```diff
@@ -17,21 +17,16 @@
...
 export const HAS_WEBXR_DEVICE_API = navigator.xr != null &&
-    self.XRSession != null && self.XRDevice != null &&
-    self.XRDevice.prototype.supportsSession != null;
+    self.XRSession != null && navigator.xr.supportsSession != null;
 
 export const HAS_WEBXR_HIT_TEST_API =
     HAS_WEBXR_DEVICE_API && self.XRSession!.prototype.requestHitTest;
 
-export const HAS_FULLSCREEN_API = document.documentElement != null &&
-    document.documentElement.requestFullscreen != null;
```
elalish: I think there's some more fullscreen stuff that hasn't been reverted yet.

src/features/ar.ts
```diff
@@ -153,32 +140,9 @@ export const ARMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-        this.shadowRoot!.querySelector('#default-exit-fullscreen-button') as
-        HTMLElement;
-
-    // NOTE(cdata): We use a second, separate "fallback" click handler in
-    // order to work around a regression in how Chrome on Android behaves
-    // when requesting fullscreen at the same time as triggering an intent.
-    // As of m76, intents could no longer be triggered successfully if they
-    // were dispatched in the same handler as the fullscreen request. The
-    // workaround is to split both effects into their own event handlers.
-    // @see https://github.com/GoogleWebComponents/model-viewer/issues/693
```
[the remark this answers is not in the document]
klausw: I've restored that code in https://github.com/GoogleWebComponents/model-viewer/pull/757/commits/45c4a9b132ffee949c67f998f565a3ab57397474 , I had misunderstood how this works and didn't realize that AR_VIEWER mode also used requestFullscreen.

src/constants.ts
klausw: This constant only appeared to be used for WebXR mode. It was used in utilities.ts's assertIsArCandidate which is only used in the WebXR ARRenderer.ts, and I had removed it there since immersive-ar mode doesn't require use of the Fullscreen API. (I think that was part of the reason I was under the impression that the other AR code didn't need fullscreen.)

src/three-components/ARRenderer.ts
klausw: The "const viewport" assignment above specifically gets the viewport size for use with the current view, either the whole framebuffer or an appropriate subsection. See https://immersive-web.github.io/webxr/#dom-xrwebgllayer-getviewport :
> getViewport() queries the XRViewport the given XRView should use when rendering to the layer.

In this case only one view is expected, but the view loop should work for the more general case also.

src/three-components/ARRenderer.ts
```diff
@@ -52,93 +47,66 @@ export class ARRenderer extends EventDispatcher {
...
    * Given an inline Renderer, construct an ARRenderer and return it
    */
   static fromInlineRenderer(renderer: Renderer) {
-    return new ARRenderer(renderer.canvas, renderer.context!);
+    return new ARRenderer(renderer);
   }
 
-  constructor(
-      inputCanvas: HTMLCanvasElement, inputContext: WebGLRenderingContext) {
+  constructor(parentRenderer: Renderer) {
```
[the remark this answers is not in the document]
klausw: I've added some clarifying comments, and replaced the fromInlineRenderer call with using the constructor directly, both here and in the ARRenderer-spec test.

src/three-components/ARRenderer.ts
```diff
@@ -52,93 +47,66 @@ export class ARRenderer extends EventDispatcher {
...
-        .setFramebuffer((session.baseLayer as XRWebGLLayer).framebuffer);
+    // The render state update takes effect on the next animation frame. Wait
+    // for it so that we get a framebuffer.
+    let waitForAnimationFrame = new Promise((resolve, _reject) => {
+      session.requestAnimationFrame(() => resolve());
+    });
+    await waitForAnimationFrame;
 
+    (this.renderer as any)
+        .setFramebuffer(
```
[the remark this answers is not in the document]
klausw: Done.

src/three-components/ARRenderer.ts
klausw: I've added a comment, this triggers a resize in the render loop. The parent Renderer code does this also, I was following its example, but I'm not sure if this is the best way to do so.

src/types/webxr.d.ts
klausw: I've added more specific type info using XRRenderStateInit.

src/three-components/ARRenderer.ts
klausw: I added a comment in postSessionCleanup. Renderer.render auto-adjusts the device pixel ratio as needed already.

klausw:
PTAL, and I've updated the gh-pages at  https://klausw.github.io/model-viewer/examples/augmented-reality.html , though github is sometimes slow in consistently serving that.

src/three-components/ARRenderer.ts
klausw: Sorry, I had described the way it should work, but the implementation didn't match. Should be fixed by https://github.com/GoogleWebComponents/model-viewer/pull/757/commits/311706df0e21924830be4c5c0a7ff18e8ed6171d

elalish:
Thanks, this is looking great! I want to do a bit more testing before merging, but that's it.

elalish:
@klausw Would it be possible to add an extra WebXR button in this PR? When I test I can only launch SceneViewer, unless I'm missing something?

klausw:
The button should currently switch to WebXR mode if the prerequisites are
met - you need to enable WebXR Device API and WebXR AR Module in
chrome://flags, and use a reasonably current Chrome Canary on an
ARCore-supported phone with Android >= 8.0 / Oreo.

I'll look into adding a separate button.

klausw:
On Fri, Sep 27, 2019 at 10:55 AM Klaus Weidner <notifications@github.com>
wrote:

> The button should currently switch to WebXR mode if the prerequisites are
> met - you need to enable WebXR Device API and WebXR AR Module in
> chrome://flags, and use a reasonably current Chrome Canary on an
> ARCore-supported phone with Android >= 8.0 / Oreo.
>

Correction, it also needs WebXR Hit Test API enabled in chrome://flags.
This also works on the stable Chrome release 77 with the same flag changes,
except that this version doesn't have a "WebXR AR module" flag. (It's
auto-activated by enabling hit test).

klausw:
> The main thing is to put back the fullscreen stuff so that SceneViewer still works. We'll have to test it manually on a Pixel 3.

I had reverted those changes to restore the SceneViewer fullscreen path, AFAIK the changes should now only affect WebXR mode. I've tested this by disabling WebXR AR Module (removing one of the other prerequisites should also work), and verifying that the AR button then launches Scene Viewer.

(Note to self and other testers: scene viewer does NOT appear to work for a port-forwarded localhost mode such as "npm run dev" using localhost:8000. It gets launched but can't load the model. It should work for the github-hosted version at https://klausw.github.io/model-viewer/examples/augmented-reality.html .)

PR #760, opened 2019-09-17
Add fidelity tests

### LINKED ISSUE #510: <model-viewer> doesn't support KHR_materials_pbrSpecularGlossiness
There is a good chance we are breaking this when cloning a parsed glTF. According to @mrdoob :

> GLTFLoader produces a custom material when parsing KHR_materials_pbrSpecularGlossiness and clone may not be copying that

elalish:
Fixes #510 

I added most of the khronos glTF sample models to our fidelity test suite. I didn't add a few of the really simple ones and two immersive scenes because our camera doesn't apply very well to them. 

It looks like `<model-viewer>` has errors on:
-BrainStem
-Monster
-MultiUVTest
-NormalTangentMirrorTest
-NormalTangentTest

And Filament seems to have errors on:
-AnimatedMorphCube
-AnimatedMorphSphere
-EnvironmentTest
-MorphPrimitivesTest
-SimpleSparseAccessor
-SpecGlossVsMetaRough
-TextureSettingsTest


src/three-components/ModelUtils.ts
```diff
@@ -87,6 +79,20 @@ export const cloneGltf = (gltf: Gltf): Gltf => {
...
+   * shader program.
+   */
+  const cloneAndPatchMaterial = (material: Material): Material => {
+    const clone = (material as any).isGLTFSpecularGlossinessMaterial ?
+        specularGlossiness.cloneMaterial(material) :
+        material.clone();
+    clone.onBeforeCompile = updateShader;
+    return clone;
+  };
+
```
cdata: Since we are moving this in here anyway, might as well move it inside the condition on line 99 so that we don't bother with it unless there is a scene in the glTF.

elalish: SGTM

elalish:
Above I was referring to rendering errors, so yeah it's graceful. The exception is Filament's errors on EnvironmentTest and SimpleSparseAccessor, which I had to exclude because they seem to crash.

PR #761, opened 2019-09-18
Revert mixin type inference

cdata:
Unfortunately, we must revert this part of #747 because it is not compatible with generating external declaration files.

See also https://github.com/Microsoft/TypeScript/issues/30355#issuecomment-473060190

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,9 +12,9 @@
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
     "outDir": "./lib",
-    "allowJs": true,
     "experimentalDecorators": true,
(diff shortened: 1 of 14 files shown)


tsconfig.json
```diff
@@ -12,9 +12,9 @@
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
     "outDir": "./lib",
-    "allowJs": true,
     "experimentalDecorators": true,
-    "noErrorTruncation": true
+    "noErrorTruncation": true,
+    "declaration": true
```
cdata: This should have been enabled previously. Now that it is on, we will see the category of errors this inference pattern raised in google3.

PR #762, opened 2019-09-18
Make sure license headers exist and have an annotation

cdata:
I have also updated [our license header templates in the wiki](https://github.com/GoogleWebComponents/model-viewer/wiki/Code-conventions#license-headers).


src/three-components/shader-chunk/encodings_pars_fragment.glsl.ts
```diff
@@ -1,3 +1,18 @@
+/* @license
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
elalish: Technically this file is from three.js, so shouldn't it use their license? Same with normalmap below.

PR #763, opened 2019-09-18
Convert RGBELoader to TS

elalish:
Well, that was ugly. I got it to compile, but it still doesn't run. Pretty sure I've missed something about how this is supposed to extend a class. Pointers?


elalish:
No need for this one, as I just removed the file from #750

PR #769, opened 2019-09-24
[Refactor] Unified logic for single touch and mouse pointer

mikkoh:
## What?
This is a small refactor which unifies single touch and mouse interaction logic.

## Why?
The main reason I'd like to do this is that it's much easier to write code for both single touch and mouse interactions when these logic paths are combined. 

For instance in a subsequent PR I'd like to explore adding a new `interaction-policy` which would toggle interactions on and off whenever a `<model-viewer>` is tapped/touched. So basically tap to enable interactions, tap to disable interactions. (think play pause button for video) It would help with this [issue](https://github.com/GoogleWebComponents/model-viewer/issues/479#issuecomment-490560705) mentioned by @elalish.

During user testing that @pushmatrix ran it seemed that this was the best UX to solve this problem.

## Things to consider
- I changed the type of `$lastPointerPosition` from `Vector2` to a `Pointer`. I think `Vector2` is fantastic but it actually wasn't being used for anything right now.
- I haven't implemented `$handleSinglePointerUp` right now since the logic is the same whether using touch/mouse. Although in a subsequent PR I likely will do this.

### DIFF
--- a/src/three-components/SmoothControls.ts
+++ b/src/three-components/SmoothControls.ts
@@ -13,13 +13,18 @@
  * limitations under the License.
  */
 
-import {Event as ThreeEvent, EventDispatcher, PerspectiveCamera, Quaternion, Spherical, Vector2, Vector3} from 'three';
+import {Event as ThreeEvent, EventDispatcher, PerspectiveCamera, Quaternion, Spherical, Vector3} from 'three';
 
 import {clamp} from '../utilities.js';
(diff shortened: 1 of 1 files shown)


mikkoh:
@cdata 👌 

I'm gonna put up a PR and we can chat about it more there. If it's heading in a bad direction we can just kill it.

PR #772, opened 2019-09-26
Implemented interaction-policy 'allow-toggle'

mikkoh:
## What?

This PR creates a new `interaction-policy` called `allow-toggle`. When this interaction policy is enabled users must tap/click on the `<model-viewer>` to turn on the ability interact (orbit / zoom). Once the initial tap/click is performed then the user can interact with the `<model-viewer>` as per normal for instance by dragging to rotate the model.

## Why?

On mobile there are a some cases where `<model-viewer>` interferes with normal interaction. For instance:

- When a `<model-viewer>`  takes up the entire page and the user is trying to scroll below the fold but cannot since the `<model-viewer>` fills the entire page
- When a `<model-viewer>` is in a carousel and you can swipe/drag on the carousel to go to the next slide `<model-viewer>` will prevent this interaction

## What to focus on?

- The `click` event is used to toggle interaction in `controls.ts`
- `controls.ts` aims to still have logic to enable/disable interactions on `SmoothControls` vs all logic for toggling interaction living in `SmoothControls`
- `click` is fired after a `mousemove` on desktop which means that if the user has dragged we should assume they want to continue interacting with the `<model-viewer>` and not toggle off/disable interaction
- On desktop the cursor is a `pointer` when the `<model-viewer>` interaction has not been toggled. Once toggled it becomes `grab`. When dragging the cursor is `grabbing`

## Further work

- There's a piece of UX missing which is some sort icon/indicator that shows whether the `<model-viewer>` can be interacted with (think a play pause button on a video player)
- Likely some sort of event must be fired to let the hosted application/page know whether interaction is enabled/disabled
- It maybe advantageous to refactor `cursor` logic out of `SmoothControls`

### DIFF
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -33,7 +33,7 @@ export interface SphericalPosition {
 }
 
 export type InteractionPromptStrategy = 'auto'|'when-focused'|'none';
-export type InteractionPolicy = 'always-allow'|'allow-when-focused';
+export type InteractionPolicy = 'always-allow'|'allow-when-focused'|'allow-toggle';
 
 const InteractionPromptStrategy:
     {[index: string]: InteractionPromptStrategy} = {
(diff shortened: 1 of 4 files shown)


pushmatrix:
Just an FYI that we have this functionality as part of the <model-viewer> UI library we put ontop of <model-viewer>. This is no longer a blocker for us.

elalish:
I think I'd rather solve this along the lines of my comment in #1424.

PR #777, opened 2019-09-27
Add quick-look-browsers to Augmented Reality page

cwervo:
Many times when referring to the `quick-look-browsers` documentation I would open up the [**Augmented Reality**](https://googlewebcomponents.github.io/model-viewer/examples/augmented-reality.html) page & was surprised to see `ios-src` listed but not `quick-look-browsers`. This rectifies that & lessens the amount people need to scroll to find the feature documentation.

### Reference Issue
No reference issue.


cdata:
👍 sorry for the omission, and thanks for the fix!

PR #779, opened 2019-09-27
Change logic for detecting iOS 13 Safari on iPad

### LINKED ISSUE #758: AR button is not showed in Safari on iPad Pro on iOS 13
### Description

AR button is not showed in Safari on iPad Pro on iOS 13. According to Apple's list of supported devices [here down in the page](https://www.apple.com/ios/augmented-reality/) this device has support for AR on Safari.

### Browser Affected

- [ ] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [X] Safari

### OS

- [ ] Android
- [X] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### Versions

- model-viewer: v0.6.0
- three.js: r105

cdata:
Fixes #758 

In #775 we addressed incorrect detection of iOS 13 Safari on iPad. Unfortunately, the logic was subtly flawed, and we discovered the flaw when testing the fix on an iPad Pro this morning. This change addresses the flaw. It has been tested on an iOS 13 iPad Pro, iOS 12 iPad Air and MacOS Safari to ensure that the AR button display cases are what we expect.

Once this goes out, we will also back-port this as v0.6.2.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -54,19 +54,6 @@ export const OFFSCREEN_CANVAS_SUPPORT_BITMAP =
 
 export const IS_ANDROID = /android/i.test(navigator.userAgent);
 
-export const IS_IOS =
-    /iPad|iPhone|iPod/.test(navigator.userAgent) && !(self as any).MSStream;
-
-export const IS_AR_QUICKLOOK_CANDIDATE = (() => {
-  const tempAnchor = document.createElement('a');
-
-  return Boolean(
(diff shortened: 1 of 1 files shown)


cdata:
The logic for `IS_IOS` passes through to `IS_IOS_*` for both Chrome and Safari. Strictly speaking, we want this test to apply in both cases.

PR #780, opened 2019-09-27
Fix over-spin reversal

### LINKED ISSUE #778: Pointer interaction can cause model to rotate in unintended ways
This is must likely a bug that arises from #756 . Basically, it's very easy to cause the model to sort-of glitch out and rotate in unintended ways when interacting with it.

ccontrol

elalish:
Fixes #778 

This change means adjustOrbit should not be used to attain an absolute theta (since it will clamp), but that seems fine since we already have a setOrbit method for that purpose.

### DIFF
--- a/src/test/three-components/SmoothControls-spec.ts
+++ b/src/test/three-components/SmoothControls-spec.ts
@@ -158,6 +158,22 @@ suite('SmoothControls', () => {
           expect(controls.getCameraSpherical().theta)
               .to.be.closeTo(-QUARTER_PI, 0.0001);
         });
+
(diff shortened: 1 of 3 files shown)


src/test/three-components/SmoothControls-spec.ts
```diff
@@ -158,6 +158,22 @@ suite('SmoothControls', () => {
           expect(controls.getCameraSpherical().theta)
               .to.be.closeTo(-QUARTER_PI, 0.0001);
         });
+
+        test(
+            'adjustOrbit does not move the goal theta more than pi past the current theta',
```

PR #785, opened 2019-09-30
Force canvas onto its own render layer

cdata:
Although we fixed computation-based visibility errors in #755 , Chrome 76 apparently has a rendering disparity that causes render layers to be messed up without any interventions by us. This is what the render layers look like without / with the intervention:

Without intervention | With forced `transform`
---------------------|-------------------------
image|image

### DIFF
--- a/src/template.ts
+++ b/src/template.ts
@@ -45,6 +45,12 @@ template.innerHTML = `
       width: 100%;
       height: 100%;
       display: none;
+      /* NOTE(cdata): Chrome 76 and below apparently have a bug
+       * that causes our canvas not to display pixels unless it is
+       * on its own render layer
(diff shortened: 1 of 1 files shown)


pushmatrix:
I was not aware of the existence of this translateZ hack. Interesting

cdata:
@pushmatrix basically any style that forces the element onto a render layer would work. The problem with this is that it only works because of how leaky an abstraction browser rendering is :(

pushmatrix:
So was this the only fix for that rendering issue? Or was it a combination of this + the framing math fix.

cdata:
@pushmatrix I expect both fixes to be required to address this issue comprehensively. The math issue (where we computed a poisonous aspect ratio that permanently borked the camera) should have affected all browsers.

PR #786, opened 2019-09-30
Fix autorotate

### LINKED ISSUE #784: Model framing is incorrect after model begins to autorotate
Per discussions in https://github.com/GoogleWebComponents/model-viewer/pull/755#issuecomment-536597893 it is possible for some models to become incorrectly framed when `auto-rotate` is enabled.

Test case demonstrating the issue here:  https://congruous-freesia.glitch.me/

Screen recording of the issue:

autorotate bug

elalish:
Fixes #784 

This was an error introduced when we added configurable camera-target support that is here fixed. I added auto-rotate to the camera target examples to act as a visual regression test.


src/features/staging.ts
```diff
@@ -65,7 +66,7 @@ export const StagingMixin = <T extends Constructor<ModelViewerElementBase>>(
       super.updated(changedProperties);
 
       if (changedProperties.has('autoRotate')) {
-        (this as any)[$scene].pivot.rotation.set(0, 0, 0);
```
cdata: 💯

src/features/staging.ts
```diff
@@ -65,7 +66,7 @@ export const StagingMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+        this[$scene].pivot.rotation.set(0, 0, 0);
```
cdata: Should this be done using the new `setRotation` method? Seems like it would be better to let the `ModelScene` have full authority over the `pivot`'s rotation...

elalish: Good call, I missed that one.

Issue #708, opened 2019-08-14
Animate model along with interaction prompt


pushmatrix:
A better interaction prompt would be to move the model along with the prompt hand indicator.
This way it's clearer that the object is not just a photo.

![loop](https://user-images.githubusercontent.com/190692/63053249-579d6d80-beaf-11e9-9816-2e5a2366522b.gif)

We can fine tune the easing & speed of it, but the general idea is something like the above.

As soon as you interact with the model, the hand would disappear and the object would stop rocking back and forth.

pushmatrix:
I might be able to find time to make a PR on it, but would like some guidance into what files something like this should live in. Also, what do you think would be the best approach for keeping the hand css in sync with the model's theta value. Seems like there could be some framerate issues animating both a webgl model and a DOM element each frame.

pushmatrix:
Not sure how this would work with the auto-rotate feature though

elalish:
I think the most straight-forward thing is to simply superimpose the motions of this prompt and auto-rotate, which would happen naturally, since this prompt would act on the camera while auto-rotate acts on the stage. Also, from a UX perspective, I wonder if it wouldn't be nicer to make this prompt intermittent: every few seconds, the hand appears, the model wiggles once, then stops.

elalish:
I'd really like to see this, because it would have a secondary benefit of replacing our animated SVG, which has terrible energy cost! Chrome is much more efficient at CSS transforms.

pushmatrix:
Yeah I agree with you on wiggling once and stopping for a bit. 

Not quite the same idea, but here's something we played around with a few months back:
![rotate](https://user-images.githubusercontent.com/190692/65075373-cd8c6d00-d964-11e9-8f13-5648198e5fcc.gif)

elalish:
That's nice, though I might prefer a longer delay. As for where this code should live, I would say controls.ts seems like the best candidate, since that's where all the interaction prompt logic is, and it has easy access to SmoothControls. As for CSS coordination, I'll defer to @cdata.

Issue #738, opened 2019-09-03
Allow manipulation of model scene graph via "Render Worklet"


cdata:
Requests to directly manipulate the model scene graph have come up several times, so I'm filing an issue for us to have a canonical place to discuss the feature.

The idea behind this feature is that we would allow users to specify a script that will run as a kind of "render worklet" that has imperative access to the model's underlying scene graph. The goal of the worklet is two-fold:

 1. Ensure that user scripts can be transparently moved between the main thread and a worker thread, depending on what set of optimizations `<model-viewer>` applies in the given environment
 2. Ensure that user scripts do not depend on underlying Three.js APIs, in case we decide to offer alternative rendering backends in the future.

The fundamental details of this API are still very much up in the air, but the idea is that a Render Worklet script would have imperative access to a scene graph that is modeled after the glTF JSON schema. Users would be able to change things like PBR properties of materials or vertex information and see the changes reflected in the model in real time.

cdata:
Some related discussions on this topic include: https://github.com/GoogleWebComponents/model-viewer/issues/186#issuecomment-439303818 and https://github.com/GoogleWebComponents/model-viewer/issues/685#issuecomment-516614906

fselcukcan:
Plus one on that!

ataylor09:
Would like to voice my support for this feature.

soraryu:
Same here. Would that also allow adding objects? In our case we're looking at options to have something like "model annotations" (essentially small billboard buttons in 3D that might drive a camera and have a callback to react to).

cdata:
@soraryu yes, this would eventually include the ability to add objects, although it may take multiple iterations before the worklet has every possible capability.

Although we anticipate that you could implement annotations with the worklet alone, we also wish to offer a high-level API for annotations / hotspots / billboard buttons (see #334 ).

aiursrage2k:
I would like to request having multiple animations. Imagine something like an ikea furniture, so when you click a button it will move onto the the next step in the animation

cdata:
@aiursrage2k this is actually possible with our current animation system, I think, although there is probably stuff we could do to make it easier. We already support multiple animations, and we reveal the current time of the playing animation. If you pair that with `play` and `pause` methods, you can construct a UX like what you are describing.

The thing that we could do to make it easier would be to let you control the iteration count, like with an `animation-iteration-count` attribute. Currently, animations play on a loop forever, so you would have to watch `currentTime` and manually `pause()` after the first loop.

[...]

PR #788, opened 2019-09-30
Model animate interaction prompt

maaslalani:
### Reference Issue
Fixes: https://github.com/GoogleWebComponents/model-viewer/issues/708

Removes SVG animations from interaction-prompt and uses a CSS based animation for the interaction prompt and slides it back and forth, if the interaction prompt is visible then the position of the prompt is monitored and the model rotates accordingly.

### General Approach
The approach was to use CSS keyframe animations to move the interaction prompt cursor and set the `pivot.rotation.y` based on the offset of the center of the cursor in relation to the center of the model viewer. While the keyframe animation is on the `.controls-prompt` the animation is paused and plays once the element is `visible`.

### Examples

<table>
<tbody>
<td>
<br>

model-op
</td>
<td>
<br>

model2
</td>
</tbody>
</table>

#### Animated models
<details>
<summary>Animated Robot Model Example</summary>

animated
</details>

#### Auto rotate
If `auto-rotate` is enabled, interaction prompt will display if the user has not interacted with the model yet, after the model has been interacted with the prompt will go away and `auto-rotate` will work as normal. 
<details>
<summary>Auto Rotate Interaction Example</summary>

model-auto-rotate-op
</details>


maaslalani:
@googlebot I signed it!

src/features/controls.ts
```diff
@@ -290,17 +287,37 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+        const {
+          left: promptLeft,
+          width: promptWidth,
+        } = this[$promptElementSVG].getBoundingClientRect();
+
+        const modelViewerCenter = modelViewerLeft + (modelViewerWidth / 2);
+        const promptCenter = promptLeft + (promptWidth / 2);
+        const promptOffset = promptCenter - modelViewerCenter;
+
+        (this as any)[$scene].pivot.rotation.y = promptOffset * OFFSET_ROTATION_MULTIPLIER;
```
elalish: It looks like promptOffset is in units of pixels and that it moves by a fixed number of pixels. I wonder if it wouldn't be better to have the motion scale to a % of the canvas size, but keep the angular rotation constant? What I notice now is that the prompt motion relative to the model changes based on how I adjust the window size.

src/template.ts
```diff
@@ -165,6 +165,18 @@ template.innerHTML = `
...
 
+    @keyframes wiggle {
+      0%, 50%, 100% {
+        transform: translateX(0%);
+      }
+      25% {
+        transform: translateX(-5%);
+      }
+      75% {
+        transform: translateX(5%);
```
elalish: Could we change this animation to be intermittent? I think it would be more pleasing to the eye if it were a bit shorter with a few seconds of space in between that didn't show the SVG.

src/template.ts
```diff
@@ -179,6 +191,11 @@ template.innerHTML = `
       transform-origin: center center;
       transform: scale(0.9);
       transition: transform 0.3s, opacity 0.3s;
+      animation-name: wiggle;
+      animation-duration: 3s;
+      animation-iteration-count: infinite;
+      animation-timing-function: linear;
```
elalish: Would it be possible to use a different function here to make the direction change less sudden?

src/features/controls.ts
```diff
@@ -68,12 +68,16 @@ const PHI = 2.0 * Math.PI;
 const AZIMUTHAL_QUADRANT_LABELS = ['front', 'right', 'back', 'left'];
 const POLAR_TRIENT_LABELS = ['upper-', '', 'lower-'];
 
+const OFFSET_ROTATION_MULTIPLIER = 0.005;
+
 export const DEFAULT_INTERACTION_PROMPT_THRESHOLD = 3000;
```
elalish: I see a small bug where the model seems to jerk when the prompt first comes up. I believe this is because the model starts to auto-rotate just before the prompt comes up, and then the prompt resets the rotation to zero, which could get worse if this threshold attribute is increased.

src/template.ts
maaslalani: Agreed with the intermittent animations, after a bit of searching it doesn't seem there's a way to pause animation between iterations through CSS.

There's a few alternatives:
* [placing a pause within the animation keyframes](https://css-tricks.com/css-keyframe-animation-delay-iterations/)
* [performing a few iterations of the animation each with a delay on them](http://www.standardista.com/animation-iteration-delay/)

In terms of the UX, what do you think about this animation?

model

src/features/controls.ts
maaslalani: Will update to a relative percentage 👍

src/template.ts
maaslalani: Yep, will switch to `ease-in-out`.

src/features/controls.ts
maaslalani: What do you think the best way to handle auto-rotate is?

We could:
* Not play the auto-rotate animation if the interaction prompt is going to be displayed.
* Have the auto-rotate animation play as normal and animate the model back to the centre once the interaction prompt animation begins.
* The interaction prompt animation animates the model at whatever rotation it happens to be at.

elalish: Perhaps an even simpler answer? How about make the auto-rotation delay (which is not currently configurable) equal to the interaction prompt threshold plus enough delay to ensure the prompt always fires first. @cdata Does this sound good to you?

src/template.ts
maaslalani: Here's a preview of what we were thinking about the fine tuned adjustments for the animation: https://translucent-lunch.glitch.me

pushmatrix: @elalish We're feeling pretty good about the latest animation that @maaslalani just posted.

The symmetry is really important to us as it shows a good range of motion, and makes for a good flow.

In a subsequent PR we also have some changes to the hand svg based on some usability research we did.

elalish: That animation looks really good to me! Thanks for the effort.

pushmatrix:
@maaslalani 

One thing we need to check for is what happens when you put in a custom interaction prompt via a `slot`.

src/features/controls.ts
```diff
@@ -290,18 +287,38 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+          width: modelViewerWidth,
+        } = this.getBoundingClientRect();
+        const {
+          left: promptLeft,
+          width: promptWidth,
+        } = this[$promptElementSVG].getBoundingClientRect();
+
+        const modelViewerCenter = modelViewerLeft + (modelViewerWidth / 2);
+        const promptCenter = promptLeft + (promptWidth / 2);
+        const promptOffsetRelative = (promptCenter - modelViewerCenter) / modelViewerWidth;
```
pushmatrix: We need to guard against when modelViewerWidth = 0 to avoid divide by zero errors.

This can happen if the model viewer starts hidden on a page.

maaslalani: Good call! Fixed 👍

src/template.ts
cdata: @pushmatrix if we changed the animation in a subsequent change (perhaps the timing function, or the duration), would you consider that a breaking change?

src/features/controls.ts
cdata: First, I would like to state plainly that I increasingly desire for us to deprecate `auto-rotate` in favor of #738 because the way it works is kind of annoying and I think everyone would be happier without it.

Imagining the user intention of someone literally writing this HTML:

```html
<model-viewer auto-rotate></model-viewer>
```

They have explicitly asked for auto-rotation, regardless of interaction-prompt behavior. I think bullet #3 from @maaslalani is what I expect in that case. Auto-rotation should happen as normal, and the wiggle position should be additive to whatever the turntable rotation value happens to be in a given frame.

elalish: A potentially easier way to achieve bullet #3 would be to animate the camera instead of the pivot, since this way they would be independent and their effects would add automatically.

cdata: This would be a breaking change, and also make auto-rotate even more annoying and useless (because it's just sugar over `camera-orbit` at that point). Orbiting the camera != rotating the pivot.

cdata: I may have misunderstood @elalish . I now understand that he was referring to the interaction prompt, not `auto-rotate`.

So, yeah, I agree, that would be a reasonable path forward as well.

src/template.ts
```diff
@@ -165,6 +165,27 @@ template.innerHTML = `
       transform: translateY(-100%);
     }
 
+    @keyframes wiggle {
+      10%, 12% {
+        transform: translateX(-5%);
```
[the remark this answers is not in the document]
maaslalani: Building off the 3rd option, and offering a CSS custom property to override the `animation-name` property. Instead of `animation-name`, we can simply add the override for `animation` which is the shorthand to all the animation properties. This way the user can specify their own animation keyframes (name), animation duration, delay, iteration count, etc... by overriding this exposed property.

CSS Syntax:
```css
animation: name duration timing-function delay iteration-count direction fill-mode play-state;
```

src/features/controls.ts
```diff
@@ -296,18 +293,40 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
 
           this[$promptElement].classList.add('visible');
         }
       }
 
+      if (this[$promptElementVisible]) {
+        const {
+          left: modelViewerLeft,
+          width: modelViewerWidth,
+        } = this.getBoundingClientRect();
```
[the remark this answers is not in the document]
maaslalani: https://github.com/GoogleWebComponents/model-viewer/pull/788/commits/89f9040cdb77727335f302dad012c26c7abc7f34

src/template.ts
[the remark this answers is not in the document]
cdata: @maaslalani it's a nice idea, but unfortunately shadow root encapsulation will not allow for this. If I specify a keyframe animation in the global document scope, it won't be available in the local shadow scope. The user would be able to change the name that is sued, but CSS won't be able to find the keyframes that correspond to that name (except in legacy polyfill scenarios, where encapsulation is just a facade). [Here is an example that demonstrates the problem](http://lopsided-motion.glitch.me/animation-encapsulation.html).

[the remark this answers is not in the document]
pushmatrix: I suppose there is precedence for the attribute approach. We have the whole `allow-when-focused` and `always-allow` which are opinionated ways of interacting with the viewer. 

So it comes with `basic` and `wiggle` out of the box, but it would be nice to be able to add your own.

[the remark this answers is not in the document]
cdata: > So it comes with basic and wiggle out of the box, but it would be nice to be able to add your own.

@pushmatrix agreed, and this is why this change is relatively difficult to land as-is. The UX is quite opinionated and difficult to customize. Ideally we would lean in to something like #738 to accomplish specialty interaction prompts like this.

As mentioned above, CSS parts would make this effect a lot easier to customize. For now, we'll have to wait for future web platform features to land, and try to get to #738 as quickly as we can.

src/features/controls.ts
```diff
@@ -143,7 +146,10 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
 
     protected[$promptElement] =
         this.shadowRoot!.querySelector('.controls-prompt')!;
+    protected[$promptElementSVG] =
```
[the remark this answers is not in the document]
maaslalani: Changed so that we no longer depend on the SVG element (https://github.com/GoogleWebComponents/model-viewer/pull/788/commits/c7dfc339dcdfe09cfe894cd9b02a784127e1e885), Since the opacity is being animated on the slot itself removing the class would suddenly fade out the prompt element. I added a `div.wrapper` to solve this. If there's a better way to do this let me know! Tested out with an `img` and works correctly, so the icon is customizable via slots.

cdata:
@maaslalani great work on the updates! In the interest of time, I'm going to make a proposal for some final tweaks to this PR. My hope is that we can land it ASAP and cut a release on Wednesday or Thursday.

pushmatrix:
We had an accessibility audit done, and the main result we found with this (and the previous CTA) is that continuous animations (even with a pause) can be problematic. This is for someone with a cognitive disability, who's easily distracted, or someone having difficulty reaching and interacting with the model in order to cease the animation completely.

It was suggested a maximum of 3 rotations and then stopping. I'd be tempted to make that a parameter, but then it feels like we have parameter bloat.

Curious about your thoughts on this.

src/template.ts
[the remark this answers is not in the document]
cdata: The WebKit folks have blessed us with some promising news on this topic: https://webkit.org/blog/9609/release-notes-for-safari-technology-preview-94/

> Added support for the ::part() pseudo element from CSS Shadow Parts

cdata:
These changes were landed via #834 . Thanks again @maaslalani !

PR #793, opened 2019-09-30
Update Filament

elalish:
Updating to Filament.js 1.4.0, which they were nice enough to roll for me. Fixes many of the fidelity issues we were seeing on their side. Also, they showed me how to set the exposure to be identical to ours, so no more guesswork. I need to update the screenshots once more from my mac to get rid of the scrollbars.


src/test/fidelity/components/renderers/filament-viewer.ts
```diff
@@ -250,5 +250,6 @@ export class FilamentViewer extends LitElement {
         verticalFoV, aspect, near, far, Fov!.VERTICAL);
     const up = [0, 1, 0];
     this[$camera].lookAt(eye, center, up);
+    this[$camera].setExposure(1.0, 1.2, 100);
```
cdata: Mysterious!

PR #795, opened 2019-10-02
Stop event bubbling when handling behavior is 'prevent-all'

### LINKED ISSUE #794: Model-viewer does not prevent scroll in horizontal carousels
### Description

By default the model viewer hijacks mouse / touch movements. Meaning you can't scroll down the page when your finger is on it. This is intended.

However in carousels, it seems like the model-viewer is not preventing horizontal movement:

carousel

It should act the same horizontally as it does vertically to be consistent.

#### Live Demo
Check it out here: https://steep-possum.glitch.me

### Browser Affected
All browsers

### OS
All OS

### Versions
- model-viewer: v0.6.2
- three.js: r108

pushmatrix:
Fixes #794

It seems like we are doing `preventDefault()` but we should also be doing `stopPropagation()`. Do you know if this would have any ill-intended events?

I also couldn't quite figure out how to test this.
https://github.com/GoogleWebComponents/model-viewer/blob/master/src/test/three-components/SmoothControls-spec.ts#L304-L308

Maybe we need to register another event on it and make sure it doesn't fire?

### DIFF
--- a/src/three-components/SmoothControls.ts
+++ b/src/three-components/SmoothControls.ts
@@ -638,6 +638,7 @@ export class SmoothControls extends EventDispatcher {
     if ((handled || this[$options].eventHandlingBehavior === 'prevent-all') &&
         event.cancelable) {
       event.preventDefault();
+      event.stopPropagation();
     };
   }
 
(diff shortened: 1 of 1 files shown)


cdata:
@pushmatrix for the horizontal carousel case, is scrolling otherwise handled by the browser, or is there manual horizontal scrolling being performed by content scripts?

cdata:
See my comment in the related issue https://github.com/GoogleWebComponents/model-viewer/issues/794#issuecomment-538588798

PR #808, opened 2019-10-09
Fixed autorotate interpolation

### LINKED ISSUE #806: Model jumps around when camera-target changes
The model jumps around in unexpected ways when changing `cameraTarget`. See the attached screen recording, taken from our deployed docs site.

cameratargetwat

elalish:
Fixes #806 

The first fix was a one-liner to force the pivot to recalculate when the target changes. This caused me to also notice it wasn't smoothly interpolating along with the camera, so I fixed that too.

### DIFF
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -306,7 +306,7 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
       const target = this.getCameraTarget();
       if (!this[$scene].pivotCenter.equals(target)) {
         this[$scene].pivotCenter.copy(target);
-        this[$scene].setRotation();
+        this[$scene].setRotation(this[$scene].pivot.rotation.y);
       }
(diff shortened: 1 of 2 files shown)


src/features/controls.ts
```diff
@@ -302,6 +303,11 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
       }
 
       this[$controls].update(time, delta);
+      const target = this.getCameraTarget();
+      if (!this[$scene].pivotCenter.equals(target)) {
+        this[$scene].pivotCenter.copy(target);
+        this[$scene].setRotation();
```
cdata: Here we fall back to the argument default, which is tricky to understand without reading the method signature. But, it's equivalent to what we are doing on line 280 because the default we just added is the same value. Let's just drop the default and specify the value here instead.

PR #809, opened 2019-10-09
fixed shadow fighting

elalish:
I guess I need to make an issue for this. Anyway, this makes our shadows not fight with models that have shadows baked in.


cdata:
SGTM. Although, we should probably also talk about the floor plane removal hack. I'll file an issue for that.

cdata:
FYI https://github.com/GoogleWebComponents/model-viewer/issues/810

PR #822, opened 2019-10-11
fix shadow Z-fight on mobile

elalish:
This comes down to precision, which varies a lot by GPU, so while I've tested this on Pixel2 XL, we should ideally try to verify on a few more phones. I didn't want to make the gap any larger, because you can start to see it when you rotate to floor level. I suppose this is another reason to remove the baked-in shadow entirely, but I don't see a robust way to identify it in general.


cdata:
If we make the removal opt-in, it removes the burden on us to provide a super robust solution (at least initially). Thanks for the the testing!

PR #831, opened 2019-10-15
Introduce auto-rotate-delay attribute

### LINKED ISSUE #823: Add auto-rotate-delay to set time before auto-rotate starts
We already have the built-in notion of a delay before auto-rotate begins, but we don't expose it as an attribute. We should expose it as an attribute so that users can configure the model to do things like auto-rotate immediately, or auto-rotate after a longer period of time.

### LINKED ISSUE #827: Don't reset the auto-rotate rotation when autoRotate is disabled
### Description

I am trying to disable auto-rotation after the user interacts with the model and I'm doing this by setting autoRotate to false in the event handler for 'camera-change' when source is 'user-interaction'.  The problem is that turning autoRotate off causes the rotation to reset to 0 which does an immediate snap of the rotation back to unrotated.

#### Live Demo
Attached zip file with repro.
[jump.zip](https://github.com/GoogleWebComponents/model-viewer/files/3726933/jump.zip)

### Browser Affected
- [x] Chrome - tested
- [x] Edge
- [x] Firefox
- [x] Helios
- [x] IE
- [x] Safari

### OS
- [x] Android - tested
- [x] iOS
- [x] Linux
- [x] MacOS - tested
- [x] Windows

### Versions
- model-viewer: head on 10/14 - testing from https://unpkg.com/@google/model-viewer/dist/
- three.js: rXX

cdata:
This change introduces an `auto-rotate-delay` attribute, which allows users to configure the a previously internal-only delay value that defaults to 3 seconds. After the delay has passed, the model will start rotating.

Additionally, this change revises our previous behavior for when `auto-rotate` is toggled. Previously, we would reset the "turntable" rotation when `auto-rotate` is disabled, causing the model to appear to jump back to its initial rotation. Now, the rotation is not reset when `auto-rotate` is disabled. In order to achieve the previous effect, a new `resetTurntableRotation` method has been introduced that sets the turntable's rotation back to 0.

Fixes #823 
Fixes #827

### DIFF
--- a/index.html
+++ b/index.html
@@ -131,7 +131,12 @@
             </li>
             <li>
               <div>auto-rotate</div>
-              <p>Enables the auto rotation of the model.</p>
+              <p>Enables the auto-rotation of the model.</p>
+            </li>
+            <li>
+              <div>auto-rotate-delay</div>
+              <p>Sets the delay before auto-rotation begins. The format of the value
+              is a number in milliseconds. The default is 3000.</p>
             </li>
             <li>
               <div>autoplay</div>
@@ -438,6 +443,13 @@
               <p>Causes animations to be paused. If you want to reset the
               current animation to the beginning, you should also set the
               currentTime property to 0.</p>
+            </li>
+            <li>
+              <div>resetTurntableRotation()</div>
+              <p>Resets the turntable that rotates the model when auto-rotate
+              is enabled. The new value of the turntable rotation will be 0 after
+              this method is invoked, but the model may not update until the next
+              render frame.</p>
             </li>
             <li>
               <div>toDataURL(type, encoderOptions)</div>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -131,7 +131,12 @@ <h3 class="grouping-title">Attributes</h3>
             </li>
             <li>
               <div>auto-rotate</div>
-              <p>Enables the auto rotation of the model.</p>
+              <p>Enables the auto-rotation of the model.</p>
+            </li>
+            <li>
+              <div>auto-rotate-delay</div>
+              <p>Sets the delay before auto-rotation begins. The format of the value
+              a number in milliseconds. The default is 3000.</p>
```
elalish: +is

PR #833, opened 2019-10-15
Only load when in render tree

### LINKED ISSUE #815: Models are loaded even when the model is not in the render tree
If a `<model-viewer>` element is `display: none`, we still load the model. We shouldn't. Let's fix that!

cdata:
Fixes #815

### DIFF
--- a/src/features/loading.ts
+++ b/src/features/loading.ts
@@ -15,7 +15,7 @@
 
 import {property} from 'lit-element';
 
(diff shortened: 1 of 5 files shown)


cdata:
@elalish there are configurable thresholds for this. We haven't done a lot of study to figure out what  would be ideal, but it might be worth our time to experiment. You can configure both distance from the viewport before you are considered to be intersecting, and also percentage of intersection that you care about (e.g., 25% of the element is intersecting, vs just 10%).

pushmatrix:
🥂

PR #834, opened 2019-10-16
New interaction prompt final touches

### LINKED ISSUE #708: Animate model along with interaction prompt

### LINKED ISSUE #817: Update controls prompt icon hand SVG to work with all types of models
### Description

The current hand icon svg isn't easily visible for some models. For example, a model that has a lot of white will make the prompt hard to see:

https://abalone-diplodocus.glitch.me/

One solution would be to allow the hand colour to be customized, but that won't work in cases where the model-viewer doesn't know what the input glb will be in advance.

A better solution is to have an icon that works in all scenarios, regardless of background and model materials. 

What do you think of something like this:

image

<table>
<tr>
<th>Proposal</th>
<th>Current</th>
</tr>
<td>
Screen Shot 2019-10-10 at 3 38 05 PM
</td>
<td>
Screen Shot 2019-10-10 at 3 38 58 PM
</td>
</tr>
<tr>
<td>
Demo: <a href="https://elite-pincushion.glitch.me/">https://elite-pincushion.glitch.me/</a>
</td>
<td>
Demo: <a href="https://translucent-lunch.glitch.me/">https://translucent-lunch.glitch.me/</a>
</td>

</tr>
</table>

cdata:
This change ties up the last required changes for us to land #788 

A brief summary of changes over #788 :

 - New prompt SVG via #817 
 - Added attribute to revert to non-animated prompt (`interaction-prompt-style="basic"`)
 - Fixed transitions between configuration states
 - Makes prompt animation additive with `auto-rotate`
 - Changed styles so that focus ring doesn't appear to float with animation
 - Avoids invoking `getBoundingClientRect` on `<model-viewer>`
 - Rebased/reconciled with latest changes on master

Fixes #708 
Fixes #817


maaslalani:
@googlebot I consent.

src/features/controls.ts
```diff
@@ -327,11 +348,31 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+                .getPropertyValue('transform');
+        // Parse the fifth term of computed transform style
+        // which is in form of matrix(n0, n1, n2, n3, n4, n5)
+        // @see https://www.w3.org/TR/css-transforms-1/#serialization-of-the-computed-value
+        const offset = parseFloat(transformString.split(',')[4]);
+        const delta = offset - this[$lastPromptOffset];
+
+        if (isFinite(offset)) {
+          this[$lastPromptOffset] = offset;
+          this[$scene].pivot.rotation.y +=
```
elalish: I think this should use the setRotation method instead. I think if you move the camera target you'll see incorrect behavior.

src/test/features/controls-spec.ts
```diff
@@ -361,6 +361,38 @@ suite('ModelViewerElementBase with ControlsMixin', () => {
...
+
+        test('plays css animation only when visible', async () => {
+          element.interactionPrompt = 'auto';
+
+          const computedStyle =
+              getComputedStyle((element as any)[$promptAnimatedContainer]);
+          expect(computedStyle.animationPlayState)
+              .to.be.match(/^paused(\, paused)?$/);
+
+          await timePasses(element.interactionPromptThreshold + 100);
```
elalish: This doesn't seem to directly test visibility.

src/test/features/controls-spec.ts
```diff
@@ -426,9 +458,9 @@ suite('ModelViewerElementBase with ControlsMixin', () => {
 
                 canvas.blur();
 
+                const modelLoads = waitForEvent(element, 'load');
                 element.src = ASTRONAUT_GLB_PATH;
-
-                await waitForEvent(element, 'load');
+                await modelLoads;
```
elalish: Is this just for code clarity, or does this affect flakiness?

cdata: It was a hopeful attempt at the later... but I don't think it worked.

src/test/features/controls-spec.ts
cdata: Will fix

src/features/controls.ts
cdata: Will fix

cdata:
Just an update since this has taken a while: IE11 tests are timing out. It's not obvious why. They pass just fine seemingly 100% of the time when run locally:

image

Sauce could be having a bad day, or it could be because it is IE11's 6th birthday today.

pushmatrix:
AhhhhHHhhHHhhhhHhhhHhhhhh Huge props @maaslalani @cdata @elalish for getting this in

Issue #835, opened 2019-10-16
Refactor controls[$updateCamera], move to Model or ModelScene


cdata:
Prior to #739 it made sense for the "controls" feature implementation to compute the ideal field of view value. Now, it probably makes more sense for the `Model` or `ModelScene` to hold this value, to be acted on by the higher-level mixin.

PR #836, opened 2019-10-17
Fix FOV intrinsics

cdata:
This change address an oversight from #739 , where intrinsics for `field-of-view` were not being updated appropriately.

Filed #835 for future refactor work here.

### DIFF
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -54,9 +54,12 @@ export const DEFAULT_CAMERA_ORBIT = '0deg 75deg 105%';
 const DEFAULT_CAMERA_TARGET = 'auto auto auto';
 const DEFAULT_FIELD_OF_VIEW = 'auto';
 
-export const fieldOfViewIntrinsics = {
-  basis: [numberNode(DEFAULT_FOV_DEG * Math.PI / 180, 'rad')],
-  keywords: {auto: [null]}
+export const fieldOfViewIntrinsics = (element: ModelViewerElementBase) => {
+  return {
+    basis: [numberNode(
+        (element as any)[$zoomAdjustedFieldOfView] * Math.PI / 180, 'rad')],
+    keywords: {auto: [null]}
+  };
 };
 
 export const cameraOrbitIntrinsics = (() => {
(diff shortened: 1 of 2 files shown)


src/features/controls.ts
```diff
@@ -385,7 +390,8 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
 
         const maximumFieldOfView = this[$framedFieldOfView]!;
         controls.applyOptions({maximumFieldOfView});
-        controls.setFieldOfView(zoom * this[$framedFieldOfView]!);
+        // TODO(#835): Move computation of this value to Model or ModelScene
+        this[$idealFieldOfView] = this[$framedFieldOfView]! * zoom;
```
elalish: I think I might call this $currentFieldOfView, since zoom is coming from user interaction. $framedFieldOfView is as ideal as it gets.

src/features/controls.ts
```diff
@@ -54,9 +54,12 @@ export const DEFAULT_CAMERA_ORBIT = '0deg 75deg 105%';
 const DEFAULT_CAMERA_TARGET = 'auto auto auto';
 const DEFAULT_FIELD_OF_VIEW = 'auto';
 
-export const fieldOfViewIntrinsics = {
-  basis: [numberNode(DEFAULT_FOV_DEG * Math.PI / 180, 'rad')],
-  keywords: {auto: [null]}
+export const fieldOfViewIntrinsics = (element: ModelViewerElementBase) => {
+  return {
+    basis: [numberNode(
+        (element as any)[$idealFieldOfView] * Math.PI / 180, 'rad')],
```
elalish: Why not skip the pi/180 and just call it degrees?

src/features/controls.ts
cdata: But this is only applied when the `field-of-view` is set to `auto`. In theory, this number should not vary for a given model and aspect ratio, right?

src/features/controls.ts
cdata: All numbers in the style evaluator are currently normalized to either radians or meters. We can change that, but I would rather change it later.

elalish: gotcha, I missed that.

src/features/controls.ts
elalish: $framedFieldOfView is exactly as you've described. $idealFieldOfView is here based on zoom, which is dependent on how the user has zoomed, which makes it no longer ideal.

cdata: Okay, I may have misunderstood what was happening here then. I'll try a different approach.

cdata: I changed some of the names. I *think* the logic is sound, and comparable to what we had before, but I would appreciate any feedback that suggests otherwise!

PR #843, opened 2019-10-18
Fix wiggle in Safari and IE11

### LINKED ISSUE #839: Model does not "wiggle" correctly in Safari
Our new model interaction prompt seems to be very glitchy in Safari. Sometimes it works fine, sometimes the model doesn't rotate, sometimes the model rotation jumps and sometimes the hand icon disappears entirely (but the model is still rotating / glitching). The problem seems to be exacerbated by tabbing-away from Safari and then returning, or resizing it.

glitch

### LINKED ISSUE #842: <model-viewer> does not display in docs in IE11
All you see is an empty white space where the model should be. It looks like we might be getting multiple element definitions for the same tagname somehow.

cdata:
This change addresses messed up wiggling in Safari 13 (and similar glitches in IE11).

As of this change, we now orbit the camera rather than rotate the pivot node while wiggling. The upside of this is that it now looks more like what the user sees when they interact with the model. The possible downside is that the camera smoothing looks aesthetically different from the original implementation.

I tried "jumping" the camera to its goal when wiggling. This achieves a similar look to the original implementation, but also causes the wiggle effect to layer poorly with other effects that change the camera orbit (such as the recently added scroll position tracking in #739 ).

The new implementation was manually tested and is known to work as intended IE11 and Safari 13.

Fixes #839 
Fixes #842

### DIFF
--- a/index.html
+++ b/index.html
@@ -249,7 +249,7 @@
               <div>interaction-prompt-style</div>
               <p>Configures the presentation style of the interaction-prompt when
               it is raised. The two allowed values are "wiggle" and "basic". When
-              set to "wiggle", the prompt will animated from horizontally and the
+              set to "wiggle", the prompt will animate from horizontally and the
               model will appear to be rotated as though the prompt is interacting
               with it. When set to "basic", the prompt is not animated, and instead
               simply appears until it is dismissed by user interaction.
--- a/index.html
+++ b/index.html
@@ -61,7 +61,7 @@
           <template>
 <!-- Import the component -->
 <script type="module" src="https://unpkg.com/@google/model-viewer/dist/model-viewer.js"></script>
-<script nomodule src="https://unpkg.com/@google/model-viewer/dist/model-viewer-legacy.js"></script>
+<script type="noexecute" nomodule src="https://unpkg.com/@google/model-viewer/dist/model-viewer-legacy.js"></script>
 <!-- Use it like any other HTML element -->
 <model-viewer src="examples/assets/Astronaut.glb" alt="A 3D model of an astronaut" auto-rotate camera-controls background-color="#455A64"></model-viewer>
           </template>
@@ -244,6 +244,16 @@
               prompt will only be activated if the element has first received
               focus. The interaction prompt will only display if camera-controls
               are enabled. Defaults to "auto".</p>
+            </li>
+            <li>
+              <div>interaction-prompt-style</div>
+              <p>Configures the presentation style of the interaction-prompt when
+              it is raised. The two allowed values are "wiggle" and "basic". When
+              set to "wiggle", the prompt will animated from horizontally and the
+              model will appear to be rotated as though the prompt is interacting
+              with it. When set to "basic", the prompt is not animated, and instead
+              simply appears until it is dismissed by user interaction.
+              Defaults to "wiggle".</p>
             </li>
             <li>
               <div>interaction-prompt-threshold</div>
--- a/src/features/controls.ts
+++ b/src/features/controls.ts
@@ -388,23 +388,24 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
         }
       }
 
+
       if (isFinite(this[$promptElementVisibleTime]) &&
           this.interactionPromptStyle === InteractionPromptStyle.WIGGLE) {
         const scene = this[$scene];
         const animationTime =
-            ((time - this[$promptElementVisibleTime]) % PROMPT_ANIMATION_TIME) /
-            PROMPT_ANIMATION_TIME;
+            ((time - this[$promptElementVisibleTime]) / PROMPT_ANIMATION_TIME) %
+            1;
         const offset = wiggle(animationTime);
         const opacity = fade(animationTime);
 
         const xOffset = offset * scene.width * 0.05;
-        const deltaTheta = (offset - this[$lastPromptOffset]) * Math.PI / 8;
+        const deltaTheta = (offset - this[$lastPromptOffset]) * Math.PI / 16;
 
         this[$promptAnimatedContainer].style.transform =
             `translateX(${xOffset}px)`;
         this[$promptAnimatedContainer].style.opacity = `${opacity}`;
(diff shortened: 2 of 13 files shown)


src/utilities/animation.ts
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Creates a TimingFunction that uses a given ease to interpolate between
+ * two configured number values.
+ */
+export const interpolate =
+    (start: number, end: number, ease: TimingFunction = easeInOutQuad):
+        TimingFunction => (time: number) => start + (end - start) * ease(time);
+
+/**
+ * Creates a TimingFunction that that interpolates through a weighted list
```
pushmatrix: double `that`

pushmatrix:
This this new system will it be possible to fade off the cursor on interaction like with the css version?
# CSS fade (old implementation)
fade1

# Current
fade2

cdata:
@pushmatrix yes, good catch. That was an unintended omission and I'll add it back in.

pushmatrix:
This is the behaviour I'm seeing in IE11. Are you seeing the same?
ie

cdata:
@pushmatrix nope! That looks a lot like what I was seeing *before* this change though. Can you share a repro case I can try? I'm using our docs page.

cdata:
Actually, judging by the changing lighting and missing smoothing, I'm 99% sure you are testing master in that recording.

pushmatrix:
@cdata Shoot you're right, I keep forgetting `npm run build` doesn't build legacy.

pushmatrix:
Ok tested with proper legacy, works great 👍

index.html
```diff
@@ -245,6 +245,16 @@ <h3 class="grouping-title">Attributes</h3>
               focus. The interaction prompt will only display if camera-controls
               are enabled. Defaults to "auto".</p>
             </li>
+            <li>
+              <div>interaction-prompt-style</div>
+              <p>Configures the presentation style of the interaction-prompt when
+              it is raised. The two allowed values are "wiggle" and "basic". When
+              set to "wiggle", the prompt will animated from horizontally and the
```
pushmatrix: `animated` -> `animate`

src/features/controls.ts
```diff
@@ -356,32 +382,32 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+        const opacity = fade(animationTime);
+
+        const xOffset = offset * scene.width * 0.05;
+        const deltaTheta = (offset - this[$lastPromptOffset]) * Math.PI / 8;
+
+        this[$promptAnimatedContainer].style.transform =
+            `translateX(${xOffset}px)`;
+        this[$promptAnimatedContainer].style.opacity = `${opacity}`;
+
+        this[$controls].adjustOrbit(deltaTheta * Math.PI / 6, 0, 0, 0);
```
pushmatrix: What's the difference here with `Math.PI / 6` vs `Math.PI / 8` above?

cdata: That's a bug! Will fix.

src/features/controls.ts
```diff
@@ -356,32 +382,32 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-          this[$lastPromptOffset] = offset;
-          scene.setPivotRotation(
-              scene.getPivotRotation() +
-              delta / scene.width * OFFSET_ROTATION_MULTIPLIER);
-          this[$needsRender]();
-        }
+        const scene = this[$scene];
+        const animationTime =
+            ((time - this[$promptElementVisibleTime]) % PROMPT_ANIMATION_TIME) /
+            PROMPT_ANIMATION_TIME;
```
elalish: Might be simpler as `((time - this[$promptElementVisibleTime]) / PROMPT_ANIMATION_TIME) % 1`

src/utilities/animation.ts
```diff
@@ -0,0 +1,85 @@
+// Adapted from https://gist.github.com/gre/1650294
+export const easeInOutQuad: TimingFunction = (t: number) =>
+    t < .5 ? 2 * t * t : -1 + (4 - 2 * t) * t;
```
elalish: This seems a little complicated; would we be better off just wiggling as the `sin(time)` intermittently and skipping most of the keyframing stuff? Then again, maybe that's not as configurable...

src/features/controls.ts
cdata: Will fix

cdata:
@elalish @pushmatrix It turned out that the fade-out-upon-interaction wasn't omitted. It was a bug caused by contention between user-interaction and the prompt implementation trying to control the camera. The fix for the contention that I landed on was to always dispatch a `change` event in the frame when user interaction was detected (even though the camera technically hasn't moved yet). This is kind of a half-measure, so I'm open to suggestions for alternative approaches.

src/features/controls.ts
```diff
@@ -24,6 +24,31 @@ import {IdentNode, NumberNode, numberNode, parseExpressions} from '../styles/par
...
+import {timeline} from '../utilities/animation.js';
+
+// NOTE(cdata): The following "animation" timing functions are deliberately
+// being used in favor of CSS animations. In Safari 12.1 and 13, CSS animations
+// would cause the interaction prompt to glitch unexpectedly
+// @see https://github.com/GoogleWebComponents/model-viewer/issues/839
+const PROMPT_ANIMATION_TIME = 5000;
+
+const wiggle = timeline(0, [
+  {frames: 6, value: 0},
```
elalish: I know you explain this down in animation.ts, but it might be nice to have a short overview comment here. I guess the frames are portions of the overall animation time, so more like frame lengths than frame times, yeah?

src/utilities/animation.ts
pushmatrix: Could css & part be reintroduced down the line when the bug is gone? I guess that'd be hard to do though cause not everyone would update their browsers.

src/features/controls.ts
cdata: Will fix

PR #850, opened 2019-10-23
Fix progress state thrashing on first configuration

### LINKED ISSUE #847: Loading progress state thrashes on initial configuration
You can see from our progress bar that the reported loading state is thrashing on first load. It jumps from 0 to 1 really fast and then jumps back down to 0 and loads normally. We should fix this to remove the thrashing.

Here is a slowmo GIF of the visual artifact created by the thrashing state:

progressbar

cdata:
When I implemented #833 I introduced the bug described in #847. We did defer loading models as intended, but since `IntersectionObserver` announces intersections somewhat late compared to property updates we were still performing other side-effects early. Among these effects was environment map generation.

With this change, environment map generation is also deferred until we are in the render tree. The rationale for this is that being `display: none` or outside of the viewport should be considered "no cost" in the typical case.

Also, a new, one-time progress activity has been introduced that measures the time before we get the first signal from `IntersectionObserver` that we are visible. This should account for any additional side-effects that measure progress early.

Fixes #847

### DIFF
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -219,8 +219,7 @@ export default class ModelViewerElementBase extends UpdatingElement {
         for (let entry of entries) {
           if (entry.target === this) {
             const oldValue = this[$isInRenderTree];
-            this[$isInRenderTree] = this[$scene].isVisible =
-                entry.isIntersecting;
+            this[$isInRenderTree] = this[$scene].visible = entry.isIntersecting;
             this.requestUpdate($isInRenderTree, oldValue);
 
             if (this[$isInRenderTree]) {
@@ -241,7 +240,7 @@ export default class ModelViewerElementBase extends UpdatingElement {
     } else {
       // If there is no intersection obsever, then all models should be visible
       // at all times:
-      this[$isInRenderTree] = this[$scene].isVisible = true;
+      this[$isInRenderTree] = this[$scene].visible = true;
       this.requestUpdate($isInRenderTree, false);
     }
   }
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -44,6 +44,7 @@ const $clearModelTimeout = Symbol('clearModelTimeout');
 const $onContextLost = Symbol('onContextLost');
 const $contextLostHandler = Symbol('contextLostHandler');
 
+export const $isInRenderTree = Symbol('isInRenderTree');
 export const $resetRenderer = Symbol('resetRenderer');
 export const $ariaLabel = Symbol('ariaLabel');
 export const $loadedTime = Symbol('loadedTime');
@@ -99,8 +100,9 @@ export default class ModelViewerElementBase extends UpdatingElement {
   @property({converter: {fromAttribute: deserializeUrl}})
   src: string|null = null;
 
-  protected[$loaded]: boolean = false;
-  protected[$loadedTime]: number = 0;
+  protected[$isInRenderTree] = false;
+  protected[$loaded] = false;
+  protected[$loadedTime] = 0;
   protected[$scene]: ModelScene;
   protected[$container]: HTMLDivElement;
   protected[$canvas]: HTMLCanvasElement;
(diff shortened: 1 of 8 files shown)


src/model-viewer-base.ts
```diff
@@ -211,11 +213,24 @@ export default class ModelViewerElementBase extends UpdatingElement {
...
-            this.requestUpdate();
+            const oldValue = this[$isInRenderTree];
+            this[$isInRenderTree] = this[$scene].isVisible =
+                entry.isIntersecting;
+            this.requestUpdate($isInRenderTree, oldValue);
+
+            if (this[$isInRenderTree]) {
+              // Wait a microtask to give other properties a change to respond
+              // to the state change, then resolve progress on entering the
+              // rendre tree:
```
elalish: nit: rendre -> render, change -> chance

src/model-viewer-base.ts
```diff
@@ -211,11 +213,24 @@ export default class ModelViewerElementBase extends UpdatingElement {
...
     if (HAS_INTERSECTION_OBSERVER) {
+      const enterRenderTreeProgress = this[$progressTracker].beginActivity();
+
       this[$intersectionObserver] = new IntersectionObserver(entries => {
         for (let entry of entries) {
           if (entry.target === this) {
-            this[$scene].isVisible = entry.isIntersecting;
-            this.requestUpdate();
+            const oldValue = this[$isInRenderTree];
+            this[$isInRenderTree] = this[$scene].isVisible =
```
elalish: Can we remove this[$scene].isVisible? It seems to be duplicate.

src/test/features/environment-spec.ts
```diff
@@ -163,6 +163,20 @@ suite('ModelViewerElementBase with EnvironmentMixin', () => {
...
+    const environmentChangeHandler = () => environmentChangeCount++;
+    element.addEventListener('environment-change', environmentChangeHandler);
+    element.style.display = 'none';
+    document.body.appendChild(element);
+    await rafPasses();
+    expect(environmentChangeCount).to.be.equal(0);
+    element.style.display = 'block';
+    await waitForEvent(element, 'environment-change');
+    expect(environmentChangeCount).to.be.equal(1);
+    element.addEventListener('environment-change', environmentChangeHandler);
```
elalish: What does this extra addEventListener do?

cdata: Sorry, typo. Fixed.

src/model-viewer-base.ts
cdata: No, I think it should stay. The reason why is that it controls the notion of visibility on the Three.js side, while `$isInRenderTree` is a detail of the DOM element. If we relied only `$isInRenderTree`, we would be building-in more dependencies on a DOM node in `Renderer`. And then, when we try to move Three.js stuff into a Worker, we have more work to separate those things back out.

src/model-viewer-base.ts
cdata: Done

src/model-viewer-base.ts
elalish: Fair, but ModelScene extends Scene, which already has a visible member. Can we just use that instead of creating isVisible on top of it? Or do they still have separate meanings?

cdata: That's a good point. My guess is that we can re-use `scene.visible`. I'll look into it.

cdata: Switched over. AFAIK `scene.visible` will always be `true` in the code path where we actually render, so it should be fine.

src/test/three-components/Renderer-spec.ts
```diff
@@ -103,15 +103,15 @@ suite('Renderer', () => {
       expect(!scene.isDirty).to.be.ok;
     });
 
-    test('does not render scenes marked as !isVisible', async function() {
-      scene.isVisible = false;
+    test('does not render scenes marked as not visible', async function() {
```
elalish: This is really just a test of three.js at this point, but I guess it isn't hurting anything.

PR #855, opened 2019-10-25
Preserve query string when intenting

### LINKED ISSUE #742: "Couldn't load Object" when using a src Firebase URL
### Description

I'm using Firebase for my app, but setting up `model-viewer` to reference the storage is proving problematic. 

When attempting to launch Scene Viewer on Android I get "Couldn't load Object - Looks like there's something wrong with this object." 

couldnt load obj

The model loads fine when the URL is relative or points to some other external source. When I include an `ios-src`, iOS correctly handles the Firebase urls and displays the model.

To simplify and to make sure it wasn't my model, I uploaded the Astronaut and replicated a basic setup on glitch. A snippet like this should replicate the issue (spaces added to the URL so you don't have to scroll horizontally):

```
<model-viewer src="https://firebasestorage.googleapis.com/v0/b
                   /<proj>.appspot.com/o/model%2FAstronaut.glb
                   ?alt=media&token=<token>"
              camera-controls ar
              interaction-prompt="auto">
</model-viewer>
```

#### Live Demo

[Demo on Glitch](https://shining-foxglove.glitch.me/)

### Browser Affected

- [x] Chrome (tested on version 76.0.3809.132)

### OS

- [x] Android (tested on version 9)

### Versions

- model-viewer: v0.6.0

cdata:
Although this is not currently a common case, we have been breaking Scene Viewer integration whenever the model URL contains sensitive query parameters. A common case for this is when models are hosted on static storage that allows token-based access. In such scenarios, the token is often added to the query string.

Fixes #742

### DIFF
--- a/examples/augmented-reality.html
+++ b/examples/augmented-reality.html
@@ -59,7 +59,7 @@
         </div>
         <example-snippet stamp-to="demo-container-1" highlight-as="html">
           <template>
-<model-viewer src="assets/Astronaut.glb?link" ar camera-controls alt="A 3D model of an astronaut" background-color="#222" ios-src="assets/Astronaut.usdz" magic-leap unstable-webxr></model-viewer>
+<model-viewer src="assets/Astronaut.glb" ar camera-controls alt="A 3D model of an astronaut" background-color="#222" ios-src="assets/Astronaut.usdz" magic-leap unstable-webxr></model-viewer>
           </template>
         </example-snippet>
 
--- a/examples/augmented-reality.html
+++ b/examples/augmented-reality.html
@@ -59,7 +59,7 @@
         </div>
         <example-snippet stamp-to="demo-container-1" highlight-as="html">
           <template>
-<model-viewer src="assets/Astronaut.glb" ar camera-controls alt="A 3D model of an astronaut" background-color="#222" ios-src="assets/Astronaut.usdz" magic-leap unstable-webxr></model-viewer>
+<model-viewer src="assets/Astronaut.glb?link" ar camera-controls alt="A 3D model of an astronaut" background-color="#222" ios-src="assets/Astronaut.usdz" magic-leap unstable-webxr></model-viewer>
           </template>
         </example-snippet>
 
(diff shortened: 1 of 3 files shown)


src/features/ar.ts
```diff
@@ -56,9 +60,12 @@ export const openARViewer = (() => {
 
     title = encodeURIComponent(title);
     modelUrl.protocol = 'intent://';
+    // It's possible for a model URL to have meaningful query parameters
+    // already. Sure hope they aren't called 'link' or 'title' though 😅
+    modelUrl.search +=
+        (modelUrl.search ? '&' : '') + `link=${link}&title=${title}`;
```
elalish: Can we log a console warning if it does find a link or title query param?

cdata: Good idea!

cdata: Done.

examples/augmented-reality.html
```diff
@@ -59,7 +59,7 @@ <h4>This demonstrates several augmented reality features, including <code>unstab
         </div>
         <example-snippet stamp-to="demo-container-1" highlight-as="html">
           <template>
-<model-viewer src="assets/Astronaut.glb" ar camera-controls alt="A 3D model of an astronaut" background-color="#222" ios-src="assets/Astronaut.usdz" magic-leap unstable-webxr></model-viewer>
+<model-viewer src="assets/Astronaut.glb?link" ar camera-controls alt="A 3D model of an astronaut" background-color="#222" ios-src="assets/Astronaut.usdz" magic-leap unstable-webxr></model-viewer>
```
elalish: Is this change intentional?

cdata: Oop

holaworlds:
I am still having the same issue, is there any update for Firebase Storage issue? I was following other [issue](https://github.com/google/model-viewer/issues/742) as well but seems not fully covered yet.

elalish:
I'm not sure what your issue is, specifically. Can you file a new one and fill out the issue template completely so that we can diagnose the problem?

PR #856, opened 2019-10-28
Add minimum viable debug mode

cdata:
This change introduces a very light "debug" mode to `<model-viewer>`. Debug mode can be activated in two ways:

 1. By adding a query parameter to the URL:
    - `http://example.com/foo.html?model-viewer-debug-mode`
 2. By including a global signal on the page with script:
```html
<script>
ModelViewerElement = ModelViewerElement || {};
ModelViewerElement.debugMode = true;
</script>
```

Activating debug mode currently does two things:

 1. Enables verbose shader error messages on the GL context
 2. Exposes renderer internal details with a globally dispatched `model-viewer-renderer-debug` event

An example of using this event looks like:

```html
<!-- Before model-viewer.js is loaded: -->
<script>
self.addEventListener('model-viewer-renderer-debug', (event) => {
  const {
    THREE, // Limited subset of Three.js constructors
    renderer // The singleton instance of Renderer
  } = event.detail;

  // Do debugging stuff...
});
</script>
```

The `Debugger` will also advertise scenes as they are registered and unregistered. You can listen for global `model-viewer-scene-added-debug` and `model-viewer-scene-removed-debug` events, which contain a reference to the related scene as `event.detail.scene`. The scenes in this case will be instances of `ModelScene`.

If you wanted to turn on debugging for most tests, you could manually run them with a URL like: `http://localhost:8000/test/?model-viewer-debug-mode`.

Note that this will currently only enable debug mode for tests that create a `Renderer` via `ModelViewerElementBase`. Tests that instantiate a `Renderer` directly would need to pass a debug flag e.g., `new Renderer({ debug: true });`.

### DIFF
--- a/src/three-components/Debugger.ts
+++ b/src/three-components/Debugger.ts
@@ -1,4 +1,4 @@
-import {Mesh, OrthographicCamera, PlaneBufferGeometry, ShaderMaterial, Texture, WebGLRenderTarget} from 'three';
+import {Mesh, OrthographicCamera, PlaneBufferGeometry, Scene, ShaderMaterial, Texture, WebGLRenderTarget} from 'three';
 
 import {Renderer} from './Renderer';
 
@@ -10,6 +10,7 @@ export interface ModelViewerRendererDebugDetails {
     OrthographicCamera: Constructor<OrthographicCamera>;
     WebGLRenderTarget: Constructor<WebGLRenderTarget>;
     Texture: Constructor<Texture>;
+    Scene: Constructor<Scene>;
     Mesh: Constructor<Mesh>;
   };
 }
@@ -25,7 +26,7 @@ export interface ModelViewerRendererDebugDetails {
  */
 export class Debugger {
   constructor(renderer: Renderer) {
-    renderer.renderer.debug = {checkShaderErrors: false};
+    renderer.renderer.debug = {checkShaderErrors: true};
     Promise.resolve().then(() => {
       self.dispatchEvent(new CustomEvent<ModelViewerRendererDebugDetails>(
           'model-viewer-renderer-debug', {
@@ -35,6 +36,7 @@ export class Debugger {
                 ShaderMaterial,
                 Texture,
                 Mesh,
+                Scene,
                 PlaneBufferGeometry,
                 OrthographicCamera,
                 WebGLRenderTarget
--- a/src/three-components/Debugger.ts
+++ b/src/three-components/Debugger.ts
@@ -1,0 +1,46 @@
+import {Mesh, OrthographicCamera, PlaneBufferGeometry, ShaderMaterial, Texture, WebGLRenderTarget} from 'three';
+
+import {Renderer} from './Renderer';
+
+export interface ModelViewerRendererDebugDetails {
+  renderer: Renderer;
+  THREE: {
+    ShaderMaterial: Constructor<ShaderMaterial>;
+    PlaneBufferGeometry: Constructor<PlaneBufferGeometry>;
+    OrthographicCamera: Constructor<OrthographicCamera>;
+    WebGLRenderTarget: Constructor<WebGLRenderTarget>;
+    Texture: Constructor<Texture>;
+    Mesh: Constructor<Mesh>;
+  };
+}
+
+/**
+ * This Debugger exposes internal details of the <model-viewer> rendering
+ * substructure so that external tools can more easily inspect and operate on
+ * them.
+ *
+ * It also activates shader debugging on the associated GL context. Shader
+ * debugging trades performance for useful error information, so it is not
+ * recommended to activate this unless needed.
+ */
+export class Debugger {
+  constructor(renderer: Renderer) {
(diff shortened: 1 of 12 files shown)


src/three-components/Debugger.ts
```diff
@@ -0,0 +1,46 @@
...
+export class Debugger {
+  constructor(renderer: Renderer) {
+    renderer.renderer.debug = {checkShaderErrors: false};
+    Promise.resolve().then(() => {
+      self.dispatchEvent(new CustomEvent<ModelViewerRendererDebugDetails>(
+          'model-viewer-renderer-debug', {
+            detail: {
+              renderer,
+              THREE: {
+                ShaderMaterial,
```
cdata: We should probably include `Scene` here, I think. Any others, @elalish ? My hope is to limit it to things that we already import for rendering, and are necessary to do essential debug rendering outside of the element.

src/three-components/Renderer.ts
```diff
@@ -96,6 +102,10 @@ export class Renderer extends EventDispatcher {
       this.renderer.physicallyCorrectLights = true;
       this.renderer.setPixelRatio(resolveDpr());
 
+      if (options == null || options.debug === false) {
+        this.debugger = new Debugger(this);
```
elalish: Is this condition backwards? Seems like you make a debugger if debug is false.

src/three-components/Debugger.ts
```diff
@@ -0,0 +1,46 @@
...
+                Texture,
+                Mesh,
+                PlaneBufferGeometry,
+                OrthographicCamera,
+                WebGLRenderTarget
```
elalish: What's the rationale behind these particular things to include?

src/three-components/Renderer.ts
cdata: Ah, yes, good catch. Thanks!

src/three-components/Debugger.ts
cdata: These are intended to be the minimum viable set of things needed to do debug rendering outside of the element. The actual set we should use may be different, have any suggestions?

src/three-components/Debugger.ts
elalish: Agreed on Scene. Renderer already includes TextureUtils, so that's good. We might want to reevaluate what is private, for instance TextureUtils has a private member PMREMGenerator that it might be nice to access. Likewise perhaps with Renderer's private member arRenderer.

src/three-components/Debugger.ts
elalish: So basically what's needed to display a texture? Yeah that sounds good. I'm not sure what other kinds of debugging we'd do here.

cdata: We can add more as we need it

src/three-components/Renderer.ts
cdata: Fixed

src/three-components/Debugger.ts
cdata: I added `Scene`, and I added a read-only accessor for the `PMREMGenerator` in `TextureUtils`. Let's add `arRenderer` later if and when we need it.

elalish: When you said Scene, I was actually thinking [$scene], since that owns things like the model and the shadow. The renderer has the set of all scenes, but it might be hard to tell which one relates to this element.

cdata: Yah, currently there is no offered mechanism to distinguish the scenes. Originally, I added a special event to advertise each scene as it is registered with the `Renderer`, but I don't have a use case for it yet so I took it out.

I'm happy to add it back in if you think we should land it now.

elalish: I don't have a strong opinion, but I like the idea of being able to access all of the textures. Many of those are in the Model's materials and the shadow lives in the ModelScene, so having hooks to that seems important.

src/three-components/Renderer.ts
```diff
@@ -96,6 +102,12 @@ export class Renderer extends EventDispatcher {
       this.renderer.physicallyCorrectLights = true;
       this.renderer.setPixelRatio(resolveDpr());
 
+      if (options != null && options.debug) {
+        this.debugger = new Debugger(this);
+      } else {
+        this.renderer.debug = {checkShaderErrors: false};
```
elalish: if this condition is going to live here, maybe turning it on should live here too instead of buried in the Debugger.ts?

cdata: Yeah, it's funky because it's technically on by default, I just wanted the `Debugger` to assert that it be on when it is used.

src/three-components/Debugger.ts
cdata: Added advertisement of scenes as they are registered/unregistered. Details in the PR description.

src/three-components/Renderer.ts
cdata: I changed the logic here to make it more explicit. LMKWYT.

PR #861, opened 2019-10-31
Implement toBlob function on ModelViewerElementBase

### LINKED ISSUE #858: Improving the screenshot API using toBlob instead of toDataURL
### Description

`<model-viewer>` added the `toDataURL` API method as part of https://github.com/GoogleWebComponents/model-viewer/issues/596 to make it easier to extract the shadow DOM's canvas element but one of the issues with `toDataURL` is that it is sync and blocks the page while generating the data.

Canvas's [toBlob API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob) can do the same thing as `toDataURL` with the benefit of being async and prevent page blocking. The only wrinkle is that `toBlob` isn't supported on Edge browsers. Would it be possible to have the `toDataURL` API make use of `toBlob` if available with an async interface while still falling back on`toDataURL` for unsupported (cough edge) browsers?

/cc @pushmatrix

maaslalani:
### Reference Issue
Fixes #858 

Adds `toBlob` to the `ModelViewerElementBase`.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob

### General Approach
The general approach was to delegate the `toBlob` on the model-viewer to the canvas. Instead of using callbacks like `HTMLCanvasElement` does with `toBlob`, I used promises so that it is possible to use `async/await` syntax. We can change this back if needed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -475,8 +475,9 @@
               <p>Returns a promise that resolves into a Blob object in the format
               specified by the <i>mimeType</i> (defaults to image/png). A Blob object
               represents a file-like object of immutable, raw data. You can also
-              specify a value between 0 and 1 for <i>qualityArgument</i> which defaults
-              to 0.92 and 0.8 for image/png and image/webp respectively.</p>
+              specify a value between 0 and 1 for <i>qualityArgument</i> (Currently
+              only available on Chrome desktop and Firefox) which defaults to 0.92
+              and 0.8 for image/png and image/webp respectively.</p>
             </li>
           </ul>
 
--- a/index.html
+++ b/index.html
@@ -470,6 +470,14 @@
               value for <i>encoderOptions</i> between 0 and 1 (<i>encoderOptions</i>
               defaults to 0.92 otherwise).</p>
             </li>
+            <li>
+              <div>toBlob(mimeType, qualityArgument)</div>
+              <p>Returns a promise that resolves into a Blob object in the format
+              specified by the <i>mimeType</i> (defaults to image/png). A Blob object
+              represents a file-like object of immutable, raw data. You can also
+              specify a value between 0 and 1 for <i>qualityArgument</i> which defaults
+              to 0.92 and 0.8 for image/png and image/webp respectively.</p>
+            </li>
           </ul>
 
         </div>
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -328,8 +328,17 @@ export default class ModelViewerElementBase extends UpdatingElement {
 
   async toBlob(mimeType?: string, qualityArgument?: number): Promise<Blob> {
     return new Promise(async (resolve, reject) => {
+      if ((this[$canvas] as any).msToBlob) {
+        // NOTE: msToBlob only returns image/png
+        // so ensure mimeType is not specified (defaults to image/png)
+        // or is image/png, otherwise fallback to using toDataURL on IE.
+        if (!mimeType || mimeType === 'image/png') {
+          return resolve((this[$canvas] as any).msToBlob());
+        }
+      }
+
       if (!this[$canvas].toBlob) {
-         return resolve(await dataUrlToBlob(this[$canvas].toDataURL(mimeType)));
+        return resolve(await dataUrlToBlob(this[$canvas].toDataURL(mimeType, qualityArgument)));
       }
 
       this[$canvas].toBlob((blob) => {
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -24,7 +24,7 @@ import {ModelScene} from './three-components/ModelScene.js';
 import {ContextLostEvent, Renderer} from './three-components/Renderer.js';
 import {debounce, deserializeUrl, isDebugMode, resolveDpr} from './utilities.js';
 import {ProgressTracker} from './utilities/progress-tracker.js';
-import {dataURLToBlob} from './utilities/data-conversion.js';
+import {dataUrlToBlob} from './utilities/data-conversion.js';
 
 let renderer = new Renderer({debug: isDebugMode()});
 
@@ -329,7 +329,7 @@ export default class ModelViewerElementBase extends UpdatingElement {
   async toBlob(mimeType?: string, qualityArgument?: number): Promise<Blob> {
     return new Promise((resolve, reject) => {
       if (!this[$canvas].toBlob) {
-         return resolve(dataURLToBlob(this[$canvas].toDataURL(mimeType)));
+         return resolve(dataUrlToBlob(this[$canvas].toDataURL(mimeType)));
       }
 
       this[$canvas].toBlob((blob) => {
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -24,6 +24,7 @@ import {ModelScene} from './three-components/ModelScene.js';
 import {ContextLostEvent, Renderer} from './three-components/Renderer.js';
 import {debounce, deserializeUrl, isDebugMode, resolveDpr} from './utilities.js';
 import {ProgressTracker} from './utilities/progress-tracker.js';
+import {dataURLToBlob} from './utilities/data-conversion.js';
 
 let renderer = new Renderer({debug: isDebugMode()});
 
@@ -323,6 +324,22 @@ export default class ModelViewerElementBase extends UpdatingElement {
 
   toDataURL(type?: string, encoderOptions?: number): string {
     return this[$canvas].toDataURL(type, encoderOptions);
+  }
+
+  async toBlob(mimeType?: string, qualityArgument?: number): Promise<Blob> {
+    return new Promise((resolve, reject) => {
+      if (!this[$canvas].toBlob) {
+         return resolve(dataURLToBlob(this[$canvas].toDataURL(mimeType)));
+      }
+
+      this[$canvas].toBlob((blob) => {
+        if (!blob) {
+          return reject(new Error('Unable to retrieve canvas blob'));
+        }
+
+        resolve(blob);
+      }, mimeType, qualityArgument);
+    });
   }
 
   get[$ariaLabel]() {
--- a/src/model-viewer-base.ts
+++ b/src/model-viewer-base.ts
@@ -327,9 +327,9 @@ export default class ModelViewerElementBase extends UpdatingElement {
   }
 
   async toBlob(mimeType?: string, qualityArgument?: number): Promise<Blob> {
-    return new Promise((resolve, reject) => {
+    return new Promise(async (resolve, reject) => {
       if (!this[$canvas].toBlob) {
-         return resolve(dataUrlToBlob(this[$canvas].toDataURL(mimeType)));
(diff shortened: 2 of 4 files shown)


src/test/model-viewer-base-spec.ts
```diff
@@ -208,6 +208,32 @@ suite('ModelViewerElementBase', () => {
...
+          expect(objectUrlMatcher.test(objectUrl)).to.be.true;
+        });
+
+        test('has size', async () => {
+          const blob = await element.toBlob();
+          expect(blob.size).to.be.greaterThan(0);
+        });
+
+        test('uses fallbacks on unsupported browsers', async () => {
+          // Emulate unsupported browser
```
maaslalani: What's the best way to do this?

robin-drexler: Maybe you can do 

```ts
const canvas = element[$canvas]
canvas.toBlob = undefined;
```
?

cdata: We wrote a small `spy` helper that would probably be appropriate here:

https://github.com/GoogleWebComponents/model-viewer/blob/0e6582b6d6e30ceeb51576435b605425d95bf803/src/test/helpers.ts#L160-L192


Here is an example of a test that uses it:
https://github.com/GoogleWebComponents/model-viewer/blob/0e6582b6d6e30ceeb51576435b605425d95bf803/src/test/features/ar-spec.ts#L45-L63

The basic idea is it lets you easily replace a property descriptor on an object, and returns a function that restores the original property descriptor when called. You could use it to temporarily replace `toBlob` on `HTMLCanvasElement`:

```typescript
const restoreCanvasToBlob = spy(HTMLCanvasElement.prototype, 'toBlob', { value: undefined });
// do your test...
restoreCanvasToBlob();
```

index.html
```diff
@@ -470,6 +470,14 @@ <h3 class="grouping-title">Methods</h3>
               value for <i>encoderOptions</i> between 0 and 1 (<i>encoderOptions</i>
               defaults to 0.92 otherwise).</p>
             </li>
+            <li>
+              <div>toBlob(mimeType, qualityArgument)</div>
```
cdata: It might be worth noting that `qualityArgument` is currently only available in Chrome (desktop only) and Firefox ([caniuse](https://caniuse.com/#feat=mdn-api_htmlcanvaselement_toblob_image_quality)).

src/model-viewer-base.ts
```diff
@@ -325,6 +326,22 @@ export default class ModelViewerElementBase extends UpdatingElement {
     return this[$canvas].toDataURL(type, encoderOptions);
   }
 
+  async toBlob(mimeType?: string, qualityArgument?: number): Promise<Blob> {
+    return new Promise((resolve, reject) => {
+      if (!this[$canvas].toBlob) {
+         return resolve(dataURLToBlob(this[$canvas].toDataURL(mimeType)));
+      }
+
+      this[$canvas].toBlob((blob) => {
```
cdata: [Allegedly IE11 has `HTMLCanvasElement.prototype.msToBlob`](https://caniuse.com/#feat=mdn-api_htmlcanvaselement_toblob) - could you try incorporating that?

src/utilities/data-conversion.ts
```diff
@@ -0,0 +1,28 @@
+export function dataURLToBlob(base64DataURL: string) {
```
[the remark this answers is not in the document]
cdata: Please add a descriptive comment for this utility function.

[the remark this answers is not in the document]
maaslalani: Fixed!

Also, I adjusted the style of the function to match the style of some of the functions found in other utility functions:
https://github.com/GoogleWebComponents/model-viewer/blob/6f4875528cee6db308bfc20037802e58821c1419/src/utilities/animation.ts#L81-L100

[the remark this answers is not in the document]
maaslalani: https://github.com/GoogleWebComponents/model-viewer/pull/861/commits/5a1d85436c8841fb0fa6cc4c0620cc82105a8d52

src/test/model-viewer-base-spec.ts
maaslalani: Awesome! Used this ^

src/model-viewer-base.ts
maaslalani: I tried checking for `msToBlob` within the `if (!this[$canvas].toBlob)` statement but it results in a Typescript error:

```
Property 'msToBlob' does not exist on type 'HTMLCanvasElement'.
```

Is this safe to ignore?

cdata: The easiest thing to do would be to cast it away, kind of like what we do here: https://github.com/GoogleWebComponents/model-viewer/blob/6f4875528cee6db308bfc20037802e58821c1419/src/constants.ts#L78-L79

src/model-viewer-base.ts
```diff
@@ -325,6 +326,31 @@ export default class ModelViewerElementBase extends UpdatingElement {
     return this[$canvas].toDataURL(type, encoderOptions);
   }
 
+  async toBlob(mimeType?: string, qualityArgument?: number): Promise<Blob> {
+    return new Promise(async (resolve, reject) => {
+      if ((this[$canvas] as any).msToBlob) {
+        // NOTE: msToBlob only returns image/png
+        // so ensure mimeType is not specified (defaults to image/png)
+        // or is image/png, otherwise fallback to using toDataURL on IE.
+        if (!mimeType || mimeType === 'image/png') {
```
maaslalani: Let me know if we want to avoid nested if statements and whether the note is necessary/useful.

maaslalani: Also should `'image/png'` be a constant?

cdata: Exceeds expectations! 🙌

cdata:
Push build running now: https://travis-ci.org/GoogleWebComponents/model-viewer/jobs/606219986

src/test/model-viewer-base-spec.ts
```diff
@@ -208,6 +208,46 @@ suite('ModelViewerElementBase', () => {
...
+          const restoreCanvasToBlob = spy(HTMLCanvasElement.prototype, 'toBlob', { value: undefined });
```
cdata: This was intended to help us emulated unsupported browsers. Unfortunately, in Edge this had an unintended consequence (throwing when we can't find the property being spied on).

I would wrap this call in a try/catch, and ignore exceptions thrown here.

src/test/model-viewer-base-spec.ts
```diff
@@ -208,6 +208,46 @@ suite('ModelViewerElementBase', () => {
...
+
+        test('blobs on supported and unsupported browsers are equivalent', async () => {
+          const restoreCanvasToBlob = spy(HTMLCanvasElement.prototype, 'toBlob', { value: undefined });
+          const unsupportedBrowserBlob = await element.toBlob();
+
+          restoreCanvasToBlob();
+
+          const supportedBrowserBlob = await element.toBlob();
+
+          expect(unsupportedBrowserBlob).to.eql(supportedBrowserBlob)
```
cdata: I think you would probably have to compare the result of `await blob.arrayBuffer()` in order to be sure that the results are equal (unless you are just checking that both results are in fact `Blob`s).

src/test/model-viewer-base-spec.ts
cdata: Alternatively, we could change `spy` to define properties even if they are missing. I would accept a change that causes `spy` to define the property on the originally-passed-in object when no property descriptor for the property is found.

In this case, when "restoring" the property, we should effectively delete it from the object.

src/test/model-viewer-base-spec.ts
maaslalani: Fixed!

src/test/model-viewer-base-spec.ts
maaslalani: This seems reasonable, currently I have the `try/catch` in place but if we decide to implement this I can update these tests.

cdata: `try/catch` is fine for now, we can do the other change another time.

src/test/model-viewer-base-spec.ts
```diff
@@ -208,6 +208,63 @@ suite('ModelViewerElementBase', () => {
...
+          }
+
+          const unsupportedBrowserBlob = await element.toBlob();
+
+          restoreCanvasToBlob();
+
+          const supportedBrowserBlob = await element.toBlob();
+
+          const supportedBrowserResponse = new Response(supportedBrowserBlob);
+          const unsupportedBrowserResponse = new Response(unsupportedBrowserBlob);
```
cdata: This looks like it is intended to work around missing `Blob.prototype.arrayBuffer()` in Safari/Edge 👍 

Unfortunately, IE11 doesn't have Fetch API, so it doesn't have the `Response` constructor, and we don't currently attempt to polyfill it.

I think the best course of action here would probably be to just skip IE11 for now (`if (IS_IE11) { return; }`).

src/test/model-viewer-base-spec.ts
```diff
@@ -208,6 +208,63 @@ suite('ModelViewerElementBase', () => {
...
+      suite('toBlob', () => {
+        test('produces a blob', async () => {
+          const blob = await element.toBlob();
+          expect(blob).to.not.be.null;
+        });
+
+        test('can convert blob to object URL', async () => {
+          const blob = await element.toBlob();
+          const objectUrl = URL.createObjectURL(blob);
+          const objectUrlMatcher = /^blob\:https?:\//;
```
cdata: FYI this is the format of the object URL that I get in IE11 when doing these steps manually: `blob:B0C5923D-DC2D-4666-9362-3AFDD6BADC30`

src/test/model-viewer-base-spec.ts
cdata: (also please leave comment about why we are using `Response` here 🙏)

maaslalani: Found [this stack overflow answer](https://stackoverflow.com/a/55204517) that  goes from `Blob` to `arrayBuffer`.

maaslalani: Should I just use: https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer?

cdata: It looks like what you did is required, as `Blob.prototype.arrayBuffer` is apparently not available in Edge/Safari. Just leave a comment to that effect so that someone else reading the test knows why 👍

maaslalani: Sweet, will do

src/test/model-viewer-base-spec.ts
maaslalani: Should we just match: `/^blob\:/` then?

cdata: That's fine with me

cdata:
Thanks for the great work on this @maaslalani LGTM 💯

PR #867, opened 2019-11-01
Zoom canvas properly

### LINKED ISSUE #862: Size not updating when zooming or changing to new display
I saw a co-worker move `<model-viewer>` from one display to another while it was loading, and the size did not update properly. I tested out zooming, and sure enough this bug (#298, #314) is back to haunt us (appropriately, it is Halloween today):

zoomfail

elalish:
I've reverted the issue fix (since the fix didn't entirely work), but figured we could merge what's left of this simply for the minor cleanup. 

I was so confused by our many different canvases and contexts that I did a little rename here to disambiguate. I also noticed a few `as any` that weren't needed so I removed them.


src/three-components/Renderer.ts
```diff
@@ -237,13 +237,13 @@ export class Renderer extends EventDispatcher {
...
-          this.canvas.height - heightDPR,
+          this.canvas3D.height - heightDPR,
           widthDPR,
           heightDPR,
           0,
           0,
-          widthDPR,
-          heightDPR);
+          width,
+          height);
```
cdata: I think this will be fine for effective DPRs of `1`, but on my MBP we get this predictable result:

Canvas pixels w/ transparent space |
---|
image|

You have to take DPR into account when drawing to the 2D canvas context, as the base canvas width/height in pixels is scaled by DPR.

elalish: Reverted.

Issue #605, opened 2019-06-05
Host documentation at https://modelviewer.dev


cdata:
Currently we host our documentation at a Github pages URL. We registered the domain name modelviewer.dev. We should host our documentation under that domain instead!

smalls:
Fixed

PR #869, opened 2019-11-02
Add CNAME pointing at modelviewer.dev to the root

smalls:
Per the docs, this needs to live in the root of our deployed source.

If we don't want to include the actual file here we could have it as a step in the deploy, but that seemed more complex.


smalls:
Fixes #605

PR #880, opened 2019-11-09
Rendering refactor for anti-aliasing

### LINKED ISSUE #724: Textures are blurring when in the distance in 0.6.0
### Description

0.6.0 added improvements to fix issues like the Moiré issue:  https://github.com/GoogleWebComponents/model-viewer/issues/632

It does a pretty good job:
Screen Shot 2019-08-27 at 1 39 45 PM

However on the left hand photo above, you notice some banding happening.

Looking at it closer, it looks as though the texture is blurring the further away from the camera it is. Seems like anisotropic filtering issue.
Screen Shot 2019-08-27 at 1 40 14 PM

Here's a gif of it happening (notice the detail in the stitching and zippers)
details

#### Live Demo
Compare both new and old viewers here: https://saber-soup.glitch.me/

### Browser Affected

- [ x ] Chrome 76.0.3809.100
- [ x ] Safari (desktop and mobile)  12.1.1 (14607.2.6.1.1)

### OS

- [x ] iOS (12.3)
- [x ] MacOS (10.14.5)

### Versions

- model-viewer: v0.6.0
- three.js: r105

elalish:
Fixes #724 

This is the big one I've been working on for weeks now, and I believe it solves the rest of this umbrella issue that has represented our various rendering artifacts. The normal map detail is back, we have a new roughness mipmapper to anti-alias the normal map, as well as anti-aliasing on the geometry normals. I've also updated PMREM to simplify its shader, remove some blur artifacts, and remap the roughness. I've also made a new fidelity test, DirectionalLightTest, which compares Filament's directional light to our PMREM using a single white pixel as input, scaled to be the same 1 lux of illuminance. This is a very hard test, since a directional light can make much more precise calculations, and I'm quite happy with the result. Note some of the difference is also due to modeling multiscattering differently in three vs filament. 

You can still find aliasing here and there, but it's much improved and some I can't do much about (like where triangles don't share normals; MSAA handles this pretty well, but it's not always enough).


elalish:
I can see in the diff this still needs a little cleanup; I'll fix it up Monday since I doubt anyone will be looking at it between now and then anyway.

elalish:
This is ready for review now. I had to bump the three.js version since the roughnessMipmapper is dependent on a fix a pushed into three last month.

pushmatrix:
hot diggity dog! 🌭 
 
It's beautiful! 👏 

🎉 Incredible work with this. I just tried it with several models and it's amazing.

package.json
```diff
@@ -83,6 +83,7 @@
     "@types/puppeteer": "^1.11.1",
     "@types/resize-observer-browser": "^0.1.2",
     "@types/three": "^0.93.12",
+    "@types/webgl2": "^0.0.5",
```
cdata: That's interesting. What are these being used for?

elalish: I remember you found this fix for me when I upgraded three and got type errors. Three added it to their dependencies; maybe I can just inherit it directly now?

cdata: Oh yah, forgot about that. If it isn't in r108, then we can't benefit from their fix. Of course, we could also update to r110...

elalish: I did update to r110 here, as I needed my own bugfix. I'll see if I can remove this now.

elalish: I get this error when building if I remove it: node_modules/three/src/renderers/webgl/WebGLUtils.d.ts:3:43 - error TS2304: Cannot find name 'WebGL2RenderingContext'.

3  constructor( gl: WebGLRenderingContext | WebGL2RenderingContext, extensions: any, capabilities: any );

cdata:
I'm going to submit a separate code review, but I wanted to share some renders for comparison so that we can reason about them.

Overall, I think that rendering looks amazing now in many respects. We have come a long way from our humble beginnings thanks to your work on this topic @elalish 🙌

The following is a series of comparisons that compare older versions to this version. I think they generally reflect the great progress our rendering has seen. However, for some models there are cases where details may be considered subjectively worse.

First, let's look at what seem like stand-out awesome cases:

Backpack | Backpack (extreme angle) | Damaged Helmet details | Musculoskeletal System (bone)
----------|----------------------------|-------------------------|--------------------------------
image|image|image|image


I want to also consider some potentially controversial (but arguably correct) changes:

Cerebellum | Brain | Musculoskeletal System (muscle) | 
------------|------|-----------------------------------|
image|image|image

Some observations (offered without suggesting goodness or badness) about the potentially controversial changes:

 - From our starting point v0.5.1, default lighting has progressively become brighter over time
 - Surfaces that looked "wet" or "slick" in previous releases look substantially less so by the time we arrive at this change
 - Many fine details now look much more clear, with fewer undesired artifacts; some fine details appear less "busy" (possibly due to removed artifacts; see Musculoskeletal System)

pushmatrix:
Thanks for making that compilation, @cdata. 

The scene has definitely gotten brighter. You can see he difference even in the astronaut.

0.5.1
Screen Shot 2019-11-15 at 2 13 15 PM

master
Screen Shot 2019-11-15 at 2 13 24 PM

Some of our merchants have started playing around with exposure parameters, and for Rebecca Minkoff we needed to change the env map to something that provided less hard highlights.

cdata:
@pushmatrix thanks for that feedback. I definitely expect that merchants will want to adjust the default lighting conditions to better suit their products, even in the ideal case. The question I hope we can start asking is:

Do our default lighting conditions seem harsh now because of a rendering issue, or because the default environment is actually representative of harsh lighting?

One thing I will note for context about the lighting we are modeling our generated environment after is that it was originally designed and calibrated in Unity, with tweaks after the fact to ensure it looked good with a Filament-based renderer, and then was finally touched up in Photoshop.

src/three-components/PMREMGenerator.ts
```diff
@@ -208,10 +225,12 @@ export class PMREMGenerator {
       generateMipmaps: false,
       type: UnsignedByteType,
       format: RGBEFormat,
-      encoding: RGBEEncoding
+      encoding: RGBEEncoding,
+      stencilBuffer: false
     };
     const cubeUVRenderTarget = this[$createRenderTarget](params);
-    this[$pingPongRenderTarget] = this[$createRenderTarget](params);
+    this[$pingPongRenderTarget] =
```
cdata: This is kind of a meta-comment, but it is tricky to reason about when this gets disposed from the place where it is allocated or assigned.

src/three-components/PMREMGenerator.ts
```diff
@@ -45,6 +45,7 @@ const $roughness = Symbol('roughness');
 const $sigma = Symbol('sigma');
 const $sizeLod = Symbol('sizeLod');
 const $lodPlanes = Symbol('lodPlanes');
+const $axisDirections = Symbol('axisDirections');
 const $blurMaterial = Symbol('blurMaterial');
 const $flatCamera = Symbol('flatCamera');
 const $pingPongRenderTarget = Symbol('pingP$pingPongRenderTarget');
```
cdata: `pingP$pingPong...`

elalish:
I'd like to call out my best piece of evidence that as of now we have more accurate PBR IBL than Filament. 
First is the baseline, Filament rendering our directional light test with an actual directional light:
filament-direct
Second is model-viewer's golden, using the IBL that represents said directional light (a single bright pixel):
model-viewer-golden
Third, we have Filament using the same IBL:
filament-indirect
Note the middle row demonstrates how Filament's diffuse term is covering most of the sphere due to the inaccuracy of spherical harmonics. The top row in model-viewer gets a large haze compared to Filament because three does multiscattering differently. I'd like to know which is more accurate. Finally, note that model-viewer gets down to a proper single-pixel reflection (which does not blink as the view is turned, due to successful anti-aliasing) for the smooth sphere, which Filament does not manage for either directional light or IBL, due to numerical stability issues. This gives us notably clearer reflections in mirrored objects. Also note that the roughness mapping (size of reflection at a given roughness) is not consistent between the directional light and the IBL for Filament.

src/three-components/PMREMGenerator.ts
```diff
@@ -467,31 +503,22 @@ void main() {
...
-          vec2 xzY = R * vec2(xz, sampleDirection.y);
-          if (xzY.x < 0.0) {
-            sampleDirection = vec3(0.0, sign(sampleDirection.y), 0.0);
-          } else {
-            sampleDirection.xz *= xzY.x / xz;
-            sampleDirection.y = xzY.y;
-          }
-        }
+        vec3 axis = latitudinal ? poleAxis : cross(poleAxis, vOutputDirection);
+        if (all(equal(axis, vec3(0.0))))
```
cdata: TIL about `all` and `equal` 💎

src/three-components/RoughnessMipmapper.ts
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {LinearMipMapLinearFilter, Mesh, MeshStandardMaterial, NoBlending, OrthographicCamera, PlaneBufferGeometry, RawShaderMaterial, Scene, Vector2, WebGLRenderTarget} from 'three';
+import {_Math} from 'three/src/math/Math';
```
cdata: Include the file extension for package sub-modules like this e.g., `three/src/math/Math.js`

cdata: For readability's sake, I would suggest renaming the import along the lines of `import {_Math as ThreeMath} ...`

src/three-components/RoughnessMipmapper.ts
```diff
@@ -0,0 +1,181 @@
...
+
+import {sceneRenderer} from './Renderer';
+import {roughness2variance, variance2roughness, varianceDefines} from './shader-chunk/common.glsl';
```
cdata: `roughnessToVariance`, `varianceToRoughness` (unless this is copying some upstream convention from Three.js)

src/three-components/RoughnessMipmapper.ts
```diff
@@ -0,0 +1,181 @@
...
+const $scene = Symbol('scene');
+const $flatCamera = Symbol('flatCamera');
+const $tempTarget = Symbol('tempTarget');
+
+/**
+ * The RoughnessMipmapper class allows for the custom generation of mipmaps for
+ * the roughness map of a MeshStandardMaterial. This custom mipmapping is based
+ * on the normal map, so that as normal variation is lost at deeper mip levels,
+ * that loss is encoded as increased roughness to keep reflections consistent
+ * across zoom levels and reduce aliasing.
```
cdata: Great doc 💯

src/three-components/shader-chunk/lights_physical_fragment.glsl.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * @license MIT
+ * @see https://github.com/mrdoob/three.js/blob/dev/LICENSE
+ */
+
+export const lightsChunk = /* glsl */ `
```
cdata: Did we have to make modifications to this implementation?

src/three-components/shader-chunk/common.glsl.ts
```diff
@@ -135,3 +132,72 @@ vec4 linearToOutputTexel(vec4 value){
     }
 }
 `;
+
+export const varianceDefines = /* glsl */ `
```
cdata: Looks like some docs would offer some helpful context here

src/three-components/WebGLUtils.ts
```diff
@@ -75,3 +78,31 @@ export const applyExtensionCompatibility = (gl: WebGLRenderingContext) => {
     return extension;
   };
 };
+
+/**
+ * Debug method to save an offscreen render target to an image; filename should
+ * have a .png extension to ensure lossless transmission.
+ */
+export const saveTarget = (target: WebGLRenderTarget, filename: string) => {
```
cdata: Can we move this to a separate module?

If we do that, it would be possible to trivially omit it from the deployed bundle in the typical case. Then, when we want it we can just import the module that has it during dev.

src/model-viewer-base.ts
```diff
@@ -21,13 +21,11 @@ import {HAS_INTERSECTION_OBSERVER, HAS_RESIZE_OBSERVER} from './constants.js';
 import {makeTemplate} from './template.js';
 import {$evictionPolicy, CachingGLTFLoader} from './three-components/CachingGLTFLoader.js';
 import {ModelScene} from './three-components/ModelScene.js';
-import {ContextLostEvent, Renderer} from './three-components/Renderer.js';
-import {debounce, deserializeUrl, isDebugMode, resolveDpr} from './utilities.js';
+import {ContextLostEvent, sceneRenderer} from './three-components/Renderer.js';
```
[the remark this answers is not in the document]
elalish: I changed it because I kept getting confused by lines like `const renderer = renderer.renderer` Is there a better way to disambiguate?

src/test/features/environment-spec.ts
```diff
@@ -119,12 +119,6 @@ const waitForLoadAndEnvMap =
     };
 
 suite('ModelViewerElementBase with EnvironmentMixin', () => {
-  suiteTeardown(() => {
-    // Reset the renderer once at the end of this spec, to clear out all
-    // of the heavy cached image buffers:
-    ModelViewerElementBase[$resetRenderer]();
```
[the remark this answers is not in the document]
elalish: I haven't seen this cause any test flakiness; I think it was a red herring. Can we give it a try and I'll put it back if there's a problem?

src/test/fidelity/components/renderers/filament-viewer.ts
```diff
@@ -167,10 +167,21 @@ export class FilamentViewer extends LitElement {
 
     await fetchFilamentAssets([modelUrl, iblUrl, skyboxUrl]);
 
-    this[$ibl] = this[$engine].createIblFromKtx(iblUrl);
-    this[$scene].setIndirectLight(this[$ibl]);
-    this[$ibl].setIntensity(1.0);
-    this[$ibl].setRotation([0, 0, -1, 0, 1, 0, 1, 0, 0]);  // 90 degrees
+    if (lightingBaseName === 'spot') {
```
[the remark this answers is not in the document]
elalish: I'll put a comment. It's a special case for DirectionalLightTest.

src/three-components/CachingGLTFLoader.ts
```diff
@@ -150,6 +153,18 @@ export class CachingGLTFLoader {
...
+        if (!(node as Mesh).isMesh) {
+          return;
+        }
+        const mesh = node as Mesh;
+        const materials =
+            Array.isArray(mesh.material) ? mesh.material : [mesh.material];
+        materials.forEach(material => {
+          if ((material as any).isMeshStandardMaterial) {
+            this.roughnessMipmapper.generateMipmaps(
+                material as MeshStandardMaterial);
```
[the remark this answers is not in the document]
elalish: I've done a little work to check this. The actual render calls are very fast, all together only a few ms. It looks like it takes a long time, but that's just because it makes it upload the textures, which otherwise happens on the first animation frame. Then there's that DPR bug, where on a macbook it takes a long time to setDPR, so I intend to fix that.

src/three-components/ModelUtils.ts
```diff
@@ -74,9 +72,9 @@ export const cloneGltf = (gltf: Gltf): Gltf => {
           specularGlossiness.cloneMaterial(material) :
           material.clone();
       clone.onBeforeCompile = updateShader;
-      // TODO(elalish): remove this when we upgrade three.js to a version with
-      // this fix: mrdoob/three.js#17795
-      clone.vertexTangents = material.vertexTangents;
+      if (!clone.vertexTangents && clone.normalScale) {
```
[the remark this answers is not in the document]
elalish: This is a fix for NormalTangentMirrorTest. Three is considering it, which is where I got the idea.

src/three-components/PMREMGenerator.ts
elalish: oops, thanks.

src/three-components/shader-chunk/lights_physical_fragment.glsl.ts
elalish: Yes, it's where I added the geometry normal anti-aliasing; all the stuff about specularRoughness.

src/three-components/WebGLUtils.ts
elalish: K; do you have a filename in mind? And is there anything else similar that should go in the same one?

romainguy:
@elalish I'd love to hear more about your changes. A couple of things to note: Filament uses 2 band SH for Web/mobile by default, switching to 3 band SH should help (to the extent that SH can be helped with this particularly tricky case). Geometric/specular anti-aliasing is also available but need to be turned on per material for Web/mobile.

src/three-components/PMREMGenerator.ts
elalish: I refactored it; see if this looks better.

src/three-components/shader-chunk/lights_physical_fragment.glsl.ts
elalish: Added docs.

src/three-components/shader-chunk/common.glsl.ts
elalish: Added docs.

src/three-components/ModelUtils.ts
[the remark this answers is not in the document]
elalish: Added docs.

src/three-components/CachingGLTFLoader.ts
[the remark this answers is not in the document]
cdata: Can we file an issue the DPR hot path? Otherwise SGTM

src/model-viewer-base.ts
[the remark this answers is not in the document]
cdata: My main issue is that `sceneRenderer` is redundant and doesn't align with the original class (`Renderer`). It's actually a smell that everything accesses the Three.js `WebGLRenderer` as a sub-property of this `Renderer`. It suggests that `Renderer` should now "own" its own `WebGLRenderer` but rather be borrowing it from elsewhere.

As a compromise, would you be willing to rename the `Renderer.renderer` to `Renderer.threeRenderer`? This would add missing context, as well as establish a distinction that will be helpful if we ever decide to look into using alternative renderers in the future.

src/three-components/CachingGLTFLoader.ts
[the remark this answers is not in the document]
elalish: #886

src/model-viewer-base.ts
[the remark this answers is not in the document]
elalish: SG.

src/three-components/WebGLUtils.ts
cdata: Let's throw them in `utilities/debug.ts`. The shadow helper could also go there.

cdata:
Marked as a breaking change due to Three.js version bump, and also rendering changes.

src/three-components/Renderer.ts
```diff
@@ -273,3 +275,5 @@ export class Renderer extends EventDispatcher {
         {type: 'contextlost', sourceEvent: event} as ContextLostEvent);
   }
 }
+
+export let renderer = new Renderer({debug: isDebugMode()});
```
cdata: Moving `renderer` into here has consequences, but we will only realize them in tests for now. The move means `new Renderer` happens (every) first time that this module gets imported from somewhere, regardless of why it was imported. So, in our test that tests `Renderer`, we will create an extra instance of it and that instance will just sit around and do nothing.

Just putting this out there, since it is bound to come up down the road. No action requested.

src/utilities/debug.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Renders a box representing the shadow camera, which is helpful in
+ * debugging.
+ */
+export const showShadowHelper =
+    (scene: ModelScene) => {
+      if (scene.shadow != null) {
+        const helper = new CameraHelper(scene.shadow.shadow.camera);
+        scene.add(helper);
+      }
+    }
```
cdata: Missing semi-colon here

mrdoob:
🔥🔥🔥🔥🔥🔥

pushmatrix:
🎉 🎉 🎉 🎉 🎉

mrdoob:
Spectacular work Emmett!

For comparing:
https://model-viewer-tester.glitch.me/
https://model-viewer-tester.glitch.me/0.7.2.html

This looks much better with all my local models.
0.7.2 was in fact a bit too bright sometimes and a bit too dark other times.

mrdoob:
@pushmatrix 

Comparing [master](https://model-viewer-tester.glitch.me/) with [0.5.1](https://model-viewer-tester.glitch.me/0.5.1.html)...  😁

0.5.1
Screen Shot 2019-11-19 at 1 08 31 PM

master
Screen Shot 2019-11-19 at 1 08 46 PM

mrdoob:
Actually, that's not fair. I should add `experimental-pmrem` to 0.5.1...

mrdoob:
0.5.1
Screen Shot 2019-11-19 at 1 15 07 PM

master
Screen Shot 2019-11-19 at 1 08 46 PM

mrdoob:
`WatterBottle` is the perhaps the most obvious one.

0.5.1
Screen Shot 2019-11-19 at 1 22 04 PM

master
Screen Shot 2019-11-19 at 1 22 36 PM

cdata:
It's pretty incredible to see how far we have come 😮

BTW it might be better in the comparison to also set `stage-light-intensity="0"`, because that is what is causing that bright spot in the top-center of the model. Of course, since our calibration was relatively messed up in v0.5.1, you might have to up other values to compensate...

pushmatrix:
I'm so happy the models have had time to "dry off". 0.5.1 had lots of models look really wet.

@mrdoob Check out this model in your tool:
https://github.com/GoogleWebComponents/model-viewer/files/3273850/fabricball.gltf.zip

This is what it used to look like:
image

mrdoob:
@pushmatrix Nice! Is it okay if I add that model to `model-viewer-tester`?

pushmatrix:
@mrdoob Of course! Just hope people don't get confused why they are looking at a sphere :P

mrdoob:
Done!

soraryu:
@mrdoob great to see that there's a simple testing page for model-viewer! After trying it out I found a slight bug:

https://model-viewer-tester.glitch.me/ does not allow changing the environment map (nothing happens when selecting something in the dropdown)
https://model-viewer-tester.glitch.me/0.7.2.html does allow changing the env map

elalish:
@soraryu @mrdoob This is because of a breaking change (API rename): #923. Should be easy to fix the glitch. Also, we might want to consider moving that glitch into our docs, as it's quite useful!

PR #890, opened 2019-11-14
Seemless poster transitions

### LINKED ISSUE #599: Seamless poster transitions: configure model to fit desired CSS pixel rect
One of my favourite 3D examples on the web is from Panic: https://panic.com/transmit/

transmit

It's the perfect progressive enhancement. It starts off as an image of the truck, and then when the 3D model is loaded it swaps out the image for the 3D and you don't notice it at all. I'd like to explore what that would look like for model-viewer.

The main issue I see is with supporting different aspect ratios. The aspect ratio can make it tricky to line the poster up seamlessly with the rendered model.

@cdata suggests that it is theoretically possible with a combination of "background-size: cover" and "camera-orbit". You would need to adjust the "camera-orbit" attribute to ensure that the model is the same pixel size regardless of aspect ratio.

Let's see if it's possible to add a public method to make adjusting the camera orbit based on desired CSS pixel size a thing.

elalish:
Fixes #599 

I already did the math work to enable this when we updated our framing method. The key here is the change to the template and the addition of an idealAspect flag for toBlob() that allows for easily generating a poster of the proper aspect ratio. The template change is breaking, so we could remove it and tell people to do it manually with their own poster `<div>`, but honestly I think contain is better than cover anyway, since it won't cut off the image. 

You can see this in action here: https://glitch.com/edit/#!/seemless-poster?path=index.html:3:8


src/model-viewer-base.ts
```diff
@@ -332,30 +333,44 @@ export default class ModelViewerElementBase extends UpdatingElement {
...
-
-      this[$canvas].toBlob((blob) => {
-        if (!blob) {
-          return reject(new Error('Unable to retrieve canvas blob'));
-        }
-
-        resolve(blob);
-      }, mimeType, qualityArgument);
-    });
+  async toBlob(options: {
```
cdata: Let's make `options` optional (`options?`)

src/test/model-viewer-base-spec.ts
```diff
@@ -211,66 +211,76 @@ suite('ModelViewerElementBase', () => {
...
+
+              let restoreCanvasToBlob = () => {};
+              try {
+                restoreCanvasToBlob = spy(
+                    HTMLCanvasElement.prototype, 'toBlob', {value: undefined});
+              } catch (error) {
+                // Ignored...
+              }
+
+              const unsupportedBrowserBlob = await element.toBlob({});
```
cdata: Remove object literal from here and elsewhere after making options optional

elalish:
I noticed a couple of bugs with the above Glitch, which I don't believe are related to this change. First, our docs claim that --poster-color should default to background-color, but that doesn't seem to be happening. 

Second, resizing when height-limited seems to work fine, but when width-limited is janky. Try making the window tall and skinny, then slide the height up and down. I see the object's size vibrate, and sometimes it stops in a state that's cut off (not framed properly). This also makes the poster download wrong when the window is tall.

src/test/model-viewer-base-spec.ts
```diff
@@ -211,66 +211,76 @@ suite('ModelViewerElementBase', () => {
...
-          const supportedBrowserResponse = new Response(supportedBrowserBlob);
-          const unsupportedBrowserResponse = new Response(unsupportedBrowserBlob);
-
-          const supportedBrowserArrayBuffer = await supportedBrowserResponse.arrayBuffer();
-          const unsupportedBrowserArrayBuffer = await unsupportedBrowserResponse.arrayBuffer();
-
-          expect(unsupportedBrowserArrayBuffer).to.eql(supportedBrowserArrayBuffer);
-        });
+        test(
+            'blobs on supported and unsupported browsers are equivalent',
```
cdata: It will require a little setup, but it would be nice to see a test added for `idealAspect`. I think it could roughly look like:

 1. Load a model with `<model-viewer>`
 2. Get `Blob`s with and without `idealAspect`
 3. Create object URLs for each blob
 4. Create `img` tags and assign object URLs as `img.src`
 5. Wait for images to load
 6. Compare the `width` and `height` of the images to expectations

src/model-viewer-base.ts
elalish: I tried to do that, except `const {mimeType, qualityArgument, idealAspect} = options!;` says can't destructure undefined. I was hoping it would just destructure them to undefined as well, but apparently not. Is there a clean way to resolve that?

elalish:
Cool, apparently Edge 17 doesn't support .finally(). Any idea what to use instead?

src/test/model-viewer-base-spec.ts
elalish: Done.

src/model-viewer-base.ts
elalish: Done.

src/test/model-viewer-base-spec.ts
```diff
@@ -15,14 +15,30 @@
...
-import {Constructor} from '../utilities.js';
+import {Constructor, resolveDpr} from '../utilities.js';
 
 import {assetPath, spy, timePasses, until, waitForEvent} from './helpers.js';
 import {BasicSpecTemplate} from './templates.js';
 
 
 const expect = chai.expect;
 
+const expectBlobDimensions =
```
cdata: 💎

src/template.ts
```diff
@@ -107,9 +107,10 @@ canvas.show {
   width: 100%;
   height: 100%;
   position: absolute;
-  background-size: cover;
+  background-size: contain;
+  background-repeat: no-repeat;
   background-position: center;
-  background-color: var(--poster-color, #fff);
+  background-color: var(--poster-color, inherit);
```
cdata: I don't think this is actually correct. `inherit` should just apply the `background-color` of the parent, which is most likely not set because the parent is a `<slot>`.

src/features/environment.ts
```diff
@@ -115,6 +115,15 @@ export const EnvironmentMixin = <T extends Constructor<ModelViewerElementBase>>(
 
       const {backgroundImage, environmentImage} = this;
       let {backgroundColor} = this;
+      if (!backgroundColor) {
```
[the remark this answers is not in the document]
elalish: Good call; I'd just blindly copied it up from below, but you're right, it's not actually needed at all.

src/template.ts
elalish: Interesting; this fixed my glitch, but you're right, it also broke our lazy loading examples (by break I mean the background is the wrong color). I think some other fix is still necessary...

elalish: Fixed.

src/template.ts
```diff
@@ -88,11 +88,13 @@ canvas.show {
 .slot.poster {
   opacity: 0;
   transition: opacity 0.3s 0.3s;
+  background-color: inherit;
 }
 
 .slot.poster.show {
   opacity: 1;
   transition: none;
+  background-color: inherit;
```
cdata: I think this one is redundant.

pushmatrix:
@elalish You're on :fire:!

PR #892, opened 2019-11-20
Renderer cleanup

### LINKED ISSUE #886: Startup is slow on macbooks because setDPR is slow
### Description

As noted here: https://github.com/GoogleWebComponents/model-viewer/issues/490#issuecomment-552510648 and I've repro'ed myself, our default environment map generation is slow on macbooks, and a significant contributor is setting the renderer DPR, which in three calls setSize, so it reallocates the default framebuffer. On Chrome linux this is nearly instantaneous, but for some reason takes up to 90ms on a macbook pro. This also affects our other offscreen renderings, like PMREM and as of #880, roughness mipmapping. 

### Browser Affected

- [x] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [x] MacOS
- [ ] Windows

### LINKED ISSUE #891: Resizing when width-bound is janky
When the window is resized while tall, the model's width vibrates, and can result in the scene being zoomed in after resize.

elalish:
Fixes #886 

This is mostly a continuation of the renderer cleanup and rename to separate our renderer from the threeRenderer. I also removed the setPixelRatio calls I had put in for my offscreen rendering because it turns out that can be really slow on a macbook (even though it's instantaneous on my Linux box). Our startup performance is notably better now on macbooks.


src/features/controls.ts
```diff
@@ -498,6 +499,20 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+      // side-effect of clearing it, and also if we wait for the next rAF to
+      // render again we might get hit with yet-another-resize, or worse we
+      // may not actually be marked as dirty and so render will just not
+      // happen. Queuing a render to happen here means we will render twice on
+      // a resize frame, but it avoids most of the visual artifacts associated
+      // with other potential mitigations for this problem. See discussion in
+      // https://github.com/GoogleWebComponents/model-viewer/pull/619 for
+      // additional considerations.
+      Promise.resolve().then(() => {
+        renderer.render(performance.now());
```
cdata: `this[$renderer]`

src/features/controls.ts
```diff
@@ -498,6 +499,20 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+      // necessary because setting the width and height of the canvas has the
+      // side-effect of clearing it, and also if we wait for the next rAF to
+      // render again we might get hit with yet-another-resize, or worse we
+      // may not actually be marked as dirty and so render will just not
+      // happen. Queuing a render to happen here means we will render twice on
+      // a resize frame, but it avoids most of the visual artifacts associated
+      // with other potential mitigations for this problem. See discussion in
+      // https://github.com/GoogleWebComponents/model-viewer/pull/619 for
+      // additional considerations.
+      Promise.resolve().then(() => {
```
cdata: It looks like moving this doesn't have a practical difference other than changing where the implementation lives. Both methods (`ModelViewer.prototype[$onResize]` and `ModelScene.prototype.setSize`) are invoked in the same turn of the event loop, and a microtask will always happen after the callstack for that turn unwinds.

`ModelScene` is always a member of a `ModelViewerElementBase`-derived class, `ControlsMixin` is optional implementation. So, with this change, any class that does not use `ControlsMixin` will inherit the render issue described in the comment above.

elalish: Interesting; this was how I fixed the resize jank. It seemed to be drawing a new frame before it figured out the the zoom level. It still happens a little though, even after this change, so there must be a better way. I need to find a better way to debug the event loop.

src/features/animation.ts
```diff
@@ -74,14 +75,14 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
       }
 
       this[$paused] = true;
-      this[$renderer].threeRenderer.shadowMap.autoUpdate = false;
+      renderer.threeRenderer.shadowMap.autoUpdate = false;
```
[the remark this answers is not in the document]
elalish: I see; it was unclear to me what the purpose of this indirection was. I'll see if I can add useful comment when I revert it.

src/features/controls.ts
cdata: It's possible that there are cases where one method is called and the other isn't. I'll dig deeper and see if that's what is going on.

elalish:
I think I've addressed your general feedback, but please let me know if I missed anything. I'm still trying to figure out what's up with the event loop.

elalish:
I went ahead and removed my attempt at fixing #891; we can look into that again in another PR.

src/three-components/RoughnessMipmapper.ts
```diff
@@ -105,8 +105,8 @@ export class RoughnessMipmapper {
         texelSize.set(0.0, 0.0);
       }
 
-      threeRenderer.setRenderTarget(this[$tempTarget]);
-      threeRenderer.setViewport(position.x, position.y, width, height);
+      threeRenderer.setViewport(
+          position.x, position.y, width / dpr, height / dpr);
```
cdata: 💯

pushmatrix:
Nice work!

deers:
Hi, is there a way that I can still access the renderer? or is there a way to change gammaInput and gammaOutput value without accessing the renderer?

PR #904, opened 2019-11-27
turned off glTF light shadows

### LINKED ISSUE #903: glTF lights cause erroneous shadows
Note the weird extra shadow: https://narrow-sulfur.glitch.me/

This is caused by a light included in the glTF file.

elalish:
Fixes #903 

Only meshes should cast shadows, not lights (our shadow light is added later). I believe this means when lights are included in the glTF they are still illuminating the model (in addition to our environment light). I assume this is expected behavior, but if anyone feels otherwise, this is the place to change it.

### DIFF
--- a/src/three-components/CachingGLTFLoader.ts
+++ b/src/three-components/CachingGLTFLoader.ts
@@ -142,7 +142,6 @@ export class CachingGLTFLoader {
 
     if (gltf.scene != null) {
       gltf.scene.traverse((node: Object3D) => {
-        node.castShadow = true;
         // Three.js seems to cull some animated models incorrectly. Since we
         // expect to view our whole scene anyway, we turn off the frustum
         // culling optimization here.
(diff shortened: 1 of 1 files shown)


mrdoob:
Wouldn't it be better to remove lights at load time?

elalish:
That’s what I thought at first, but this is a one liner. Also, this removes
the shadows but not the lights, so it depends what the goal is.

mrdoob:
Removing "user" lights is less work for `WebGLRenderer` 😇

cdata:
Up until this point, we have side-stepped this question. It seems like the best thing we could do (until we know what we want to do) is just remove user-supplied lights, because they probably won't look very good in a lot of cases. But, I would also want to do a survey of what Scene Viewer and Quick Look do here. Maybe we can remove them as a separate change?

PR #915, opened 2019-12-04
Add configurable camera extents

### LINKED ISSUE #458: Configuration for min/max extents of camera orbit & field of view
In #288 , we introduced the ability to specify a camera spherical position by user interaction. As part of this change, we chose what we deemed to be [good defaults for the min/max extents of the camera's spherical position](https://github.com/GoogleWebComponents/model-viewer/blob/master/src/three-components/SmoothControls.ts#L51-L63).

As of #452 , we will have public API for configuring the spherical position of the camera.

We should also add the ability for content authors to configure the valid extents of the camera's spherical position. This means adding the ability to specify min/max azimuth, inclination and radius.

elalish:
Fixes #458


index.html
```diff
@@ -221,9 +235,21 @@ <h3 class="grouping-title">Attributes</h3>
               this value changes from its initially configured value, the camera
               will interpolate from its current value to the new value. Defaults
               to "auto", which sets either the vertical or horizontal field of
-              view to 45 degrees depending on the dimenions of the model and the
+              view to 45 degrees depending on the dimensions of the model and the
               aspect ratio of the canvas.</p>
             </li>
+            <li>
+              <div>field-of-view-min</div>
```
cdata: This is very nit-picky, but let's do `min-*`/`max-*` rather than `*-min`/`*-max` to align with CSS-style e.g., `min-width`/`max-width`.

src/features/controls.ts
```diff
@@ -217,6 +227,54 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
         {type: String, attribute: 'field-of-view', hasChanged: () => true})
     fieldOfView: string = DEFAULT_FIELD_OF_VIEW;
 
+    @style({
+      intrinsics: {
+        basis: [
+          numberNode(-Infinity, 'rad'),
```
cdata: The basis value is the number that is multiplied for the purposes of scalar values like a percentage. So, if I specify `200%` for the minimum camera theta, it will be the same as `100%` (since both percentages get applied to `-Infinity`).

In this case that might actually be what you want 🤷‍♂

src/features/controls.ts
```diff
@@ -217,6 +227,54 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+          numberNode(Math.PI / 8, 'rad'),
+          numberNode(0, 'm')
+        ],
+        keywords: {auto: [null, null, null]}
```
cdata: It would probably be useful to add an `infinity` keyword here for convenience. Maybe:

```javascript
keywords: {
  auto: [null, null, null],
  infinity: [numberNode(-Infinity, 'rad'), numberNode(0, 'rad'), numberNode(0, 'm')]
}
```

WDYT?

src/features/controls.ts
```diff
@@ -217,6 +227,54 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+      },
```
cdata: Let's pull the intrinsics declarations out of the decorator so that it is a little less busy here (similarly to what we did with the other ones).

src/features/controls.ts
```diff
@@ -350,6 +408,32 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
       this[$controls].setOrbit(style[0], style[1], style[2]);
     }
 
+    [$syncCameraOrbitMin](style: EvaluatedStyle<SphericalIntrinsics>) {
+      this[$controls].applyOptions({
+        minimumAzimuthalAngle: style[0],
+        minimumPolarAngle: style[1],
+        minimumRadius: style[2]
+      })
```
cdata: Missing semi-colons at the end of these expressions.

src/features/controls.ts
```diff
@@ -470,7 +554,7 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
 
         if (azimuthalQuadrant !== lastAzimuthalQuadrant ||
             polarTrient !== lastPolarTrient) {
-          const {canvas} = (this as any)[$scene];
+          const {canvas} = this[$scene];
```
cdata: 💯

src/three-components/SmoothControls.ts
```diff
@@ -356,13 +356,10 @@ export class SmoothControls extends EventDispatcher {
     // Re-evaluates clamping based on potentially new values for min/max
     // polar, azimuth and radius:
     this.setOrbit();
+    this.setFieldOfView(Math.exp(this[$goalLogFov]));
```
cdata: Can you offer more detail for this change?

src/features/controls.ts
elalish: Yes, that's what made the most sense to me.

src/features/controls.ts
elalish: I dunno, specifying Infinity and getting zero seems pretty weird to me. The other reason I didn't go this route is because it goes by so many names: Inf, inf, infinity, -Inf, etc. Since auto does what we want anyway, it seemed simpler.

src/three-components/SmoothControls.ts
elalish: Sure, I had to update it since goalSpherical didn't include field of view. Then I realized that this was just doing exactly what our jumpToGoal method was intended for; I must have just forgotten to update it when I added that method. It already contains the isStationary check, as well as all the field of view behavior.

src/features/controls.ts
cdata: Fair enough, we can always add it later

index.html
```diff
@@ -221,7 +221,7 @@ <h3 class="grouping-title">Attributes</h3>
               this value changes from its initially configured value, the camera
               will interpolate from its current value to the new value. Defaults
               to "auto", which sets either the vertical or horizontal field of
-              view to 45 degrees depending on the dimenions of the model and the
+              view to 45 degrees depending on the dimensions of the model and the
```
cdata: 💯

plut0nist:
Hello and thank you for this. I've been trying to use the min-field-of-view and max-field-of-view attributes, but nothing seems to be happening. Other attributes work fine. Any ideas what's going on?

elalish:
@plut0nist Sorry, I'm guessing you're using a release version? Our docs get updated automatically, but this is only available in master at the moment. It will be in our 0.8 release slated for next week though. If you're having trouble with this on master, please leave some more details and I'll help figure it out.

plut0nist:
@elalish That explains it– I am using a release version, and since the attributes are already listed on [this page](https://modelviewer.dev/), I thought they were already in there. Looking forward to the release, thanks!

Issue #218, opened 2018-11-15
Inspection mode for debugging and diagnostics


cdata:
It would be useful to be able to force any given `<model-viewer>` into an inspection mode when debugging. This could be as simple as having the element dump information to the console.

theprojectsomething:
Would be great to have some visual helpers available as well, e.g:

- bounding boxes for models
- bounding spheres (max and min), current coords / normals for camera orbit (imo more useful than model BBs)
- planes for annotations

```
<model-viewer data-debug="camera model">
  <button slot="hotspot-0" data-debug></button>
</model-viewer>
```

elalish:
@theprojectsomething I like your example! It's just the two of us on this project and we have a ton of work to do; if you'd be interested in contributing a PR for this, we'd love to work with you!

theprojectsomething:
@elalish absolutely, happy to contribute a PR. Will try to find some time in the next few weeks to go over the code base properly. Cheers to both of you for all your work!

theprojectsomething:
I have zero typescript so have put together add-on package instead, for inclusion in development environments: https://www.npmjs.com/package/model-viewer-debugger

elalish:
@theprojectsomething Awesome, thank you! We're busy upgrading our docs, so we should link to this as we stabilize them.

cdata:
@theprojectsomething That's awesome! You might find the debug mode we added in https://github.com/google/model-viewer/pull/856 (and released in v0.8.0) to be useful.

Basically it gives you access to the internal renderer and a few critical Three.js constructors from the version of Three.js that is bundled w/ `<model-viewer>`.

theprojectsomething:
Ah great, thanks @cdata ... as you can see I opted for research-lite / brute force 👍💣 Will refactor with these in mind as soon as I get a chance.

PR #917, opened 2019-12-06
Hotspots

### LINKED ISSUE #334: Support annotations a la Sketchfab's model viewer
Sketchfab has a cool feature where users can annotate positions in 3D space with labels and markers that allow users to get detail about highlights of the model, and also change the camera perspective / controls scheme as they cycle through active annotations.

elalish:
Fixes #334 

Well, 'fixes' might be going a little far, but this is a first step. All I have here is the ability to put blue spots in space, but they work and they become partially transparent when viewed from behind (which the UX studies seem to be favoring). Obviously we'll want to make these more CSS customizable and interactive. For animated models we'll also want a mesh-relative way to specify them, rather than global coordinates. 

Check out examples/annotations.html to see how this works so far. Mostly I'd like feedback on the direction this is going and general API shape.


src/features/annotation.ts
```diff
@@ -0,0 +1,137 @@
...
+                  ThreeMath.smoothstep(
+                      view.angleTo(object.normal),
+                      this[$hiddenAngle] - TRANSITION_ANGLE,
+                      this[$hiddenAngle]);
+          object.element.style.opacity = `${opacity}`;
+        }
+      }
+    }
+
+    addHotspot(position: Vector3, normal: Vector3) {
```
cdata: Why not let the user provide the DOM node?

elalish: Sure, figured that would be the next thing. Should we provide a nice default one, or always make them do it themselves?

cdata: One idea would be to create a `<slot>` dynamically for each added hotspot. The `<slot>` could have an opinionated default element. If the user wanted to bring their own node for the hotspot, then they would create a child of `<model-viewer>` that targets that slot. The hierarchy would look something like this:

```html
<model-viewer>
  #shadow-root
    <slot name="hotspot-1">
      <div class="default-hotspot-style"></div>
    </slot>
  <div slot="hotspot-1"><!-- anything or nothing goes here--></div>
</model-viewer>
```

elalish: Cool, I was envisioning something along those lines. A common use case is probably having just one custom hotspot style; Can they reuse their element? Or should they be able to change our default slot CSS?

lokiiarora: Wouldn't it be great to have all hotspots to be part of one slot and then we could observe mutations or additions on top it, rather than making slots for each one of them? And also have a declarative api to define the hotspots .

cdata: @lokiiarora 

> And also have a declarative api to define the hotspots .

Definitely, we're just using the imperative API as a starting point.

> Wouldn't it be great to have all hotspots to be part of **one** slot

(emphasis mine) this is also possible, but there would be a trade-off. If we use 1 slot per distributed node, then all transforms of the nodes can be opaque to the user of `<model-viewer>`. OTOH if all nodes are distributed to a single slot, we would have to transform the nodes in the light DOM, which leaks style details and risks clobbering user styles.

cdata: > Can they reuse their element?

This is an interesting question. I think the best we could do (without delving into hacks) is let them specify a `<template>` element that contains DOM that we could "stamp" for each hotspot. There is no easy way to "instance" a DOM node in its raw form without copying it or rendering it somehow (and rendering is a huge hack with caveats).

cdata: > Or should they be able to change our default slot CSS

Yeah, we don't need to decide this right away, but it would be easy enough to expose the styles with CSS custom properties.

lokiiarora: @cdata 
oh sure..
what's the advantage of putting the hotspot as a dom node rather than a 3D object?

cdata: > what's the advantage of putting the hotspot as a dom node rather than a 3D object?

There are advantages and disadvantages.

The advantages are that a DOM node naturally has a declarative style API via CSS, participates in DOM event dispatch (`myHotspot.addEventListener('click', () => { /* custom logic */ });`) and can be replaced with any node type including custom elements.

The disadvantages include worse performance qualities and low-fidelity composition with the 3D scene. We don't get occlusion, we have to fake it. It might be hard to make this work right for complex scenes.

We aren't married to the current approach, but it's the one we want to explore first.

lokiiarora: Nice thought..
would love to contribute on this feature ! @cdata :)

elalish: @cdata I've been playing with this a bit and I think I'll need a bit more specifics. The way these hotspots work is they are not children of a slot, but are referenced from the scene graph and CSS2DRenderer takes care of moving them around as needed (both in terms of position and hierarchy) to get them rendered. So I think they are not supposed to be in a slot? Only CSS2DRenderer is in a slot as of now.

cdata: Here is the pseudo-code I'm imagining:

```javascript
class ModelViewerWithHotspots {
  addHotspot(name: string, position: Vector3, normal: Vector3) {
    const wrapper = document.createElement('div');
    wrapper.classlist.add('hotspot-slot');

    const slot = document.createElement('slot');
    slot.setAttribute('name', `hotspot-${name}`);

    const defaultHotspot = document.createElement('button');
    defaultHotspot.classlist.add('hotspot');

    wrapper.appendChild(slot);
    slot.appendChild(defaultHotspot);

    const hotspot = new Hotspot(wrapper, position, normal);
    this[$scene].pivot.add(hotspot);
  }
}
```

The hierarchy this creates looks something like this:

```html
<div class="hotspots-container">
  <div class="hotspot-slot">
    <slot name="hotspot-foo">
      <!-- The default if no other element is assigned to this slot -->
      <button class="hotspot"></button>
    </slot>
  </div>
  <div class="hotspot-slot">
    <slot name="hotspot-bar">
      <button class="hotspot"></button>
    </slot>
  </div>
</div>
```

Note that we have to use the "wrapper" div because in browsers without native ShadowDOM the `<slot>` elements will dissolve away (this detail is required to preserve the topology of the main document in such scenarios) so the wrappers are necessary evil.

So, Three.js' CSS renderer should be moving around the wrapper nodes. Since the `<button>` is a descendant of these nodes, they should be positioned correctly, and so will any custom alternatives assigned to the slot.

elalish:
@cdata After playing with this awhile, I began to feel that it was more important for the hotspot styles to work cleanly like the rest of HTML and CSS, and that this trumped the concern of exposing their positions to the light DOM. This lead me to the design here, which has a pretty clean API at least (see the annotation example). I still need to add a mutation observer to handle changes. Feedback welcome, especially about what kinds of edge cases this might break.

cdata:
TL;DR the current approach creates the potential for style contention between a content author and the component; we can avoid this and also be more coherent with built-in DOM/style behavior by using `<slot>`s.

The main problem with the current approach is that the DOM node is essentially transitioned from author control (it is part of the author-controlled "light" DOM) to component control (it is a detail of the components "shadow" DOM), but without any contract or guarantee for who will control the element moving forward. Shadow roots are opaque to authors who work in the host document. In handling a "light" DOM element directly in the component, we are piercing the shadow veil and creating a surface area for style contention.

Importantly, if the goal is to behave coherently with HTML/CSS, there should only be one canonical "owner" for a given node and its styles/attributes. Anything that the browser does to a DOM node that relates to layout (particularly) can and does allow author styles and configuration to layer on top, in an additive fashion or as a way for the user to override built-in behavior.

Consider the following contrived case:

```html
<style>
#hotspot {
  transform: rotate(15deg); // Doesn't seem to work 🤔
  transition: opacity 0.3s;
}

model-viewer > div {
  transition: opacity 0.3s;
}

model-viewer.loaded > div {
  opacity: 0;
}
</style>
<model-viewer id="modelViewer">
  <img id="hotspot" style="border-radius: 2px" src="pin.png" data-position="...">
  <div id="loadingIndicator"></div>
</model-viewer>
<script>
  modelViewer.addEventListener('load', () => {
    modelViewer.classList.add('loaded'); // Hotspots all fade away!?
    console.log(document.querySelectorAll('model-viewer > img').length); // 0 😳
  });

  window.onload = () => {
    
  };
</script>
```

There a few things about this that will probably violate this user's expectations:

 - The `transform` style that they provided is not applied to the hotspot
 - The inline `border-radius` style is overwritten
 - New hierarchy is created in the "light" DOM ("where did that come from?") and the hotspot itself is moved, so the query cannot find it
 - The newly introduced element is not encapsulated from author-specific styles, so it accidentally fades with the loading indicator

Why are these things true? Because when `<model-viewer>` connected to the document, it did a number of things that DOM elements don't typically do:

 - [It created and appended a wrapper element to itself in the host document](https://github.com/GoogleWebComponents/model-viewer/blob/a7a350cef464cde73e2d12090987941e0518f46e/packages/model-viewer/src/features/annotation.ts#L88), depriving the element (a `<model-viewer>` internal detail) of shadow root encapsulation from content author CSS
 - It directly assigned a [transform](https://github.com/mrdoob/three.js/blob/45418089bd5633e856384a8c0beefced87143334/examples/js/renderers/CSS2DRenderer.js#L86) and an [opacity](https://github.com/GoogleWebComponents/model-viewer/blob/a7a350cef464cde73e2d12090987941e0518f46e/packages/model-viewer/src/features/annotation.ts#L127) to the element, superseding and overwriting user style configuration respectively

Once all the DOM effects have played out, the live DOM will look something like this:

```html
<model-viewer id="modelViewer">
  <div id="loadingIndicator"></div>
  <div> <!-- CSS2DRenderer.domElement -->
    <img id="hotspot" style="position: absolute; transform: ..." src="pin.png" data-position="...">
  </div>
</model-viewer>
```

We can avoid all of these problems by using `<slot>`s (and wrappers) in the shadow root. With `<slot>`s, no elements need to be moved, no hierarchy needs to be changed, and no styles need to be overwritten. Positional styles provided by the content author will layer additively with positional styles assigned to a `<slot>` wrapper. If the author sets the `opacity` of the hotspot, it will be multiplied by the `opacity` set by `<model-viewer>` on the `<slot>` wrapper. The `` (in the example) doesn't need to move around the hierarchy (which is what the content author expects). And perhaps most importantly: if the `CSS2DRenderer.domElement` is in the shadow root, content author styles do not apply to it (so no danger of accidentally messing with it).

elalish:
@cdata Thank you for the detailed example! I'll address your first four bullets as though they're numbered:
1) True, though I believe you can accomplish relative transforms by making a child node under it. This would certainly require documentation.
2) I just tested your example and it actually works fine. What would overwrite it? In fact, this is the primary reason I like this approach better than slots; all my CSS styles apply just as they would to any other nodes I put on my page.
3) Valid con, though I think documentation would suffice.
4) Akin to 2), I actually view this as a pro, not a con. It behaved exactly as I told it to!

The trouble I have with slots is the convoluted nature of styling them, and potentially having state that needs to be synced between two places (the light and shadow DOMs). But I agree, it's not a clear slam dunk either way.

cdata:
I'll put together a live example and then we can discuss more.

cdata:
Please consider this live example demonstrating a handful of the ways that user expectations could be violated with this approach: http://lopsided-motion.glitch.me/hotspot-exposure.html

In the first example, I try to style the transform of the hotspot and it doesn't seem to work.
In the second example, I try something many who are familiar with CSS might try: add an `!important` expression to bump up the precedence of my style. This breaks in a totally new way.
In the third example, I break annotations all in one go by accidentally styling the wrapper.

Note that the third example is particularly important because users don't always have a say in what a stylesheet targets. They may be consuming a stylesheet from a third party that violates the expectations of `<model-viewer>` or Three.js. These examples illustrate the perils of a shared style scope, and explain why shadow roots exist: you need strong style encapsulation in order to guarantee the portability of a component.

You were right about `border-radius`, because the `style` attribute is not being replaced wholesale (I made a mistake in my example), so I used another style property to demonstrate the contention (`opacity`). As for asking users to just make more nodes to work around our implementation details, we don't *need* to ask that of them; what do we gain from doing so?

HTML is a sort of high-level, functional programming language. Each tag is a function invocation, and attributes and the child list are the arguments. When an element manipulates its own child list, it is analogous to re-invoking a function with different arguments. It turns out that almost no DOM elements spawn new hierarchy and/or re-parent nodes. One notable exception to this is `<table>`, but it is the exception that proves the rule: it is one of the oldest HTML elements, and we do not create elements in its likeness today because what it does is unintuitive and confusing.

Assuming the current implementation, let's say we want to support removing a hotspot node in order to hide a hotspot. We cannot simply observe for nodes being removed from the host (where the user told them to go, and where the user will intuitively expect them to be); after all, the nodes are re-parented to a whole new element. If the user reads our docs and understands this, then they will know to remove the hotspot node from the Three.js-specific wrapper. But, if they want to dynamically add a new hotspot, should they add it to the Three.js-specific wrapper? Or to `<model-viewer>`? And for implementation purposes, `<model-viewer>` will need not one but two mutation observers: one for the host `<model-viewer>` (to observe when nodes are added) and one for the Three.js `<div>` wrapper (to observe when nodes are removed).

On the flip side, let's say that hotspots stay exactly where they are when they are created (and that we are using `<slot>`s internally). The user can inspect the live DOM and see that the node is where they said it should be when they authored their HTML. No docs are necessary to explain this to them, because by-and-large this is how HTML/DOM works. And `<model-viewer>` implementation is simpler because there is only one node that we care about observing: `<model-viewer>`. When hotspot nodes are added to `<model-viewer>`'s child list, we create hotspots for them. When hotspot nodes are removed from `<model-viewer>`'s child list, we cleanup/remove corresponding hotspot state.

The most important thing, ultimately, is that if we used slots, we would not need to ask our users to read our docs in order to understand how the hierarchy of the DOM changes when they configure `<model-viewer>`. Instead, those details will be opaque to the user, and `<model-viewer>` will do exactly what they asked for (which, ultimately, is best represented by the HTML they have written themselves). Styles will merge in reasonable ways. There will be no danger of the user manipulating `<model-viewer>`-internal details by accident.

elalish:
@cdata Updated to use slots, but I must be missing something stupid because no hotspots are showing up. Everything seems to be working properly in dev tools; I think the CSS2DRenderer element is just hidden, but I can't figure out why.

elalish:
@cdata Nevermind, fixed!

elalish:
@cdata I've tested enough that I think the mutationObserver is now doing all the things we want it to. I need to add some automated tests for that. The only weird thing I see is that the hotspots will appear briefly at the top corner before the progress bar appears, which is unsightly. Do you have any idea how to fix that?

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,208 @@
...
+    })
+    @property({type: String, attribute: 'hidden-angle'})
+    hiddenAngle: string = 'auto';
+
+    private[$annotationRenderer] = new CSS2DRenderer();
+    private[$hiddenAngle] = DEFAULT_HIDDEN_ANGLE;
+    private[$nextHotspotIndex] = 0;
+    private[$hotspotMap] = new Map();
+    private[$observer] =
+        new MutationObserver((mutations) => {mutations.forEach((mutation) => {
```
cdata: This formatting seems pathologically bad :(

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,208 @@
...
+    const normalNodes = parseExpressions(element.dataset.normal!)[0].terms;
+    for (let i = 0; i < 3; ++i) {
+      this.normal.setComponent(
+          i, normalizeUnit(normalNodes[i] as NumberNode<'m'>).number);
+    }
+  }
+}
+
+export declare interface AnnotationInterface {
+  hiddenOpacity: number;
```
cdata: This seems like a good use case for a CSS custom property. Consider this [example](https://glitch.com/edit/#!/lopsided-motion?path=slotted-style-test.html) ([live version](http://lopsided-motion.glitch.me/slotted-style-test.html)).

With this approach, you would add or remove a class from the `<slot>`'s wrapper element depending on whether or not that hotspot is "hidden".

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,208 @@
...
+      style.position = 'absolute';
+      style.top = '0';
+      this.shadowRoot!.querySelector('.container')!.appendChild(domElement);
+
+      for (let i = 0; i < this.children.length; ++i) {
+        this.addHotspot(this.children[i]);
+      }
+
+      this[$observer].observe(
+          this, {childList: true, attributes: true, subtree: true});
```
cdata: `subtree: true` is both slow and probably not what you want; it makes the observer recursive, so it will notify you of all such changes (child list and attribute changes) to both the observed node and each of its descendants.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,208 @@
...
+    connectedCallback() {
+      super.connectedCallback();
+      const {domElement} = this[$annotationRenderer];
+      const {style} = domElement;
+      style.pointerEvents = 'none';
+      style.position = 'absolute';
+      style.top = '0';
+      this.shadowRoot!.querySelector('.container')!.appendChild(domElement);
+
+      for (let i = 0; i < this.children.length; ++i) {
```
cdata: This approach will attempt to add every child, including nodes that are possibly not intended to be hotspots, as hotspots. For example, consider this configuration:

```html
<model-viewer>
  <img slot="poster" src="foo.png" data-position="..." data-normal="...">
</model-viewer>
```

Now, it is fair to say that the user annotating this node as though it were a hotspot and then slotting it as the poster is a bit fishy. However, the point is that we are _inferring_ that the user meant to create a hotspot when, in fact, the user's intention is ambiguous. `data-*` attributes are, after all, intended for authors to decorate their nodes in any way for any reason.

For a declarative API, I would recommend that we have an explicit signal. For this change, we could punt on that and focus on the imperative API. OTOH if we do it now I would probably suggest a more explicit signal. For example, the author could pre-populate the slot attribute:

```html
<model-viewer>
  <div id="some-hotspot" slot="hotspot-foo" data-position="..." data-normal="..."></div>
</model-viewer>
```

Then, we would create hotspots out of all nodes that have a `slot="hotspot-*"` attribute.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,208 @@
...
+    for (let i = 0; i < 3; ++i) {
+      this.normal.setComponent(
+          i, normalizeUnit(normalNodes[i] as NumberNode<'m'>).number);
+    }
+  }
+}
+
+export declare interface AnnotationInterface {
+  hiddenOpacity: number;
+  hiddenAngle: string;
```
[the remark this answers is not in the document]
elalish: it's the angle between the viewer and the hotspot normal at which it begins to be transparent. Honestly we could just remove it and use the default of 90 degrees.

packages/model-viewer/src/features/annotation.ts
elalish: Yeah, the reason I put this in is because I wanted to be notified when the immediate children's attributes changed, but all I was getting was the containers' attributes.

cdata: Yah, for the first pass let's skip observing changes to the attributes. There isn't really a good way to do it without either A) requiring hotspots to be custom elements or B) observing each node individually with a second mutation observer.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,177 @@
...
+const $observer = Symbol('observer');
+const $addHotspot = Symbol('addHotspot');
+const $removeHotspot = Symbol('removeHotspot');
+
+interface HotspotConfiguration {
+  name: string;
+  position?: string;
+  normal?: string;
+}
+class Hotspot extends CSS2DObject {
```
cdata: Please add some documentation for this class

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,177 @@
...
+import {Constructor} from '../utilities.js';
+
+const $annotationRenderer = Symbol('annotationRenderer');
+const $updateHotspots = Symbol('updateHotspots');
+const $hotspotMap = Symbol('hotspotMap');
+const $observer = Symbol('observer');
+const $addHotspot = Symbol('addHotspot');
+const $removeHotspot = Symbol('removeHotspot');
+
+interface HotspotConfiguration {
```
cdata: Please add some documentation for this interface

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,177 @@
...
+    const slot = document.createElement('slot');
+    slot.name = config.name;
+    wrapper.appendChild(slot);
+    super(wrapper);
+    this.normal = new Vector3(0, 1, 0);
+    this.updatePosition(config.position);
+    this.updateNormal(config.normal);
+  }
+
+  updatePosition(position?: string) {
```
cdata: Please add documentation for all public API in this class

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,177 @@
...
+          i, normalizeUnit(normalNodes[i] as NumberNode<'m'>).number);
+    }
+  }
+}
+
+export declare interface AnnotationInterface {
+  updateHotspot(config: HotspotConfiguration): void;
+}
+
+export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
```
cdata: Please document this mixin

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,177 @@
...
+
+      this[$observer].observe(this, {childList: true});
+    }
+
+    disconnectedCallback() {
+      super.disconnectedCallback();
+      this[$observer].disconnect();
+    }
+
+    updated(changedProperties: Map<string, any>) {
```
cdata: Looks like we can remove this

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,177 @@
...
+            object.element.classList.add('hide');
+          } else {
+            object.element.classList.remove('hide');
+          }
+        }
+      }
+    }
+
+    [$addHotspot](node: Node) {
+      if (!(node instanceof HTMLElement && node.slot.includes('hotspot')))
```
cdata: For the purposes of IE11 support, we should avoid `includes`. Consider `indexOf('hotspot') === 0` as an alternative.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,177 @@
...
+    [$removeHotspot](node: Node) {
+      const hotspot = this[$hotspotMap].get(node);
+      if (!hotspot)
+        return;
+      this[$scene].pivot.remove(hotspot);
+      this[$hotspotMap].delete(node);
+    }
+
+    updateHotspot(config: HotspotConfiguration) {
+      const element = this.querySelector(`[slot="${config.name}"]`);
```
cdata: Let's just keep a map of hotspots by name and validate with that. Names should be unique identifiers for hotspots anyway, so this map will be useful elsewhere.

In fact, you might not need to keep track of slotted elements at all.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+  }
+}
+
+export declare interface AnnotationInterface {
+  updateHotspot(config: HotspotConfiguration): void;
+}
+
+export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
+    ModelViewerElement: T): Constructor<AnnotationInterface>&T => {
```
cdata: It would be nice to see a token test suite added for this new mixin.

packages/model-viewer/src/features/annotation.ts
elalish: Done.

packages/model-viewer/src/features/annotation.ts
elalish: Done.

packages/model-viewer/src/features/annotation.ts
elalish: Done.

packages/model-viewer/src/features/annotation.ts
elalish: Added.

packages/model-viewer/src/features/annotation.ts
elalish: Fixed.

elalish:
@cdata Can you make heads or tails of why our Sauce labs tests are suddenly failing? The output is a bit baffling.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * The Hotspot object is a reference-counted slot. If decrement() returns true,
+ * it should be removed from the tree so it can be garbage-collected.
+ */
+export class Hotspot extends CSS2DObject {
+  public normal: Vector3;
+  private referenceCount: number;
+
+  constructor(config: HotspotConfiguration) {
+    const wrapper = document.createElement('div');
```
cdata: Let's set `pointer-events: all` on the wrapper element, so that users don't have to opt-in to pointer-events in their hotspots.

packages/model-viewer/src/features/annotation.ts
cdata: Docs are for the person reading the code; I usually imagine a hapless contributor trying to make a change and hoping that hovering over a method call will tell them something about what that method does.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,245 @@
...
+    };
+    private[$observer] = new MutationObserver(this[$mutationCallback]);
+
+    connectedCallback() {
+      super.connectedCallback();
+      const {domElement} = this[$annotationRenderer];
+      const {style} = domElement;
+      style.pointerEvents = 'none';
+      style.position = 'absolute';
+      style.top = '0';
```
cdata: Let's just add a class to this element (like `domElement.classList.add('annotation-container')`) and put the styles for it in `template.ts`. We can do something similar for each slot wrapper.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -0,0 +1,264 @@
...
+    private[$observer] = new MutationObserver(this[$mutationCallback]);
+
+    constructor(...args: Array<any>) {
+      super(...args);
+
+      const {domElement} = this[$annotationRenderer];
+      domElement.classList.add('annotation-container');
+      this.shadowRoot!.querySelector('.container')!.appendChild(domElement);
+
+      for (let i = 0; i < this.children.length; ++i) {
```
cdata: I think this loop needs to be done in the `connectedCallback`, otherwise there are conditions where the timing would cause some children not to be detected. Consider this:

```typescript
const mv = new ModelViewerElement(); // Loop runs to look for children, none found

for (let i = 0; i < numHotspots; ++i) {
  mv.appendChild(hotspotNode); // MutationObserver is not active yet
}

document.body.appendChild(mv); // Element is "connected"; MutationObserver is activated now
```

theprojectsomething:
Congrats on the merge - have been waiting for this one. Currently testing with some interesting models and no major issues. Some very minor notes / considerations:
- the modelviewer.dev site seems to be in sync with the master branch, so your annotations example page is live, however the latest release is v0.8.1 (and being served by npm/unpkg)
- it wasn't immediately clear how to utilise `data-normal` (or even what is was initially) .. assume this is adapted from sketchfab as per original issue but a line in the examples page would help some (a 'helper' plane would help a lot)
- this seems like a good opportunity to include a debug / helper feature to show e.g. bounding boxes, camera coords, annotation visibility planes, etc. Assume a feature request along these lines has been opened already but will follow up if not

elalish:
@theprojectsomething Thank you! The first is a known issue; we have a PR in the works to heavily upgrade our docs. I'll also make a note to better document the normal attribute. As for the third, I don't believe we have a feature request along those lines, so please write one!

theprojectsomething:
That's great @elalish - re. no3 I've made some notes in #218 (seemingly opened by @cdata as an aside back in the early days). Happy to create a new feature request if needed. Cheers!

theprojectsomething:
Sorry, some more hotspot questions / suggestions .. I've found myself listening to 'camera-change' events in order to notify individual hotspots as to their visibility (i.e. to do more than just toggle opacity via css variables). Is it possible to toggle a non-intrusive class / data attribute on the slotted node, in addition to the `.hide` class you toggle on its container in the component?

Also in looking into this I noticed that the `$updateHotspots` logic is running on `tick` and not taking into account the previous state of the hotspot element. Taking this and the above into account, can I suggest the following?

```js
// from annotation.ts line 208
[$updateHotspots]() {
  // get the camera position outside of the loop
  const view = this[$scene].activeCamera.position.clone();

  // check current position against previous position, continue if camera has moved
  // note - this.view is defined after first run
  if (this.view && view.distanceTo(this.view) === 0) return;

  // clone the current position before it's subtracted (below)
  this.view = view.clone();
  for (let i = 0, l = children.length; i < l; i++) {
    const object = children[i];
    if (object instanceof Hotspot) {
      view.sub(object.position);

      // check if the element is currently hidden / occluded
      const hidden = object.element.classList.contains('hide');

      // if element should be hidden / occluded (and isn't already)
      if (view.dot(object.normal) < 0) {
        if (!hidden) {
          object.element.classList.add('hide');

          // also update [data-occluded="1"] on the slotted nodes
          // data attribute is suggested over a class as per normal and position attrs
          object.element.children[0].assignedNodes()
          .forEach($el => ($el.dataset.occluded = 1));
        }
      }
      
      // if element shouldn't be hidden / occluded (but is)
      else if (hidden) {
        object.element.classList.remove('hide');
        
        // also update [data-occluded="0"] on the slotted nodes
        object.element.children[0].assignedNodes()
        .forEach($el => ($el.dataset.occluded = 0));
      }
    }
  }
}
```

Issue #894, opened 2019-11-21
Disable resize on AR


aviggiano:
### Description

This is a feature request to provide an option to disable content resize on Augmented Reality, such as with [`allowsContentScaling`](https://developer.apple.com/documentation/arkit/arquicklookpreviewitem/3152993-allowscontentscaling?language=objc) on AR Quick Look.

cdata:
@aviggiano thanks for the feature request! We'll see what we can do :)

PR #918, opened 2019-12-09
Adding ar-scale to control scaling behavior in Scene Viewer.

eugenecys:
Fixes: #894

### DIFF
--- a/examples/augmented-reality.html
+++ b/examples/augmented-reality.html
@@ -59,7 +59,7 @@
         </div>
         <example-snippet stamp-to="demo-container-1" highlight-as="html">
           <template>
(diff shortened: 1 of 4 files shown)


examples/augmented-reality.html
```diff
@@ -71,6 +71,14 @@ <h3 style="margin-top: 40px; padding-bottom: 8px;" class="border-bottom">iOS, Ma
...
             on supported devices.</p>
           </li>
+          <li>
+            <div>ar-scale</div>
+            <p>Controls the scaling behavior in AR mode in
+            <a href="https://developers.google.com/ar/develop/java/scene-viewer">Scene Viewer</a>.
+            Set to "fixed" to disable scaling of the model, which sets it to
+            always be at 100% scale. Defaults to "auto" which allows the model to
+            be resized.</p>
+          </li>
```
elalish: Do we need this here since it is also in index.html?

cdata: Yeah, I think it's safe to omit the docs here in favor of the docs in `index.html` for now.

PR #919, opened 2019-12-09
Access the loaded model's scene graph

### LINKED ISSUE #738: Allow manipulation of model scene graph via "Render Worklet"

cdata:
This change proposes an MVP API for accessing the scene graph of a `<model-viewer>` element. The API is designed to be renderer-agnostic, with a taxonomy modeled after [glTF 2.0](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0).

For the purposes of design discussion, we are calling the API 3D Object Model, or **3DOM**.

The change is still a work in progress. While it is being finished, here are some highlights of the design:

 - The core API for scene graph primitives is defined in [a stand-alone declaration file](https://github.com/GoogleWebComponents/model-viewer/blob/scene-graph-access/src/3dom/api.ts)
 - The current implementation runs scene graph scripts in an isolated worker called a `scene-graph-worklet`
 - Only one concrete mutation is allowed at this time: material PBR base color factor
   - This is for the sake of demonstrating the principle; in time, most of the graph will be mutable
 - All mutations to the scene graph are asynchronous
   - This will incur notable limitations on what effects/experiences can be achieved, but is necessary for now in order to enforce script isolation
 - APIs available to `scene-graph-worklet` scripts are carefully controlled a la [WorkerDOM](https://github.com/ampproject/worker-dom) in order to maximize portability
 - A "capability" API is planned that will work as a form of [Feature Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy) for `scene-graph-worklet` scripts
 - Most of the implementation here could be feasibly extracted and distributed as its own library for use in general Three.js projects
 - Currently ESM scripts are not supported ([they don't work in workers right now anyway](https://chromestatus.com/features/5761300827209728)) and `importScripts` is restricted behind an unimplemented feature policy

Example usage:

```html
<model-viewer src="any.gltf">
  <script type="scene-graph-worklet">
self.addEventListener('model-change', () => {
  // Change base color factor to red:
  self.model.materials[0].pbrMetallicRoughness.setBaseColorFactor([1, 0, 0, 1]);
});
  </script>
</model-viewer>
```

Fixes #738


elalish:
@cdata Can this be closed in favor of #992?

cdata:
Yes, closing this in favor of #992

PR #933, opened 2019-12-17
Update Model-Viewer to use OffscreenCanvas for rendering

yiyix:
Update Model-Viewer to use OffscreenCanvas for rendering, which will help the performance in ModelViewer.


cdata:
Looks great @yiyix thanks for the contribution!

elalish:
@yiyix I found a serious issue, which is that this breaks our best WebGL debugging tool: spector.js. They apparently want to support OffscreenCanvas, but [haven't managed to yet](https://github.com/BabylonJS/Spector.js/issues/90). If you want more adoption of OffscreenCanvas, you might consider helping them solve this.

Issue #381, opened 2019-02-20
Bad render results for Alpha Blend Mode test model


cdata:
As of #360 , we will be including the [Khronos Alpha Blend Mode Test model](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AlphaBlendModeTest) in our fidelity tests. Unfortunately, `<model-viewer>` seems to produce incorrect render results in four out of five cases tested by the model:

`<model-viewer>`|Filament `gltf_renderer`|Khronos' example
-------------------|------------------------|------------------
![image](https://user-images.githubusercontent.com/240083/53075029-1482b480-34a1-11e9-8b12-de60d65b2df5.png)|![image](https://user-images.githubusercontent.com/240083/53075049-1f3d4980-34a1-11e9-9ba1-cd6a4c47de63.png)|![image](https://user-images.githubusercontent.com/240083/53075082-31b78300-34a1-11e9-9e08-9503b8cd8eb0.png)

donmccurdy:
I'm confused by this result – it shows up correctly in my threejs-based viewer: https://gltf-viewer.donmccurdy.com/#model=https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/AlphaBlendModeTest/glTF/AlphaBlendModeTest.gltf

<img width="1017" alt="screen shot 2019-02-20 at 7 23 57 am" src="https://user-images.githubusercontent.com/1848368/53102482-98a85c80-34e0-11e9-8df7-21dafe8aa4a3.png">

donmccurdy:
Result also looks more correct in the https://truthful-roof.glitch.me demo –

<img width="1218" alt="screen shot 2019-03-06 at 12 51 58 pm" src="https://user-images.githubusercontent.com/1848368/53912882-c9bf7b80-400e-11e9-9a5d-4245d36adb84.png">

I'm wondering if the `scene.clone()` call used inside `<model-viewer/>` has lost some information...

cdata:
If it's useful, the live test using `<model-viewer>` is here: https://googlewebcomponents.github.io/model-viewer/test/fidelity/khronos-AlphaBlendModeTest/

It does look like the automation screenshot results in artifacts that are not present in the live render. Screenshot of live render as of now looks like https://truthful-roof.glitch.me:

![image](https://user-images.githubusercontent.com/240083/53914378-57e93100-4012-11e9-8bcb-6bb8b64f386c.png)

[...]

PR #949, opened 2020-01-03
Add second instance of Renderer to manage transparent background.

### LINKED ISSUE #874: Make it possible to configure the background to be transparent
Sorry if this is not a relevant question but I couldn't find anything regarding this topic on the documentation/Internet. Is it possible to set the background-color attribute of <model-viewer> to transparent/ none? It appears to default to a white background if the CSS string is not parsed correctly.

nicolas-daures:
### Reference Issue
Fixes #874

Management of the "transparent" value for the background-color property.

To allow the property value to change (transparent -> opaque, opaque -> transparent), we create two instances of the Renderer class. The first to manage a background color, the second to manage transparency.


cdata:
Thanks for the change suggestion @nicolas-daures !

Would it be possible to structure this so that the transparent renderer is created on-demand? This way folks only pay a cost in the case that they actually want to use transparency.

Also for the sake of cross-referencing things, we have had issues with transparent rendering in the past (see #381 ).

src/three-components/Renderer.ts
```diff
@@ -247,12 +260,15 @@ export class Renderer extends EventDispatcher {
...
       this.threeRenderer.setRenderTarget(null);
-      this.threeRenderer.clearDepth();
+      this.threeRenderer.clear(this.isTransparent, true, false);
       this.threeRenderer.setViewport(0, 0, width, height);
       this.threeRenderer.render(scene, camera);
 
       const widthDPR = width * dpr;
       const heightDPR = height * dpr;
+      if (this.isTransparent) {
+        context.clearRect(0, 0, widthDPR, heightDPR);
```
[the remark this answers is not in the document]
nicolas-daures: If I don't clear the context, the transparent pixels of the image to draw in the canvas will not overwrite the pixels of the previous image.

If I don't clear the color buffer of the three renderer, the transparent pixels of the new rendering will not overwrite those of the previous rendering.

rendering_without_clear

nicolas-daures:
@cdata The renderer is now created on-demand.

About transparency issues, I have the same rendering as on this comment: https://github.com/GoogleWebComponents/model-viewer/issues/381#issuecomment-470281211

@elalish I added an example in customize-ui with a linear gradient background to see the transparency.

packages/model-viewer/examples/customizing-ui.html
```diff
@@ -74,6 +74,25 @@ <h4></h4>
...
+    <div id="demo-container-2" class="demo"></div>
+    <div class="content">
+      <div class="wrapper">
+        <div class="heading">
+          <h2 class="demo-title">Transparent Background</h2>
+          <h4></h4>
+        </div>
+        <example-snippet stamp-to="demo-container-2" highlight-as="html">
+          <template>
+            <model-viewer ar camera-controls ios-src="assets/Astronaut.usdz" src="assets/Astronaut.glb" background-color="transparent" alt="A 3D model of an astronaut">
```
elalish: nit: Everything inside the `<template>` should be indented starting over from the left. If you look at this page in your browser you'll see why. Also, can we add an image or some text or something behind this so that it's clear the transparency is working? These dev pages serve as manual regression tests for this project.

elalish: Ah, I just saw the linear gradient; the trouble is that it's not inside the `<template>`, so it's not a visible part of the demo code.

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -55,22 +55,39 @@ const $singletonWithTransparency = Symbol('singletonWithTransparency');
...
  */
 export class Renderer extends EventDispatcher {
-  static[$singleton] = new Renderer({debug: isDebugMode()});
-  static[$singletonWithTransparency] = new Renderer({debug: isDebugMode(), isTransparent: true});
+  protected static[$singleton]: Renderer|null = null;
+  protected static[$singletonWithTransparency]: Renderer|null = null;
+
+  static get singleton(): Renderer {
+    let singleton = this[$singleton];
+    if (!singleton) {
```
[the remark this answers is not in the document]
nicolas-daures: I modified the code to adapt to the project's style. I deleted the temporary variable. It just used to avoid the cast of Renderer.

packages/model-viewer/examples/customizing-ui.html
nicolas-daures: I modified the example to integrate the gradient and a text in the template. Thanks for all the feedback.

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -53,15 +55,33 @@ const $singleton = Symbol('singleton');
...
+    }
+    return this[$singletonWithTransparency] as Renderer;
   }
 
   static resetSingleton() {
-    this[$singleton].dispose();
+    if (this[$singleton] != null) {
+      (this[$singleton] as Renderer).dispose();
+    }
     this[$singleton] = new Renderer({debug: isDebugMode()});
```
valdrinkoshi: Should this be created only if the previous value was defined (aka inside the if statement)? Same for the singleton with transparency below.

nicolas-daures: Good remark, thank you.

elalish:
@nicolas-daures I pulled your branch, but I don't seem to be getting any transparency on Chrome on my Macbook Pro:
Screen Shot 2020-01-16 at 10 00 49 AM

cdata:
@nicolas-daures thanks so much for your diligent work on this change. I want to give you a friendly heads up that we are trying to start the year off on the right foot by landing a major source restructuring change (see #962 ). Your changes are still good, but I will likely ask you to reconcile before this lands and everything will have been moved two directories deep.

Anyway, I am happy to help you reconcile the changes when that time comes. Just wanted to let you know what is going on.

elalish:
@nicolas-daures This seems to work, except when switching between the renderers. You can repro by scrolling up and down on the customize-ui page. It seems to switch back and forth between rendering transparently and non each time it comes into view. Can you repro?

nicolas-daures:
@cdata I merged master in my pull request to get the packages and lerna. It works. I would update the pull request with your changes regularly.

@elalish I can reproduce the bug and I fixed it. Thanks for the feedback.

elalish:
@nicolas-daures Thanks! That big change @cdata was talking about just landed, so once you merge or rebase I'll take another look. Note the build instructions have changed too, so check out the updated READMEs.

nicolas-daures:
@elalish It's merged.

packages/model-viewer/src/features/environment.ts
```diff
@@ -126,6 +126,27 @@ export const EnvironmentMixin = <T extends Constructor<ModelViewerElementBase>>(
         this[$cancelEnvironmentUpdate] = null;
       }
 
+      // Update background color before to get texture utils because renderer singleton depends on background transparency
+      if (skyboxImage == null)
+      {
```
cdata: `clang-format` needs to be run on this change

packages/model-viewer/src/features/environment.ts
```diff
@@ -126,6 +126,27 @@ export const EnvironmentMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+        if (backgroundColor === 'transparent') {
+          if (this[$scene].background != null) {
+            this.disconnectedCallback();
+            delete this[$scene].background;
```
cdata: If this isn't significantly different from assigning `null`, prefer to assign `null` instead

packages/model-viewer/src/features/environment.ts
```diff
@@ -126,6 +126,27 @@ export const EnvironmentMixin = <T extends Constructor<ModelViewerElementBase>>(
         this[$cancelEnvironmentUpdate] = null;
       }
 
+      // Update background color before to get texture utils because renderer singleton depends on background transparency
+      if (skyboxImage == null)
+      {
+        if (backgroundColor === 'transparent') {
+          if (this[$scene].background != null) {
+            this.disconnectedCallback();
```
cdata: These (`connectedCallback`/`disconnectedCallback`) are custom element lifecycle methods, and are meant to be invoked by the browser based on the semantics defined in the custom element spec. If we are calling them manually, then it's a sign that we may need to factor out part of their implementation to discrete methods.

Why are we invoking these methods manually?

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -53,15 +55,33 @@ const $singleton = Symbol('singleton');
  * the texture.
  */
 export class Renderer extends EventDispatcher {
-  static[$singleton] = new Renderer({debug: isDebugMode()});
+  protected static[$singleton]: Renderer|null = null;
+  protected static[$singletonWithTransparency]: Renderer|null = null;
 
-  static get singleton() {
-    return this[$singleton];
+  static get singleton(): Renderer {
```
cdata: It seems weird to call this a singleton now that we are creating up to two of them. Maybe we should just call them `opaqueRenderer` and `transparentRenderer`?

packages/model-viewer/src/features/environment.ts
```diff
@@ -126,6 +126,27 @@ export const EnvironmentMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+            delete this[$scene].background;
+            this.connectedCallback();  
+          }
+        }
+        else {
+          const mustUpdateRenderer = this[$scene].background == null;
+          mustUpdateRenderer && this.disconnectedCallback();
```
cdata: If these lifecycle methods are being invoked in order to get the renderer to update, then we could use `this[$needsRender]()` to provoke a re-render and skip calling the lifecycle methods.

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -128,7 +128,7 @@ export default class ModelViewerElementBase extends UpdatingElement {
   }
 
   get[$renderer]() {
-    return Renderer.singleton;
+    return this[$scene].background ? Renderer.singleton : Renderer.singletonWithTransparency;
```
cdata: It seems like we should only use the transparent renderer if the background color is, in fact, `"transparent"`. One way to accomplish this would be to specialize the `[$renderer]` accessor for the `EnvironmentMixin` implementation, so that it overrides the default ("always use opaque renderer") with a version that returns a renderer based on the value of the `backgroundColor` property.

koch-crsd:
hi guys, great work! :)

I have a small question: I've noticed that if the `background-color` attribute is set as `"transparent"` it works. But if for example additionally also the `shadow-intensity` attribute is set, then the transparency isn't applied and it gets the default white background.?

it would be really awesome if the transparency would still work even when having the `shadow-intensity` attribute set...

packages/model-viewer/src/features/environment.ts
nicolas-daures: ok, that's fixed.

packages/model-viewer/src/three-components/Renderer.ts
nicolas-daures: ok, that's renamed.

packages/model-viewer/src/features/environment.ts
nicolas-daures: These callbacks are called to unregister the scene from the renderer and to unsubscribe from the contextlost event. I change the code to call only what is necessary.

packages/model-viewer/src/features/environment.ts
nicolas-daures: The objective is not really to update the rendering but to unsubscribe listeners and unregister the scene from the old renderer before switching to the new renderer. So the accessor $renderer must return the old renderer to unsubscribe the listener, then the same accessor must return the new renderer to subscribe the listener. This is why there is the test to null on the background performed in the accessor.

elalish:
@nicolas-daures @valdrinkoshi @koch-crsd I have an alternative PR that I'd love feedback on: #975. It draws heavily from this one, but also simplifies our API.

elalish:
Closing in favor of #975. Thank you @nicolas-daures for pushing us on this and contributing to the landed change!

PR #954, opened 2020-01-07
Monorepo

### LINKED ISSUE #927: Restructure repo as multi-project mono-repo
As `<model-viewer>` has developed, there have been many `<model-viewer>`-adjacent projects added to the source tree. Currently the project source tree includes:

 - The `<model-view>` source
 - The "raw" source of our docs page
 - The source for our fidelity testing infrastructure

And the list is growing, with 3DOM set to be introduced as well via #919 .

This issue proposes that we re-organize our code as a mono-repo managed by https://lerna.js.org/. Each of the named projects would be split out and managed as its own package, but all of them would live in the same Git repository. This will allow us to manage compatibility across all of these projects as we do today (by keeping them versioned in the same git repo) while also giving us the flexibility to publish and distribute them separately.

cdata:
This change is the minimum viable introduction of lerna.js into our project hierarchy. It moves the core `<model-viewer>` source tree into a subfolder (`packages/model-viewer`) and introduces lerna.js at the top-level of the repository for managing multiple sub-packages.

I originally planned to split out docs and fidelity testing as part of this change, but the change is already quite significant without that, so in the interest of keeping scope down I would like to submit this change first and split out projects as a follow-up.

Fixes #927


cdata:
Yah, let's wait. It probably is broken in some ways. Please do play with it. In theory, everything should just work like it did before (but of course, you will have to cd into the subdirectory).

cdata:
Just FYI when I land this, there is a good chance that some CI-related stuff will break. I'll try to keep an eye on it and fix as needed.

PR #961, opened 2020-01-13
Fixing Transparency

### LINKED ISSUE #960: Transparent material rendering
We render transparent glTF materials, but not terribly well. Here is an example of two nested transparent materials. Front:
image
Back:
image
Two problems:
1) Some surfaces are not getting rendered at all due to depth testing: note the vanishing notch on the second image.
2) Triangle seams are visible due to triangles being rendered in different orders.
These are problems many renderers have, as sorting triangles is not real-time feasible, but I think we can at least do better than this. Also note that DRACO encoding tends to make these problems more visible, as it orders the triangles in a fairly arbitrary way.

elalish:
Fixes #960 

This is two fixes, the first is based on Don's open [PR](https://github.com/mrdoob/three.js/pull/18235) to three.js. However, the majority of the artifacts were fixed by the larger change which I found hidden in a three.js [example](https://threejs.org/examples/#webgl_materials_physical_transparency). By rendering the back-facing side before the font-facing side, it properly orders all triangles on convex objects, and so vastly reduces the visibility of artifacts. This is especially important with DRACO compression, as this makes the rendering order of triangles non-spatially consistent. 

I also had to bump our version of puppeteer, as our fidelity tests were getting flaky. 

Note our new renders:
image
image


packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -161,6 +162,7 @@ export class CachingGLTFLoader {
 
     const gltf = await cache.get(url)!;
 
+    const duplicate: Mesh[] = [];
```
cdata: The name here is a little ambiguous to me. Would you consider something a bit more descriptive like `meshesToDuplicate`? 🙏

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -178,17 +180,51 @@ export class CachingGLTFLoader {
...
 
+    // We duplicate transparent, double-sided meshes and render the back face
+    // before the front face. This creates perfect triangle sorting for all
+    // convex meshes. Sorting artifacts can still appear when you can see
+    // through more than two layers of a given mesh, but this can usually be
+    // mitigated by the author splitting the mesh into mostly convex regions.
+    // The performance cost is not too great as the same shader is reused and
+    // the same number of fragments are processed; only the vertex shader is run
+    // twice.
+    for (let i = 0; i < duplicate.length; i++) {
```
cdata: If the only thing you are using `i` for is to index into the array, consider rewriting as:

```typescript
for (const mesh of duplicate) {
  // ...
}
```

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -178,17 +180,51 @@ export class CachingGLTFLoader {
...
     }
 
+    // We duplicate transparent, double-sided meshes and render the back face
+    // before the front face. This creates perfect triangle sorting for all
+    // convex meshes. Sorting artifacts can still appear when you can see
+    // through more than two layers of a given mesh, but this can usually be
+    // mitigated by the author splitting the mesh into mostly convex regions.
+    // The performance cost is not too great as the same shader is reused and
+    // the same number of fragments are processed; only the vertex shader is run
+    // twice.
```
cdata: Excellent comment 🍻 

Maybe add an `@see https://threejs.org/examples/webgl_materials_physical_transparency.html` at the end.

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -178,17 +180,51 @@ export class CachingGLTFLoader {
...
+    for (let i = 0; i < duplicate.length; i++) {
+      const mesh = duplicate[i];
+      const material = Array.isArray(mesh.material) ?
```
cdata: I think it's a bit easier to follow the traversal scheme used above on line 184. There, we wrap `mesh.material` in an array if it is not an array. This is arguably wasteful because material is usually _not_ an array, but at the end of the day it is cheap enough that you won't notice it in a profile, and it has the advantage of simplifying the subsequent traversal code.

elalish: Unfortunately three does not accept a material array when it doesn't expect one, so I need a little extra tweaking, but I still like the look of this way better than how it was. :+1:

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -178,15 +180,48 @@ export class CachingGLTFLoader {
...
+
+    // We duplicate transparent, double-sided meshes and render the back face
+    // before the front face. This creates perfect triangle sorting for all
+    // convex meshes. Sorting artifacts can still appear when you can see
+    // through more than two layers of a given mesh, but this can usually be
+    // mitigated by the author splitting the mesh into mostly convex regions.
+    // The performance cost is not too great as the same shader is reused and
+    // the same number of fragments are processed; only the vertex shader is run
+    // twice. @see https://threejs.org/examples/webgl_materials_physical_transparency.html
+    for (const mesh of meshesToDuplicate) {
```
cdata: 💯

PR #962, opened 2020-01-15
Separate out docs, fidelity tests and shared assets

cdata:
This change splits out the source tree into four distinct packages:

 - The `<model-viewer>` web component
 - Our documentation site modelviewer.dev
 - Render fidelity tests / tools
 - Bulk shared assets like models and environment images

As part of this change, we are moving to [lerna.js](https://lerna.js.org/) in order to manage our repo as a collection of distinct packages. [The new top-level README](https://github.com/GoogleWebComponents/model-viewer/blob/split-docs-fidelity-assets/README.md) contains instructions for getting set up with this new arrangement:

```sh
git clone git@github.com:GoogleWebComponents/model-viewer.git
cd model-viewer
npm install
npm run bootstrap
```

These `npm` commands wrap `lerna` calls, so hopefully direct interaction with `lerna` will not be needed for typical dev work.

`lerna` gives us the ability to fan-out calls to `npm run *` commands, so from the root of the project you can invoke commands like `npm run build` and `npm run test` to run the corresponding commands in all sub-packages that implement them.

Most of the `npm run *` commands still work as they used to when run from the `packages/model-viewer` folder. In addition, there is a new `npm run serve` command that you can use from the root of the project to get a project-wide server.

## Important note about `npm install`

TL;DR prefer `npm run bootstrap` at the project root over `npm install` in individual sub-projects.

In order to "cross-link" packages with each other during local dev, `lerna` creates symlink within the `node_modules` folder of each package wherever there happens to be a dependency on another local package. This means that when you make a change in one package, you don't need to do anything to keep other packages up to date. However, it also means that if you run `npm install` (and a few other `npm` commands) directly inside of a given sub-package (as opposed to the root of the project), `npm` will delete the symlinks and try to replace them with real packages downloaded from the NPM registry. This is typically just annoying, but in the case of at least one of our packages it will outright fail because it is a "private" package and never published to the NPM registry.

The solution to this problem is to always prefer `npm run bootstrap` at the root of the project repository in order to install dependencies from NPM, and let `lerna` do its thing.

## Notable differences
 
In no particular order:

 - `npm run check-fidelity` is now `npm run test` for the `rendering-fidelity-tools` project
 - The main `README.md` has been cleaned up and moved to the `model-viewer` project; all other READMEs include links to that one
 - Published fidelity test results will be at https://modelviewer.dev/fidelity/results-viewer.html
 - All build steps are expected to run under the latest stable Node.js release (contemporaneously v12.x.x)


cdata:
I think the reason for the megadiff is that we went from 1 to 4 package-lock.json files...

cdata:
Rebased on master

PR #973, opened 2020-01-29
Fix poster alignment when text-align is set

### LINKED ISSUE #965: Poster image is offset if parent element has text-align: center
### Description
I noticed that on 0.8.0, Bumbleride's posters appears offset:
Screen Shot 2020-01-20 at 10 03 35 PM

The loaded model is in the right spot though.

This must have something to do with the seamless transition work that was done in https://github.com/GoogleWebComponents/model-viewer/pull/890.

I narrowed it down to a parent element on the page having `text-align: center`. It seems like when that is set, the poster gets offset.

I made a Glitch of it here: https://mousy-target.glitch.me/

It's just a model-viewer wrapped in a div with `text-align: center`. Is this expected behaviour?
### Browser Affected
All

### OS
All

### Versions
- model-viewer: v0.8.0

cdata:
Fixes #965

### DIFF
--- a/packages/model-viewer/src/template.ts
+++ b/packages/model-viewer/src/template.ts
@@ -126,6 +126,9 @@ canvas.show {
 #default-poster {
   width: 100%;
   height: 100%;
+  /* The default poster is a <button> so we need to set display
+   * to prevent it from being affected by text-align: */
+  display: block;
   position: absolute;
   border: none;
   padding: 0;


elalish:
@cdata Once #917 lands you can pull master to fix the fidelity CI failure.

cdata:
Rebased

PR #975, opened 2020-01-29
Transparent rendering

### LINKED ISSUE #182: Background color does not accept all valid CSS strings

### LINKED ISSUE #381: Bad render results for Alpha Blend Mode test model

### LINKED ISSUE #492: Make it possible to set background color with CSS custom property
As of #487 we will support a few new CSS custom properties so that authors can style `<model-viewer>` controls with CSS. Initially, we wanted to add `--background-color` as well, but it's a bit complicated because the background color isn't actually configured with CSS (it's done in WebGL).

We should do the work to support configuring the color with a CSS property, though!

### LINKED ISSUE #874: Make it possible to configure the background to be transparent

elalish:
Fixes #874 
Fixes #492 
Fixes #182

This a possible alternative approach to #949, to avoid a second renderer, as I'm concerned about how that will complicate some upcoming performance improvement work. This pulls ideas from #949 (thank you @nicolas-daures!), but makes the singleton renderer transparent and instead fixes #381 directly. 

This also does away with our background-color attribute, since with a properly transparent rendering context, the standard CSS background-color property works as expected. Also, is it just me or does this render faster?


elalish:
@googlebot I fixed it.

packages/model-viewer/src/three-components/ModelUtils.ts
```diff
@@ -36,10 +32,35 @@ export const cloneGltf = (gltf: GLTF): GLTF => {
    */
   const cloneAndPatchMaterial = (material: Material): Material => {
     const clone = material.clone();
+    // This allows us to patch three's materials, on top of patches already
+    // made, for instance GLTFLoader patches SpecularGlossiness materials.
+    // Unfortunately, three's program cache differentiates SpecGloss materials
+    // via onBeforeCompile.toString(), so these two functions do the same thing
+    // but look different in order to force a proper recompile.
+    const oldOnBeforeCompile = material.onBeforeCompile;
+    clone.onBeforeCompile = (material as any).isGLTFSpecularGlossinessMaterial ?
```
cdata: Is it possible for both material types to use the the SpecularGlossiness path here?

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -246,12 +246,12 @@ export class Renderer extends EventDispatcher {
       // clearing the depth from a different buffer -- possibly
       // from something in
       this.threeRenderer.setRenderTarget(null);
-      this.threeRenderer.clearDepth();
       this.threeRenderer.setViewport(0, 0, width, height);
       this.threeRenderer.render(scene, camera);
 
       const widthDPR = width * dpr;
       const heightDPR = height * dpr;
+      context.clearRect(0, 0, widthDPR, heightDPR);
```
cdata: 👍

packages/model-viewer/src/three-components/ModelUtils.ts
```diff
@@ -36,10 +32,35 @@ export const cloneGltf = (gltf: GLTF): GLTF => {
...
     clone.shadowSide = FrontSide;
+    // This improves transparent rendering and can be removed whenever
+    // https://github.com/mrdoob/three.js/pull/18235 finally lands.
     if (clone.transparent) {
       clone.depthWrite = false;
     }
+    // This little hack ignores alpha for opaque materials, in order to comply
+    // with the glTF spec.
+    if (!clone.alphaTest && !clone.transparent) {
+      clone.alphaTest = -0.5;
```
cdata: Does `-0.5` have meaningful significance?

elalish: No, any value less than zero has the same behavior.

packages/model-viewer/src/three-components/ModelUtils.ts
elalish: Alas no; I tried to describe it in the comment. If the toString() version of onBeforeCompile is the same for the two materials, the second one will not be compiled and will instead try to use the first, which does not work. Took me all morning to diagnose this bug...

cdata: Yah, that's a really funky way to test for shader equivalence.

It would be nice if there was a more hygienic / extensible API for patching Three.js materials. Oh well.

elalish: Indeed, though I'm just glad there's a way to patch Three.js materials at all. The onBeforeCompile method hasn't been around that long.

koch-crsd:
@elalish @nicolas-daures 
Thanks so much guys 👍 this is a really great solution!
tested and working fine this way for what I need :)

PR #978, opened 2020-01-30
Switch model viewer's experimental WebXR to use new hit test API

bialpio:
This PR will make the ModelViewer use new hit test API instead of the legacy, promise-based one.


packages/model-viewer/src/three-components/Reticle.ts
```diff
@@ -47,42 +49,45 @@ export default class Reticle extends Object3D {
...
+    if(!this.hitTestSourceRequest) {
+      this.hitTestSourceRequest = _session.requestHitTestSource({space: _viewerReferenceSpace as any})
+                                          .then(hitTestSource => {
+                                            this.hitTestSource = hitTestSource;
+                                          });
+    } else if(this.hitTestSource){
+      const hitTestResults = _frame.getHitTestResults(this.hitTestSource);
+      if (hitTestResults.length) {
+        const hit = hitTestResults[0];
+        this._hitMatrix = new Matrix4().fromArray(hit.getPose(_frameOfRef as any)!.transform.matrix);
```
elalish: I'm noticing a number of `as any`; can we get proper TS types for these somewhere? I know they weren't there before either, but we are trying to come out of unstable.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -151,6 +153,7 @@ export class ARRenderer extends EventDispatcher {
     // @TODO #293, handle WebXR API changes
```
elalish: Have we handled this TODO yet? And any others in there for that matter?

packages/model-viewer/src/three-components/Reticle.ts
bialpio: We already have typings for what those objects are (`XRReferenceSpace`s) in webxr.ts - the problem here is that ARRenderer.ts has both the $viewerRefSpace and $refSpace typed as `XRCoordinateSystem`. If you think that those can be safely changed (I have not tried doing so) from XRCoordinateSystem into XRReferenceSpace, then there should be no need for so many casts to `any` (both here and in ARRenderer.tx).

packages/model-viewer/src/three-components/ARRenderer.ts
bialpio: From what I've seen, webxr.ts already was fairly up-to-date, I'll make another pass at this file and try to remove the `any` casts - they are probably the best way to determine whether the API changes are already handled.

packages/model-viewer/src/three-components/Reticle.ts
elalish: Let's give it a shot. If it doesn't work, that'll be good info to have. This stuff is all pretty separate from the rest of the code base, so I'm not worried about far-reaching implications.

bialpio: Removed all WebXR-related `any` casts from Reticle.ts and ARRenderer.ts.

packages/model-viewer/src/three-components/ARRenderer.ts
bialpio: This TODO is mostly resolved but I'd expect this to be an ongoing effort as WebXR is adding new features and typings will probably always be slightly behind it.

PR #985, opened 2020-02-04
shortcut rendering for single instance

### LINKED ISSUE #845: Model view is very slow on Firefox in demo
### Description

Steps to reproduce:
* Open up Firefox on Mac.
* Go to: https://model-viewer-tester.glitch.me/
* Observe the slow frame rate

I took a profile here: https://perfht.ml/31EpEhu

It appears the slowness comes from the following area:

https://github.com/GoogleWebComponents/model-viewer/blob/52b15951165bba67c6ea2f36446ffa780952bc50/src/three-components/Renderer.ts#L209-L218

Specifically the call to `context.drawImage`:
`context.drawImage(this.renderer.domElement, 0, this.canvas.height - heightDPR, widthDPR, heightDPR, 0, 0, widthDPR, heightDPR);` is very slow in Firefox making the model viewer not very usable.

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1590515 in Firefox, but wanted to pass it along to this project, as I'm not sure when that will be prioritized.

### Browser Affected

- [ ] Chrome
- [ ] Edge
- [x] Firefox (71-72)
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [x] MacOS
- [ ] Windows

### Versions

Unknown

elalish:
Fixes #845 

My method here involves placing the renderer's canvas element into the shadow root when there is only one registered scene, thus bypassing the slow drawImage call. For multiple scenes on a page, it falls back to the old behavior. Unfortunately this also meant removing OffscreenCanvas, but this is currently doing nothing for us as it hasn't been hooked up to an ImageBitmapRenderingContext yet. I'm hoping this approach can be extended to support OffscreenCanvas in the future, but it's hard to tell from the limited API documentation so far.

### DIFF
--- a/packages/model-viewer/src/constants.ts
+++ b/packages/model-viewer/src/constants.ts
@@ -42,10 +42,9 @@ export const IS_MOBILE = (() => {
   return check;
 })();
 
-export const HAS_OFFSCREEN_CANVAS = Boolean((self as any).OffscreenCanvas);
-export const OFFSCREEN_CANVAS_SUPPORT_BITMAP =
-    Boolean((self as any).OffscreenCanvas) &&
-    Boolean((self as any).OffscreenCanvas.prototype.transferToImageBitmap);
+export const USE_OFFSCREEN_CANVAS = Boolean((self as any).OffscreenCanvas) &&
+    Boolean((self as any).OffscreenCanvas.prototype.transferToImageBitmap) &&
+    Boolean((self as any).OffscreenCanvas.prototype.transferControlToOffscreen);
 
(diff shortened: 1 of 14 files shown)


elalish:
Hmm, on second thought, just moving the controls up to the container node has some a11y implications (and probably more). I think instead I'll make a canvas container that the two canvases can live in and put all the DOM stuff on it instead.

elalish:
I think we can close this in favor of #995.

PR #990, opened 2020-02-05
Include attribute to target annotations / hotspot elements based on visibility

### LINKED ISSUE #989: No ability to target annotations / hotspots based on visibility
### Description

Currently there is no easy way to style annotations / hotspots based on their visibility. While some CSS variables are available (`--min-opacity`), these aren't sufficient for e.g. transitions on child elements, etc.

I'd propose including a `[data-occluded]` attribute on the slotted element, as per the `[data-position]` and `[data-normal]` attributes, to allow targeting. The value should probably be binary for performance, so `1 = occluded` and `0 = visible`. Alternatively the value could update as the camera moves `1 > occluded > -1` and trigger an associated event, or remain binary but trigger a detailed event.

#### Live Demo

https://glitch.com/edit/#!/model-viewer

### Browser Affected

- [ x ] Chrome
- [ x ] Edge
- [ x ] Firefox
- [ x ] Helios
- [ x ] IE
- [ x ] Safari

### OS

- [ x ] Android
- [ x ] iOS
- [ x ] Linux
- [ x ] MacOS
- [ x ] Windows

### Versions

- model-viewer: v0.8.1 (master branch)
- three.js: r113.1

theprojectsomething:
Resolves #989 

In order to enable performant attribute updates on annotations, it was necessary to update the annotations feature to only update / render annotations when necessary, rather than on each tick.


packages/model-viewer/src/features/annotation.ts
```diff
@@ -192,30 +195,64 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-      this[$annotationRenderer].render(this[$scene], this[$scene].activeCamera);
+
+      // NOTE: checks the current annotation view against camera for
+      // position / field-of-view updates (or alternatively a forced update)
+      // QUESTION: is this already implemented elsewhere?
+      const position = this[$annotationView].get('position');
+      if (!position
+        || this[$annotationView].get('update')
+        || this[$scene].activeCamera.position.distanceTo(position) !== 0
+        || this[$scene].camera.getEffectiveFOV() !== this[$annotationView].get('fov')) {
```
cdata: I think we need to run clang-format on these files. If you rebase, there is now a pre-commit hook that will do it for you.

packages/model-viewer/package.json
```diff
@@ -36,6 +36,7 @@
     "prepare": "if [ ! -L './shared-assets' ]; then ln -s ./node_modules/@google/model-viewer-shared-assets ./shared-assets; fi",
     "build": "tsc && rollup -c --environment NODE_ENV:production",
     "build:dev": "tsc && rollup -c --environment NODE_ENV:development",
+    "build:release": "mkdir -p release; cp LICENSE POLYFILLS.md README.md package-lock.json package.json screenshot.png release; cp -r dist release/dist; cp -r lib release/lib",
```
cdata: Let's discuss changing our release process in a separate PR

packages/model-viewer/src/model-viewer.ts
```diff
@@ -34,7 +34,9 @@ export const ModelViewerElement = MagicLeapMixin(AnnotationMixin(StagingMixin(
 
 export type ModelViewerElement = InstanceType<typeof ModelViewerElement>;
 
-customElements.define('model-viewer', ModelViewerElement);
+if (!customElements.get('model-viewer')) {
```
cdata: This seems like a case where we should be warning that `model-viewer` has already been defined. In fact, this change hides a error from them that would have warned them.

`model-viewer.js` is a gigantic bundle, so we should make the user aware that it is being loaded without having been defined.

packages/model-viewer/package.json
theprojectsomething: apologies - this was a later push to the branch, wasn't intended to be included

packages/model-viewer/src/model-viewer.ts
theprojectsomething: This was primarily to resolve a hot-reloading issue in a module setup. Works on the assumption that `model-viewer` would only have been defined as a `<ModelViewerElement>`.

The current error is a bit opaque. Would it be worth type checking the already-defined element and warning specifically if there's a mismatch / the library has been double-loaded? Either way I'll move this for discussion elsewhere.

theprojectsomething:
@cdata I've created a new PR #1018 to replace this one

PR #996, opened 2020-02-07
updated posters

elalish:
This was ostensibly about updating our lazy-loading posters to be properly seamless (that had been bothering me). However in the process I also found several small bugs related to posters in our tester, so those are also fixed here.


cdata:
Could you describe the bugs? Can we add regression tests for any of them?

cdata:
Ah, nevermind. It looks like only the tester itself was affected?

PR #1008, opened 2020-02-08
Hotspot helper

### LINKED ISSUE #977: Rename `--min-opacity` to `--min-hotspot-opacity`
The hotspot custom property `--min-opacity` is too vague. We should rename it so that it is more explicit about what nodes it applies to.

### LINKED ISSUE #979: Hotspot normals need to take turntable rotation into account
It looks like normals assigned to hotspots aren't used in a way that takes turntable rotation into account. This has the effect of causing "occluded" hotspots to stay low-opacity even when they shouldn't, and for visible hotspots to remain high-opacity even when they shouldn't, whenever the model is rotating due to `auto-rotate`.

elalish:
Fixes #977 
Fixes #979 

In the spirit of getting our docs ready for release, I wanted to add a hotspot helper to our tester doc. In order to make this possible, I also had to add an API method, `getHitResult`, but I believe this is pretty generally useful anyway. Try out our interactive example and see how you can add hotspots by clicking on the model and they display the data-position and data-normal attribute values you need to put a hotspot there. 

One bug: I notice that different models are returning normal vectors seemingly in different coordinate systems. I'll resolve this before the PR lands, but I'd like some feedback in the meanwhile.

### DIFF
--- a/packages/model-viewer/src/features/annotation.ts
+++ b/packages/model-viewer/src/features/annotation.ts
@@ -214,13 +214,21 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
       this[$pixelPosition].y *= -1;
       raycaster.setFromCamera(this[$pixelPosition], this[$scene].getCamera());
       const hits = raycaster.intersectObject(this[$scene], true);
+
       if (hits.length === 0) {
         return {position: '', normal: ''};
       }
       const hit = hits[0];
+
       const {x, y, z} = hit.point;
       const hitPosition = `${x}m ${y}m ${z}m`;
-      const hitNormal = '0 0 1';
+
+      let hitNormal = '0 1 0';
+      if (hit.face != null) {
+        const {x: nx, y: ny, z: nz} = hit.face.normal;
+        hitNormal = `${- nx} ${ny} ${- nz}`;
+      }
+
       return {position: hitPosition, normal: hitNormal};
     }
 
--- a/packages/model-viewer/src/features/annotation.ts
+++ b/packages/model-viewer/src/features/annotation.ts
@@ -17,7 +17,7 @@
 import {Matrix4, Raycaster, Vector2, Vector3} from 'three';
 import {CSS2DObject, CSS2DRenderer} from 'three/examples/jsm/renderers/CSS2DRenderer.js';
 
-import ModelViewerElementBase, {$onResize, $scene, $tick, Vector3D} from '../model-viewer-base.js';
+import ModelViewerElementBase, {$onResize, $scene, $tick, toVector3D} from '../model-viewer-base.js';
 import {normalizeUnit} from '../styles/conversions.js';
 import {NumberNode, parseExpressions} from '../styles/parsers.js';
 import {Constructor} from '../utilities.js';
@@ -112,8 +112,7 @@ export class Hotspot extends CSS2DObject {
 
 export declare interface AnnotationInterface {
   updateHotspot(config: HotspotConfiguration): void;
-  positionAndNormalFromPoint(pixelX: number, pixelY: number):
-      {position?: Vector3D, normal?: Vector3D}
+  positionAndNormalFromPoint(pixelX: number, pixelY: number)
 }
 
 /**
@@ -209,8 +208,7 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
      * data-normal attributes. If the mesh is not hit, position returns the
      * empty string.
      */
-    positionAndNormalFromPoint(pixelX: number, pixelY: number):
-        {position?: Vector3D, normal?: Vector3D} {
+    positionAndNormalFromPoint(pixelX: number, pixelY: number) {
       const {width, height} = this[$scene];
       this[$pixelPosition]
           .set(pixelX / width, pixelY / height)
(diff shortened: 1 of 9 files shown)


packages/model-viewer/src/features/annotation.ts
```diff
@@ -194,6 +205,41 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
 
+    /**
+     * This method returns the world position and normal of the point on the
+     * mesh corresponding to the input pixel coordinates given relative to the
+     * model-viewer element. The position and normal are returned as strings in
+     * the format suitable for putting in a hotspot's data-position and
+     * data-normal attributes. If the mesh is not hit, position returns the
+     * empty string.
+     */
+    getHitResult(pixelX: number, pixelY: number): HitResult {
```
cdata: It seems like a `HitResult` is a descriptor for a plane. It might help folks to understand the significance of `normal` (which our users have reported is confusing at first) if we use language that is more natural to what the API does for them. e.g., `planeFromPoint` or `surfaceFromPoint` or something to that effect. Suggested naming inspired by [`elementFromPoint`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/elementFromPoint).

Also: it would be great to see some tests for this 👍

packages/model-viewer/src/features/annotation.ts
```diff
@@ -194,6 +205,41 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+      raycaster.setFromCamera(this[$pixelPosition], this[$scene].getCamera());
+      const hits = raycaster.intersectObject(this[$scene], true);
+
+      if (hits.length === 0) {
+        return {position: '', normal: ''};
+      }
+      const hit = hits[0];
+
+      const {x, y, z} = hit.point;
+      const hitPosition = `${x}m ${y}m ${z}m`;
```
cdata: We typically report values from the imperative API as objects, not serialized string equivalents.

I recognize that it's convenient to get the serialized value for the purpose of assigning directly to an attribute (and that it is relatively cumbersome to construct a string manually), but it's possible for a user to construct a string from a structured object; it's a lot harder for them to parse the serialized string correctly if they want to access the values.

In the long run, a happy middleground might be to return specialized objects with `toString` implementations so that they can be easily serialized (and assigned to attributes directly w/o a manual serialization step).

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -61,6 +61,14 @@ export const $progressTracker = Symbol('progressTracker');
 export const $getLoaded = Symbol('getLoaded');
 export const $getModelIsVisible = Symbol('getModelIsVisible');
 
+export class Vector3D {
```
cdata: Let's make this an interface, not a class. Class construction has extra cost associated with it that isn't really called for for such a simple type.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -194,6 +201,45 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+      if (hits.length === 0) {
+        return {};
+      }
+      const hit = hits[0];
+      const worldToPivot =
+          new Matrix4().getInverse(this[$scene].pivot.matrixWorld);
+      const position = new Vector3D();
+      Object.assign(position, hit.point.applyMatrix4(worldToPivot));
+
+      if (hit.face == null) {
```
cdata: What is this condition?

packages/model-viewer/src/features/annotation.ts
```diff
@@ -194,6 +201,45 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+      this[$pixelPosition]
+          .set(pixelX / width, pixelY / height)
+          .multiplyScalar(2)
+          .subScalar(1);
+      this[$pixelPosition].y *= -1;
+      raycaster.setFromCamera(this[$pixelPosition], this[$scene].getCamera());
+      const hits = raycaster.intersectObject(this[$scene], true);
+
+      if (hits.length === 0) {
+        return {};
```
cdata: I think `null` would make a more explicit signal for this case.

packages/modelviewer.dev/index.html
```diff
@@ -511,6 +511,15 @@ <h3 class="grouping-title">Methods</h3>
...
               attribute on account of user interaction or camera interpolation.</p>
             </li>
+            <li>
+              <div>positionAndNormalFromPoint(pixelX, pixelY)</div>
+              <p>Returns the world position and normal of the point on the mesh
+              corresponding to the input pixel coordinates given relative to the
+              model-viewer element. The position and normal are returned as strings in
+              the format suitable for putting in a hotspot's data-position and
+              data-normal attributes. If the mesh is not hit, position returns the
+              empty string.</p>
```
cdata: Is this last line still true?

packages/modelviewer.dev/index.html
```diff
@@ -511,6 +511,15 @@ <h3 class="grouping-title">Methods</h3>
               This value may differ from the value configured with the field-of-view 
               attribute on account of user interaction or camera interpolation.</p>
             </li>
+            <li>
+              <div>positionAndNormalFromPoint(pixelX, pixelY)</div>
+              <p>Returns the world position and normal of the point on the mesh
+              corresponding to the input pixel coordinates given relative to the
+              model-viewer element. The position and normal are returned as strings in
```
cdata: Not strings anymore; also, what about the case where there is no normal?

packages/model-viewer/src/features/annotation.ts
elalish: three can hit things that aren't faces. I think it would be rare for us, but I don't want to put an arbitrary return normal.

PR #1018, opened 2020-02-12
Annotations visibility events / performance

### LINKED ISSUE #989: No ability to target annotations / hotspots based on visibility

theprojectsomething:
### Fixes #989 to allow annotations to be targeted based on visibility
 - dispatches a bubbling `hotspot-change` event as visibility changes on a hotspot
 - includes a test / performance fix to ensure updates only occur when the scene / camera changes (as opposed to every tick); this extends to the `.hide` class being applied within the component context

**Note:** This PR replaces #990


cdata:
@theprojectsomething thanks for your hard work on this. I have bundled what I believe are equivalent changes into a refinement PR for hotspots here: https://github.com/google/model-viewer/pull/1023

Please take a look! If you are willing to give feedback, it would be welcome and appreciated.

theprojectsomething:
Thanks @cdata that looks like a much more comprehensive refactor!

Everything looks good to me, appreciate the authoritative input! My only question would be around rendering every tick .. assume this applies elsewhere in the component as well .. but it seems like a waste of horsepower?

cdata:
@theprojectsomething with my change, we avoid touching the DOM on every tick (which is an improvement over what we started with). I think it's close to as good as it is going to get.

Let me know if you have any feedback after using these improvements. In particular, I'm interested to hear how `data-visibility-attribute` works for you, as we added that based on your feedback 👍 

I'll close out this PR since it is superseded by #1023 . Thanks again for all your hard work on this!

theprojectsomething:
Thanks for following up @cdata, and for all the effort at your end! The attribute looks good, but let me come back to you once I've had a chance to test it in the wild over the next couple of days.

Totally understood on the tick callback. In my simplistic world view I was thinking that the annotations' projection should only change with the camera .. but appreciate there's a lot more going on than that. Would you be opposed to me opening an issue for future (ie. definitely not immediate) consideration? Or is the CSS2D renderer overhead super negligible / am I pushing my luck :)

cdata:
@theprojectsomething all issues are welcome and encouraged, and if you see an opportunity for improvement you should not assume that we already considered it 😅 we are only human ourselves.

Especially in the case of performance optimization, we have wonderful profilers in Chrome devtools for measuring the impact and we find areas for improvement all the time. So, do take stock if you have time and inclination!

PR #1025, opened 2020-02-13
3dom and iframes

judax:
Use case: A page contains a model-viewer instance but the functionality to control
that instance through 3dom scripts is emedded in an iframe.

* ThreeDOMExecutionContext supports different modes: Standalone, ExternalWorker, ExternalPort.

* One MessageChannel in ThreeDOMExecutionContext.

* Worker can be null in ThreeDOMExecutionContext.

* port2 can be exposed in ThreeDOMExecutionContext.

* The scene-graph feature exposes new method: setThreeDOMExecutionContext.

* New demo to show the iframe functionality.

### DIFF
--- a/packages/3dom/src/context-host.ts
+++ b/packages/3dom/src/context-host.ts
@@ -1,0 +1,108 @@
+import {ThreeDOMCapability} from './api.js';
+import {ThreeDOMExecutionContext} from './context.js';
+import {ThreeDOMMessageType} from './protocol.js';
+
+// This fake worker class is just used to capture the transfer of the message
+// port.
+class FakeWorker extends EventTarget implements Worker {
+  onerror: EventListener|null = null;
+  onmessage: EventListener|null = null;
+  onmessageerror: EventListener|null = null;
+  port2: MessagePort|null = null;
+  hostExecutionContext: HostThreeDOMExecutionContext;
+
+  constructor(hostExecutionContext: HostThreeDOMExecutionContext) {
+    super();
+    this.hostExecutionContext = hostExecutionContext;
+  }
+
+  postMessage(message: any, transfer: Array<Transferable>): void;
+  postMessage(message: any, options?: PostMessageOptions|undefined): void;
+  postMessage(message: any, secondArg: any): void {
+    // When the handshake message is sent, capture and store the port.
+    if (message.type == ThreeDOMMessageType.HANDSHAKE) {
+      const transfer = secondArg as Array<Transferable>;
+      this.port2 = transfer[0] as MessagePort;
+      // Indicate the host execution context that the port has been captured.
+      this.hostExecutionContext.portHasBeenSet();
+    }
+  }
+
+  terminate() {
+    this.port2 = null;
+  }
+}
+
+const $iframe = Symbol('iframe');
+const $iframeLoaded = Symbol('iframeLoaded');
+
+export class HostThreeDOMExecutionContext extends ThreeDOMExecutionContext {
+  protected[$iframe]: HTMLIFrameElement;
+  protected[$iframeLoaded] = false;
+
+  constructor(
+      capabilities: Array<ThreeDOMCapability>,
+      iframe: HTMLIFrameElement|null = null) {
+    super(capabilities);
+    // Make sure there is an iframe to connect with
+    if (!iframe) {
+      iframe = document.querySelector('iframe');
+      if (!iframe) {
+        throw new Error(
+            'Either provide an iframe or the page should contain and iframe.');
+      }
+    }
+    this[$iframe] = iframe;
+    // Wait for the iframe to load
+    const onIFrameLoaded = () => {
+      this[$iframeLoaded] = true;
+      this[$iframe].removeEventListener('load', onIFrameLoaded);
+      this.sendHandshakeToIFrame();
+    };
+    this[$iframe].addEventListener('load', onIFrameLoaded);
+  }
+
+  // Called from the FakeWorker when the postMessage is passed.
+  portHasBeenSet() {
+    this.sendHandshakeToIFrame();
+  }
+
+  protected sendHandshakeToIFrame() {
+    // Wait until the iframe is loaded AND the port has been set
+    const fakeWorker = this.worker as FakeWorker;
+    const port2 = fakeWorker.port2;
+    if (!this[$iframeLoaded] || !port2) {
+      return;
+    }
+    // Listen for messages from the iframe
+    const onMessageReceived = (event: MessageEvent) => {
+      // If the iframe send the handshake response, it will contain the 3DOM
+      // script to be loaded, so evaluate it (it will actually be received and
+      // evaluated in the iframe's worker).
+      if (event.data.action === 'handshakeResponse') {
+        // No need to listen to more messages from the iframe
+        window.removeEventListener('message', onMessageReceived);
+        // Load the script passed from the iframe
+        this.eval(event.data.payload);
+        // Indicate the iframe that the host is ready
+        const contentWindow = this[$iframe].contentWindow;
+        if (contentWindow) {
+          contentWindow.postMessage({action: 'ready'}, '*');
+        }
+        window.dispatchEvent(
+            new CustomEvent('3domready', {detail: {executionContext: this}}));
+      }
+    };
+    window.addEventListener('message', onMessageReceived);
+    // Send the hadnshake to the iframe transferring the port
+    const contentWindow = this[$iframe].contentWindow;
+    if (contentWindow) {
+      contentWindow.postMessage({action: 'handshakeRequest'}, '*', [port2]);
+    }
+  }
+
+  // Override
+  protected createWorker(_url: string): Worker {
+    return new FakeWorker(this);
+  }
+}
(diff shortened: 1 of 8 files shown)


cdata:
@judax very interesting idea. I wonder: could we spare most of the modifications required here if we abstracted at a lower level?

For example, we could allow specializations of the `ThreeDOMExecutionContext` like this:

```typescript
class ThreeDOMExecutionContext extends EventTarget {
  // ...

  // Creates the isolated context where scripts run, and returns a reference to an object
  // with a Worker-like interface to communicate that context and control its lifecycle
  // NOTE: url may or may not be used depending on the implementation of this
  // method; there are good reasons for both approaches.
  protected createIsolatedContext(url: string): Worker {
    // Default implementation:
    return new Worker(url);
  }
}
```

With the above implementation, we could write a specialization for the host frame and a counterpart class for the child `<iframe>`:

```typescript
/* === In the host frame context  === */

/**
 * A specialization of ThreeDOMExecutionContext that controls a Worker in a
 * remote <iframe>
 */
class HostThreeDOMExecutionContext extends ThreeDOMExecutionContext {
  // Override:
  protected createIsolatedContext(url: string): Worker {
    return new WorkerProxyInterfaceToIframe(url);
  }
}


/**
 * A class that looks like a Worker but actually controls execution in a remote
 * <iframe> context
 */
class WorkerProxyInterfaceToIframe implements Worker {
  protected iframe: HTMLIframeElement;

  constructor(url: string) {
    // Negotiate connection w/ iframe, pass in a dedicated MessagePort
    // Probably best to pass url into the iframe once communication is
    // established, rather than regenerate it inside the iframe
  }

  // Rest of Worker implementation goes here
  // @see https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L18821-L18830
}



/* === In the <iframe> context  === */

/**
 * Establishes connections with a host frame, creates Worker instances and
 * forwards messages to those workers
 */
class RemoteAgent {
  constructor() {
    self.addEventListener('message', (event) => {
      if (event.origin !== TRUSTED_GOOGLE_ORIGIN) {
        // Summarily ignored
        return;
      }
      // Establish connection with host context:
      this.negotiateConnection(event.ports[0]);
    });
  }

  protected onHostMessage(event: MessageEvent) {
    // Handle host commands to create workers on behalf of the
    // HostThreeDOMExecutionContext
    // Also forward messages from host to the workers
  }

  protected onWorkerMessage(event: MessageEvent) {
    // Forward message from the worker to the host context
  }
}

self.remoteAgent = new RemoteAgent();
```

In the example above, the `WorkerProxyInterfaceToIframe` class would be in an intermediary position between the `ThreeDOMExecutionContext` and the `<iframe>`. All messages passed from 3DOM intended for the `Worker` would be proxied by that class. The `WorkerProxyInterfaceToIframe` would be responsible for managing the relationship with the `<iframe>` and negotiating the connection to some special code in the `<iframe>`'s main thread.

The advantage of this approach is that 3DOM can generally remain agnostic to where the `Worker` runs. It doesn't have to care about whether the `Worker` is spawned from the same execution context, or is perhaps running in Unity or some other remote context. Instead, it just cares that the thing it is communicating with has a `Worker`-like interface.

I'm interested to hear your thoughts on this.

judax:
Thanks for the feedback Chris. I took a new stab at the proposal following your advise.

- Added an internal `createWorker` method in `ThreeDOMExecutionContext` to delegate the creation of the worker to a method that an inheriting class could override.

- Created 2 new classes that inherit from `ThreeDOMExecutionContext`, one fort the host (main) page and one for the iframe page.

  - `HostThreeDOMExecutionContext`: The worker is fake and captures the port when it is sent out to the worker. It handles all the handshake needed with the iframe page to hand over the captured port and inject the 3dom script passed from the iframe.

  - `IFrameThreeDOMExecutionContext`: The worker is functional but it also captures the port passing and delays it until the port is received from the host page.

In this new approach I maintained the modification in the scene-graph mixin to set the execution context to model viewer in the host page. I was thinking that this could be done with yet another mixin, but some symbols from the current scene-graph mixin might need to be exported. This would let anyone to create new "flavors" of model-viewer as needed. Let me know what you think of this idea of using a mixin instead or if you have any other ideas.

packages/3dom/src/context-host.ts
```diff
@@ -0,0 +1,108 @@
...
+  onmessageerror: EventListener|null = null;
+  port2: MessagePort|null = null;
+  hostExecutionContext: HostThreeDOMExecutionContext;
+
+  constructor(hostExecutionContext: HostThreeDOMExecutionContext) {
+    super();
+    this.hostExecutionContext = hostExecutionContext;
+  }
+
+  postMessage(message: any, transfer: Array<Transferable>): void;
```
lincolnfrog: what type is message? it seems like it is just an Object? Linter is complaining we're supposed to avoid any in general...

packages/3dom/src/context-host.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  terminate() {
+    this.port2 = null;
+  }
+}
+
+const $iframe = Symbol('iframe');
+const $iframeLoaded = Symbol('iframeLoaded');
+
+export class HostThreeDOMExecutionContext extends ThreeDOMExecutionContext {
```
lincolnfrog: Docstring needed here

packages/3dom/src/context-host.ts
```diff
@@ -0,0 +1,108 @@
...
+  protected[$iframe]: HTMLIFrameElement;
+  protected[$iframeLoaded] = false;
+
+  constructor(
+      capabilities: Array<ThreeDOMCapability>,
+      iframe: HTMLIFrameElement|null = null) {
+    super(capabilities);
+    // Make sure there is an iframe to connect with
```
lincolnfrog: this[$iframe] = iframe ? iframe : document.querySelector('iframe');
if (!this[$iframe]) {
  throw new Error(...
}

packages/3dom/src/context-host.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (!iframe) {
+      iframe = document.querySelector('iframe');
```
lincolnfrog: constant?

packages/3dom/src/context-host.ts
```diff
@@ -0,0 +1,108 @@
...
+    };
+    this[$iframe].addEventListener('load', onIFrameLoaded);
+  }
+
+  // Called from the FakeWorker when the postMessage is passed.
+  portHasBeenSet() {
+    this.sendHandshakeToIFrame();
+  }
+
+  protected sendHandshakeToIFrame() {
```
lincolnfrog: docstring

packages/3dom/src/context-host.ts
```diff
@@ -0,0 +1,108 @@
...
+    // Wait until the iframe is loaded AND the port has been set
+    const fakeWorker = this.worker as FakeWorker;
+    const port2 = fakeWorker.port2;
+    if (!this[$iframeLoaded] || !port2) {
+      return;
+    }
+    // Listen for messages from the iframe
```
lincolnfrog: any reason this isn't just a class method?

packages/3dom/src/context-host.ts
```diff
@@ -0,0 +1,108 @@
...
+    const onMessageReceived = (event: MessageEvent) => {
+      // If the iframe send the handshake response, it will contain the 3DOM
+      // script to be loaded, so evaluate it (it will actually be received and
+      // evaluated in the iframe's worker).
+      if (event.data.action === 'handshakeResponse') {
```
lincolnfrog: should all these literal strings be constants?

packages/3dom/src/context-iframe.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  protected postHandshakeMessage() {
+    if (this.handshakeMessage && this.port2) {
+      super.postMessage(this.handshakeMessage, [this.port2]);
+    }
+  }
+}
+
+export class IFrameThreeDOMExecutionContext extends ThreeDOMExecutionContext {
```
lincolnfrog: docstring

packages/3dom/src/context-iframe.ts
```diff
@@ -0,0 +1,79 @@
...
+  constructor(capabilities: Array<ThreeDOMCapability>) {
+    super(capabilities);
+    const onMessageReceived = (event: MessageEvent) => {
+      switch (event.data.action) {
+        case 'handshakeRequest': {
```
lincolnfrog: especially if used in multiple places, these string literals should be constants

packages/3dom/src/context-iframe.ts
```diff
@@ -0,0 +1,79 @@
...
+          break;
+        }
+        case 'ready':
+          window.removeEventListener('message', onMessageReceived);
+          window.dispatchEvent(
+              new CustomEvent('3domready', {detail: {executionContext: this}}));
+          break;
+      }
+    };
+    window.addEventListener('message', onMessageReceived);
```
lincolnfrog: this too

elalish:
We've gone a different direction with 3DOM: #1302.

Issue #1016, opened 2020-02-11
Model is black on AR example


takahirox:
### Description

I tried [model-viewer AR example](https://modelviewer.dev/examples/augmented-reality.html) on my Samsung Galaxy S8+ + Chrome Canary. I see a model black (I guess no lighting?) in AR mode while the model looks fine out of AR mode. Is this a known issue? Or can it be a platform specific issue?

![image](https://user-images.githubusercontent.com/7637832/74285350-810e4900-4cda-11ea-99fd-d31e32854477.png)

Device: Samsung Galaxy S8+
OS: Android
Browser: Chrome Canary 82.0.4054.0

Console log via remote debugger
```javascript
prismatic.min.js:6 Unable to render content: No mixed-reality browser detected.
three.module.js:16000 THREE.WebGLRenderer: OES_texture_float extension not supported.
three.module.js:16000 THREE.WebGLRenderer: OES_texture_half_float extension not supported.
three.module.js:16000 THREE.WebGLRenderer: OES_texture_half_float_linear extension not supported.
three.module.js:16000 THREE.WebGLRenderer: OES_texture_float_linear extension not supported.
three.module.js:16000 THREE.WebGLRenderer: EXT_frag_depth extension not supported.
three.module.js:16000 THREE.WebGLRenderer: WEBGL_draw_buffers extension not supported.
three.module.js:16000 THREE.WebGLRenderer: EXT_shader_texture_lod extension not supported.
three.module.js:16000 THREE.WebGLRenderer: EXT_texture_filter_anisotropic extension not supported.
Attempting to present in AR...
```

#782 might be a related issue? Model is always black here tho.

### Browser Affected

- [x] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [x] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

cdata:
Thanks for reporting, @takahirox ! It looks like this is the WebXR-based AR mode. Is that correct?

takahirox:
Yes!

cdata:
@takahirox thanks, this part of our code hasn't received much love, but we are hoping to shore it up soon. It's very interesting to see this, and I too wonder if it is a device-specific issue.

cc @elalish @bialpio

takahirox:
[...]

PR #1027, opened 2020-02-14
Fix the black model problem in WebXR AR mode

takahirox:
### Reference Issue

Fixes: #1016 

Please refer to https://github.com/google/model-viewer/issues/1016#issuecomment-586032019 for the detail

**Notes**
* My platform I'm testing on is [Windows 10 + WSL Ubuntu](https://github.com/google/model-viewer/issues/1021#issuecomment-585985485). It seems `npm run test` couldn't run correctly, it stops during the test without any messages. I'm happy if someone runs the test on my behalf.
* I'm very new to &lt;model-viewer&gt; inside. Especially I don't see out of WebXR AR code so much then I'm not really sure if this solution is best. Feedback is very welcome.

### DIFF
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -293,7 +293,9 @@ export class ARRenderer extends EventDispatcher {
       return;
     }
 
-    this.scene.environment = this[$presentedScene]!.environment;
+    if (this.scene.environment !== this[$presentedScene]!.environment) {
+      this.scene.environment = this[$presentedScene]!.environment;
+    }
 
     for (const view of frame.getViewerPose(this[$refSpace]!).views) {
       const viewport = session.renderState.baseLayer!.getViewport(view);
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -293,6 +293,8 @@ export class ARRenderer extends EventDispatcher {
       return;
     }
 
+    this.scene.environment = this[$presentedScene]!.environment;
+
     for (const view of frame.getViewerPose(this[$refSpace]!).views) {
       const viewport = session.renderState.baseLayer!.getViewport(view);
       this.threeRenderer.setViewport(
@@ -318,5 +320,7 @@ export class ARRenderer extends EventDispatcher {
       // this.threeRenderer.clearDepth();
       this.threeRenderer.render(this.scene, this.camera);
     }
+
+    this.scene.environment = null;
   }
 }
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -203,6 +203,7 @@ export class ARRenderer extends EventDispatcher {
 
     this[$refSpace] = null;
     this[$presentedScene] = null;
+    this.scene.environment = null;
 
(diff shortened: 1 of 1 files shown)


cdata:
@takahirox thanks for the contribution! This fix seems straight-forward enough. I will test it out locally and let you know if I see any issues.

cdata:
Also, don't worry about running the tests yourself. Our CI builds will take care of that for you 👍

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -293,6 +293,8 @@ export class ARRenderer extends EventDispatcher {
       return;
     }
 
+    this.scene.environment = this[$presentedScene]!.environment;
```
takahirox: It seems [`environment` texture (skybox-image attribute) can be dynamically changed anytime](https://modelviewer.dev/examples/lighting-and-environment.html) so I speculate that copying `.environment` each frame in `$onWebXRFrame()` is safer, rather than doing in `present()` at that `presentedScene` is set. If we do in `present()`, changing environment (skybox-image/) attribute during AR mode won't have an effect. Please let me know if there is any other best place where we should copy `.environment`.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -318,5 +320,7 @@ export class ARRenderer extends EventDispatcher {
       // this.threeRenderer.clearDepth();
       this.threeRenderer.render(this.scene, this.camera);
     }
+
+    this.scene.environment = null;
```
takahirox: I think clearing `scene.environment` is safer to prevent memory leak.

packages/model-viewer/src/three-components/ARRenderer.ts
cdata: I think doing it here is fine. I think for defensive purposes, we should probably check that the environment has changed e.g.,

```typescript
if (this.scene.environment !== this[$presentedScene].environment) {
  this.scene.environment = this[$presentedScene]!.environment;
}
```

That way we don't risk accidentally suggesting to Three.js that the environment is changing in cases where it actually isn't.

cdata:
This is working great for me locally. Thanks for all the investigation and contributing the fix @takahirox 🍻

packages/model-viewer/src/three-components/ARRenderer.ts
takahirox: Thanks for the comment. Updated.

BTW just in case, I release the reference to environment texture at the end of each frame, so the check is seeing whether `presentedScene.environment` is `null` rather than seeing whether the environment has changed since the previous frame.

packages/model-viewer/src/three-components/ARRenderer.ts
cdata: I think it might be better to put this in the `[$postSessionCleanup]` method. WDYT?

https://github.com/google/model-viewer/blob/e835eb3bef53a30ae9af1f3f4f43a179f80b5e37/packages/model-viewer/src/three-components/ARRenderer.ts#L186

takahirox: I think that sounds reasonable. I'll update tomorrow.

takahirox: Updated, thanks

cdata:
Thanks for the contribution @takahirox !

PR #1040, opened 2020-02-20
Update index.html

mkeblx:
Latest versions of stable Samsung Internet have FS API support.


cdata:
Thanks for the contribution @mkeblx

PR #1042, opened 2020-02-21
Fix resetting shadow intensity to 0

### LINKED ISSUE #1032: Shadow does not disappear when shadow-intensity property set to zero
The model-viewer documentation for property `shadow-intensity` states that "
Controls the opacity of the shadow. Set to 0 to turn off the shadow entirely. Defaults to 0.",  but even when set to zero, the shadow does not disappear.

cdata:
We had a small bug that prevented shadow intensity from being reset to 0 after being increased from 0.

Fixes #1032

### DIFF
--- a/packages/model-viewer/src/test/three-components/ModelScene-spec.ts
+++ b/packages/model-viewer/src/test/three-components/ModelScene-spec.ts
@@ -55,6 +55,22 @@ suite('ModelScene', () => {
     });
   });
 
+  suite('with a model', () => {
+    setup(async () => {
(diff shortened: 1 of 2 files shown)


elalish:
@cdata Good catch, but I'm wondering if this is the best solution. The reason that check was there in the first place was so that we don't pay the cost of rendering a shadow unless we need one. I believe now the shadow is being rendered always, even when invisible.

cdata:
PRs welcome 😁

PR #1046, opened 2020-02-24
Upgrade Filament and use its TypeScript bindings.

prideout:
This PR upgrades the fidelity tests to use Filament's latest release candidate. Does not update golden images.


packages/render-fidelity-tools/README.md
```diff
@@ -25,4 +25,7 @@ Command                         | Description
 `npm run build`                 | Build artifacts required to run testing and review tools
 `npm run clean`                 | Deletes all build artifacts
 `npm run test`                  | Run a render fidelity check comparing `<model-viewer>` to other renderers
-`npm run update-screenshots`    | Update the "golden" screenshots for all renderers
\ No newline at end of file
+`npm run update-screenshots`    | Update the "golden" screenshots for all renderers
+
+After running `npm run test`, you can look at the results by starting a local web server
+(e.g. `npx http-server`) in this folder, then opening `test/results-viewer.html` in a browser.
```
cdata: 👍

packages/render-fidelity-tools/package.json
```diff
@@ -31,9 +31,10 @@
     "@types/resize-observer-browser": "^0.1.2",
     "@types/rimraf": "^2.0.3",
     "@webcomponents/webcomponentsjs": "~2.1.3",
-    "filament": "1.4.5",
+    "filament": "^1.5.0-rc.1",
```
cdata: How long until you folks graduate this package from `rc`?

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -14,6 +14,7 @@
  */
 
 import {resolveDpr} from '@google/model-viewer/lib/utilities.js';
+import * as Filament from 'filament';
```
cdata: We tend to avoid blanket imports like this because it can make tree-shaking/code-splitting less efficient. Would you be willing to change this to explicitly named imports instead (e.g., `import {Engine, Scene, Renderer, ...} from 'filament'`)?

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -180,7 +179,8 @@ export class FilamentViewer extends LitElement {
     // environment map with a single bright pixel that represents a 1 lux
     // directional light.
     if (lightingBaseName === 'spot1Lux') {
-      this[$directionalLight] = self.Filament.EntityManager.get().create();
+      const entity = self.Filament.EntityManager.get().create();
```
cdata: An explicit variable name (e.g., `directionalLight`) would make this easier to read

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -238,9 +239,7 @@ export class FilamentViewer extends LitElement {
   private[$render]() {
     this[$rendering] = true;
 
-    if (this[$renderer] != null) {
-      this[$renderer].render(this[$swapChain], this[$view]);
-    }
+    this[$renderer]?.render(this[$swapChain], this[$view]);
```
cdata: We haven't set up our linter for this yet, but please prefer an explicit `null` check over optional chaining (currently `clang-format` does bad things to optional chaining expressions; I'm actually surprised that our pre-commit hook didn't break this for you).

cdata:
BTW, did you try out the Filament renderer after this update? Even if we don't update the goldens in this change, it would be worth _trying_ to update them to see what you get...

prideout:
> BTW, did you try out the Filament renderer after this update? Even if we don't update the goldens in this change, it would be worth _trying_ to update them to see what you get...

Yep, at first glance it seemed like the Filament renderings did not change significantly, even though our 1.5.0 release has some significant changes, including support for bloom and screens-space refraction.  Maybe I should update the goldens in a separate PR?

packages/render-fidelity-tools/package.json
prideout: Probably in a week or two, we were going to wait until we land our upcoming variance shadow maps feature... unless something comes up that motivates an earlier release.

prideout:
Thanks Chris, I'll fix this up based on your review comments. I also forgot to remove `src/types/filament.ts` (!)

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
prideout: Done!

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
prideout: Done!

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
prideout: This was tricky but I got it to work!

packages/render-fidelity-tools/package.json
```diff
@@ -31,9 +31,10 @@
     "@types/resize-observer-browser": "^0.1.2",
     "@types/rimraf": "^2.0.3",
     "@webcomponents/webcomponentsjs": "~2.1.3",
-    "filament": "1.4.5",
+    "filament": "^1.5.0-rc.2",
     "focus-visible": "^5.0.1",
     "fullscreen-polyfill": "^1.0.2",
+    "gl-matrix": "^3.2.1",
```
cdata: Is this dependency being used anywhere? I wasn't able to find it in the diff..

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -14,13 +14,18 @@
  */
 
 import {resolveDpr} from '@google/model-viewer/lib/utilities.js';
+import {Camera, Engine, IndirectLight, Renderer, Scene, SwapChain, View} from 'filament';
+import {Entity, EntityManager, gltfio$FilamentAsset, Skybox} from 'filament';
+import {Camera$Fov, LightManager, LightManager$Type} from 'filament';
+import {Aabb, init} from 'filament';
+import {fetch as filamentFetch} from 'filament';
```
cdata: Let's merge all of these into a single import. I know it's kind of a mess with all of these individual imports, but it's more consistent with what we do for e.g., `three`.

packages/render-fidelity-tools/package.json
prideout: Good catch, this is an indirect dependency so should not be listed here.  Fixed.

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
prideout: Fixed.

PR #1055, opened 2020-02-27
fix shadow lazy creation

elalish:
This ensures a shadow in not created until needed, fixing a small regression from #1042.


packages/model-viewer/src/test/three-components/ModelScene-spec.ts
```diff
@@ -68,6 +68,12 @@ suite('ModelScene', () => {
         scene.setShadowIntensity(0);
         expect(scene.shadow!.getIntensity()).to.be.equal(0);
       });
+
+      test('shadow is only created when intensity is greater than zero', () => {
```
cdata: 💯

PR #1063, opened 2020-02-28
Express material names in the 3DOM scene graph

### LINKED ISSUE #1062: 3DOM material names are not being expressed
When accessing materials in a 3DOM scene graph worker, names are not readable in the public API for `Material`. Names need to be expressed in the public API for `Material`.

cdata:
Fixes #1062

### DIFF
--- a/packages/3dom/src/context.ts
+++ b/packages/3dom/src/context.ts
@@ -161,7 +161,6 @@ export class ThreeDOMExecutionContext extends EventTarget {
     const port = await this[$workerInitializes];
     const {port1, port2} = new MessageChannel();
 
-    console.log(modelGraft?.model.toJSON());
     port.postMessage(
         {
           type: ThreeDOMMessageType.MODEL_CHANGE,
--- a/packages/3dom/src/context.ts
+++ b/packages/3dom/src/context.ts
@@ -161,6 +161,7 @@ export class ThreeDOMExecutionContext extends EventTarget {
     const port = await this[$workerInitializes];
     const {port1, port2} = new MessageChannel();
 
+    console.log(modelGraft?.model.toJSON());
     port.postMessage(
         {
           type: ThreeDOMMessageType.MODEL_CHANGE,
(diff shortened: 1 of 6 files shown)


packages/3dom/src/context.ts
```diff
@@ -161,6 +161,7 @@ export class ThreeDOMExecutionContext extends EventTarget {
     const port = await this[$workerInitializes];
     const {port1, port2} = new MessageChannel();
 
+    console.log(modelGraft?.model.toJSON());
```
elalish: Sure you want to keep this?

packages/3dom/src/end-to-end-spec.ts
```diff
@@ -71,4 +71,27 @@ suite('end-to-end', () => {
...
+    executionContext.changeModel(graft);
+
+    const messageEventArrives =
+        waitForEvent<MessageEvent>(executionContext.worker, 'message');
+    executionContext.eval('self.postMessage(model.materials[0].name)');
+
+    const messageEvent = await messageEventArrives;
+    console.log(messageEvent.data);
+
+    expect(messageEvent.data).to.be.equal(material.name);
```
elalish: Does the astronaut have a material name, or are we just checking that empty string equals empty string?

packages/3dom/src/facade/three-js/three-dom-element.ts
```diff
@@ -71,12 +71,17 @@ export class ThreeDOMElement implements ThreeDOMElementInterface {
...
-          null;
+
+    // NOTE: Some Three.js object names are modified from the names found in the
+    // glTF. Special casing is handled here, but might be better moved to
+    // subclasses down the road:
+    if ((relatedObject as Material).isMaterial) {
+      // Material names can be safely referenced directly from the Three.js
+      // object.
+      // @see: https://github.com/mrdoob/three.js/blob/790811db742ea9d7c54fe28f83865d7576f14134/examples/js/loaders/GLTFLoader.js#L2162
+      return (relatedObject as Material).name;
```
elalish: :+1:

packages/3dom/src/context.ts
cdata: Thanks, will fix

packages/3dom/src/end-to-end-spec.ts
cdata: It has a name, I checked... `Astronaut_mat`, I think

elalish: SG; you could add a check that the name is not empty here in the test as documentation.

PR #1070, opened 2020-03-03
Fix up WebXR

### LINKED ISSUE #582: Fix "ModelViewerElementBase with StagingMixin » with a loaded model » auto-rotate » pauses rotate after user interaction"

### LINKED ISSUE #583: Fix "ModelViewerElementBase with ControlsMixin » when registered » camera-orbit » can independently adjust azimuth"

### LINKED ISSUE #1057: Animation doesn't work in WebXR
I tried putting the `unstable-webxr` attribute on our animation samples and found the model is rendered statically in AR mode. The problem is the logic we have in our Renderer class regarding animation, exposure, and shadows, which does not exist in the ARRenderer class. Looks like we need a bit of refactoring to share this code.

elalish:
Fixes #1057 
Fixes #582 
Fixes #583 

In addition to fixing the above and several related bugs, this also enables hotspots and annotations within WebXR if you flip the WebXR Incubations flag in Chrome Canary. We're still working through some bugs there, both here and in Chrome.

I've also refactored to simplify ARRenderer by using our pivot directly instead of an additional dolly and to remove the extra scene. You can verify that animation and shadows are fixed in WebXR, as I've added the button to our animation doc examples. Once WebXR is default, should we add the button to all our docs pages?


packages/model-viewer/src/features/annotation/hotspot.ts
```diff
@@ -142,6 +145,10 @@ export class Hotspot extends CSS2DObject {
     }
   }
 
+  orient(radians: number) {
```
cdata: Is there a test case covering the behavior for this?

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -191,10 +209,20 @@ export class ARRenderer extends EventDispatcher {
...
     // necessary and sufficient?
-    if (this[$presentedScene] != null) {
-      this.dolly.remove(this[$presentedScene]!);
-      this[$presentedScene]!.isDirty = true;
+    const scene = this[$presentedScene];
+    if (scene != null) {
+      scene.setCamera(scene.camera);
+      scene.remove(this.reticle);
+      scene.pivot.visible = true;
+      (scene.element as any)[$setHotspotsVisibility](true);
```
cdata: Let's clean up this `any` cast.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -191,10 +209,20 @@ export class ARRenderer extends EventDispatcher {
...
+
+      scene.pivot.position.set(0, 0, 0);
+      scene.setPivotRotation(this[$turntableRotation]!);
+      scene.setShadowIntensity(this[$oldShadowIntensity]!);
+      (scene.element as any)[$orientHotspots](0);
```
cdata: Let's clean up this `any` cast.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -137,22 +143,34 @@ export class ARRenderer extends EventDispatcher {
       console.warn('Cannot present while a model is already presenting');
     }
 
-    scene.model.scale.set(1, 1, 1);
+    scene.setCamera(this.camera);
```
cdata: There is a lot of AR-specific state that has to be tracked on a shared object (the `scene`). Is there a way for us to make these state changes more easily reversible? Maybe we need a different `scene` object specifically for AR?

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -230,17 +253,19 @@ export class ARRenderer extends EventDispatcher {
...
       const camPosition = vector3.setFromMatrixPosition(this.camera.matrix);
-      this.dolly.lookAt(camPosition.x, this.dolly.position.y, camPosition.z);
-      this.dolly.rotateY(-presentedScene.pivot.rotation.y);
+      pivot.lookAt(camPosition.x, pivot.position.y, camPosition.z);
+      pivot.updateMatrixWorld();
+      shadow!.setRotation(pivot.rotation.y);
 
-      this.dolly.add(presentedScene);
+      pivot.visible = true;
+      (element as any)[$setHotspotsVisibility](true);
```
cdata: Let's clean up this `any` cast

packages/model-viewer/src/features/annotation.ts
```diff
@@ -156,40 +165,63 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-            hotspot.hide();
-          } else {
-            hotspot.show();
-          }
+          func(hotspot);
         }
       }
     }
 
+    [$updateHotspots]() {
```
cdata: Let's make sure that we keep implementation like this `private` or `protected`.

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -207,6 +207,23 @@ export class Renderer extends EventDispatcher {
     return this[$arRenderer] != null && this[$arRenderer].isPresenting;
   }
 
+  preRender(scene: ModelScene, t: number, delta: number) {
```
cdata: What is the intended purpose of this new public API? (please add docs)

packages/model-viewer/src/three-components/Reticle.ts
```diff
@@ -53,28 +53,31 @@ export default class Reticle extends Object3D {
     return this._hitMatrix;
   }
 
+  reset() {
```
cdata: What is the intended purpose of this new public API? (please add docs)

packages/model-viewer/src/features/annotation.ts
elalish: I'm happy to follow our conventions, but I think in this case I'm not quite sure what they are. Looking at our other Mixins, we don't use `private` for Symbol methods very much (though we do occasionally). Perhaps I just missed what the rule was, or are we trying to update them to a new standard?

cdata: We may not have consistently applied standards, but I encourage you to use your judgement as a software engineer and apply the appropriate API access level. The rules, to the extent that we have any, are outlined in https://github.com/google/model-viewer/wiki/Code-conventions#api-access

packages/model-viewer/src/features/annotation/hotspot.ts
elalish: No, it seemed like putting one on hotspot would be pretty tautological. I'd like to have one testing AR, but I'm not sure we have it sufficiently mocked out yet. What would you like to see?

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: That's the way it was before; that had symmetric problems with not bringing state along, like lighting, hotspots, pivot. I think overall it's better this way, since we just make the changes in present() and $postSessionCleanup().

packages/model-viewer/src/three-components/Renderer.ts
elalish: Done.

packages/model-viewer/src/three-components/Reticle.ts
elalish: Done.

elalish:
@cdata I refactored the hotspot control logic out of AnnotationMixin and into ModelScene. I also moved hotspot.ts from features to three-components. You're right, this is a better layout and addresses `any` casts and circular dependencies.

packages/model-viewer/karma.conf.js
```diff
@@ -160,6 +160,18 @@ module.exports = function(config) {
         // domain so we go there directly instead:
         url: 'http://bs-local.com:9876'
       },
+      'iOS Safari (iOS 12)': {
```
cdata: Weird, I thought I put this in there 😵

packages/model-viewer/src/test/features/controls-spec.ts
```diff
@@ -110,8 +110,7 @@ suite('ModelViewerElementBase with ControlsMixin', () => {
         expect(element.getCameraOrbit().radius).to.be.equal(defaultRadius);
       });
 
-      // TODO(#583)
-      test.skip('can independently adjust azimuth', async () => {
+      test('can independently adjust azimuth', async () => {
```
cdata: 💯

packages/model-viewer/src/test/features/staging-spec.ts
```diff
@@ -44,11 +43,6 @@ suite('ModelViewerElementBase with StagingMixin', () => {
   BasicSpecTemplate(() => ModelViewerElement, () => tagName);
 
   suite('with a loaded model', () => {
-    if (IS_IE11) {
```
cdata: 💯

packages/model-viewer/karma.conf.js
elalish: I removed it and put it back in this PR; you're not crazy.

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -271,4 +278,103 @@ export class ModelScene extends Scene {
...
+
+  /**
+   * This method returns the world position and normal of the point on the
+   * mesh corresponding to the input pixel coordinates given relative to the
+   * model-viewer element. The position and normal are returned as strings in
+   * the format suitable for putting in a hotspot's data-position and
+   * data-normal attributes. If the mesh is not hit, position returns the
+   * empty string.
+   */
+  positionAndNormalFromPoint(pixelX: number, pixelY: number):
```
cdata: Use your judgement, but I would recommend adding tests for all of this new API

elalish: This method isn't new, it just moved; it's still tested by the annotation mixin. The new functions are very simple, so it would be more valuable to test them end-to-end. Once I get WebXR up to snuff, I'll look at enabling an Android device and then remake the AR tests to actually run and test these things.

elalish:
@cdata If you want to review this you're welcome to, but I plan to merge it tomorrow morning. Now that we have Karma and debugging is actually possible, I decided to dig into these flaky staging tests. There were some legit problems, which became more reproducible once I lowered the delay time. I believe I have removed the flakiness (and a couple files worth of code), and all without having to skip any tests!

This took awhile because I was getting used to our new debugging and testing process, but I'm quite happy now that I understand it better. Definitely a major improvement, so thanks for that!

christianpugger:
do annotations work in AR? If you apple and android?

thank you

elalish:
@christianpugger In our latest release, v0.10.0, annotations do work in AR, but only `webxr` mode and they require Chrome 83, which is currently Beta. You can try them out here on an ARCore-capable Android phone: https://modelviewer.dev/examples/annotations.html

There is not yet support for annotations in SceneViewer or QuickLook, and they are unlikely to support full HTML ever. WebXR is not supported on iOS for any browser.

christianpugger:
> @christianpugger In our latest release, v0.10.0, annotations do work in AR, but only `webxr` mode and they require Chrome 83, which is currently Beta. You can try them out here on an ARCore-capable Android phone: https://modelviewer.dev/examples/annotations.html
> 
> There is not yet support for annotations in SceneViewer or QuickLook, and they are unlikely to support full HTML ever. WebXR is not supported on iOS for any browser.

too bad, I wish I could use the AR annotations on my table.

But the same limit is for animations?

elalish:
@christianpugger Yes, please bug Apple about supporting WebXR! Once they do, annotations will work in AR everywhere. I believe animations are supported in all of our AR platforms; do you have a model with animations that aren't working?

christianpugger:
> @christianpugger Yes, please bug Apple about supporting WebXR! Once they do, annotations will work in AR everywhere. I believe animations are supported in all of our AR platforms; do you have a model with animations that aren't working?

for a client I have to evaluate whether to use annotations or animations, so I was asking if and which worked in AR and not only in webxr

PR #1072, opened 2020-03-03
Reduce to unique materials

### LINKED ISSUE #1061: 3DOM materials are not reduced to sparse set
Currently if you enumerate the materials on a model, you will see duplicates of a given material if it occurs multiple times in a scene. The top-level set of materials should only include the reduction of unique materials in the scene.

cdata:
This started out as a fix for the problem where 3DOM scene graphs expressed a unique material for every location in the scene graph where a material occurred. It turned out that this was an implementation problem caused in `CachingGLTFLoader` and related to how models are cloned. So, to address the issue I made a light refactor of the related code (which was overdue anyway). Some highlights from the change:

 - Refactored `CachingGLTFLoader` and `cloneGltf`
   - Separate loading/caching from model preparation/cloning
   - Introduced `GLTFInstance`, a conveniently clone-able, dispose-able, API-compatible alternative to `GLTF`
   - `CachingGLTFLoader` now loads a full GLTF-like, rather than just a scene
   - `RoughnessMipmapper` is created once, rather than per-instance of `CachingGLTFLoader`
   - Ensure that discrete materials are only cloned once

Depends on #1063 

Fixes #1061

### DIFF
--- a/packages/model-viewer/src/three-components/Model.ts
+++ b/packages/model-viewer/src/three-components/Model.ts
@@ -13,13 +13,14 @@
  * limitations under the License.
  */
 
-import {AnimationAction, AnimationClip, AnimationMixer, Box3, Object3D, Scene, Vector3} from 'three';
+import {AnimationAction, AnimationClip, AnimationMixer, Box3, Object3D, Vector3} from 'three';
 
-import {$releaseFromCache, CacheRetainedScene, CachingGLTFLoader} from './CachingGLTFLoader.js';
+import {CachingGLTFLoader} from './CachingGLTFLoader.js';
+import {ModelViewerGLTFInstance} from './gltf-instance/ModelViewerGLTFInstance.js';
 import {moveChildren, reduceVertices} from './ModelUtils.js';
 
 const $cancelPendingSourceChange = Symbol('cancelPendingSourceChange');
-const $currentScene = Symbol('currentScene');
+const $currentGLTF = Symbol('currentGLTF');
 
 export const DEFAULT_FOV_DEG = 45;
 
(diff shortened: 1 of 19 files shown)


packages/model-viewer/src/three-components/Model.ts
```diff
@@ -101,41 +102,36 @@ export default class Model extends Object3D {
...
-    this[$currentScene] = scene as CacheRetainedScene;
+    this[$currentGLTF] = gltf;
 
-    if (scene != null) {
-      moveChildren(scene, this.modelContainer);
+    if (gltf != null) {
+      moveChildren(gltf.scene, this.modelContainer);
     }
 
-    this.modelContainer.traverse(obj => {
```
cdata: This was duplicate work, and now consolidated in `ModelViewerGLTFInstance` implementation.

packages/model-viewer/src/three-components/Model.ts
```diff
@@ -231,11 +227,12 @@ export default class Model extends Object3D {
...
     this.url = null;
     this.userData = {url: null};
+    const gltf = this[$currentGLTF];
     // Remove all current children
-    if (this[$currentScene] != null) {
-      moveChildren(this.modelContainer, this[$currentScene]!);
-      this[$currentScene]![$releaseFromCache]();
-      this[$currentScene] = null;
+    if (gltf != null) {
+      moveChildren(this.modelContainer, gltf.scene);
```
cdata: Moving children around like this is probably not necessary; Three.js allows a `Scene` to be a child of another `Object3D`, so we could probably just take advantage of that and insert the `gltf.scene` directly into the graph. Punting for the future, though.

PR #1073, opened 2020-03-04
filament-renderer: bump version and disable bloom

prideout:
I don't see any significant differences.


prideout:
Maybe I should trim this PR to include only the first commit, since the second is a no-op.

elalish:
@prideout If you've verified that, that's probably a good idea.

prideout:
Done.

prideout:
The FIlament team has released 1.5.0, 1.5.1, and 1.5.2 in a slew so I've updated this PR to 1.5.2.

NOTE: while examining screenshots I noticed that the model-viewer screenshot of `SimpleMorph` needs to be updated, but the FIlament screenshot does not need to be updated.  This seems unrelated to this PR so should perhaps be updated separately.

prideout:
I've now reviewed the screenshots so I'll mark this as ready for review.  As previously noted, the filament-specific screenshots have remained as-is.

Issue #691, opened 2019-08-05
Re-enable Android tests via Browserstack


cdata:
This is straight-forward now since we migrated our tests to Browserstack already.

elalish:
This was fixed by #2955

PR #1075, opened 2020-03-05
Use Karma/BrowserStack to run <model-viewer> tests

cdata:
This change grew substantially out of scope, but one thing led to another and now our test toolchain and CI pipeline is totally different ✨

This change updates our pipeline to run in Github Actions instead of Travis. This is good because it brings us closer to Github's well-tread cowpaths, and moves us away from Travis which seems to be more-or-less frozen in time feature/support-wise (probably because of [their recent sale](https://news.ycombinator.com/item?id=18978251)).

Secondly, this change moves us from SauceLabs to BrowserStack for our x-browser testing needs. This has generally yielded an increase in overall performance for our x-browser tests, as well as improved reliability for iOS Safari tests and unlocks the potential to re-introduce our Android Chrome tests in the future. We might even get crazy and add tests for Samsung Internet someday. Future work on this tracked in #691 .

Thirdly, this change moves our test toolchain from WCT (which is [deprecated](https://github.com/Polymer/tools/issues/3493)) to Karma (which has an uncertain future, but is way better than WCT and [has active support for now](https://github.com/karma-runner/karma/pulse/monthly)). Thankfully, only minimal changes were required to our test suites to support this shift. Most of the time spent on this change went to ensuring that the combination of Karma, BrowserStack and Github Actions all worked in harmony and as intended.

And the coup de grâce: `npm run dev` does something useful for the `@google/model-viewer` package now (more or less what it does for `@google/3dom`), so running all the build tools simultaneously in a single terminal just got easier.

A final note on Travis: this PR continues to run Travis builds even though the Travis config has been removed. Travis has to be manually turned off for our repository before those status lines will go away. I will do that after this PR is merged.

An example 3P PR has been set up here to demonstrate how the build changes: https://github.com/google/model-viewer/pull/1082


.github/workflows/unit-tests-legacy.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Generate legacy bundles
+      run: npm run build:legacy-support
+
+    - name: Unit tests
+      env:
+        BROWSER_STACK_USERNAME: ${{ secrets.BROWSER_STACK_USERNAME }}
+        BROWSER_STACK_ACCESS_KEY: ${{ secrets.BROWSER_STACK_ACCESS_KEY }}
+      run: |
+        # Don't run Sauce from a forked repo, since the code may not be trusted
```
elalish: If this is still true, we should probably at least update that it's not Sauce anymore.

.github/workflows/unit-tests.yml
```diff
@@ -39,15 +39,13 @@ jobs:
 
     - name: Unit tests
       env:
-        SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
-        SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
+        BROWSER_STACK_USERNAME: ${{ secrets.BROWSER_STACK_USERNAME }}
+        BROWSER_STACK_ACCESS_KEY: ${{ secrets.BROWSER_STACK_ACCESS_KEY }}
       run: |
         # Don't run Sauce from a forked repo, since the code may not be trusted
```
elalish: Same as above; maybe just do a search for 'Sauce'.

.travis.yml
```diff
@@ -26,7 +26,7 @@ install:
 - npm run bootstrap
 - npm run build
 script:
-- npm run test:ci
+- npm run wct-local-chrome
```
elalish: WCT is gone now, right?

packages/3dom/karma.conf.js
```diff
@@ -60,65 +60,117 @@ module.exports = function(config) {
...
+        const browser = args[2];
+        const query = config.url.split('?')[1];
+        browser.url = `${browser.url}?${query}`;
+      }
+      return assign.apply(this, args);
+    };
+    // Something in Karma deps actually asserts the sub-keys of Object.assign,
+    // so make sure to copy those over too:
+    assign.call(Object, newAssign, assign);
+    Object.assign = newAssign;
```
elalish: I love JavaScript! 😈

packages/model-viewer/karma.conf.js
```diff
@@ -0,0 +1,206 @@
...
+        os: 'iOS',
+        os_version: '12',
+        device: 'iPhone 8',
+        browser: 'iPhone',
+        real_mobile: 'true',
+        // BrowserStack seems to drop the port when redirecting to this special
+        // domain so we go there directly instead:
+        url: 'http://bs-local.com:9876'
+      },
+      'iOS Safari (iOS 12)': {
```
elalish: Should we add an android device someday? Not saying today is that day...

packages/model-viewer/package.json
```diff
@@ -76,35 +73,34 @@
...
     "@types/pngjs": "^3.4.0",
     "@ungap/event-target": "^0.1.0",
     "@webcomponents/webcomponentsjs": "~2.1.3",
     "chai": "^4.1.2",
-    "chokidar-cli": "^1.2.1",
-    "concurrently": "^4.0.1",
-    "cross-os": "^1.3.0",
     "focus-visible": "^5.0.1",
     "fullscreen-polyfill": "^1.0.2",
+    "http-server": "^0.12.1",
```
elalish: Better than local-web-server?

packages/model-viewer/package.json
```diff
@@ -76,35 +73,34 @@
...
     "intersection-observer": "^0.5.1",
     "lit-html": "^1.1.2",
-    "local-web-server": "^2.6.0",
-    "make-dir": "^1.3.0",
     "mocha": "^5.2.0",
```
elalish: Does karma-mocha not include mocha?

packages/model-viewer/scripts/ci-run-tests.sh
```diff
@@ -20,11 +20,8 @@ set -x
 
 if [ "${TEST_TYPE}" = "unit" ]; then
 
-  npm run test
+  npm run wct-local-chrome
 
-  if  [[ "${TRAVIS_PULL_REQUEST}" = "false" && "${TRAVIS_BRANCH}" != "master" ]]; then
```
elalish: Does this mean third parties start browserStack tests, or did this logic end up elsewhere?

packages/model-viewer/src/test/decorators-spec.ts
```diff
@@ -50,7 +50,7 @@ suite('decorators', () => {
       customElements.define(tagName, class extends StyleableElement {});
 
       element = document.createElement(tagName) as StyleableElement;
-      document.body.appendChild(element);
+      document.body.insertBefore(element, document.body.firstChild);
```
elalish: What's all this about? Something browser-specific?

packages/model-viewer/src/test/model-viewer-base-spec.ts
```diff
@@ -338,10 +338,10 @@ suite('ModelViewerElementBase', () => {
 
       test('sets a model within viewport to be visible', async () => {
         await until(() => {
-          return elements[0][$scene].visible;
+          return elements[2][$scene].visible;
```
elalish: Why this change?

packages/model-viewer/karma.conf.js
cdata: Yes, let's make it another day though. Tracked in #691

packages/model-viewer/package.json
cdata: Yes, and consistent with the other packages

packages/model-viewer/package.json
cdata: It does not, it manages the integration between the two.

packages/model-viewer/scripts/ci-run-tests.sh
cdata: This got moved elsewhere

packages/model-viewer/src/test/decorators-spec.ts
cdata: Now that we are running in vanilla mocha, without additional intervention all of our elements get appended after the Mocha test results. This ensures that our test elements are inserted before those Mocha DOM elements.

packages/model-viewer/src/test/model-viewer-base-spec.ts
cdata: Because we reversed the order that the elements are inserted to the document.

cdata:
I'm happy to share that "full" unit test runs (including Chrome, 2 x Firefox, 2 x Desktop Safari, 2 x MS Edge, 2 x iOS Safari, and IE11 across both @google/model-viewer and @google/3dom) are finishing in ~8 minutes. Unit tests and fidelity tests both complete in under 10 minutes. This is a 55-65% reduction in test time compared to our Travis/WCT/Sauce builds ([example build](https://travis-ci.org/github/google/model-viewer/builds/661719760)).

packages/model-viewer/scripts/ci-run-tests.sh
```diff
@@ -20,11 +20,8 @@ set -x
 
 if [ "${TEST_TYPE}" = "unit" ]; then
 
-  npm run test
+  npm run wct-local-chrome
```
elalish: Still confused by this name; are we really still using WCT?

elalish: Oh wait, we don't actually need this file anymore, right? Since we're not using Travis anymore?

cdata: Good catch, this script is vestigial and has been removed 👍

PR #1077, opened 2020-03-10
Refactor visibility

### LINKED ISSUE #773: Re-factor / re-define modelIsVisible
The concept of "visibility" was overloaded in the DOM before `<model-viewer>` was invented, but despite this we have been somewhat careless when using that term in our public. `modelIsVisible` was intended to describe cases when the model is considered renderable. However, it does not address the technical nuances involved in our context. For example, the `<model-viewer>` may report `modelIsVisible` when the model is not being rendered if the model is fully loaded and the `<model-viewer>` is not present in the viewport.

This issue asks that we carefully define what we mean by "visibility" here, and redesign this API to avoid the potential confusion that arises from the overloaded concept of "visibility."

elalish:
Fixes #773 

Most of this change is just internal renaming, to avoid overloaded terms like visible and render where possible. The big change to the external API is that modelIsVisible and the model-visibility event are now tied to both the poster and the IntersectionObserver (if one exists). I believe the only behavior change is that models now only start the autorotate timer when they scroll into view, which I think is an improvement.


packages/model-viewer/src/test/features/annotation-spec.ts
```diff
@@ -132,11 +132,11 @@ suite('ModelViewerElementBase with AnnotationMixin', () => {
         let wrapper: HTMLElement;
 
         setup(async () => {
+          await rafPasses();
```
cdata: I'm sure I put this here in the first place, but can we leave a note as to its significance?

elalish: Done.

PR #1081, opened 2020-03-12
AR attributes update

### LINKED ISSUE #541: Simplify AR attributes
We now have several attributes that each individually control our AR experience:

* `ar` for Scene Viewer
* `ios-src` for Quick Look
* `magic-leap` for Magic Leap devices
* `unstable-webxr` for browsers that support the new WebXR API

Currently, a given platform/browser only supports a single AR experience - therefore, there's no overlap between these individual settings. Therefore we currently have the opportunity to simplify and consolidate on a single attribute (`ar`).

This means removing the other attributes (`ios-src`, `magic-leap`, and `unstable-webxr`), and gating those features by a mix of the `ar` attribute and other parameters like:
* included additional source (in the case of `magic-leap`) 
* other configuration (for `ios-src`, nested `source` or `src` as discussed in #429).

In the future a given platform may support several of these experiences, and a developer may want to express an opinion about which ones to prefer. However, as those become available we can assess if that's needed (or if there are always best options available).

This is a further refinement of the `ar` attribute, documented as part of #540. We'll want to do this after moving away from `ios-src` as part of #429.

### LINKED ISSUE #584: Fix "ModelViewerElementBase with ControlsMixin » when registered » camera-controls » a11y » does not prompt if user already interacted"

### LINKED ISSUE #838: Re-enable interaction prompt / loading test in IE11
In #834 , we disabled the following test for IE11 only: `does not prompt users to interact before a model is loaded`.

The reason for failure in IE11 is unclear. The core problem was that the test would run indefinitely, presumably because we are waiting for a state that is never reached towards the end of the test. However, when running in IE11 on local machines the test never seems to flake.

elalish:
Fixes #541 
Fixes #584
Fixes #838

As discussed, here I've removed the `unstable-webxr` attribute and introduced the `ar-modes` attribute, which defaults to 'webxr scene-viewer quick-look fallback', and represents a prioritized list of the types of AR experiences you want to enable. The `ar` attribute now globally turns them all on and off.

I've updated the docs based on Chrome 81 being released, which is scheduled for March 17th.


cdata:
This is blocked pending #1070 , since we cannot default to WebXR until suitably improved UX has landed.

elalish:
@cdata True, though I think the bulk of this PR is mergeable. It's a one-liner to remove (or add) webxr as a default. Then again, we could argue it's also blocked on Chrome 81, so maybe revisit next week?

elalish:
I've updated the docs and removed `webxr` from the default since Chrome 81's release has been delayed.

elalish:
Now that our multi browser tests are easier to debug, I dug into our test flakes some more. Managed to close a couple of issues and get this all working reliably.

PR #1087, opened 2020-03-13
Attempt to shard BrowserStack via project/build name

cdata:
Browser stack builds appear to be establishing overlapping tunnels. This may be due to misconfiguration on our part. I'm testing out amendments to our configuration in this PR.

We landed on forcing a unique tunnel ID into each `karma-browserstack-launcher` run. This ID is required to run BrowserStack tunnels in parallel, according to all documentation that I can find on the topic. This appears to work as described, but we'll need to keep our eyes on the build after landing this change just in case.

Original approach:

The new approach creates a distinct BrowserStack build (which, for the sake of pedantry, is different but related to the Github Actions build) for each group of connections we make during a Github Actions run. In theory, this will result in two BrowserStack builds for each full unit test run. You can see the difference between the original approach and this new approach here:

image

In the original approach, we weren't specifying a build name or a project name, and all BrowserStack browsers were lumped into the same build.

My naive hope is that now that the builds are split apart, individual builds will be isolated from each other.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -45,7 +45,7 @@ jobs:
       env:
         BROWSER_STACK_USERNAME: ${{ secrets.BROWSER_STACK_USERNAME }}
         BROWSER_STACK_ACCESS_KEY: ${{ secrets.BROWSER_STACK_ACCESS_KEY }}
-        BROWSER_STACK_BUILD_NAME: ${{ github.ref + '/' + github.run_id }}
+        BROWSER_STACK_BUILD_NAME: ${{ format('{0} / {1}', github.ref, github.run_id) }}
       run: npm run test:ci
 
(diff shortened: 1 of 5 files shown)


cdata:
Force push to kick off the build again

cdata:
No dice, unfortunately. You can still see browsers connecting across overlapping sessions:

image

cdata:
It appears as though BrowserStack tunnels require a unique ID in order to run in parallel (evidence here and here). Unfortunately, `karma-browserstack-launcher` does not support specifying a unique tunnel ID unless you run the tunnel separately. Luckily, its dependency `browserstack-local` does support configuring a unique ID. I have hacked up `karma-browserstack-launcher` in an attempt to forward this currently unsupported configuration.

cdata:
Force pushed to kick of a build

PR #1114, opened 2020-03-27
Update model animation and shadow in pause mode

### LINKED ISSUE #1113: .currentTime is ignored when paused
### Description

We have a couple of usecases where we
- are playing back synced audio with model-viewer animation
- are pausing model-viewer animations at specific points in time
- allow jumping to other points in time, both in play and pause mode

While it's possible to achieve I think I found a bug -
currentTime is ignored when model-viewer is currently paused. This means that to get the behaviour, we currently have to keep model-viewer playing all the time and run a requestAnimationFrame to force it to a specific time. This doesn't feel right and makes this more complicated than it should be.

Note that the Live Demo has some other interesting things we found for audio support - so this is related to
#425, https://github.com/google/model-viewer/issues/483#issuecomment-579449685. 

Also note that autoplay on mobile is handled by browser policy (if you have often pressed play on a website it allows autoplay, if not then not) and is handled in this example by showing a screen overlay.

I started a PR for this but am unsure about some details, see here: #1114 

#### Live Demo
https://glitch.com/edit/#!/chatter-sleet-restaurant

As-is, this shows the broken behaviour - when you pause the audio, you can neither scrub the timeline nor press the buttons to jump to specific seconds, despite modelViewer currentTime being updated (and logged in the console).

If you comment in lines 103..114, this "fixes" it by forcing the currentTime.

soraryu:
I started with the obvious changes but unsure how best to implement a better check that we're really only updating the animation when needed.

Should we defer setting `this[$scene].model.animationTime` until the next tick (store it in a temporary variable when currentTime setter is called)?

### Reference Issue
Fixes #1113

### DIFF
--- a/packages/model-viewer/src/features/animation.ts
+++ b/packages/model-viewer/src/features/animation.ts
@@ -66,6 +66,7 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
 
     set currentTime(value: number) {
       this[$scene].model.animationTime = value;
+      this[$renderer].threeRenderer.shadowMap.needsUpdate = true;
     }
 
     pause() {
@@ -104,11 +105,7 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
       super[$tick](_time, delta);
 
       if (this[$paused]) { 
-        // TODO we should return here if currentTime has not actually changed
-        // should we defer setting this[$scene].model.animationTime until here
-        // (store it in a temporary variable when currentTime setter is called?)
-        // return; 
-        this[$renderer].threeRenderer.shadowMap.needsUpdate = true;
+        return;
       }
 
       const {model} = this[$scene];
--- a/packages/model-viewer/src/features/animation.ts
+++ b/packages/model-viewer/src/features/animation.ts
@@ -103,8 +103,12 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
     [$tick](_time: number, delta: number) {
       super[$tick](_time, delta);
 
-      if (this[$paused]) {
-        return;
+      if (this[$paused]) { 
+        // TODO we should return here if currentTime has not actually changed
+        // should we defer setting this[$scene].model.animationTime until here
+        // (store it in a temporary variable when currentTime setter is called?)
+        // return; 
+        this[$renderer].threeRenderer.shadowMap.needsUpdate = true;
       }
 
       const {model} = this[$scene];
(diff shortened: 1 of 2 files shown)


soraryu:
@googlebot I signed it!

packages/model-viewer/src/features/animation.ts
```diff
@@ -103,8 +103,12 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
       super[$tick](_time, delta);
 
-      if (this[$paused]) {
-        return;
+      if (this[$paused]) { 
+        // TODO we should return here if currentTime has not actually changed
+        // should we defer setting this[$scene].model.animationTime until here
+        // (store it in a temporary variable when currentTime setter is called?)
+        // return; 
+        this[$renderer].threeRenderer.shadowMap.needsUpdate = true;
```
elalish: I think you don't need any changes here. Instead, move this line up into the `set currentTime` function. Then [here](https://github.com/google/model-viewer/blob/007ae3b9405a2dabd9a4e37b3361c7cf6729285a/packages/model-viewer/src/three-components/Model.ts#L159) you'll want to add a call to this [method](https://threejs.org/docs/index.html#api/en/animation/AnimationMixer.setTime). Let me know if that helps.

soraryu: Thanks for pointing me directly to the right places :) I think what I wasn't aware is that setting the mixer will update the animation directly without the updateAnimation call.

packages/model-viewer/src/three-components/Model.ts
```diff
@@ -157,6 +157,7 @@ export default class Model extends Object3D {
   set animationTime(value: number) {
     if (this.currentAnimationAction != null) {
       this.currentAnimationAction.time = value;
```
elalish: I believe you can remove this line and the corresponding null check, since I think the method you added does this internally. Though, please test it to make sure I'm right :)

elalish:
Also, if you could write a test for this pausing bug, that would be great for preventing regressions. 🙏

soraryu:
I'll try! Good opportunity to learn a bit of model-viewer development and hopefully fix more of my own issues in the future.

I followed the set-up steps, however I'm hitting a blocker - any idea on how to proceed? Tried running the command in Powershell, Command Line, Git Bash and node.js command prompt but get the same error everywhere:
image

elalish:
@soraryu, thanks for all your work! Unfortunately none of us use a windows
development environment; we’re all Mac and Linux. It looks like our scripts
are not windows friendly. I know modern windows includes a Linux command
line; maybe that could help?

Sorry for not being as portable as we could be.

soraryu:
I'll try to do the changes inside a VM but it's gonna take a bit longer.
Thanks for the explanation!

elalish:
Closing in favor of #1139; Thanks for the contribution @soraryu !

PR #1135, opened 2020-04-08
update three.js to r115

### LINKED ISSUE #1043: Model Viewer crashing Google Chrome on some devices
### Description

We have implemented the Model Viewer on an e-commerce's website and some users (2 out of 10) are experiencing a weird bug where the browser seems to crash (black screen) and recover after a short while. The model does not appear after downloading, but we can see only the "drag to rotate" icon indicating that the 3D should've been there.

I am not sure how to reproduce it, since our QA have tested the same URL on multiple devices (using browserstack and crossbrowsertesting) and none of them had this problem.

#### Live Demo
- [Live demo](https://loja.electrolux.com.br/geladeira-french-door-electrolux-com-dispenser-de-agua-e-gelo-na-porta-538l-inox-dm85x/p) - click the "Visualizar 3D" button and wait for it to appear
- [Recording of the bug](https://drive.google.com/file/d/1fAIwNkdhoF5dukRE4j3BgEO-JIXHMg7c/view?usp=sharing)

### Browser Affected

- [x] Chrome - version 79.0.3945.130
- [x] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [x] Windows

### Versions

- model-viewer: v0.6.2
- three.js: v0.105.2

### LINKED ISSUE #1089: model viewer test page crashes Chrome, Edge
### Description

When I visit the page (https://modelviewer.dev/examples/tester.html), my screen flashes black and I see this:
image

My DXDIAG:
image

#### Live Demo
https://modelviewer.dev/examples/tester.html

### Browser Affected

- [x ] Chrome
- [x ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [x ] Windows

### Versions
Live deployed at: https://modelviewer.dev/examples/tester.html

elalish:
Fixes #1089 
Fixes #1043 

Actually, I have no idea if this fixes #1043, but it pulls in an upstream change in three.js that fixed #1089, and I'm just hoping they are the same bug. @aviggiano If you can find a way to verify this one way or the other, that would be fantastic. 

The only breaking change we got from this three.js upgrade is that a GLTF scene is now a Group instead of a Scene, and a Group is just an Object3D, annoyingly with an isGroup() method added that angers TS. I did a very quick pass to make TS happy and everything seems to run just fine, but I'm open to better typing if anyone has a good idea.


aviggiano:
@elalish thank you for the heads up, I'll check it out!

cdata:
Just a note that this is a breaking change for both `@google/model-viewer` and `@google/3dom`

cdata:
I think it's just pedantry (because `Group` is API-compatible with `Object3D`), but this bit should be updated: https://github.com/google/model-viewer/blob/master/packages/3dom/src/test-helpers.ts#L104-L116

packages/3dom/src/test-helpers.ts
```diff
@@ -102,7 +102,7 @@ export class FakeModelKernel implements ModelKernel {
 }
 
 export const createFakeGLTF = () => {
-  const scene = new Scene();
+  const scene = new Group();
```
elalish, line 105: I think I did that change here; or are you looking for something else?

cdata: Sorry, you're right. LGTM.

PR #1139, opened 2020-04-10
Fix current time when paused

### LINKED ISSUE #1113: .currentTime is ignored when paused

elalish:
Fixes #1113 

Thanks @soraryu, I'm taking over your #1114 so we can get it merged for this release. Let me know if any problems remain!

### DIFF
--- a/packages/model-viewer/src/features/animation.ts
+++ b/packages/model-viewer/src/features/animation.ts
@@ -66,6 +66,7 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
 
     set currentTime(value: number) {
       this[$scene].model.animationTime = value;
+      this[$renderer].threeRenderer.shadowMap.needsUpdate = true;
     }
 
     pause() {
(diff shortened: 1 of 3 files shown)


elalish:
Sorry @soraryu can you sign the CLA again for our googlebot?

soraryu:
@googlebot I consent

soraryu:
@googlebot I consent.

soraryu:
Thanks for taking this over. After trying multiple things (in a Mac VM, in a live Linux, through WSL, ...) without success for running the tests a fresh drive for properly setting up a Linux dev environment arrives today - so in the future I can hopefully actually follow through with my PRs.

elalish:
@soraryu Thanks, and sorry Windows isn't working as a dev environment. I appreciate your dedication!

PR #1142, opened 2020-04-13
Update SceneViewer intent

### LINKED ISSUE #1095: Scene Viewer doesn't load Cloudfront pre-signed urls
### Description
When i'm trying to view a model from **a signed cloudfront url**
Screenshot_20200324-123259_Google
It's Dutch for the famous **'Can't load object'** message.

I've created a scenario to test this.
[**The first url**](https://api.i-see-3d.com/viewer/test) displays the Astronaut model and loads the files from S3
[**The second url**](https://api.i-see-3d.com/viewer/5e79f0b02230e306a289a923) loads the exact same model only from cloudfront using a presigned URL

The first one loads happily in Scene Viewer, The second displays the message above.
The ThreeJS viewer loads everything just fine.

iOS seems to show a blank screen.

So i'm wondering if:
- You can't have query parameters in the file's url?
- Presigned is off limits?

### OS

- [ x ] Android
- [ x ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### Versions

- model-viewer: latest

### LINKED ISSUE #1097: Update to the latest Scene Viewer intent scheme
[Scene Viewer has an updated intent scheme](https://developers.google.com/ar/develop/java/scene-viewer#3d-or-ar). We should update to use it. In addition to being the likely path forward for configuring new Scene Viewer features, it appears to offer the ability to specify parameters like `title` and `link` without using the query string in the model URL.

### LINKED ISSUE #1143: Remove link and title from SceneViewer
I propose removing the optional link and title from our SceneViewer intent, as all they do is create a more busy UX (the extra bar at the bottom), and add a third button (visit) to the already existing top-left X and bottom-left back arrow that all do exactly the same thing (return to model-viewer, as mentioned in #1085). Additionally, if someone neglects to set the alt text, displaying it in SceneViewer makes the error more glaring. 

image

elalish:
Fixes #1097 
Fixes #1143 
Fixes #1095 

This is working on my Pixel3, but more checks would be great, especially for #1095. It now goes directly into AR. I also removed the title and link to clean up the UI, as these seem unnecessary, but they can be easily put back if anyone objects. It also seems like our fallback is broken, but not because of this change.


packages/model-viewer/src/features/ar.ts
```diff
@@ -229,7 +221,7 @@ export const ARMixin = <T extends Constructor<ModelViewerElementBase>>(
           await this[$enterARWithWebXR]();
           break;
         case ARMode.SCENE_VIEWER:
-          openSceneViewer(this.src!, this.alt || '', this.arScale);
+          openSceneViewer(this.src!, this.arScale);
```
cdata, line 224: Why did we change this behavior?

I'm all for us making this customizable, but short of a clearly spelled out design goal (in the form of a filed issue) I would say this should remain unchanged.

elalish: Okay, here's the spelled out design goal: #1143. Still open for discussion.

cdata: I left some feedback. This probably needs more discussion before we can move on it, so for now I would recommend reverting the change.

elalish: I registered my disagreement, but the change has been reverted here. I'm also not sure what existing query parameters will do in this new intent scheme. I think they won't work anymore, but maybe there's a way to fix that? I got into some trouble with `?` and `&`.

PR #1149, opened 2020-04-15
Ensure built-in poster has default background color

cdata:
I think this may have regressed as part of the seamless poster change in #890 . I don't think it's necessary for the poster to have a fallback color of `inherit` for seamless poster transitions to work, and also it means that in most cases, the built-in poster won't have any background color at all.

### DIFF
--- a/packages/modelviewer.dev/examples/lazy-loading.html
+++ b/packages/modelviewer.dev/examples/lazy-loading.html
@@ -81,6 +81,11 @@
           <h4></h4>
         </div>
         <example-snippet stamp-to="demo-container-1" highlight-as="html">
+          <style>
+            model-viewer#reveal {
+              --poster-color: transparent;
+            }
+          </style>
           <template>
 <model-viewer id="reveal" preload camera-controls auto-rotate poster="../assets/poster-shishkebab.png" src="../shared-assets/models/shishkebab.glb" alt="A 3D model of a shishkebab"></model-viewer>
           </template>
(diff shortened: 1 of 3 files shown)


elalish:
Hmm, I think the issue is less seamless posters and more transparent rendering (and therefore transparent posters). Look at modelviewer.dev/examples/lazy-loading.html with this change. You'll see that the background color changes from white to grey now, which is why I put `inherit` IIRC.

cdata:
Ah, so the poster color needs to be transparent in the case that there is a configured poster image with transparency in it? I think it's fair to say that this should be configured by the user as needed to achieve the desired effect.

packages/modelviewer.dev/examples/lazy-loading.html
```diff
@@ -81,6 +81,11 @@ <h2 class="demo-title">Display a poster until loaded</h2>
           <h4></h4>
         </div>
         <example-snippet stamp-to="demo-container-1" highlight-as="html">
+          <style>
+            model-viewer#reveal {
+              --poster-color: transparent;
+            }
+          </style>
```
elalish, line 88: I think we should move this inside the template to make it a visible part of our documentation. We should also add this to the example below. The third example seems to already work without needing this, even though it is also a transparent poster. Do you know why?

cdata:
Many docs and an example have been updated

packages/modelviewer.dev/examples/lazy-loading.html
cdata: Oh, sorry. This was a mistake. It should be in the template.

cdata: The third example replaces the default poster (via a `slot`-ed alternative), so `--poster-color` no longer factors into the equation.

elalish: I think we still want to add this change to the second example, as it is like the first.

PR #1153, opened 2020-04-15
Webxr cleanup

elalish:
This fixes a couple of issues where we were not cleaning up webxr properly in the case that the session was cancelled before it could start.


cdata:
Looks good. I'm going to test it out. I filed https://github.com/google/model-viewer/issues/1154 for potential future refinement in this area.

cdata:
Infinitely better 💯 nice fix!

PR #1155, opened 2020-04-16
Document our versioning strategy

### LINKED ISSUE #473: Discuss (and document) deprecation cycle
We've had a few recent issues where we want to change attributes (#452 and #429). In #452 we changed the attribute name, in #429 (at least currently) we may preserve the old attribute for a period of time.

Ideally, expectations around when we'll deprecate attributes, and how long a deprecated attribute will survive for would be documented.

However, before we can document this, I'm interested in what cycles we'd all be comfortable with or would prefer.

smalls:
Fixes #473

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,3 +44,38 @@ We recommend making your pull request from a fork. See [Creating a pull
 request from a
 fork](https://help.github.com/articles/creating-a-pull-request-from-a-fork/)
 for more information.
+
+## Stability
+
+It’s important that our users can depend on our product, and not to worry
+about changes in model-viewer causing regressions in their use of the
+component. However, it’s also important that we continue to improve
+model-viewer, making changes to improve ergonomics and rendering quality.
+
+In general, <model-viewer> follows semver. Major versions (X.y.z) are used for
+the largest features and backwards-incompatible changes (more below on that),
+minor versions (x.Y.z) are used for new features, and patch versions (x.y.Z)
+are used for bug fixes.
+
+When possible please keep your changes backwards-compatible - including
+keeping both existing APIs and rendering consistent. When that’s not possible,
+we’ll follow a three-step process across several [semver](semver.org) major
+versions (in the `0.x.y` versions leading up to 1.0, we’ll use the minor
+version - `x` - for the same purpose): 
+
+1) Add a new optional API to enable the new behavior; in the release notes,
+document the new behavior, as well that it will become the default in the
+next version (2.0.0)
+1) Make the new API default enabled (allow it to be explicitly disabled for
+the old behavior); in the release notes, document this change, and that the
+API will be removed in the next version (3.0.0)
+1) Remove the API (4.0.0)
+
+Breaking changes mostly center around anything that the developer would have
+to take action on. This includes changes in APIs, or changes to the model
+staging/lighting that they might want to undo in order to achieve the effect
+that they currently enjoy.
+
+Additive changes (a new API) aren’t breaking, nor are smaller changes (a fix
+to the correctness of reflections, or a 1% change in padding).
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ approach.
 
 Please include tests in your pull request.
 
-We recommend making your pull request from a fork. See [Creating a pull
+We recommend making your pull request from a fork. See [creating a pull
 request from a
 fork](https://help.github.com/articles/creating-a-pull-request-from-a-fork/)
 for more information.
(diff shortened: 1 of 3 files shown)


packages/model-viewer/README.md
```diff
@@ -41,6 +41,31 @@ It can also be used directly from various free CDNs such as [unpkg.com](https://
...
+upgrade when you run [`npm update`](https://docs.npmjs.com/cli/update.html).
+
+If you’re using [unpkg.com](https://unpkg.com), you can pin to a specific
+version by specifying it in the URL (the sample below pins to version 0.10.0):
+
+```html
+<!-- Loads <model-viewer> for modern browsers: -->
+<script type="module"
+    src="https://unpkg.com/@google/model-viewer@0.10.0/dist/model-viewer.js">
+</script>
```
cdata, line 65: I'm a little nervous about keeping a fixed version in an example in the README because we need to remember to update it with each release. I know it's tricky to convey the concept without doing that, but also I know from our cookbook results that folks are just going to blindly copy/paste example snippets like this.

packages/model-viewer/README.md
```diff
@@ -55,7 +80,7 @@ desktop and mobile browsers. `<model-viewer>` is also supported on IE11.
 capabilities and compatibility of the library get better as the web evolves.
 
 However, not all browsers support all of these features today. **Check out
-[POLYFILLS.md](https://github.com/PolymerLabs/model-viewer/blob/master/POLYFILLS.md) to
+[POLYFILLS.md](https://github.com/google/model-viewer/blob/master/packages/model-viewer/POLYFILLS.md) to
```
cdata, line 83: 💯

packages/model-viewer/README.md
smalls: Agreed! I had the same problem. For a bit I had a <version> marker in the URL - I could switch back to that if you think it'd be better?

cdata: Personally I would be happier if we included a generic snippet that will fail. Maybe we can strive towards some automation that will auto-update this part of the README for us? Github Actions seems like a very suitable solution for this.

CONTRIBUTING.md
```diff
@@ -40,7 +40,42 @@ approach.
...
+versions (in the `0.y.z` versions leading up to 1.0, we’ll use the minor
+version - `y` - for the same purpose): 
+
+1) Add a new optional API to enable the new behavior; in the release notes,
+document the new behavior, as well that it will become the default in the
+next version (in for instance the `2.0.0` release)
+1) Make the new API default enabled (allow it to be explicitly disabled for
+the old behavior); in the release notes, document this change, and that the
+API will be removed in the next version (the `3.0.0` release)
+1) Remove the API (the `4.0.0` release)
```
elalish, line 72: If we say the API will be removed in the 3.0.0 release, then what is 4.0.0 for?

CONTRIBUTING.md
```diff
@@ -40,7 +40,42 @@ approach.
...
+
+Breaking changes mostly center around anything that the developer would have
+to take action on. This includes changes in APIs, or changes to the model
```
elalish, line 75: I think we could use a more specific definition of what we mean by public APIs, as in what is vs. isn't covered by this deprecation plan. We should also mention Three.js updates, which are breaking in a sense, but I believe also do not follow this process.

CONTRIBUTING.md
smalls: Actually removing it - in 3.0.0 we document that it will be removed, and then it's not removed until 4.0.0.

CONTRIBUTING.md
smalls: Would "documented API" rather than "public API" be clearer?

Ideally three.js would follow the same process, acknowledging that they're not committing to the same sort of thing so we're doing our best to absorb their changes into our process.

When they make a small rendering fix, that falls into the small tweaks category that doesn't require cross-version documentation.

When we've made bigger rendering changes we've done our best to communicate them ahead of time or at the latest in the release notes. I think that's a miss here - changes that aren't part of the documented API, but are significant, will be added to the release notes under the "breaking changes" heading.

Let me know how that sounds.

elalish: Would love to have @cdata weigh in here. I believe the three.js thing had more to do with transitive dependencies. +1 for "documented API".

CONTRIBUTING.md
elalish: I think I'm having some trouble groking whether the named version refers to the step you're writing about or the step after. I'm pretty sure one way or another there are only two named versions involved. Adding doesn't break, so doesn't increment. Changing the default and removing the old behavior both do increment. Something in this wording isn't consistent with that.

PR #1167, opened 2020-04-22
Add minified bundle build for model-viewer

### LINKED ISSUE #819: Publish a minified version of the bundled build
### Description
model-viewer currently doesn't have minified releases online (or does it?).
Would be great if it follows the same pattern as other production libraries and just allows to turn "model-viewer.js" into "model-viewer.min.js" on the unpkg url.
(minifying it with online tools goes from 1.4MB to 0.7MB)

#### Live Demo
https://unpkg.com/@google/model-viewer/dist/
for comparison, three.js
https://cdnjs.com/libraries/three.js/

### Versions

- model-viewer: v0.6.2

cdata:
Fixes #819


stevesan:
> Does this also gzip the bundle? Should we for the same reason?

I think that's usually left up to the server

elalish:
@cdata I'm talking with the BrowserStack folks about the Edge test failing to connect. They've reproed it and say it's taking them longer to fix than they anticipated. We may want to disable the Edge tests for now, since every build is failing with the same error.

cdata:
+1 to what @stevesan said.

I guess we can disable Edge. Lame.

packages/model-viewer/rollup.config.js
```diff
@@ -80,6 +87,25 @@ if (NODE_ENV !== 'development') {
...
   );
+
+  plugins = [terser()];
+
+  outputOptions.push(
+      {
+        input: './dist/model-viewer.js',
+        output: {
+          file: './dist/model-viewer.min.js',
+          sourcemap: true,
```
elalish, line 98: I managed to get a basic source map working here, but it's not great; it just points back to the single huge model-viewer.js file. I tried building it from lib instead with all the plugins, but it would only find two sourcemap files: constants.js and template.js. Maybe it couldn't follow deeper paths? I tried a bunch of options for sourcemap, but nothing made any difference.

cdata: I would be interested to see if simplifying the pipeline makes a difference. I recall recently integrating terser in another application and having full source maps back to `.ts` files without much extra work (although in that case, it also required an additional plug-in to get from non-bundled `.js` to `.ts`).

Does it make a difference if you remove `cleanup(...)`?

elalish: Thanks, I just tried, but it has the same results I've been seeing; take a look. It creates a 1.9 MB model-viewer.min.js.map file, but only two source files show up:
Screen Shot 2020-06-02 at 7 49 08 AM

packages/modelviewer.dev/scripts/ci-before-deploy.sh
```diff
@@ -105,6 +105,12 @@ mv ../render-fidelity-tools/test/results $DEPLOY_ROOT/fidelity/results
 cp ../render-fidelity-tools/test/results-viewer.html $DEPLOY_ROOT/fidelity/index.html
 cp ../render-fidelity-tools/dist/* $DEPLOY_ROOT/dist/
 
+FILES_TO_PATCH_WITH_MINIFIED_BUNDLE=($(find $DEPLOY_ROOT \( -type d -name node_modules -prune \) -o -type f | grep \.html))
+
+for file_to_patch in "${FILES_TO_PATCH_WITH_MINIFIED_BUNDLE[@]}"; do
+  sed -i '' 's/model-viewer\.js/model-viewer\.min\.js/g' $file_to_patch
```
elalish, line 111: Is this replacing non-minified with minified? Don't the pages have the minified version now?

PR #1169, opened 2020-04-24
enable high-performance GPUs

### LINKED ISSUE #1110: Memory Leaks and GPU Crashes in recent model-viewer versions
### Description

I noticed that there seems to be new issues in terms of memory handling in model-viewer (or in Chrome, I don't know). I'm not 100% sure but I think this is a regression.

When working on models and doing frequent refreshs (e.g. through Glitch), memory leaks and ultimately leads to browser crash / system freeze.
The same seems to happen for graphics memory, I've seen black screens, GPU crashes recently.
This is not only on Chrome Windows, but also on Chrome Android (I've had cases of black screens and fully glitched out system-wide GPU issues which I've never seen on Android, needing a full phone restart to resolve).

Might be related to #1074 

#### Live Demo
I've seen this when working on regular 1-model pages, but it reproduces faster with this stress test here: https://glitch.com/edit/#!/model-viewer-stresstest
(these are duplicate model files, not the same one loading multiple times).

1. open glitch page
2. open "In A New Window"
3. open the glitch editor and the new window next to each other (so that both are visible)
4. open Task Manager to watch RAM
5. insert space anywhere in glitch, let page refresh
6. wait for loading done
7. repeat repeat repeat
8. watch RAM all the time

The same seems to be true when just refreshing the page, but I had the feeling it's faster when glitch causes the update (not sure).
Also, I had the feeling that the speed of leaking is increased when the page is in "mobile preview" mode (not sure).
image

### Browser Affected

- [X] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [X] Android
- [X] iOS
- [ ] Linux
- [ ] MacOS
- [X] Windows

### Versions

- model-viewer: >= v0.9.0 (not sure if a regression or I just now noticed it)

**Edit:** Checked "iOS" as well as I'm getting the same black-model issues on iOS with the same "strategy" of many refreshes after each other. Sometimes after page refresh, sometimes after coming back from QuickLook.

elalish:
Fixes #1110 

Well, that bug is a little vague, but seems to be related to dual GPU performance in recent Chrome versions, so I'm hoping this is most of the problem. It turns out in Chrome 80 the default was changed to use the "low-power" GPU instead of the "high-performance" GPU. This PR explicitly sets high-performance for model-viewer so that we use the proper GPU. 

Something I noticed: `HAS_WEBXR_HIT_TEST_API` returns true now for even desktop Chrome, so I've modified `IS_WEBXR_AR_CANDIDATE` to narrow its scope.


packages/model-viewer/src/constants.ts
```diff
@@ -70,6 +68,8 @@ export const IS_AR_QUICKLOOK_CANDIDATE = (() => {
       tempAnchor.relList.supports('ar'));
 })();
 
+export const IS_WEBXR_AR_CANDIDATE = IS_ANDROID && HAS_WEBXR_HIT_TEST_API;
```
cdata, line 71: Principally speaking, it is an anti-pattern to base the notion of web platform feature support on characteristics such as device or brother version. We do it where we have to when there are corners of the world that are actually proprietary to a device, browser or manufacturer. But, in general we should avoid qualifications like this one unless absolutely necessary.

If there is a case where a device partially (or erroneously) implements a spec, then I recommend doing feature detection for that particular case instead.

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -86,11 +86,15 @@ export class Renderer extends EventDispatcher {
   constructor(options?: RendererOptions) {
     super();
 
-    const webGlOptions = {alpha: true, antialias: true};
+    const webGlOptions = {
+      alpha: true,
+      antialias: true,
+      powerPreference: 'high-performance' as WebGLPowerPreference
```
cdata, line 92: Is it always the case that we want `'high-performance'`? What are the distinguishing characteristics between the different profiles? Surely there is a trade-off being made here.

packages/model-viewer/src/constants.ts
elalish: Agreed, though this is not erroneously implemented; it's just that the presence of the hit test API does not actually say anything about AR being supported or not. The proper way to test this is `await navigator.xr.isSessionSupported("immersive-ar")`, which is async because it will try to download ARCore if it's not already on the device (and there is some associated system UI). Right now we only attempt this when someone tries to enter AR. However, our constant is there so we can choose whether to set the renderer to preserve its drawing buffer, which we have to decide when we create it. 

Should we move the possible ARCore download to page load in order to get more accurate knowledge? Or just set preserve draw buffer all the time and not worry about any performance effect that might have? I think all three options are valid enough for discussion.

packages/model-viewer/src/three-components/Renderer.ts
elalish: It only has an effect on dual-GPU systems, telling Chrome to use the faster GPU for this tab. This was the default until Chrome 80, and now it defaults to `low-power`. Based on the feedback I've had from dual-GPU users who were surprised <model-viewer> was not using their good GPU, I think this is a reasonable choice for websites using <model-viewer>.

packages/model-viewer/src/constants.ts
elalish: Hey, nevermind! Turns out the preservedrawingbuffer is vestigial anyway, so I can remove this whole code path. It was for an old version of WebXR.

packages/model-viewer/src/three-components/Renderer.ts
elalish: Note, the UA still has authority to override, so user's power preferences still take precedence over this.

soraryu:
Thank your for investigating this. So I wasn't going crazy in my feeling that the performance had strongly regressed - if this was a relatively revent change with Chrome 80.

PR #1179, opened 2020-04-28
Refactor DPR

### LINKED ISSUE #1173: Page zoom causes oversized renderer
Using page zoom (+/- buttons in Chrome), when zoom decreases, everything works fine, but when zoom increases, even though the rendered output looks correct, the underlying canvas is becoming extra high-resolution, which hurts performance significantly. This is true even when simply returning to 100% zoom. The underlying issue appears to be double-counting DPR in this situation.

### LINKED ISSUE #1174: In non-chromium browsers, multiple model-viewer elements get rendered in low DPR.
Our drawImage code path is not properly taking DPR into account, so the resolution drops when you put an extra model-viewer element on the page in browsers like Firefox, Safari, or anything on iOS.

elalish:
Fixes #1173 
Fixes #1174 

This is at least the third time I've tried to fix handling of device pixel ratio, and obviously the other times have failed (given the above bugs). This time I opted to refactor it extensively, since I had a lot of trouble reasoning about it. The renderer now owns keeping track of DPR instead of the element and I removed the ModelScene's dependency on the renderer and pulled some associated functions up a level. I've tested all four of our canvas rendering paths for both resizing and page zooming and I believe everything is working properly and performantly now, but any help checking edge cases I may not have thought of is always appreciated. 

A behavior change here is that when DPR changes (as in via page zoom), or the elements resize it will now both upsize and downsize the renderer appropriately.


elalish:
@cdata It turns out the Safari tests were failing from a hotspot race condition, where we would fail to fire a visibility event because the slots hadn't been added to the shadow DOM yet. I must have triggered it by streamlining our rendering process a bit. Anyway, I refactored hotspots and realized several things could be removed once I got it working properly, so it should be in better shape now.

cdata:
@elalish if the renderer owns DPR, and not the element, what happens when you have multiple displays and their DPRs are not the same?

elalish:
I'm very glad to disregard that! You almost gave me a heart attack...

PR #1180, opened 2020-04-29
Add recommendations regarding verifying glTF->USDZ output

### LINKED ISSUE #559: Please add remark about differing size units between glTF/glb and USDZ
The **README** recommends to use `meter` as unit size. [Note about units and model size in AR](https://github.com/GoogleWebComponents/model-viewer/blame/master/README.md#L184). Please add a note that USDZ exports should be in `centimeter` instead and may need scaling, or iOS users will see models 100 times smaller.

smalls:
This is a reminder to test things like size units, although without going into too much detail as a recommendation depends on target use case (including if you need to support only iOS 13).
 
Fixes #559

### DIFF
--- a/packages/modelviewer.dev/examples/augmented-reality.html
+++ b/packages/modelviewer.dev/examples/augmented-reality.html
@@ -90,7 +90,9 @@
           </li>
           <li>
             <div>ios-src</div>
-            <p>Provides augmented reality on supported iOS 12+ devices via AR Quick Look, but requires an additional USDZ model.</p>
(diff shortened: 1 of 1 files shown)


elalish:
@smalls @cdata It seems your PR came in as third-party and didn't run all of our tests. Not that it's a concern for this PR, but we should probably get that fixed. I was surprised it passed so easily given the state our test infra is in, but it's just because it skipped most of it.

smalls:
I'll merge this, and will file an issue to look into the test issue.

PR #1185, opened 2020-04-30
Fix adding element

### LINKED ISSUE #1184: Appending a second model-viewer element with the same src causes the first to stop rendering.
It's only true the first time; adding more works just fine, and removing one causes the first to start working properly. See the [Glitch](https://glitch.com/edit/#!/model-viewer-leak). This appears to be a regression in v0.10.0.

elalish:
Fixes #1184 

Turns out I introduced this regression when I optimized us to use direct rendering when there is only one model-viewer instance. When adding the second, I didn't properly update the first. 

I also put in an unlinked example page, add.html, which I was using to diagnose this. I'm leaving it in because it's also useful for checking memory leaks, which I'm still chasing down.


packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -154,6 +154,9 @@ export class Renderer extends EventDispatcher {
 
   registerScene(scene: ModelScene) {
     this.scenes.add(scene);
+    this.selectCanvas();
+    scene.isDirty = true;
```
cdata, line 158: Leave a comment

PR #1187, opened 2020-05-04
match webXR spec update

elalish:
There was a change that prevented webXR from launching in Chrome Canary due to a small spec change. This fixes it.


cdata:
Is this change backwards compatible with non-Canary?

elalish:
Yes, that's why they marked it as a non-breaking change. I just tested to make sure.

PR #1191, opened 2020-05-05
fixed multiple webXR issues

### LINKED ISSUE #1190: Problems with multiple webXR demos on one page
When entering webXR mode with the second object on this [Glitch](https://model-viewer-blog.glitch.me/), the annotations disappear. When exiting webXR, the object is not properly centered.

elalish:
Fixes #1190 

The first problem was caused by a line of code I didn't write, but I'm ashamed I read it so many times without noticing the fairly glaring error. The second problem was that we were failing to call resize when cleaning up webXR. I think I missed this because our examples all use a %width CSS which causes the resize observer to fire when leaving fullscreen, but it does not for fixed-width elements.

### DIFF
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -15,7 +15,7 @@
 
 import {EventDispatcher, Matrix4, PerspectiveCamera, Ray, Vector3, WebGLRenderer} from 'three';
 
-import {$needsRender, $onResize} from '../model-viewer-base.js';
+import {$onResize} from '../model-viewer-base.js';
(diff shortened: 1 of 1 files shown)


packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -150,8 +150,8 @@ export class ARRenderer extends EventDispatcher {
           requiredFeatures: ['hit-test'],
           optionalFeatures: ['dom-overlay'],
           domOverlay: {
-            root: document.querySelector('model-viewer')!.shadowRoot!
-                      .querySelector('div.annotation-container')
```
cdata, line 154: 😳

PR #1193, opened 2020-05-06
Serve tests as a bundle

elalish:
Another experiment to see if this helps with the test outage.

### DIFF
--- a/packages/model-viewer/karma.conf.js
+++ b/packages/model-viewer/karma.conf.js
@@ -51,13 +51,7 @@ module.exports = function(config) {
       },
       {pattern: 'node_modules/@ungap/event-target/min.js', watched: false},
       {pattern: 'shared-assets/**/*', included: false},
-      {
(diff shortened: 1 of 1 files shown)


packages/model-viewer/karma.conf.js
```diff
@@ -51,13 +51,7 @@ module.exports = function(config) {
...
       {pattern: 'node_modules/@ungap/event-target/min.js', watched: false},
       {pattern: 'shared-assets/**/*', included: false},
-      {
-        pattern: 'lib/test/**/*-spec.js',
-        included: false,
-        watched: true,
-        type: 'module'
-      },
-      {pattern: 'lib/test/index.js', watched: true, type: 'module'}
+      {pattern: 'dist/unit-tests-umd.js', watched: true, type: 'module'}
```
cdata, line 54: `type` shouldn't be `'module'` (you explicitly specified UMD)

It shouldn't break anything, but there is a chance that it could...

PR #1195, opened 2020-05-07
fixed example performance

elalish:
If you have the second and third examples on this [page](https://modelviewer.dev/examples/lighting-and-environment.html) both wiggling at once, the performance is terrible. Turns out this is because of the way Three.js handles backgrounds. There is only one background for the renderer we share among all elements; it will happily swap out texture images, but if those images are in different formats (.hdr vs .jpg), it causes the background shader to recompile, in this case recompiling twice per frame. It looks like major surgery to change's three's behavior and I don't expect the need for LDR and HDR skyboxes on the same page to be great anyway (especially since environment-image is unaffected). Instead I changed the example to explain.

### DIFF
--- a/packages/modelviewer.dev/examples/lighting-and-environment.html
+++ b/packages/modelviewer.dev/examples/lighting-and-environment.html
@@ -75,40 +75,6 @@ window.oscillate = function(min, max, period, time) {
   </div>
 
   <div class="sample">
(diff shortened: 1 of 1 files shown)


cdata:
Good find @elalish - it might be worth filing a tracking issue against Three.js, since this will likely rear its head in the future and at least we will have a place to direct hapless users.

PR #1212, opened 2020-05-13
handle non-positive damper normalizer

### LINKED ISSUE #1210: With no src model, camera-change fires continuously
### Description

Put this in an html file:
```
<!-- Import the component -->
<script type="module" src="https://unpkg.com/@google/model-viewer/dist/model-viewer.js"></script>
<script nomodule src="https://unpkg.com/@google/model-viewer/dist/model-viewer-legacy.js"></script>
<!-- Use it like any other HTML element -->
<model-viewer id='mv'></model-viewer>

<script>
  const mv = document.getElementById("mv");
  mv.addEventListener("camera-change", (ev)=>{
    console.log(`hello`);
  });
</script>
```

#### Live Demo
https://jsfiddle.net/62odkf1y/

### Browser Affected

- [tested ] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ tested] Windows

### Versions

- model-viewer: v0.10.0
- three.js: r??

elalish:
Fixes #1210

The issue is that with no `src`, the ideal camera distance is zero, which makes maximum radius zero, which is the normalization factor for the radius damper. That made it asymptotically approach zero forever because it was never close enough to stop. Fixed it by making the damper handle non-positive normalization values, which should be a general fix. I believe this problem could also have come up if someone set max-camera-orbit to zero or negative radius, but this'll handle those cases as well.

### DIFF
--- a/packages/model-viewer/src/test/index.ts
+++ b/packages/model-viewer/src/test/index.ts
@@ -27,6 +27,7 @@ import './three-components/ARRenderer-spec.js';
 import './three-components/GLTFInstance-spec.js';
 import './three-components/gltf-instance/ModelViewerGLTFInstance-spec.js';
 import './three-components/SmoothControls-spec.js';
+import './three-components/Damper-spec.js';
 import './three-components/TextureUtils-spec.js';
 import './three-components/CachingGLTFLoader-spec.js';
 import './three-components/ModelUtils-spec.js';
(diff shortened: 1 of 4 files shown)


cdata:
Seems like an easy one to add a test for..

PR #1214, opened 2020-05-15
WebXR update

### LINKED ISSUE #519: Exit fullscreen button doesn't always exit fullscreen

### LINKED ISSUE #1098: Fullscreen AR fallback is currently broken in Chrome

elalish:
Fixes #519 
Fixes #1098 

This adds the X exit button to the top-left of webXR; it takes the place of the exit fullscreen button as I've also removed the fullscreen fallback, which was broken anyway. It also switches webxr to the default ar experience. There is also some cleanup of vestigial codepaths and more complete post-session state reset.

A breaking change (though it was never documented) is that the `exit-fullscreen-button` slot is now the `exit-webxr-button` slot.

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -16,7 +16,7 @@
 import {property} from 'lit-element';
 
 import {IS_ANDROID, IS_AR_QUICKLOOK_CANDIDATE, IS_IOS_CHROME, IS_IOS_SAFARI, IS_WEBXR_AR_CANDIDATE} from '../constants.js';
-import ModelViewerElementBase, {$container, $renderer, $scene} from '../model-viewer-base.js';
+import ModelViewerElementBase, {$renderer, $scene} from '../model-viewer-base.js';
 import {enumerationDeserializer} from '../styles/deserializers.js';
 import {Constructor, deserializeUrl} from '../utilities.js';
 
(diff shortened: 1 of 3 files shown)


packages/model-viewer/src/features/ar.ts
```diff
@@ -98,12 +98,12 @@ export type QuickLookBrowser = 'safari'|'chrome';
 const deserializeQuickLookBrowsers =
     enumerationDeserializer<QuickLookBrowser>(['safari', 'chrome']);
 
-export type ARMode = 'quick-look'|'scene-viewer'|'webxr'|'fallback'|'none';
+export type ARMode = 'quick-look'|'scene-viewer'|'webxr'|'none';
 
 const deserializeARModes = enumerationDeserializer<ARMode>(
-    ['quick-look', 'scene-viewer', 'webxr', 'fallback', 'none']);
```
klausw, line 104: Drive-by question: if I'm understanding it right, this is removing 'fallback' from the list of recognized values. If a site is still using this in an attribute, will it be ignored silently, or does it cause an error?

packages/model-viewer/src/template.ts
```diff
@@ -344,6 +340,17 @@ canvas.show {
 
   <div class="slot default">
     <slot></slot>
+    
+    <div class="slot exit-webxr-button">
```
cdata, line 344: Just my two cents, but I would give this a name that describes its semantic purpose. WebXR is a big umbrella, and doesn't necessarily capture what the user is trying to do when configuring the slot (it just adds ambiguity to the purpose of the button).

packages/model-viewer/src/features/ar.ts
cdata: It will be ignored silently

packages/model-viewer/src/template.ts
elalish: Can you be more specific (or recommend a name)? `exit-webxr-button` seems pretty precise to me, but I'm probably just looking at it differently.

cdata: I think this button is specifically used for exiting AR. If you were to support VR some day, you might do it via WebXR, but I suspect that this slot wouldn't apply (or at least should be interpreted differently). Maybe I'm wrong.

Just speaking generally, WebXR is a body of API around which many, various UX can be realized (not necessarily AR-specific UX).

PR #1217, opened 2020-05-19
Add getters for FOV limits

stevesan:
It may be useful to know the current FOV limits so user code can change them sensibly. And, to do this without needing to parse the attributes (duplicating work).


stevesan:
recreating with non-master branch: https://github.com/google/model-viewer/pull/1218

PR #1218, opened 2020-05-19
Add getters for FOV limits

stevesan:
It may be useful to know the current FOV limits so user code can change them sensibly. And, to do this without needing to parse the attributes (duplicating work).

### DIFF
--- a/packages/model-viewer/src/features/controls.ts
+++ b/packages/model-viewer/src/features/controls.ts
@@ -363,6 +363,14 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
       return this[$controls].getFieldOfView();
     }
 
+    getMinimumFieldOfView(): number {
+      return this[$controls].options.minimumFieldOfView;
+    }
+
+    getMaximumFieldOfView(): number {
+      return this[$controls].options.maximumFieldOfView;
+    }
+
     jumpCameraToGoal() {
       this[$jumpCamera] = true;
       this.requestUpdate($jumpCamera, false);
--- a/packages/model-viewer/src/features/controls.ts
+++ b/packages/model-viewer/src/features/controls.ts
@@ -367,19 +367,11 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
 
     // Provided so user code does not have to parse these from attributes.
     getMinimumFieldOfView(): number {
-      const value = this[$controls].options.minimumFieldOfView;
-      if(value === undefined) {
-        throw new Error('minimumFieldOfView was undefined');
-      }
-      return value;
+      return this[$controls].options.minimumFieldOfView!;
     }
 
     getMaximumFieldOfView(): number {
-      const value = this[$controls].options.maximumFieldOfView;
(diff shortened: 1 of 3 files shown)


packages/model-viewer/src/features/controls.ts
```diff
@@ -240,6 +240,8 @@ export declare interface ControlsInterface {
   getCameraOrbit(): SphericalPosition;
   getCameraTarget(): Vector3D;
   getFieldOfView(): number;
+  getMinimumFieldOfView(): number|undefined;
+  getMaximumFieldOfView(): number|undefined;
```
elalish, line 244: Our types are a little weird, but this is actually always defined (it has default values).

packages/model-viewer/src/features/controls.ts
```diff
@@ -363,6 +365,15 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
       return this[$controls].getFieldOfView();
     }
 
+    // Provided so user code does not have to parse these from attributes.
+    getMinimumFieldOfView(): number|undefined {
+      return this[$controls].options.minimumFieldOfView;
+    }
+
+    getMaximumFieldOfView(): number|undefined {
+      return this[$controls].options.maximumFieldOfView;
```
stevesan, line 374: Should I rewrite these as ```return this...maxFov ?? DEFAULT_OPTIONS.maxFov``` then?

packages/model-viewer/src/features/controls.ts
stevesan: See my comment in the impl

stevesan:
> Are you also planning to add getMinimumCameraOrbit and such? Or do you want to leave that for a separate PR?

Yep, will do others later.

packages/model-viewer/src/features/controls.ts
elalish: No, it already returns that if nothing was defined; you can add a test to verify. I'm just saying the type is lying; it will always return a number.

stevesan:
OK see latest commit. Checking undefined and throwing if it is.

packages/model-viewer/src/features/controls.ts
```diff
@@ -366,12 +366,20 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
     }
 
     // Provided so user code does not have to parse these from attributes.
-    getMinimumFieldOfView(): number|undefined {
-      return this[$controls].options.minimumFieldOfView;
+    getMinimumFieldOfView(): number {
+      const value = this[$controls].options.minimumFieldOfView;
+      if(value === undefined) {
+        throw new Error('minimumFieldOfView was undefined');
```
elalish, line 372: I understand the sentiment here, but we don't do this anywhere else in our code base. The fact is this is always defined, and if you're concerned, it's better to write a test for it.

stevesan:
OK would you rather use ! instead?

elalish:
Yes. Also, if you reply to the comment instead of the whole PR, the conversations are much easier to keep track of. :)

PR #1221, opened 2020-05-22
Fixes for changing src in webXR mode

elalish:
While constructing a demo for a carousel of AR models I found some bugs, which are fixed here. Now the progress bar shows up in AR (using DOM Overlay) and the scene's placement box and height are updated when src is updated. I also did some cleanup, simplifying some of our event handlers and making ARRenderer a public member of Renderer so we don't have to forward every method. 

I also fixed a bug where the first model appeared to be not cached (it was reloaded when switched back to), which turned out to be because a src set as an attribute was converted to a full URL, while ones set in JS stayed relative, which confused the cache. Now they all stay relative, but if anyone knows why a full URL is better, please chime in.


ronpadz:
@elalish Can you share your demo for a carousel of AR models?

elalish:
@ronpadz Yes, I plan to add a demo like that to our docs pages. Honestly it's not really that fancy; just a CSS slider of thumbnails with a little JS function to switch the `src` on click. Once v1.0 ships and Chrome 83 has rolled out completely, it should all work easily.

packages/model-viewer/src/features/ar.ts
```diff
@@ -198,7 +198,7 @@ configuration or device capabilities');
       console.log('Attempting to present in AR...');
 
       try {
-        await this[$renderer].present(this[$scene]);
+        await this[$renderer].arRenderer.present(this[$scene]);
```
cdata, line 201: I would generally recommend against this kind of pattern, where you reach several objects deep to perform some action. The fact that `ARRenderer` is a member of `Renderer` implies that there is shared business logic or state between the two. However, by directly affecting the state of `ARRenderer` in this way, you are creating surface area for contention over that state. Even if it works fine today, it is ripe for bugs in the future.

I would recommend either continuing to wrap the calls into `ARRenderer`, or else factoring it out of `Renderer` entirely so that the contract between `ARRenderer` and `Renderer` can be expressed in terms of implementation accessed in common by this module and `Renderer`.

cdata:
Good catch on the cache URL issue. Although I think the `deserializeUrl` helper is vestigial, this will technically constitute a change in API behavior so I would recommend marking it as a breaking change.

packages/model-viewer/src/features/ar.ts
elalish: Agreed; I filed #1228 to remind me to pull ARRenderer out of the Renderer class entirely.

elalish:
The Fidelity tests are still hanging in Github Actions for unknown reasons, but I'm less concerned about them from the perspective of merging. I'm going to merge PRs now when the unit tests pass since I don't want to be blocked by Github. I'll just run fidelity locally to ensure nothing is broken.

PR #1223, opened 2020-05-24
Path-tracing render backend for render-fidelity-tools (WIP)

bsdorra:
Adds https://github.com/DassaultSystemes-Technology/dspbr-pt as render backend to render-fidelity-tools.

update-screenshots script will fail at some point as not all of the test scenes are supported by the renderer yet.


elalish:
@bsdorra We actually have an affordance to handle problem fidelity tests: you can [exclude](https://github.com/google/model-viewer/blob/55b20c5faa5a726de7091e0e660574b703c50c09/test/fidelity/config.json#L128) certain renderers in the config of a given test. Can you update the config so that it passes? That will give us a clear TODO list as far as gaps in the upstream project.

elalish:
@bsdorra When you update the dependencies in package.json, you also need to run `npm run update:package-lock`. This will allow it to start building properly. You should also pull master, as I updated out testing config for newer browser versions that helps out tests to pass.

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
```diff
@@ -0,0 +1,185 @@
...
+      orbit.radius * Math.cos(phi) + target.y,
+      radiusSinPhi * Math.cos(theta) + target.z
+    ];
+    if (orbit.radius <= 0) {
+      center[0] = eye[0] - Math.sin(phi) * Math.sin(theta);
+      center[1] = eye[1] - Math.cos(phi);
+      center[2] = eye[2] - Math.sin(phi) * Math.cos(theta);
+    }
+
+    const bbox = this[$renderer].getBoundingBox();
```
elalish, line 163: This function is not found when we try to run locally. Is there a chance our dependency is not up to date?

bsdorra: > @bsdorra When you update the dependencies in package.json, you also need to run `npm run update:package-lock`. This will allow it to start building properly. You should also pull master, as I updated out testing config for newer browser versions that helps out tests to pass.

Ok, I'll fix this. 

> This function is not found when we try to run locally. Is there a chance our dependency is not up to date?

The function was indeed missing. I pushed it yesterday when i was updating the config. It should work now.

elalish: Thanks! Do you think you can push dspbr-pt up to npm? That would make tracking version updates like this more robust.

sun765:
@bsdorra  Thank you for fixing the bounding box bug! I run your branch on my Chromebook and want to report a problem to you. There's no problem rendering scenarios on the path tracing render's page, but when I run npm run test or npm run update-screenshots, it always stuck on rendering the first path tracing golden. Sometimes it hang on the command line for a while and my IDE just exit, but most of times my chromebook just died or restarted by it self. Here's a screenshot when it stuck. I tried to update other scenarios but it stuck in the same way. Also, it always stucks when the console start to log "Frame finished"

image
(i have to take a photo using phone because the chromebook stopped working completely)

I and Emmett made some guess about the reasons of this problem
1. Since update-screenshots work in the context of headless chrome, so may it's a problem of headless chrome?
2. A problem of Chrome OS?  Here is information of my Chromebook
Screenshot 2020-06-12 at 11 28 07 AM

3. A memory problem or GPU problem? I took a photo of the task manager when it stopped .About GPU, although i haven't found a way to inspect GPU information of chromebook, but if it can render scenario in the renderer's page, It doesn't sound like a GPU's problem. 
image

That's all information I have about this problem. If there's anything i can do to help solve this problem, feel free to ask. Thanks!

bsdorra:
@sun765 Hm, this is the 2CylinderEngine scene, right? Seems like the transformation parsing for that scene is still broken which might result in several issue during intersection tests. This could cause a gpu timeout. Could you please try to exclude this scene from the config and check if you run into similar issues with the other scenes. 

If others fail similarly, you could try to render the tests in a smaller resolution. The renderer allocates quite some structs in fragment shader which might be a bit heavy on your gpu. 
Another guess would be a gpu timeout because shader execution takes too long. Not sure about chromebooks, but Windows has this TDR (Timeout Detection and Recovery) thing which resets your gpu if shaders won't return in time.

elalish:
@sun765 I tried this on my macbook and it appears to be working. I had to pull master and then `npm run update:package-lock`. This may be a ChromeOS GPU issue or some timeout as mentioned. 

@bsdorra Would you prefer we took over this PR or do you want to take it over the finish line yourself? The one thing that would be really helpful from you is if you could publish your gltfpt to npm (even if it's just a v0.1), so that we can keep track of the versioning better. Is that doable?

elalish:
@bsdorra Assuming the tests pass, I think all we need is:
1) sign the CLA
2) switch to an npm dependency
3) run `update-screenshots` and commit these new goldens so we can look at them here. This will also allow the Fidelity tests to run, since currently they fail trying to find those images for comparison.

bsdorra:
@googlebot I signed it!

bsdorra:
The latest commit adds golden images for some of the test scenarios. I further filtered the scenarios, as some of them failed due to the transformation issue I mentioned above. 

I set the default sample count to 1024 to get some reasonably smooth results, so grab a pot of coffee before running "update-screenshots", it might take a while. For quick tests, the default sample count needs to be changed in the viewer code directly or it can be configured on a per scenario level in the config file:
```json
"pt": {
    "numSamples": 16
}
```

In its current configuration the pathtracer render full global illumination (restricted to max. 8 bounces), so it includes all indirect lighting/shadowing and reflections. This doesn't necessarily make sense when comparing against the rasterizers (with the current scenarios). Carefully designed and simplified test scenarios and/or intentionally disabling specific features in the pathtracer should be considered for more meaningful tests results (and faster convergence).

- CLA is signed
- dspbr-pt is now available via npm registry

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
```diff
@@ -77,41 +76,35 @@ export class PathtracingViewer extends LitElement {
...
-    };
-
-    this[$renderer] =
-        new PathtracingRenderer({'canvas': this[$canvas]!}, settings);
+    const enableControls = false;
+    this[$renderer] = new PathtracingRenderer(this[$canvas]!, enableControls);
 
     const renderer = this[$renderer];
+    renderer.autoScaleOnImport(false);
+    renderer.setPixelRatio(0.5);
```
elalish, line 84: Why not set this the same as the other renderers?

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
```diff
@@ -77,41 +76,35 @@ export class PathtracingViewer extends LitElement {
...
+    renderer.setMaxBounceDepth(8);
```
elalish, line 85: This seems like a lot, but you know path tracers better than I do. Have you found particular sample models where you need this many bounces to get good convergence?

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
```diff
@@ -77,41 +76,35 @@ export class PathtracingViewer extends LitElement {
...
+    console.log('Rendering ' + numSamples + ' samples');
+    this[$renderer].render(
         numSamples,
-        () => {
-          console.log('frame finished');
+        (frame: number) => {
+          console.log('frame finished ' + frame);
         },
         () => {
           // Wait two rAFs to ensure we rendered at least once //TODO: clarify
```
elalish, line 110: We recently found that two rafs was unnecessary in our other renderers, so it likely is here too. I wonder if this is actually causing you to render the whole scene twice, which would be unfortunate.

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
```diff
@@ -169,17 +160,10 @@ export class PathtracingViewer extends LitElement {
     const far = 2 * Math.max(modelRadius, orbit.radius);
     const near = far / 1000;
 
-    const camera = this[$renderer].getCamera();
-    camera.position.set(eye[0], eye[1], eye[2]);
-    camera.up.set(0, 1, 0);
-    camera.lookAt(center[0], center[1], center[2]);
-    camera.updateMatrixWorld();
+    this[$renderer].setLookAt(eye, center, [0, 1, 0]);
```
elalish, line 163: 👍

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
```diff
@@ -169,17 +160,10 @@ export class PathtracingViewer extends LitElement {
...
     console.log(eye, center);
 
-    camera.fov = verticalFoV;
-    camera.near = near;
-    camera.far = far;
-    camera.updateProjectionMatrix();
+    this[$renderer].setPerspective(verticalFoV, near, far);
```
elalish, line 166: 👍

packages/render-fidelity-tools/package.json
```diff
@@ -32,7 +32,7 @@
     "@types/resize-observer-browser": "^0.1.2",
     "@types/rimraf": "^2.0.3",
     "@webcomponents/webcomponentsjs": "~2.1.3",
-    "dspbr-pt": "git+https://github.com/DassaultSystemes-Technology/dspbr-pt.git",
+    "dspbr-pt": "^0.1.0",
```
elalish, line 35: 💯

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
bsdorra: I needed to reduce pixel ratio because because my notebook couldn't handle higher resolutions for some of the scene. gpu memory limitation I guess. This should be set to 1.0 otherwise.

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
bsdorra: The reason why I set this to 8 is realted to this issue
https://github.com/DassaultSystemes-Technology/dspbr-pt/issues/6

In general, having a low bounce depth usually intrduces error compared to a fully unbiased integration method, e.g. with probabilistic ray-termination or just very high bounce depth. 
But you're right, with the current scenarios, with the issue linked above being resolved, less bounces should be enough to capture relevant features with an acceptablel error.

The question which arises when thinking about rasterizer validation with pathtracing is "how much global illumination features do actually makes senses to include?".  I mentioned this before with thoughts on artificially excluding some of the full light transport features, e.g. all indirect diffuse bounce light > 1 (that's were most rasterizers will stop faking gi), or no direct shadowing from IBL, ...

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
bsdorra: Ok, that was really just copy&paste, didn't think much about it. I'll check.

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
elalish: Thanks for the info; sounds like we should reevaluate when that bug is fixed. As for exclusion, I'm not sure, but I'm generally against it. I don't expect the rasterizers to ever attain the quality of the path tracer, but I'd rather see the errors than hide them. I prefer these tests to demonstrate the known shortcomings of our rasterizers, and make it known that these differences are places where users should expect rendering changes to occur over time as we build more features in and improve.

bsdorra:
> Thanks, this is looking great! I notice that despite 1024 samples, many of the renders are still pretty noisy. Am I right that the run time scales as roughly numSamples * maxBounces? If so, would we get better results by increasing samples and decreasing bounces?

Yes, sorry for the noisy images. I'm stuck with the 2013 macbook to run this. Rendering those image already took quite some time. For a proper machine, with a recent gpu, rendering more samples should be no problem 

Regarding bounce depth, please also see my comment above. In general, a higher bounce depth increases the correctness of the image, whereas more samples will improve the quality (less noise). Specifically, you'd need four times as many samples per pixels to cut the estimation error in half. 

Theoretically more bounces are only expensive where needed. So when a path bounces of a surface and ends up tracing into the environment, it will just do the IBL look-up and finish with a single bounce. Exterior scenes with less need for indirect lighting won't suffer that much from a high bounce count, a complex interior scenes on the other hand will be significantly slower.

PR #1227, opened 2020-05-26
Dynamic render scaling

### LINKED ISSUE #261: User-configurable render-scale that overrides devicePixelRatio for individual model-viewers
Relevant discussion: https://github.com/GoogleWebComponents/model-viewer/pull/258#pullrequestreview-181403843

> I imagine for future customization cases, there may be a render-scale value such that devs could specify 1 for 3D content with the rest of their content using DPR/viewport meta values for performance. This is similar to WebXR's frameBufferScaleFactor. In WebXR, I think, no scale (1) is high-end, and generally scale down (~0.5) for perf. This related issue, IIRC, some vendors have defaults to 0.75, and as a dev, it's unclear if setting scale 1 is reducing or increasing the scale. General food for thought!

### LINKED ISSUE #1220: Please add an option to allow the model to render and be visible before model-viewer is in the viewport
There are ads use-cases where the ad slot starts below-the-fold. The entire ad needs to already be rendered off-screen before it comes into view. Currently, you can load the model offscreen, but there is still a delay when the model is rendered as the model first enters the viewport.

Live Demo at:
The model-viewer is at the bottom of the page. When you scroll down to the bottom (3000px), you will see the model-viewer's progress bar advance and the model come into view. If you need to see the issue again, please refresh the browser and repeat.

https://glitch.com/~peppermint-lime-hortensia

### Browser Affected
I only checked on Chrome, but this will happen with any browser that supports IntersectionObserver because model-viewer relies on IntersectionObserver to control visibility - https://github.com/google/model-viewer/blob/9d75b2564ed923e26bb34a36ec515f05c716d697/packages/model-viewer/src/model-viewer-base.ts#L252

- [X] Chrome Version 83.0.4103.61 (Official Build) (64-bit)
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS
I only tested on MacOS.
- [ ] Android
- [ ] iOS
- [ ] Linux
- [X] MacOS
- [ ] Windows

### Versions
- model-viewer: v0.10.0
- three.js: r115

elalish:
Fixes #1220 
Fixes #261

This is a major performance improvement, where we now detect dropped frames and scale down the rendering appropriately to maintain framerate. This also includes an additional performance improvement for firefox, where we use the fast, no-copy render path when there is only a single model visible in the viewport, rather than on the page in its entirety. I also found that we were needlessly rendering during the motionless part of the interaction prompt, so that's also fixed. Finally, we now render a first frame when the model loads even if it's not visible to fix #1220, since that seems like desired default behavior.

### DIFF
--- a/packages/model-viewer/src/constants.ts
+++ b/packages/model-viewer/src/constants.ts
@@ -42,8 +42,11 @@ export const IS_MOBILE = (() => {
   return check;
 })();
 
-export const USE_OFFSCREEN_CANVAS = Boolean((self as any).OffscreenCanvas) &&
-    Boolean((self as any).OffscreenCanvas.prototype.transferToImageBitmap);
+// Disabling offscreen canvas for now because it is slower and has bugs relating
+// to janky updates and out of sync frames.
+export const USE_OFFSCREEN_CANVAS = false;
+// Boolean((self as any).OffscreenCanvas) &&
+//     Boolean((self as any).OffscreenCanvas.prototype.transferToImageBitmap);
 
 export const IS_ANDROID = /android/i.test(navigator.userAgent);
 
--- a/packages/model-viewer/src/constants.ts
+++ b/packages/model-viewer/src/constants.ts
@@ -42,8 +42,9 @@ export const IS_MOBILE = (() => {
   return check;
 })();
 
-export const USE_OFFSCREEN_CANVAS = Boolean((self as any).OffscreenCanvas) &&
-    Boolean((self as any).OffscreenCanvas.prototype.transferToImageBitmap);
+export const USE_OFFSCREEN_CANVAS = false;
+// Boolean((self as any).OffscreenCanvas) &&
+//     Boolean((self as any).OffscreenCanvas.prototype.transferToImageBitmap);
 
 export const IS_ANDROID = /android/i.test(navigator.userAgent);
 
--- a/packages/model-viewer/src/constants.ts
+++ b/packages/model-viewer/src/constants.ts
@@ -42,9 +42,8 @@ export const IS_MOBILE = (() => {
   return check;
 })();
 
-export const USE_OFFSCREEN_CANVAS = false;
-// Boolean((self as any).OffscreenCanvas) &&
-//     Boolean((self as any).OffscreenCanvas.prototype.transferToImageBitmap);
+export const USE_OFFSCREEN_CANVAS = Boolean((self as any).OffscreenCanvas) &&
+    Boolean((self as any).OffscreenCanvas.prototype.transferToImageBitmap);
 
 export const IS_ANDROID = /android/i.test(navigator.userAgent);
 
--- a/packages/model-viewer/src/model-viewer-base.ts
+++ b/packages/model-viewer/src/model-viewer-base.ts
@@ -272,7 +272,12 @@ export default class ModelViewerElementBase extends UpdatingElement {
         }
       }, {
         root: null,
-        rootMargin: '10px',
+        // We used to have margin here, but it was causing animated models below
+        // the fold to steal the frame budget. Weirder still, it would also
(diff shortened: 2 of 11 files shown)


packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -89,10 +101,13 @@ export class Renderer extends EventDispatcher {
     const webGlOptions = {
       alpha: true,
       antialias: true,
-      powerPreference: 'high-performance' as WebGLPowerPreference
+      powerPreference: 'low-power' as WebGLPowerPreference
```
elalish, line 104: TODO: revert this before merging. This only makes it easier to test the dynamic resizing on a dual-GPU macbook.

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -169,46 +182,81 @@ export class Renderer extends EventDispatcher {
...
+
+    this.canvasElement.style.width = `${width}px`;
+    this.canvasElement.style.height = `${height}px`;
+    for (const scene of this.scenes) {
+      const {style} = scene.canvas;
+      style.width = `${width}px`;
+      style.height = `${height}px`;
+      scene.isDirty = true;
+    }
+    console.log('scale = ', this.scale);
```
elalish, line 244: TODO: remove this before merge; Again this helps with seeing what the dynamic resizing is doing.

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -271,7 +288,12 @@ export default class ModelViewerElementBase extends UpdatingElement {
         }
       }, {
         root: null,
-        rootMargin: '10px',
+        // We used to have margin here, but it was causing animated models below
+        // the fold to steal the frame budget. Weirder still, it would also
+        // cause input events to be swallowed, sometimes for seconds on the
+        // model above the fold, but only when the animated model was completely
+        // below. Setting this margin to zero fixed it.
+        rootMargin: '0px',
```
cdata, line 296: You mean it would cause animated models _more than 10px below the fold_ to steal frame budget? If so, this probably warrants some investigation and possibly a Chrome bug...

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -34,14 +34,20 @@ export interface ContextLostEvent extends Event {
   sourceEvent: WebGLContextEvent;
 }
 
+// Between 0 and 1: larger means the average responds faster and is less smooth.
+const DURATION_DECAY = 0.2;
+const LOW_FRAME_DURATION = 18;   // ms
+const HIGH_FRAME_DURATION = 26;  // ms
```
cdata, line 40: Just my personal preference, but I will usually suffix these kinds of constants with their unit in the name e.g., `HIGH_FRAME_DURATION_MS`.

packages/model-viewer/src/model-viewer-base.ts
elalish: No, it was only when they were within 10px of the fold. Stealing frame budget in that case was expected, but the part where it would swallow input events for long periods is a Chrome bug, I think.

packages/model-viewer/src/constants.ts
```diff
@@ -42,8 +42,11 @@ export const IS_MOBILE = (() => {
   return check;
 })();
 
-export const USE_OFFSCREEN_CANVAS = Boolean((self as any).OffscreenCanvas) &&
-    Boolean((self as any).OffscreenCanvas.prototype.transferToImageBitmap);
+// Disabling offscreen canvas for now because it is slower and has bugs relating
+// to janky updates and out of sync frames.
```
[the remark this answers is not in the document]
elalish: I have one that's been mostly ignored, but I'm going to try and write a better one.

PR #1239, opened 2020-06-01
use unmodified license text

elalish:
Turns out your copyright notice does not belong in the source license text.


cdata:
That is surprising, partially because it seems like common sense to include this information at that part of the license text, and partially since we have tens or hundreds of open source projects that do not follow that convention. But, it looks like we have many that do follow the convention 🤷

PR #1242, opened 2020-06-03
Metric

sun765:

### DIFF
--- a/packages/render-fidelity-tools/src/artifact-creator.ts
+++ b/packages/render-fidelity-tools/src/artifact-creator.ts
@@ -56,7 +56,7 @@ export class ArtifactCreator {
 
     for (const scenarioBase of scenarios) {
       const scenarioName = scenarioBase.name;
-      const scenario = this[$configReader].scenarioConfig(scenarioName)!;  // ??
+      const scenario = this[$configReader].scenarioConfig(scenarioName)!;
       const {dimensions} = scenario;
 
       if (scenarioWhitelist != null && !scenarioWhitelist.has(scenarioName)) {
@@ -70,8 +70,7 @@ export class ArtifactCreator {
       mkdirp.sync(scenarioOutputDirectory);
 
       const screenshot = await this.captureScreenshot(
-          'model-viewer',  // hard code to model viewer?  how do you load other
-                           // renders?
+          'model-viewer',
           scenarioName,
           dimensions,
           join(scenarioOutputDirectory, 'model-viewer.png'));
@@ -166,7 +165,7 @@ export class ArtifactCreator {
   async captureScreenshot(
       renderer: string, scenarioName: string, dimensions: Dimensions,
       outputPath: string = join(this.outputDirectory, 'model-viewer.png')) {
-    const devicePixelRatio = 2;  // hard code?
+    const devicePixelRatio = 2;
     const scaledWidth = dimensions.width;
     const scaledHeight = dimensions.height;
     const rendererConfig = this[$configReader].rendererConfig(renderer);
--- a/packages/render-fidelity-tools/src/artifact-creator.ts
+++ b/packages/render-fidelity-tools/src/artifact-creator.ts
@@ -56,7 +56,7 @@ export class ArtifactCreator {
 
     for (const scenarioBase of scenarios) {
       const scenarioName = scenarioBase.name;
-      const scenario = this[$configReader].scenarioConfig(scenarioName)!;
+      const scenario = this[$configReader].scenarioConfig(scenarioName)!;  // ??
       const {dimensions} = scenario;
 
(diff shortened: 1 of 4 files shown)


packages/render-fidelity-tools/src/workflows/test-fidelity.ts
```diff
@@ -22,17 +22,19 @@ const require = module.createRequire(import.meta.url);
 
 import {ArtifactCreator} from '../artifact-creator.js';
 
+
 const configPath = resolve(process.argv[2]);
 const rootDirectory = resolve(dirname(configPath));
 const config = require(configPath);
 
 const outputDirectory = join(rootDirectory, 'results');
+const portal = 9030;
```
elalish, line 31: Were these changes intended to be committed?

packages/render-fidelity-tools/test/config.json
```diff
@@ -1,4 +1,4 @@
-{
+{
```
elalish, line 1: Small white-space changes like this usually indicate that clang-format hasn't been applied. If you resave this with clang-format enabled, I think this change will disappear.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -165,7 +166,7 @@ export class ArtifactCreator {
   async captureScreenshot(
       renderer: string, scenarioName: string, dimensions: Dimensions,
       outputPath: string = join(this.outputDirectory, 'model-viewer.png')) {
-    const devicePixelRatio = 2;
+    const devicePixelRatio = 2;  // hard code?
```
elalish, line 169: This is the DPR of our macbooks. It doesn't really make any difference, except it exercises our DPR codepath, so helps to catch more bugs.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -56,7 +56,7 @@ export class ArtifactCreator {
 
     for (const scenarioBase of scenarios) {
       const scenarioName = scenarioBase.name;
-      const scenario = this[$configReader].scenarioConfig(scenarioName)!;
+      const scenario = this[$configReader].scenarioConfig(scenarioName)!;  // ??
```
elalish, line 59: Do you have a question for me?

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -70,7 +70,8 @@ export class ArtifactCreator {
       mkdirp.sync(scenarioOutputDirectory);
 
       const screenshot = await this.captureScreenshot(
-          'model-viewer',
+          'model-viewer',  // hard code to model viewer?  how do you load other
+                           // renders?
```
elalish, line 74: This is for the captureAndAnalyze function, which we currently only run on model-viewer (that's the only one we generate statistics for). As we add other renderers, and especially when we have better ground truth renders to compare to, we should revisit how these scripts should operate.

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
sun765: Yeah, do you think it's a good way to refactor code?

packages/render-fidelity-tools/src/artifact-creator.ts
sun765: just ignore it, i didn't mean to commit it. i don't understand the syntax at first so I leave a mark here. I have already learned what does this mean. I will remove this in the new push

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
elalish: maybe just call it port; that's more commonly used.

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
```diff
@@ -43,6 +44,7 @@ try {
 
 let scenarioWhitelist: Set<string>|null = null;
 
+// add additional arguments into scenariowhitelist which will not be rendered
```
elalish, line 47: These arguments are the whitelist, which means only these scenarios will be rendered. A blacklist would be a list of scenarios to not render.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -104,7 +104,7 @@ export interface ImageComparisonConfig {
 }
 
 export interface ComparableImage {
-  [index: number]: number;
+  [index: number]: number;  // what does this syntax mean?
```
elalish, line 107: This was fine as an excuse for a first PR, but is probably not the most efficient means of getting your questions answered about the code. I'd go ahead and revert this last commit and then we can merge this. Let's do a screen share and talk through code questions you have.

PR #1245, opened 2020-06-04
Initial support for configuring textures

cdata:
This change implements the first wave of capabilities related to configuring textures in the model. At this time, it supports configuring existing textures on the model. Textures can be configured by setting the URI of their associated source images. For example:

```typescript
self.addEventListener('model-change', () => {
  console.log('Got a new model:', self.model);

  model.materials[0]
      .pbrMetallicRoughness
      .baseColorTexture
      .texture
      .source
      .setURI(event.data.payload);
});
```

Any valid URI is supported. URIs for images that fail to load will result in a mutation error, and any associated mutations will not be applied to the source glTF. At this time, all material textures in the base glTF 2.0 spec can be configured.

### DIFF
--- a/packages/3dom/demo/index.html
+++ b/packages/3dom/demo/index.html
@@ -38,12 +38,43 @@ html {
   bottom: 0;
   left: 0;
   width: 100%;
-  height: 64px;
+  height: 128px;
 }
 
 button {
   font-size: 2em;
   margin: 0 0.25em;
+}
+
+#textures > div {
+  position: relative;
+  background-image: url(./add_photo_alternate-white-18dp.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: 50%;
+  border: 4px dashed white;
+  border-radius: 6px;
+  width: 64px;
+  height: 64px;
+  margin: 0 1em;
+  cursor: pointer;
+}
+
+#textures > div > input {
+  display: block;
+  opacity: 0;
+  width: 100%;
+  height: 100%;
+  cursor: pointer;
+}
+
+#textures .preview {
+  pointer-events: none;
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  background-size: 100%;
+  z-index: 1;
 }
   </style>
 </head>
@@ -54,6 +85,14 @@ button {
     <button data-color="1,0,0,1">Red</button>
     <button data-color="0,1,0,1">Green</button>
     <button data-color="0,0,1,1">Blue</button>
+
+    <section id="textures">
+      <div id="base-color">
+        <div class="preview"></div>
+        <div class="dropzone"></div>
+        <input class="input" type="file">
+      </div>
+    </section>
   </div>
   <script type="3DOM">
 console.log('Hello, 3DOM!');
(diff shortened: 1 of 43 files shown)


packages/3dom/demo/index.html
```diff
@@ -69,129 +108,18 @@
         console.log('Changing color to:', event.data.payload);
         material.pbrMetallicRoughness.setBaseColorFactor(event.data.payload);
         break;
+      case 'change-texture':
+        console.log('NEW TEXTURE URI', event.data.payload);
+        model.materials[0].pbrMetallicRoughness.baseColorTexture.texture.source.setURI(event.data.payload);
+        break;
```
elalish, line 114: Doesn't have to happen right now, but we should definitely update this demo to also be able to change the normal and emissive textures. Also, I wonder if we should note in our API that metallicRoughness usually also includes the occlusion channel? Maybe just a comment somewhere.

packages/3dom/src/api.ts
```diff
@@ -30,7 +31,7 @@
  *    found in a model's scene graph
  */
 export declare type ThreeDOMCapability =
-    'messaging' | 'material-properties' | 'fetch';
+    'messaging' | 'material-properties' | 'textures' | 'fetch';
```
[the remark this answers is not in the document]
cdata: The distinction here is between things like setting `baseColorFactor` and setting things that might cause network side-effects or changes to the binary buffer.

The names could potentially be better, but it does make sense to be that these represent two different layers of capability. Making a change with potential network side-effects may not always be allowed in cases where changes that don't cause such effects are fine.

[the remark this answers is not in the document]
elalish: Gotcha; are network side-effects also controlled by `'fetch'`?

packages/3dom/src/api.ts
```diff
@@ -173,6 +189,10 @@ export declare interface Material extends ThreeDOMElement {
    */
   readonly name?: string;
 
+  readonly normalTexture?: TextureInfo|null;
+  readonly occlusionTexture?: TextureInfo|null;
+  readonly emissiveTexture?: TextureInfo|null;
```
elalish, line 194: Is there a reason the baseColorTexture and MetallicRoughnessTexture haven't been added to pbrMetallicRoughness below?

packages/3dom/src/end-to-end-spec.ts
```diff
@@ -95,11 +106,42 @@ suite('end-to-end', () => {
...
+
+      executionContext.changeModel(graft);
+
+      executionContext.eval(`
+self.addEventListener('message', function(event) {
+  var textureURL = event.data;
+  model.materials[0].pbrMetallicRoughness.baseColorTexture.texture.source.setURI(textureURL);
+});
+
+self.postMessage('ready');
```
elalish, line 123: What does this do? It doesn't seem anyone is listening for a 'ready' message.

packages/3dom/src/api.ts
cdata: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-material

packages/3dom/src/end-to-end-spec.ts
cdata: `ready` is the `event.data` of the `message` event we are listening for in the worker.

cdata:
> Is the idea to add PBR textures as a follow-on?

I'm not sure which you are referring to. If you mean supporting specular-glossiness, then yes, that will have to be done later when we support configuring extensions.

AFAICT I added all of the textures that are defined in the glTF 2.0 base spec.

packages/3dom/src/api/sampler-spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+
+    suite('the generated class', () => {
+      let kernel: FakeModelKernel;
+      let GeneratedConstructor: SamplerConstructor;
+
+
+      setup(() => {
+        kernel = new FakeModelKernel();
+        GeneratedConstructor = defineSampler(ThreeDOMElement);
```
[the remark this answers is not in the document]
cdata: You have to pass it in because we need to be able to stringify the whole of its implementation when creating the worker. There are other ways to do this, but by using dependency injection we can also keep the code type safe.

The only condition is that it has to implement the `ThreeDOMElement` interface defined in `api.ts`.

packages/3dom/src/api.ts
cdata: Sorry, I misunderstood your question. The interface for `PBRMetallicRoughness` wasn't updated to reflect the changes in this PR. I'll fix that.

packages/3dom/src/api.ts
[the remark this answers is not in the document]
cdata: The `fetch` API is controlled by that. It does also cause network side-effects, but due to the separation of concerns here, a `fetch` network effect will originate from the 3DOM worker and a texture-changing effect will originate from wherever the renderer is running (currently on the main thread).

cdata:
Force pushed after rebase

PR #1250, opened 2020-06-05
updated Chrome and Edge to 83

### LINKED ISSUE #1207: Re-enable Chrome / Edge stable when 83.0 lands in stable
There appears to be some kind of critical issue in Chrome/Chromium 81.0 that causes our test suite to crash. This results in our test runner appearing to hang or timeout. The problem seems to affect both Chrome and Chromium-based derivatives such as Edge. Empirical observation suggests that Chrome 83.0 and Edge 83.0 are not affected by the bug, so we have disabled testing other versions of these browsers in our test configuration. The stable releases of these browsers should be re-enabled as 83.0 lands.

elalish:
Fixes #1207 

Interestingly, the `83.0 beta` we had been using started launching Chrome 84. Looks like Headless Chrome in Github actions has also been updated to 83 now, so that's reenabled. I also pushed out old Firefox version from 62 to 66, which looks to have removed a test flake I was seeing occasionally.

### DIFF
--- a/packages/3dom/karma.conf.js
+++ b/packages/3dom/karma.conf.js
@@ -93,12 +93,12 @@ module.exports = function(config) {
         browser: 'Firefox',
         browser_version: 'latest',
       },
-      'Firefox 72.0': {
+      'Firefox 76.0': {
         base: 'BrowserStack',
         os: 'Windows',
         os_version: '10',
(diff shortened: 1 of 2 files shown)


elalish:
@cdata FYI

PR #1251, opened 2020-06-05
Metric rms

### LINKED ISSUE #1246: update metric in fidelity test
### Description
We are trying to improve the metrics presented in fidelity test in the following ways:
1. When computing metrics pixel by by pixel, pixels of background are not distinguished from pixels of rendered object. Therefore, applying some techniques of detecting background may give us more accurate metrics.
2. Instead of averaging the sum of difference between pixels, it may be useful to compute the root mean square of the sum.

sun765:
### Reference Issue
1. Refactored render-screenshot script
2. Fixes #1246 
Add rmsDistanceRatio (sqrt(sum of all pixels' distance's square/ pixels) /max color distance) , mismatchingRmsDistanceRatio(sqrt(sum of mismatching pixels' distance's square/ mismatching pixels)/ max color distance)  into metrics. I ran the test and feel like it looks all right

### DIFF
--- a/packages/render-fidelity-tools/src/artifact-creator.ts
+++ b/packages/render-fidelity-tools/src/artifact-creator.ts
@@ -141,17 +141,27 @@ export class ArtifactCreator {
       for (const threshold of analysisThresholds) {
         console.log(`\n  📏 Using threshold ${threshold.toFixed(1)}`);
         const {analysis} = comparator.analyze(threshold);
-        const {matchingRatio, averageRmsDistance, mismatchingRmsDistance} =
-            analysis;
+        const {
+          matchingRatio,
+          averageDistanceRatio,
+          mismatchingAverageDistanceRatio,
+          rmsDistanceRatio,
+          mismatchingRmsDistanceRatio
+        } = analysis;
 
         thresholdResults.push(analysis);
 
         console.log(
             `  📊 Matching pixels: ${(matchingRatio * 100).toFixed(2)}%`);
         console.log(`  📊 Mean color distance: ${
-            (averageRmsDistance * 100).toFixed(2)}%`);
+            (averageDistanceRatio * 100).toFixed(2)}%`);
         console.log(`  📊 Mean color distance (mismatching pixels only): ${
-            (mismatchingRmsDistance * 100).toFixed(2)}%`);
+            (mismatchingAverageDistanceRatio * 100).toFixed(2)}%`);
+        console.log(`  📊 Root Mean Square color distance: ${
+            (rmsDistanceRatio * 100).toFixed(2)}%`);
+        console.log(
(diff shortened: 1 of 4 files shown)


packages/render-fidelity-tools/src/common.ts
```diff
@@ -170,11 +174,13 @@ export class ImageComparator {
           matched++;
         } else {
           mismatchingSum += delta;
+          mismatchingSquareSum += delta * delta;
         }
 
         const thresholdDelta = Math.max(0, delta - thresholdSquared);
 
         sum += thresholdDelta;
+        squareSum += thresholdDelta * thresholdDelta;
```
elalish, line 183: Part of what I hate about the old metrics is this threshold, which is very arbitrary. Let's not follow their pattern here; instead let's only build a single metric that is independent of a threshold.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -156,6 +156,10 @@ export class ArtifactCreator {
             (mismatchingAverageDistanceRatio * 100).toFixed(2)}%`);
       }
 
+      const {rmsDistanceRatio} = thresholdResults[0];
+      console.log(`\n  📊 Root mean square color distance (without threshold): ${
+          (rmsDistanceRatio * 100).toFixed(2)}%`);
+
```
sun765, line 162: Do you think thresholdResults[0] can cause potential error?

elalish:
Also, you should pull master; you're missing a fix to our test infrastructure, which I believe is why your units tests are failing.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -156,6 +156,11 @@ export class ArtifactCreator {
             (mismatchingAverageDistanceRatio * 100).toFixed(2)}%`);
       }
 
+      const {rmsDistanceRatio} = thresholdResults[0];
+      console.log(
+          `\n  📊 Decibels of root mean square color distance (without threshold): ${
+              (10 * Math.log10(rmsDistanceRatio * 100)).toFixed(2)}`);
```
sun765, line 162: The original db equation is 10 * Math.log10(I * 1000000000000). And here is the output using this equation:
Screenshot 2020-06-09 at 3 45 32 PM
I want to modify this equation for two reasons: 
1. the unit of the original equation is different, so I don't think it's meaningful to stick to the same number
2. Even if the rmsDistanceRatio is very small, the decibels can be very big. so the result may be less "interesting" as you said

So I modified it to 10 * Math.log10(rmsDistanceRatio * 100), and here is the result of the same scenario:
Screenshot 2020-06-09 at 3 47 30 PM
So do you agree with the equation i use here?

PR #1256, opened 2020-06-08
disable offscreenCanvas for ChromeOS

elalish:
We noticed the renders were inverted on ChromeOS, but only when offscreenCanvas was enabled. This is now disabled for CrOS as a workaround; next we need to file a bug on Chrome.

### DIFF
--- a/packages/model-viewer/src/constants.ts
+++ b/packages/model-viewer/src/constants.ts
@@ -42,10 +42,13 @@ export const IS_MOBILE = (() => {
   return check;
 })();
 
+export const IS_CHROMEOS = /CrOS\//.test(navigator.userAgent);
+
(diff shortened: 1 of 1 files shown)


packages/model-viewer/src/constants.ts
```diff
@@ -42,10 +42,14 @@ export const IS_MOBILE = (() => {
   return check;
 })();
 
+export const IS_CHROMEOS = /\bCrOS\b/.test(navigator.userAgent);
+console.log(IS_CHROMEOS);
```
elalish, line 46: Remember to remove debug logs :)

sun765: Oh sorry , I will do that now

Issue #1270, opened 2020-06-11
Regression: with WebXR now the default, it should be on par with SceneViewer


hybridherbst:
### Description

Since WebXR is now the default ar-mode, I think the functionality needs to be at least on par to the previous behaviour.

1. **2-finger rotation on the model does not work.**
  It works outside the model, with one finger only, while SceneViewer behaviour allows rotation with two fingers, which is crucial for larger models.
1. **no built-in way nor demo to show a similar label + link as SceneViewer+QuickLook.**
  I understand this can be custom-made, but given that this is built-in in both SceneViewer and QuickLook I think it should be built-in/default for the WebXR implementation as well. Ideally there's at least a demo for it, even better would in my opinion be if the default behaviour shows a title and link.
1. **no built-in way to capture pictures or videos** 
  I understand that this is also something that "can be built in js now" but I'd argue that model-viewer should provide that functionality by default to be on par with SceneViewer and QuickLook.
1. **UX regression in onboarding of users to AR**
  With SceneViewer, there's a one-time permission dialogue with a nicely crafted, nice looking onboarding flow. 
  With model-viewer WebXR, this is replaced by no onboarding + an ugly and scary browser default permission dialogue. 
  I don't have a good idea how to solve this; but I think that since model-viewer WebXR is effectively on track to replace SceneViewer onboarding UX should not be pushed onto devs implementating model-viewer.

#### Live Demo
https://modelviewer.dev/examples/webxr.html

### OS

- [X] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

elalish:
Thank you for the feedback.

1) Yes, I hadn't prioritized it because I found two-finger rotation in SceneViewer a bit awkward and not discoverable. But yes, the goal is parity.
2) Agreed, I plan to make a demo for this. Not sure about making it automatic; I'd certainly welcome a PR along those lines.
3) I've actually been hounding my webXR colleagues about this for more than a month now. We are actually missing a platform API for this; I'm hoping we can close that gap soon and as soon as one is available, I will certainly implement it here.

Anyway, remember you're welcome to set `ar-modes="scene-viewer webxr quick-look'` if you prefer the old behavior.

hybridherbst:
Thanks for replying! Good to know you're onto these.

[...]

PR #1275, opened 2020-06-12
fixed permission denial

### LINKED ISSUE #1269: Declining WebXR permission request freezes page
### Description

In the new WebXR flow users are asked for permission to use the camera for AR purposes. Declining that request, however, freezes the entire page instead of letting the user continue exploring the page and trying again at a later point.
Once the permission is declined, it seems after refreshing the page, clicking the "AR" button freezes immediately (without asking again).

#### Live Demo
pfc-writingball.glitch.me

### Browser Affected

- [X] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS

- [X] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### Versions

- model-viewer: latest dist

elalish:
Fixes #1269 

We were failing to properly clean up when the permission to enter webxr was denied.

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -219,6 +219,7 @@ configuration or device capabilities');
       } catch (error) {
         console.warn('Error while trying to present to AR');
         console.error(error);
+        await this[$renderer].arRenderer.stopPresenting();
       }
     }
 
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -319,12 +319,13 @@ export class ARRenderer extends EventDispatcher {
     // TODO: this method should be added to three.js's exported interface.
     (this.threeRenderer as any).setFramebuffer(null);
 
-    const session = this[$currentSession]!;
-    session.removeEventListener('selectstart', this[$onSelectStart]);
-    session.removeEventListener('selectend', this[$onSelectEnd]);
-
-    this[$currentSession] = null;
-    session.cancelAnimationFrame(this[$rafId]!);
+    const session = this[$currentSession];
+    if (session != null) {
+      session.removeEventListener('selectstart', this[$onSelectStart]);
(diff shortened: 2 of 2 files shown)


elalish:
@sun765 I'm going to start putting you on all my PRs as a reviewer. This is mostly so you can get used to the process. Don't worry about doing a super thorough review, but any questions you have are very much welcome. Mostly I'm just looking for help spotting stupid errors like debug code left in and confusing comments.

hybridherbst:
Thank you for fixing this! One understanding question. 
I know that for e.g. "Camera" permissions in regular apps, the flow is
- app asks user to give permission
- user can accept or decline
- accept: all is good, never asked again
- decline: answer is stored, user is never asked again (this is bad but unfortunately on most Android devices the flow is like this).

How will the flow be for these web permissions? If I declined one of them (e.g. by accident, because I want to go back first, ...), will I have the chance to give permissions again on the next click or will my response be "stuck" and only able to change in some hidden-away submenu?

elalish:
@hybridherbst Yeah, after extensive testing that's how it is here too. You have to click the lock icon on the URL, then site settings and reset permissions. I don't think there's anything I can do about that. Since it comes up when you tap the AR button, hopefully that means most people will be okay granting the permission, since it's pretty clear what it's for.

hybridherbst:
OK, that's very unfortunate. 
Is there any way to detect a "declined permission" on model-viewer side and warn about it / help the user in restoring functionality?

elalish:
@hybridherbst If I understand correctly, I think that is intentionally not exposed by the browser in order to keep sites from badgering users into accepting permissions. We take this quite seriously in Chrome, but I believe the other browsers are similar.

hybridherbst:
I assumed so. From a security standpoint, it's "better safe then sorry" but from an enduser/UX standpoint this is a regression in the WebXR vs SceneViewer implementation.

I added my notes to the other "soft regressions" (is there a better name for "implicit functionality lost with an update"?) to #1270 to keep it out of this closed issue.

PR #1290, opened 2020-06-17
3DOM metallicRoughness fix and extensions

hjeldin:
I've been checking out the new 3dom API to edit textures at runtime and i realized that metallicRoughness texture was not being serialized to the worklet, resulting in a runtime exception when trying to set a new texture.

I haven't created a new issue since this should be an open-and-shut case. Let me know if it's required.

### DIFF
--- a/packages/3dom/src/api/model-kernel-spec.ts
+++ b/packages/3dom/src/api/model-kernel-spec.ts
@@ -51,7 +51,7 @@ suite('api/model-kernel', () => {
         kernel.deactivate();
       });
 
-      suite('with a Model cntaining a Material', () => {
+      suite('with a Model containing a Material', () => {
         let kernel: ModelKernel;
 
         setup(() => {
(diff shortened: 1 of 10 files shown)


hjeldin:
I added both metalness and roughness to this pull request, i'll rename it to reflect these new changes.

packages/3dom/src/api/pbr-metallic-roughness-spec.ts
```diff
@@ -46,8 +46,11 @@ suite('api/pbr-metallic-roughness', () => {
             definePBRMetallicRoughness(ThreeDOMElement);
         const baseColorFactor: RGBA =
             [Math.random(), Math.random(), Math.random(), Math.random()];
+        const metallicFactor: number = Math.random();
+        const roughnessFactor: number = Math.random();
+
         const instance = new GeneratedConstructor(
-            new FakeModelKernel(), {id: 0, baseColorFactor});
+            new FakeModelKernel(), {id: 0, baseColorFactor, metallicFactor, roughnessFactor});
```
elalish, line 53: A couple of extra tests here for the new factors would be most welcome.

packages/3dom/src/api/model-kernel-spec.ts
```diff
@@ -51,7 +51,7 @@ suite('api/model-kernel', () => {
         kernel.deactivate();
       });
 
-      suite('with a Model cntaining a Material', () => {
+      suite('with a Model containing a Material', () => {
```
elalish, line 54: 👍

PR #1302, opened 2020-06-19
Simplified 3DOM

### LINKED ISSUE #999: Ensure 3DOM compatibility with IE11
Per discussion in the original 3DOM PR (https://github.com/google/model-viewer/pull/992#discussion_r376591004), 3DOM is not currently known to be compatible with IE11.

We should ensure that 3DOM is compatible with IE11 so that `<model-viewer>` users can get maximum flexibility when deciding to author scene graph worklet scripts.

elalish:
Fixes #999 

It turns out the temporary solution to the above problem accidentally blindsided us. In fact model-viewer has been nonfunctional on IE11 since v0.9, which we were unaware of because this test escaped too early. Therefore despite our tests all passing, we did not catch the fact that the scene-graph mixin fails to load even before it is used on IE11. In addition, though these tests pass on Chrome, we did not bother testing the legacy build in Chrome. It turns out this also fails some tests (though at least model-viewer can load). This PR will be open until we manage to fix these issues.

UPDATE:
I've fixed the above issues by tearing the execution context, worker scripts, script evals, and factories out of 3DOM. In addition to simplifying our codebase and fixing the conversions to ES5, this also greatly simplifies how 3DOM is accessed by users. Instead of creating a special worker script to operate on the scene graph, they now access it via the model property on the main thread like the rest of our API. The 3DOM API itself is unchanged, retaining a main thread copy of the scene graph for synchronous reads and returning promises for writes. It still dispatches messages internally for each mutation even though both ends now reside on the main thread; the reason is because we intend to move rendering to a worker in the future, and this architecture will ease the transition.


packages/model-viewer/src/test/features/scene-graph-spec.ts
```diff
@@ -13,27 +13,20 @@
  * limitations under the License.
  */
 
-import {Mesh, MeshStandardMaterial} from 'three';
+// import {Mesh, MeshStandardMaterial} from 'three';
```
stevesan, line 16: Get rid of commented lines?

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -300,20 +221,20 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-          onlyVisible: options?.onlyVisible || true,
-          maxTextureSize: options?.maxTextureSize || Infinity,
-          forcePowerOfTwoTextures: options?.forcePowerOfTwoTextures || false,
-          includeCustomExtensions: options?.includeCustomExtensions || false,
-          embedImages: options?.embedImages || true,
-          // NOTE: automatically set truncate draw range to true since
-          // we don't expose those parameters
-          truncateDrawRange: true
-        };
+          binary: true,
```
elalish, line 226: @hjeldin I updated this because my clang-format was angry about it. If you use VSCode for development and add the clang-format extension, our repo has a config that it reads to enforce style.

hjeldin: Got it, thanks!

hjeldin:
I wrote some use cases for the examples page but I encountered some issues:

1. Safari 13.1.1 (15609.2.9.1.2) cannot execute any scene graph manipulation. modelViewer.model is always undefined. I did some digging and managed to find out that the execution stops - with no exceptions - at the [super call in ModelGraft](https://github.com/google/model-viewer/blob/ed77f6e97615e47c4991f3fa08797b15af9a3e99/packages/3dom/src/facade/three-js/model-graft.ts#L66)

2. Applying textures to models that don't have any textures already applied at load time yields a `Cannot read property 'texture' of null`. Specifically `material.pbrMetallicRoughness.baseColorTexture` is undefined - which makes sense. Are we planning to expose Texture/TextureInfo/Sampler or am I missing something?

I'll dig around some more to see if I can understand what's going on with the first issue.

elalish:
@hjeldin 2) is just not implemented yet, but it's a priority. Probably not going to make it into this PR. 1) Sounds interesting; any info you can give would definitely be appreciated. I'm adding tests and fixing bugs, so watch for new commits.

PR #1309, opened 2020-06-23
Draft - add anchors support to ARRenderer

bialpio:
Early draft that adds support for WebXR Anchors API (if it's available).

There is still an issue of updated anchor position fighting with user's input.

@elalish - PTAL.


elalish:
After much discussion, it does not appear this is necessary when using local space as we are.

PR #1311, opened 2020-06-24
Add 'textures' to allowed capabilities set

stevesan:


elalish:
Thanks! How much work would it be to add a test on this?

stevesan:
> Thanks! How much work would it be to add a test on this?

I think I could add to scene-graph-spec.ts - tho I kinda wonder why this wasn't done already. is the API not really ready for use?

elalish:
No, I think this was just done in a rush the first time around.

PR #1319, opened 2020-06-30
replace directly loaded hdr with prefiltered hdr

### LINKED ISSUE #1295: Using prefiltered HDR map instead of loading hdr directly
### Description

Right now in fidelity test , babylonjs is loading hdr directly and the performance is not very good. Therefore it should use prefiltered HDR map instead just as filament did. This will reduce the time for running fidelity test.

sun765:
### Reference Issue
Fixes #1295

I updated babylonjs to the latest version because it has an option to prefilter the HDR file when loading it, and the performance looks pretty good.


packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -114,17 +114,18 @@ export class BabylonViewer extends LitElement {
 
     this[$scene].stopAllAnimations();
 
-    // load hdr directly (the size of cubmap is set to be 256 for all renderers)
+    // load and prefilter HDR texture (the size of cubmap is set to be 256 for
+    // all renderers)
     const environment = new HDRCubeTexture(
-        scenario.lighting, this[$scene], 256, false, false, false);
+        scenario.lighting, this[$scene], 256, false, false, false, true);
```
sun765, line 120: In latest version of Babylon, the HDRCubeTexture's constructor has some more options: 
noMipmap?: boolean, generateHarmonics?: boolean, gammaSpace?: boolean, prefilterOnLoad?: boolean.
I don't know if i need to set the other three variables to true or not, so i just leave them to be false.

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -114,17 +114,18 @@ export class BabylonViewer extends LitElement {
...
     this[$scene].environmentTexture = environment;
-    // rotate both skybox and hdr texture for 180 deg to match other renderers
+    // rotate environment for 90 deg, skybox for 270 deg to match the rotation
+    // in other renderers
```
sun765, line 123: Babylonjs seems to change their default orbit in their latest version.

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -114,17 +114,18 @@ export class BabylonViewer extends LitElement {
...
     environment.setReflectionTextureMatrix(
-        Matrix.RotationY(Tools.ToRadians(180)));
+        Matrix.RotationY(Tools.ToRadians(90)));
     const skybox =
         this[$scene].createDefaultSkybox(this[$scene].environmentTexture!);
-    skybox!.rotate(Axis.Y, Math.PI, Space.WORLD);
-    skybox!.infiniteDistance = true;
```
sun765, line 127: Since not setting far clip has already solved the skybox disappearing problem, this line is useless.

packages/render-fidelity-tools/package.json
```diff
@@ -19,8 +19,8 @@
     "node": ">=12.0.0"
   },
   "dependencies": {
-    "@babylonjs/loaders": "^4.1.0",
-    "@babylonjs/core": "^4.1.0",
+    "@babylonjs/loaders": "^4.2.0-alpha.22",
+    "@babylonjs/core": "^4.2.0-alpha.22",
```
sun765, line 23: change to latest version

elalish: I take it 4.2 is pre-release?

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
elalish: Wow, you have to set the environment and skybox to different orientations to get it to match? That sounds like a bug; I'd recommend you'd file something on Babylon for that.

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
elalish: We should experiment a little with these in a following PR to see what gives the closest results.

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
elalish: You might want to keep infiniteDistance; ours are infinite and I'm concerned that otherwise the skybox might change based on camera radius (which it shouldn't).

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
sun765: Ok, i will do that after this pr is merged

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -93,6 +94,9 @@ export class BabylonViewer extends LitElement {
             target.z),
         this[$scene]);
     camera.attachControl(this[$canvas]!, true);
+    // in babylon, camera use VERTICAL_FIXED mode by default, so fov here is
+    // equal to vertical fov
+    camera.fov = this[$degToRadians](verticalFoV);
```
sun765, line 99: This solves the transformation errors of the latern and the flightHelment scenarios. BJS uses 0.8 radians as default fov value.  And for our config, we use 45 deg( about 0.78 radians).  that tiny difference has an obvious on these vertical scenarios. That's really silly of me to not using the accurate radians

packages/render-fidelity-tools/package.json
sun765: Yeah, i will update it to form release when they have it

PR #1324, opened 2020-07-02
Show metric in fidelity page

### LINKED ISSUE #1284: fidelity test metrics is not shown in fidelity page
### Description
We did compute some metrics to compare the fidelity of different renderers in fidelity test. However, you have to run the fidelity test locally to see the metrics. In order to make it more convenient for users to see the it, we are going to include it in the fidelity test page.

sun765:
### Reference Issue
Fixes #1284 

Addition information shows on hover
Screen Shot 2020-07-01 at 8 27 55 PM


packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -39,27 +41,67 @@ export class RenderingScenario extends LitElement {
...
   }
 
+  updated(changedProperties: Map<any, any>) {
+    super.updated(changedProperties);
+
+    if (this.analysis == null) {
+      this.loadAnalysis();
+    }
+  }
+
```
sun765, line 51: I tried load ayalysis in constructor because it's easier to people to understand that way, but somehow it didn't work. So i just followed the same pattern as you did in image-comparision-app.ts

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -39,27 +41,67 @@ export class RenderingScenario extends LitElement {
...
+                <span class="question-icon"> </span>
+                <span class="tooltiptext">
+                  Root mean square color distance between ${
+                            golden
+                                .name} and current version of model-viewer on rendering ${
+                            this.name} in decibels.
+                  The decibel is given by: 10 * log(root mean square color distance).
+                </span>
+              </div>`}
+  </div>
```
sun765, line 100: I think clang-format doesn't work very well here. If you agree with me, I will format it manually.

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -80,11 +122,49 @@ h1 {
...
+  width: 1em;
+  height: 1em;
+  display: inline-block;
+}
+
+.metrics{
+  text-align: center;
+  height: 2em;
+}
+
```
sun765, line 162: im really not good at css, so im not sure whether these styles are written in the best way

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -80,11 +122,49 @@ h1 {
...
 .screenshot > header {
   display: block;
   position: absolute;
   width: 100%;
-  bottom: 1em;
+  bottom: 3em;
   left: 0;
```
sun765, line 168: The bottom distance need to be larger than before because there's metrics in the bottom of the screenshot div now

packages/shared-assets/icons/help_outline-24px.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="24" viewBox="0 0 24 24" width="24"><path d="M0 0h24v24H0z" fill="none"/><path d="M11 18h2v-2h-2v2zm1-16C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm0 18c-4.41 0-8-3.59-8-8s3.59-8 8-8 8 3.59 8 8-3.59 8-8 8zm0-14c-2.21 0-4 1.79-4 4h2c0-1.1.9-2 2-2s2 .9 2 2c0 2-3 1.75-3 5h2c0-2.25 3-2.5 3-5 0-2.21-1.79-4-4-4z"/></svg>
```
sun765, line 1: i tried to use external link icons (like google icons, fontawesome), but it didn't work. Also i found it's not encouraged to do so in lit-element's document, so i just use a svg file as the background of a div

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -39,27 +41,67 @@ export class RenderingScenario extends LitElement {
...
   <header>
     <h2>${golden.description}</h2>
   </header>
   <div class="check"></div>
   <img data-id="${this.name} ${golden.name}"
        style="width:${width}px" src="${basePath}/${golden.file}">
+    <div class = "metrics">
+      ${
+                    index == 0 || this.analysis == null ?
+                        html` <span> --- </span>` :
```
sun765, line 84: I added "---" for current version model-viewer here just to make it align with the other renderers. If you don't like it im happy to change it.

packages/render-fidelity-tools/src/components/rendering-scenario.ts
elalish: Agreed, we generally skip clang-format on html.

packages/render-fidelity-tools/src/components/rendering-scenario.ts
elalish: Me neither; I was going to ask if you've tested it on mobile, but I just tried the current version and it's kind of a disaster.

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -39,27 +41,67 @@ export class RenderingScenario extends LitElement {
...
+                        html` <span>${
+                            this.toDecibels(
+                                this.analysis.analysisResults[index - 1][0]
+                                    .rmsDistanceRatio)} db
+             </span>
+              <div class="tooltip">
+                <span class="question-icon"> </span>
+                <span class="tooltiptext">
+                  Root mean square color distance between ${
```
elalish, line 93: I would show the acronym: "Root mean square (RMS) Lab color distance", so that you can use that acronym below to shorten your equation: "10 * log(RMS)".

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -39,27 +41,67 @@ export class RenderingScenario extends LitElement {
...
+                            golden
+                                .name} and current version of model-viewer on rendering ${
+                            this.name} in decibels.
+                  The decibel is given by: 10 * log(root mean square color distance).
```
elalish, line 97: For those who don't like math, let's add "More negative means a closer match".

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -39,27 +41,67 @@ export class RenderingScenario extends LitElement {
...
   <img data-id="${this.name} ${golden.name}"
        style="width:${width}px" src="${basePath}/${golden.file}">
+    <div class = "metrics">
+      ${
+                    index == 0 || this.analysis == null ?
+                        html` <span> --- </span>` :
+                        html` <span>${
+                            this.toDecibels(
+                                this.analysis.analysisResults[index - 1][0]
+                                    .rmsDistanceRatio)} db
```
elalish, line 88: decibels is dB, not db.

packages/render-fidelity-tools/src/components/rendering-scenario.ts
sun765: I found that this project is formatting every changed ts file with clang-format, so I'd better just leave the format as it is

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -39,27 +41,66 @@ export class RenderingScenario extends LitElement {
...
+  updated(changedProperties: Map<any, any>) {
+    super.updated(changedProperties);
+
+    if (this.analysis == null) {
+      this.loadAnalysis();
+    }
+  }
+
+  private toDecibels(rmsDistanceRatio: number) {
+    return (10 * Math.log10(rmsDistanceRatio)).toFixed(2)
```
prideout, line 53: nit: I would've put the `toFixed` call inside the template literal rather than here.

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -39,27 +41,66 @@ export class RenderingScenario extends LitElement {
...
-</div>`);
+    const images =
+        [{
+          name: 'model-viewer',
+          description: '<model-viewer> (this version)',
+          file: 'model-viewer.png'
+        }].concat(this.goldens)
+            .filter(golden => !this.exclude.includes(golden.name))
+            .map(
+                (golden, index) => html`
```
prideout, line 73: I wonder if this is starting to reach critical mass in terms of the amount of embedded HTML.  For Filament's material debugger, I used `<template>` in some situations:

https://github.com/google/filament/blob/main/libs/matdbg/web/index.html#L27L39

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -31,6 +31,8 @@ export class RenderingScenario extends LitElement {
 
   @property({type: Array}) exclude: Array<string> = [];
 
+  @property({type: Object}) analysis: any|null;
```
bejado, line 34: Is `analysis` always an object? Maybe instead its type could be `object|null`.

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -39,27 +41,66 @@ export class RenderingScenario extends LitElement {
...
+                <img data-id="${this.name} ${golden.name}"
+                    style="width:${width}px" src="${basePath}/${golden.file}">
+                  <div class = "metrics">
+                    ${
+                    index == 0 || this.analysis == null ?
+                        html` <span> --- </span>` :
+                        html` <span>${
+                            this.toDecibels(
+                                this.analysis.analysisResults[index - 1][0]
+                                    .rmsDistanceRatio)} dB </span>`}
```
bejado, line 88: This part was somewhat difficult for me to parse. I would maybe move this logic out into its own function.

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -80,11 +121,49 @@ h1 {
...
+  background-image: url(../../shared-assets/icons/help_outline-24px.svg);
+  background-size : 0.8em;
+  background-repeat: no-repeat;
+  background-position: 0% 100%;
+  width: 1em;
+  height: 1em;
+  display: inline-block;
+}
+
+.metrics{
```
bejado, line 157: nit:
```suggestion
.metrics {
```

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -80,11 +121,49 @@ h1 {
...
+  padding: 5px 5px;
+  position: absolute;
+  z-index: 1;
+}
+
+.tooltip:hover .tooltiptext {
+  visibility: visible;
+}
+
+.question-icon{
```
bejado, line 147: nit:
```suggestion
.question-icon {
```

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -31,7 +31,7 @@ export class RenderingScenario extends LitElement {
 
   @property({type: Array}) exclude: Array<string> = [];
 
-  @property({type: Object}) analysis: any|null;
+  @property({type: Object}) analysis: ScenarioRecord|null = null;
```
sun765, line 34: Correct me if im wrong, i think creating a new interface for this object type is a better practice in ts.

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -80,22 +95,11 @@ export class RenderingScenario extends LitElement {
...
-                        and current version of model-viewer on rendering ${
-                    this.name} in decibels.
-                        The decibel is given by: 10 * log(RMS). More negative means a closer match.
-                      </span>
-                    </div>
+                               index == 0 || this.analysis == null ?
+                                   html`<span>---</span>` :
+                                   this.metricTemplate(
+                                       this.analysis.analysisResults[index - 1],
+                                       golden.name)}
```
sun765, line 102: @prideout  I saw you use Mustache template to make html very clean(I learn a lot from your code). But i found mustache doesn't support typescript, and since this web component is based on lit-element , it's not easy to use query selector (actually i didn't find out how to use it) here. So the best way i can think of to refactor this is just use a template function here and that's also the way I found in lit-element's document. If you have better way to do it im more than happy to learn

packages/render-fidelity-tools/src/components/rendering-scenario.ts
```diff
@@ -50,27 +50,42 @@ export class RenderingScenario extends LitElement {
...
 
   private async loadAnalysis() {
     const analysisPath = `${this.basePath}/analysis.json`;
     this.analysis = await (await fetch(analysisPath)).json();
   }
 
+  private metricTemplate(
+      analysisResults: Array<ImageComparisonAnalysis>, goldenName: string) {
+    const rmsInDecibel =
+        this.toDecibels(analysisResults[0].rmsDistanceRatio).toFixed(2)
```
sun765, line 64: In next PR, im going to remove a lot of things in analysisResults, so it will look nicer then.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -83,7 +83,7 @@ export class ArtifactCreator {
       const analysisResults = await this.analyze(
           screenshot, goldens, scenario, dimensions, analysisThresholds);
 
-      const scenarioRecord = Object.assign({analysisResults}, scenario);
+      const scenarioRecord = {analysisResults, scenario};
```
sun765, line 87: This may not be necessary, but i just want to keep the analysis result consistent with the ScenarioRecord interface i just created. i run the full fidelity test and found no errors. Also, correct me if im wrong, i think the scenario property is not necessary since i did find it's used anywhere

packages/render-fidelity-tools/src/components/rendering-scenario.ts
prideout: No no, lit-element looks great, sorry for the distraction.  :)

PR #1335, opened 2020-07-08
Remove skybox from fidelity tests

### LINKED ISSUE #1339: Question mark in missing on fidelity test page
### Description

Question marks that explains the metrics are missing. It probably has something to do with our current deploy actions.

sun765:
### Reference Issue
Fixes #1339

i only render and test one scenario (since the gltfpt takes too long to render goldens). After everything else is done i will renderer all the rest. 
image

Also, although i have removed skybox, when calculating rms im still taking those background pixels into acount. Do you want to exclude the background pixels when calculating rms?


packages/modelviewer.dev/scripts/ci-before-deploy.sh
```diff
@@ -53,6 +53,7 @@ DEPLOYABLE_STATIC_FILES=( \
   shared-assets/models/glTF-Sample-Models/2.0/Suzanne \
   shared-assets/models/glTF-Sample-Models/2.0/SpecGlossVsMetalRough \
   shared-assets/environments \
+  shared-assets/icons \
```
sun765, line 56: To deploy icons file to server

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -88,7 +88,10 @@ export class BabylonViewer extends LitElement {
 
     this[$updateSize]();
 
-    const {orbit, target, verticalFoV} = scenario;
+    const {orbit, target, verticalFoV, renderSkybox} = scenario;
+
+    this[$scene].clearColor = new Color4(1, 1, 1, 1);
```
sun765, line 93: At first i put clear color into scenario config, but i have to hard code filament and gltf-pt's clear color anyway, i just make all renderers' clear color hard coded as white

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -224,6 +226,12 @@ export class FilamentViewer extends LitElement {
 
     this[$updateSize]();
 
+    /*
+    this[$renderer].setClearOptions(
```
sun765, line 230: i keep the comment here because this is the syntax to set clear color in 1.70 or higher, i will clean this in next pr where i update filament to latest version

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -224,6 +226,12 @@ export class FilamentViewer extends LitElement {
...
+        {clearColor: [r, g, b, 1], clear: true, discard: true});
+    */
+    this[$view].setClearColor([1000, 1000, 1000, 1]);
```
sun765, line 233: Set it higher because of tone mapping

packages/render-fidelity-tools/test/config.json
```diff
@@ -23,6 +19,10 @@
     {
       "name": "babylon",
       "description": "Babylon"
+    },    
+    {
+      "name": "gltfpt",
+      "description": "gltf-pt"
```
sun765, line 25: Fidelity page align goldens according to the order of renderers here. i just want to put the gltfpt goldens to right most

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
prideout: Ok, understood.

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
elalish: This is a good comment to include in the code.

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -224,6 +226,12 @@ export class FilamentViewer extends LitElement {
...
+
     // Wait two rAFs to ensure we rendered at least once:
     requestAnimationFrame(() => {
       requestAnimationFrame(() => {
```
elalish, line 237: Does it work with just one raf? If so, let's simplify; it may also speed up rendering a bit.

elalish: Same for the other renderers.

packages/render-fidelity-tools/test/config.json
```diff
@@ -437,7 +439,8 @@
       "exclude": [
         "gltfpt",
         "babylon"
-      ]
+      ],
+      "renderSkybox": true
```
elalish, line 443: Can you also add the skybox back to the furnace test and the directional light test? Their backgrounds are simple, but helpful.

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -65,8 +65,16 @@
       modelViewer.style.width = `${scenario.dimensions.width}px`;
       modelViewer.style.height = `${scenario.dimensions.height}px`;
 
+      modelViewer.style.backgroundColor = `rgb(${255},${255},${255})`;
+
       modelViewer.src = scenario.model;
-      modelViewer.skyboxImage = scenario.lighting;
+      if(scenario.renderSkybox){
```
bejado, line 71:
```suggestion
      if (scenario.renderSkybox) {
```

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -65,8 +65,16 @@
...
+        modelViewer.skyboxImage = scenario.lighting;
+      }
+      else{
```
bejado, line 74:
```suggestion
      else {
```

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -65,8 +65,16 @@
       modelViewer.style.width = `${scenario.dimensions.width}px`;
       modelViewer.style.height = `${scenario.dimensions.height}px`;
 
+      modelViewer.style.backgroundColor = `rgb(${255},${255},${255})`;
```
bejado, line 68: Use a regular string literal here?

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -135,10 +138,12 @@ export class BabylonViewer extends LitElement {
...
     environment.setReflectionTextureMatrix(
         Matrix.RotationY(Tools.ToRadians(90)));
-    const skybox =
-        this[$scene].createDefaultSkybox(this[$scene].environmentTexture!);
-    skybox!.rotate(Axis.Y, Math.PI * 1.5, Space.WORLD);
-    skybox!.infiniteDistance = true;
+    if (renderSkybox) {
+      const skybox =
+          this[$scene].createDefaultSkybox(this[$scene].environmentTexture!);
+      skybox!.rotate(Axis.Y, Math.PI * 1.5, Space.WORLD);
```
bejado, line 144: I know it was like this before, but I'm confused why you have to do `skybox!`. Is it possible `createDefaultSkybox` returns `null`?

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
prideout: As follow up to our discussion today. note that all Filament actions must take place after the init's callback function has been triggered, as per https://google.github.io/filament/webgl/reference.html#init.

elalish: Yes, and also let the record show that this was my bug from when I first integrated filament.js; apparently I was just lucky enough to never trigger a race.

sun765:
I feel like this pr is bigger than necessary, it's because i thought not including background pixels when computing metrics(the original goal of this pr) will just take a few lines of codes, but it actually need a lot of refactor. Also, it's more difficult than i thought, the method i have been trying didn't work. In case this pr becomes further messy, i just stopped right here and push it. Next time i will definitely make each pr only focus on a small job.

Here's what i did since last review:
1.  remove all old metrics and thresholds
2. seperate analyze and generateVisual function
3. remove two rafs and run the full test, no errors
4. update all goldens

One thing i notice is that there're a lot more - infinite in the fidelity page. i guess it's because i make the background all the same without excluding them in rms, which makes the rms very small.  i think it will be fine after i actually exclude background pixels in the calculation.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -25,19 +25,19 @@ export const COMPONENTS_PER_PIXEL: number = 4;
...
-  analysis: ImageComparisonAnalysis;
+  analysis: ImageComparisonAnalysis
+}
+
+export interface Visuals {
   imageBuffers: {delta: ArrayBuffer|null; blackWhite: ArrayBuffer | null;};
 }
 
 export interface ScenarioRecord {
-  analysisResults: Array<Array<ImageComparisonAnalysis>>;
```
sun765, line 40: the analysisResult used to be an array of ImageComparisonAnalysis calculated in many thresholds. Now i removed all thresholds based metrics, so it should be only one ImageComparisonAnalysis with only a rms in it.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -139,26 +140,17 @@ export class ImageComparator {
     image[position + 3] = a;
   }
 
-  analyze(threshold: number, options: AnalysisOptions = {
```
sun765, line 142: analyze and generate visual are in the same functions before, probably because they both loop through all pixels in a given threshold. Now i remove threshold in analyze function, if feel like they have less in common and just seperate them.

packages/render-fidelity-tools/src/components/renderers/gltfpt-viewer.ts
```diff
@@ -107,12 +112,9 @@ export class PathtracingViewer extends LitElement {
           console.log('frame finished ' + frame);
         },
         () => {
-          // Wait two rAFs to ensure we rendered at least once //TODO: clarify
           requestAnimationFrame(() => {
```
sun765, line 115: Run the full test, no errors found. The same for filament-viewer

sun765:
Fix metrics error by correct the dimensions in fidelity analysis.

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
sun765: That's really a good question and i looked into that after you asked. the only way skybox can be null is when this[$scene].environmentTexture is null. the created environment variable cannot be null (if there's an error on creating that there'll be error and it will just stopped at that point.). It's just when assign environment to this[$scene].environmentTexture, environment is cast from BasetTexture to Nullable<BasetTexture>, that's why i need to put! after it. Don't know why they to this cast. Also, not sure if there's better way to handle null exception here. Do you have any idea about that? Thank you!

sun765: from BaseTexture to Nullable<BaseTexture>

sun765: from BaseTexture to Nullable(BaseTexture) , not sure why <>  is not shown correctly

bejado: Okay, I see. Thanks for the explanation. Maybe we can refactor in the future, in general it's probably best to use `!` as sparingly as possible. But I'm okay with this for now.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -24,20 +24,22 @@ export const COMPONENTS_PER_PIXEL: number = 4;
 // @see http://www.progmat.uaem.mx:8080/artVol2Num2/Articulo3Vol2Num2.pdf
 export const MAX_COLOR_DISTANCE: number = 35215;
 
+export const DEVICE_PIXEL_RATIO: number = 2;
```
prideout, line 27: Is this correct on non high DPI displays?

packages/render-fidelity-tools/src/common.ts
```diff
@@ -174,80 +168,79 @@ export class ImageComparator {
...
+        maximumDeltaIntensity = Math.max(deltaIntensity, maximumDeltaIntensity);
+
+        this.drawPixel(
+            blackWhiteImage!,
+            position,
+            exactlyMatched,
+            exactlyMatched,
+            exactlyMatched);
+        this.drawPixel(
+            deltaImage!,
```
prideout, line 185: Are all the bangs after `deltaImage` still necessary, now that it is created unconditionally?

packages/render-fidelity-tools/src/common.ts
```diff
@@ -174,80 +168,79 @@ export class ImageComparator {
...
-
     return {
-      analysis: {
-        matchingRatio: matched / this.imagePixels,
-        averageDistanceRatio,
-        mismatchingAverageDistanceRatio,
-        rmsDistanceRatio
-      },
       imageBuffers: {
         delta: deltaImage ? deltaImage.buffer as ArrayBuffer : null,
```
prideout, line 212: At first glance it looks like deltaImage can never be null but maybe I'm missing something.  I'm also surprised that its buffer field needs to be casted to `ArrayBuffer`.

packages/render-fidelity-tools/src/common.ts
elalish: Since this is headless in puppeteer, the DPR is not related to a physical display. We only choose to use a non-unity value because handling DPR is bug-prone and this helps with noticing errors.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -174,80 +168,79 @@ export class ImageComparator {
...
+
+  analyze(): ImageComparisonResults {
+    const {candidateImage, goldenImage} = this;
+    const {width, height} = this.dimensions;
+
+    let squareSum = 0;
+
+    if (candidateImage.length != goldenImage.length) {
+      throw new Error(`Image sizes do not match (candidate: ${
+          candidateImage.length}, golden: ${goldenImage.length})`);
```
elalish, line 227: 💯

packages/render-fidelity-tools/src/common.ts
sun765: Thank you for pointing this out. Although I just copied this code from original project, i found it wired too after you mentioned. i removed the post-fix and casting to ArrayBuffer, no compiling errors. Thank you for pointing out this !

PR #1340, opened 2020-07-09
Configure Babylonjs with ACES tone mapping

### LINKED ISSUE #1337: Tone mapping is not configured for babylonjs in fidelity test
### Description

babylonjs supports ACES tone mapping, but it's not configured to that in the fidelity test.

sun765:
### Reference Issue
Fixes #1337 

Enable ACES tone mapping for babylonjs. However, the rendering results look darker than before, so im not sure if it's supposed to be.

Before enabling tone mapping:
Screen Shot 2020-07-09 at 4 26 38 PM

After enabling tone mapping:
Screen Shot 2020-07-09 at 4 26 12 PM

Also, for the other renderers, the metrics are not exactly the same,  im not sure if it's an issue. (The first screenshot is taken from my local machine, the second is taken in the dev pages' fidelity page)


packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -78,6 +78,10 @@ export class BabylonViewer extends LitElement {
       this[$scene].dispose();
     }
     this[$scene] = new Scene(this[$engine]);
+    this[$scene].imageProcessingConfiguration.toneMappingEnabled = true;
+    this[$scene].imageProcessingConfiguration.toneMappingType =
+        ImageProcessingConfiguration.TONEMAPPING_ACES;
```
elalish, line 83: Let's adjust the exposure as well to see if we can get the renders to line up better. You can find background on why this may be necessary here: https://github.com/mrdoob/three.js/pull/19621. From that discussion, 1/0.6 might be the right factor, but please play around a bit to minimize error (just pick a few example scenes; maybe the helmets).

elalish:
There will be slight variations in metrics due to different GPUs used, but going from -32 to -34 is pretty large; you may want to post updated model-viewer goldens, since we may be out of date (perhaps from a three.js upgrade). If you put them in this PR, we'll be able to check the difference using Github's "rich diff" tool.

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -77,7 +77,16 @@ export class BabylonViewer extends LitElement {
...
     }
+
+
     this[$scene] = new Scene(this[$engine]);
+    this[$scene].imageProcessingConfiguration.toneMappingEnabled = true;
+    this[$scene].imageProcessingConfiguration.toneMappingType =
+        ImageProcessingConfiguration.TONEMAPPING_ACES;
+    // tone mapping makes a shift on explosure, so it should be increased be
+    // some factor
+    this[$scene].environmentIntensity = 1 / 0.6;
```
sun765, line 88: I tried some different factor, but the the skybox's intensity is not changed by this value, and the existence of skybox makes it really hard to tell what's the best exposure.  i think it's better to play with the exposure in my disable-skybox pr. Is that ok for you?

sun765:
After updating the model-viewer glodens, some rms of model-viewer compared with latest mode-viewer becomes infinite. is it ok to just show infinite on metric? Or do you want to set some max value?

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
elalish: Interesting; can you ask the babylon folks if there is a different parameter than exposure that does affect the background? Clearly their tone mapping does, and often there is a separate parameter for this. Recall that we will be keeping backgrounds in at least a few of our tests. That's part of why I wanted this done in this PR, so we get the backgrounds right.

elalish:
Yes, -Infinity is great.

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -77,7 +77,15 @@ export class BabylonViewer extends LitElement {
...
       this[$scene].dispose();
     }
+
     this[$scene] = new Scene(this[$engine]);
+    this[$scene].imageProcessingConfiguration.toneMappingEnabled = true;
+    this[$scene].imageProcessingConfiguration.toneMappingType =
+        ImageProcessingConfiguration.TONEMAPPING_ACES;
+    // tone mapping makes a shift on explosure, so it should be increased be
+    // some factor
+    this[$scene].imageProcessingConfiguration.exposure = 2.0;
```
sun765, line 87: exposure = 1.8: (very simiar to 1.67, 1.4 is a little dark)


exposure = 2.0


exposure = 2.2 (2.4 is obviously bright)


Although 1.8 looks better in terms metric, but if feel like 2.0 looks closer to other renderers. Do you agree with it?

elalish: Yeah, 2.0 looks pretty good to me. We'll be updating three.js and filament soon anyway, and they both contain tone mapping changes, so we'll probably have to revisit this anyway. This looks like a good baseline.

PR #1344, opened 2020-07-10
Examples for scene-graph

hjeldin:
Added some basic examples to directly manipulate the scene graph:

- Metalness factor
- Roughness factor
- Diffuse texture
- Normal texture
- Metallic-Roughness
- Occlusion
- Emission


packages/modelviewer.dev/examples/scene-graph.html
```diff
@@ -105,11 +105,162 @@ <h2 class="demo-title">Change Material Base Color</h2>
...
+    <input id="metalness" type="range" min="0" max="1" step="0.01" value="1" />
+    <input id="roughness" type="range" min="0" max="1" step="0.01" value="0" />
+  </div>
+</model-viewer>
+<script>
+const modelViewerParameters = document.querySelector("model-viewer#sphere");
+
+document.querySelector('#metalness').addEventListener('input', (event) => {
+  const [material] = modelViewerParameters.model.materials;
+  console.log(event.target.value);
```
elalish, line 129: I think we can remove this log.

packages/modelviewer.dev/examples/scene-graph.html
```diff
@@ -105,11 +105,162 @@ <h2 class="demo-title">Change Material Base Color</h2>
...
+  material.pbrMetallicRoughness.setMetallicFactor(event.target.value);
+});
+
+document.querySelector('#roughness').addEventListener('input', (event) => {
+  const [material] = modelViewerParameters.model.materials;
```
elalish, line 134: This can be defined outside of the function scope to save duplication.

packages/modelviewer.dev/examples/scene-graph.html
```diff
@@ -105,11 +105,162 @@ <h2 class="demo-title">Change Material Base Color</h2>
...
+        <div class="heading">
+          <h2 class="demo-title">Change Material Metalness and Roughness Factors</h2>
+        </div>
+        <example-snippet stamp-to="demo-container-2" highlight-as="html">
+          <template>
+
+<model-viewer id="sphere" camera-controls interaction-prompt="none" src="../shared-assets/models/reflective-sphere.gltf" alt="A 3D model of a sphere">
+  <div id="controls">
+    <input id="metalness" type="range" min="0" max="1" step="0.01" value="1" />
+    <input id="roughness" type="range" min="0" max="1" step="0.01" value="0" />
```
elalish, line 121: Can we display names and values for these?

packages/modelviewer.dev/examples/scene-graph.html
```diff
@@ -105,11 +105,162 @@ <h2 class="demo-title">Change Material Base Color</h2>
...
+  </div>
+</model-viewer>
+<script>
+const modelViewerParameters = document.querySelector("model-viewer#sphere");
```
elalish, line 125: Let's change the base color of the sphere to something colorful (gold?), as that will make the difference between metal and dielectric more obvious.

packages/modelviewer.dev/examples/scene-graph.html
```diff
@@ -105,11 +105,162 @@ <h2 class="demo-title">Change Material Base Color</h2>
...
+          <h2 class="demo-title">Swap textures - Diffuse/Normal/MetallicRoughness</h2>
+        </div>
+        <example-snippet stamp-to="demo-container-3" highlight-as="html">
+          <template>
+
+<model-viewer id="lantern" camera-controls interaction-prompt="none" src="../shared-assets/models/glTF-Sample-Models/2.0/Lantern/glTF-Binary/Lantern.glb" alt="A 3D model of a lantern">
+  <div id="controls">
+    <div>
+      <p>Diffuse</p>
+      <input id="diffuse" type="file" accept="image/png, image/jpeg"/>
```
elalish, line 158: I'm open to push-back here, but I think this might be a clearer example if people don't have to bring their own textures. Could we just grab two or three interesting textures and make a drop-down to select between them for this and the following example?

packages/modelviewer.dev/examples/scene-graph.html
```diff
@@ -105,11 +105,162 @@ <h2 class="demo-title">Change Material Base Color</h2>
...
+  </div>
+</model-viewer>
+<script type="module">
+const modelViewerTexture = document.querySelector("model-viewer#lantern");
+
+document.querySelector('#diffuse').addEventListener('input', (event) => {
+  const [material] = modelViewerTexture.model.materials;
+  if(event.target.files.length > 0){
+    var url = URL.createObjectURL(event.target.files[0]);
+    material.pbrMetallicRoughness.baseColorTexture.texture.source.setURI(url);
```
elalish, line 177: We want to keep the sample code as concise as possible; could we combine these functions into one by using a parameter to select between texture targets? It might be simpler if we move the normal texture to the next  example.

packages/modelviewer.dev/examples/scene-graph.html
hjeldin: I've been thinking about this: giving the users a dropdown to choose from may be misinterpreted as a simple model swap under the hood...but yeah, the code should be enough to judge that it's not the case.

PR #1346, opened 2020-07-13
Add npm scripts for dev watchers, for editing-adapter and space-opera

stevesan:


packages/editing-adapter/package.json
```diff
@@ -11,7 +11,14 @@
     "clean": "rm -rf lib && rm -rf dist",
     "test": "karma start --single-run",
     "test:ci": "npm run test",
-    "update:package-lock": "rm ./package-lock.json && npm i --only=production"
+    "update:package-lock": "rm ./package-lock.json && npm i --only=production",
+    "watch:test": "karma start",
+    "watch:tsc": "tsc -w",
+    "watch:rollup": "rollup -c -w",
+    "serve": "./node_modules/.bin/http-server -a 127.0.0.1 -o /demo/ -c-1",
+    "pyserve": "echo This assumes 'python' is Python 2. && which python && echo Visit $(hostname):8000/example.html && python -m SimpleHTTPServer 8000",
```
elalish, line 19: What does pyserve do for you that http-server does not? Not a huge deal, but it's clearly pulling down a lot more dev dependencies.

elalish: Or maybe I've read this wrong? Does pyserve just assume python is installed separately from npm?

stevesan: pyserve works on my remote linux system - that's all. couldn't figure out why http-server didn't work.

stevesan: (yeah, you'd have to have python already installed for this to work, thus the "which" check)

PR #1348, opened 2020-07-13
Space-opera: Remove css files

stevesan:


stevesan:
Should of created PR after merging previous..

PR #1359, opened 2020-07-14
Add space-opera to modelviewer.dev

stevesan:


packages/modelviewer.dev/index.html
```diff
@@ -85,8 +85,10 @@ <h3 class="grouping-title border-bottom">Examples</h3>
             <li><a href="examples/customizing-ui.html"><h4>Customizing UI</h4></a></li>
 
             <li><a href="examples/scene-graph.html"><h4>Scene Graph</h4></a></li>
-			
-			      <li><a href="examples/tester.html"><h4>Interactive Example</h4></a></li>
+
+            <li><a href="examples/tester.html"><h4>Interactive Example</h4></a></li>
+
+            <li><a href="./node_modules/@google/model-viewer-space-opera/index.html"><h4>Space Opera (Editor)</h4></a></li>
```
elalish, line 91: I don't think we need to publicize the Space Opera name quite this widely :). Maybe just call it Model Editor?

packages/modelviewer.dev/index.html
```diff
@@ -132,11 +134,11 @@ <h3 class="grouping-title">Attributes</h3>
             </li>
             <li>
               <div>ar-modes</div>
-              <p>A prioritized list of the types of AR experiences to enable, if available; see 
+              <p>A prioritized list of the types of AR experiences to enable, if available; see
```
elalish, line 137: joys of auto-format...

packages/editing-adapter/src/util/create_object_url.ts
```diff
@@ -35,7 +35,11 @@ export function createSafeObjectURL(blob: Blob): SafeObjectUrl {
 
 /** Returns true if the given raw URL is an object URL. */
 export function isObjectUrl(url: string): boolean {
-  return new URL(url).protocol === 'blob:';
+  try {
+    return new URL(url).protocol === 'blob:';
+  } catch (_) {
+    return false;
+  }
```
elalish, line 42: It looks like you branched off your other PR instead of off master.

packages/modelviewer.dev/index.html
stevesan: yeah..i think i'm using clang format extension in VSCode? James said you do have .clang-format, so is that the "official format" we should use?

packages/modelviewer.dev/index.html
stevesan: Done

packages/editing-adapter/src/util/create_object_url.ts
stevesan: pulled and merged.

packages/modelviewer.dev/index.html
elalish: Yeah, that's right. It's possible no one has touched this file for awhile.

PR #1363, opened 2020-07-15
Fix "Model Editor" link on modelviewer.dev

stevesan:


packages/modelviewer.dev/scripts/ci-before-deploy.sh
```diff
@@ -53,6 +53,13 @@ DEPLOYABLE_STATIC_FILES=( \
   shared-assets/models/glTF-Sample-Models/2.0/Suzanne \
   shared-assets/models/glTF-Sample-Models/2.0/SpecGlossVsMetalRough \
   shared-assets/environments \
+  node_modules/@google/model-viewer-space-opera/node_modules/web-animations-js/web-animations-next-lite.min.js \
+  node_modules/@google/model-viewer-space-opera/node_modules/js-beautify/js/lib/beautify-html.js \
+  node_modules/@google/model-viewer-space-opera/node_modules/js-beautify/js/lib/beautify-css.js \
+  node_modules/@google/model-viewer-space-opera/index.html \
+  node_modules/@google/model-viewer-space-opera/node_modules/@google/model-viewer-shared-assets/environments/spruit_sunrise_1k_HDR.hdr \
```
stevesan, line 60: Not crazy about this duplication of HDRs. Not sure about a better solution...for resolving these paths.

elalish: This gets back to the earlier comment I left you about paths. I believe there is a shared-assets symlink in space-opera already, like the rest of our packages. I think you can use that instead of the module path. Can you see if there's a way following how the other packages reference shared-assets?

stevesan: ok lemme try referring to shared_assets directly in both settings...

stevesan: ok this seems to work fine.

packages/space-opera/src/redux/initial_environment_images.ts
```diff
@@ -22,11 +22,11 @@ import {EnvironmentImage} from './lighting_state.js';
  */
 export const INITIAL_ENVIRONMENT_IMAGES: EnvironmentImage[] = [
   {
-    uri: './node_modules/@google/model-viewer-shared-assets/environments/spruit_sunrise_1k_HDR.hdr',
+    uri: '/shared-assets/environments/spruit_sunrise_1k_HDR.hdr',
     name: 'Spruit Sunrise'
   },
   {
-    uri: './node_modules/@google/model-viewer-shared-assets/environments/aircraft_workshop_01_1k.hdr',
+    uri: '/shared-assets/environments/aircraft_workshop_01_1k.hdr',
```
elalish, line 29: Sure you don't need a `./` here?

stevesan: nope, this explicitly resolves to the root of the web server.

PR #1369, opened 2020-07-16
Make sure we remove tsbuildinfo cache upon clean

stevesan:
..otherwise, build won't work after a clean, because tsc --incremental will assume there's nothing to be done (given the tsbuildinfo is still there).

### DIFF
--- a/packages/editing-adapter/package.json
+++ b/packages/editing-adapter/package.json
@@ -8,7 +8,7 @@
     "build": "npm run build:tsc && npm run build:rollup",
     "build:tsc": "tsc --incremental && cp src/gltf/*.d.ts lib/gltf",
     "build:rollup": "rollup -c --environment NODE_ENV:production",
-    "clean": "rm -rf lib && rm -rf dist",
+    "clean": "rm tsconfig.tsbuildinfo; rm -rf lib && rm -rf dist",
     "test": "karma start --single-run",
     "test:ci": "npm run test",
     "update:package-lock": "rm ./package-lock.json; npm i --only=production",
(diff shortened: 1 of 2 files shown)


packages/editing-adapter/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "npm run build:tsc && npm run build:rollup",
     "build:tsc": "tsc --incremental && cp src/gltf/*.d.ts lib/gltf",
     "build:rollup": "rollup -c --environment NODE_ENV:production",
-    "clean": "rm -rf lib && rm -rf dist",
+    "clean": "rm tsconfig.tsbuildinfo; rm -rf lib && rm -rf dist",
```
elalish, line 11: Why is this specific to these two packages?

packages/space-opera/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "npm run build:tsc && npm run build:rollup",
     "build:tsc": "tsc --incremental",
     "build:rollup": "rollup -c --environment NODE_ENV:production",
-    "clean": "rm -rf lib && rm -rf dist",
+    "clean": "rm tsconfig.tsbuildinfo; rm -rf lib && rm -rf dist",
```
elalish, line 11: Should we also delete the tsbuildinfo files and .gitignore them?

packages/editing-adapter/package.json
stevesan: Well, the only other package that uses --incremental is modelviewer...but for some reason that's not an issue? May be a question for @cdata

packages/space-opera/package.json
stevesan: They are already in .gitignore

PR #1370, opened 2020-07-16
Add "revert" buttons to all material properties, which revert to values of the original GLB

stevesan:


packages/space-opera/src/redux/space_opera_base.ts
```diff
@@ -38,6 +38,8 @@ export interface State {
   animationNames: string[];
   gltfJsonString: string;
   edits: GltfEdits;
+  // A copy of the original, so we can revert individual properties.
+  origEdits: GltfEdits;
```
elalish, line 42: Not sure these are edits, strictly speaking, but I see that's what we call the type.

PR #1378, opened 2020-07-17
Changes to poster buttons

stevesan:


packages/space-opera/src/components/poster_controls/poster_controls.ts
```diff
@@ -79,6 +90,16 @@ export class PosterControlsElement extends ConnectedLitElement {
...
   }
+
+  onDeletePoster() {
+    dispatchSetPoster(undefined);
+  }
+
+  async onDownloadPoster() {
+    if (!this.modelViewer)
+      return;
+    safeDownloadCallback(await this.modelViewer.toBlob(), 'poster.png', '')();
```
elalish, line 101: Shouldn't this download the poster you already created rather than making a new one?

elalish: Also, for posters you need the toBlob() option `idealAspect = true`. See tester.ts in modelviewer.dev.

stevesan: both addressed

PR #1388, opened 2020-07-20
add check to radius for babylon and update goldens

sun765:
Assign a very small non-zero value to radius for babylon's camera.  Here's how it looks like in fidelity test page: 
image


packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -98,7 +98,10 @@ export class BabylonViewer extends LitElement {
         'Camera',
         alpha,
         beta,
-        orbit.radius,
+        Math.max(
+            orbit.radius,
+            0.0001),  // TODO: babylon doesn't support 0 radius for now, and
```
sun765, line 103: i just used an arbitrary small value. If you don't like this value, im happy to change

PR #1393, opened 2020-07-21
Update and refactor filament

### LINKED ISSUE #1338: Filament is not updated to the latest version
### Description

The fidelity test application is based on 1.5.2 version of filament, while it's latest version is 1.8.0.  It should be updated to latest version, and some configuration should be changed accordingly when rendering filament's golden.

### LINKED ISSUE #1382: Warning and errors on filament's rendering page
There're waring about WebAssembly and error in filament init function in filament's rendering page.  But everything works fine in this page. Also, not sure whether this is related to #1383  issue or not.

image

### LINKED ISSUE #1383: Filament-viewer sometimes fails at rendering goldens
Filament-viewer sometimes fails at rendering goldens, with the error that it's trying to call method of undefined.

Here's probable reason: according to filament's document, https://google.github.io/filament/webgl/reference.html#init,  all client side call should be put in init function's onReady callback, but this is not how the current code is written. It needs to be refactored according to filament's requirement.

sun765:
Fixes #1338 
Fixes #1382 
Fixes #1383 

1.  Updated filament to latest version (1.8.0)
2. Fixed the race condition error when update goldens 
3. TODO: use destroyAsset() to fix memory leadk

### DIFF
--- a/packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
+++ b/packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
@@ -113,11 +113,13 @@ export class FilamentViewer extends LitElement {
     const engine = Engine.create(this[$canvas]!);
     const view = engine.createView();
 
+    const entityManager = EntityManager.get();
+    const emptyEntity = entityManager.create();
     this[$engine] = engine;
     this[$scene] = engine.createScene();
     this[$swapChain] = engine.createSwapChain();
     this[$renderer] = engine.createRenderer();
-    this[$camera] = engine.createCamera();
+    this[$camera] = engine.createCamera(emptyEntity);
     this[$view] = view;
     view.setCamera(this[$camera]);
     view.setScene(this[$scene]);
@@ -142,10 +144,10 @@ export class FilamentViewer extends LitElement {
 
     if (this[$currentAsset] != null) {
       const entities = this[$currentAsset]!.getEntities();
-      const size = entities.size();
+      const size = entities.length;
 
       for (let i = 0; i < size; ++i) {
-        const entity = entities.get(i);
+        const entity = entities[i];
         this[$scene].remove(entity);
         this[$engine].destroyEntity(entity);
       }
@@ -226,14 +228,10 @@ export class FilamentViewer extends LitElement {
 
     this[$updateSize]();
 
-    /*
-    this[$renderer].setClearOptions(
-        {clearColor: [r, g, b, 1], clear: true, discard: true});
-    */
-
     // because of tone mapping, white should be higher than(1,1,1). set to 1000
     // just to make sure it's white
-    this[$view].setClearColor([1000, 1000, 1000, 1]);
+    this[$renderer].setClearOptions(
+        {clearColor: [1000, 1000, 1000, 1], clear: true, discard: true});
 
     requestAnimationFrame(() => {
       this.dispatchEvent(
--- a/packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
+++ b/packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
@@ -77,7 +77,6 @@ export class FilamentViewer extends LitElement {
   constructor() {
     super();
 
-    // is it okay to only have a resolve, not reject?
     this[$initPromise] = new Promise((resolve) => {
       init([], () => {
         this[$initialize]();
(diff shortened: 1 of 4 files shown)


packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -71,12 +72,17 @@ export class FilamentViewer extends LitElement {
...
 
+  private[$initPromise]: Promise<void>|null = null;
+
   constructor() {
     super();
 
-    init([], () => {
-      this[$initialize]();
-    });
+    this[$initPromise] = new Promise((resolve) => {
```
sun765, line 80: Not sure if i used promise correctly. Also, is it okay if a promise doesn't call reject()?

packages/render-fidelity-tools/test/config.json
```diff
@@ -766,8 +766,7 @@
       },
       "exclude": [
         "gltfpt",
-        "babylon",
-        "filament"
+        "babylon"
```
sun765, line 769: after updating filament, there's no errors on this scenario anymore.

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -201,9 +211,14 @@ export class FilamentViewer extends LitElement {
       this[$ibl] = ibl
       ibl.setIntensity(1.0);
       ibl.setRotation([0, 0, -1, 0, 1, 0, 1, 0, 0]);  // 90 degrees
+
       if (scenario.renderSkybox) {
         this[$skybox] = this[$engine].createSkyFromKtx(skyboxUrl);
         this[$scene].setSkybox(this[$skybox]);
+      } else {
+        this[$view].setBlendMode(View$BlendMode.TRANSLUCENT);
```
sun765, line 219: Transparent routes works perfectly well for me. I did try use this[$skybox].setColor([0,0,0,1]), it didn't work and the background is still skybox's color. But if you don't think this is a problem then just ignore it.

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -110,14 +116,16 @@ export class FilamentViewer extends LitElement {
...
     const view = engine.createView();
 
+    const entityManager = EntityManager.get();
+    const emptyEntity = entityManager.create();
     this[$engine] = engine;
     this[$scene] = engine.createScene();
     this[$swapChain] = engine.createSwapChain();
     this[$renderer] = engine.createRenderer();
-    this[$camera] = engine.createCamera();
+    this[$camera] = engine.createCamera(emptyEntity);
```
sun765, line 128: Not sure what entity im supposed to put into camera's constructor. Is it ok to just put an empty entity?

prideout: Yes this is correct.  We're trying to make our ECS more self-consistent; here you're adding a camera component to a blank entity.

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -140,12 +149,13 @@ export class FilamentViewer extends LitElement {
     console.log('Scenario:', scenario.name);
     console.log('Lighting:', lightingBaseName);
 
+    // TODO : replace this with destroyAsset() when it's released
```
prideout, line 152: Yup.  My fault.  Unfortunately this means you'll need to upgrade to 1.8.1 as soon as we release it.

By the way, the `destroyAsset` method will not only destroy non-ECS objects (e.g. VertexBuffer and Texture), it will also destroy entities.  So, it will allow you to remove your existing destroyEntity loop.  You will, however, still need to explicitly remove those entities from the Scene.  We'll also be exposing a `removeEntities` method to make this easier.

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
prideout: This seems fine for now.

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
elalish: This is probably fine; I think the problem is just that this'll hang if the $initialize function hangs. The robust way is to wrap it in a setTimeout that will reject if it takes too long, but hopefully we don't require that kind of hardening here.

packages/render-fidelity-tools/test/config.json
elalish: Thanks!

PR #1394, opened 2020-07-22
set timer for model-viewer's model-ready event

sun765:
A temporary workaround for model-viewer's golden is taken before the model is actually loaded.

### DIFF
--- a/packages/render-fidelity-tools/test/renderers/model-viewer/index.html
+++ b/packages/render-fidelity-tools/test/renderers/model-viewer/index.html
@@ -58,7 +58,11 @@
 
       modelViewer.addEventListener('model-visibility', (event) => {
         if (event.detail.visible) {
-          self.dispatchEvent(new CustomEvent('model-ready'));
+          //TODO: this is a temporary workaround for taking screenshots before the model has been loaded
+          //This will be fixed when the screenshot function is refactored.
+          setTimeout(()=>{
+            self.dispatchEvent(new CustomEvent('model-ready'));
+          }, 300);
         }
       }, { once: true });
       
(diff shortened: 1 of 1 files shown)


packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -58,7 +58,11 @@
 
       modelViewer.addEventListener('model-visibility', (event) => {
         if (event.detail.visible) {
-          self.dispatchEvent(new CustomEvent('model-ready'));
+          //TODO: this is a temporary workaround for taking screenshots before the model has been loaded
+          //This will be fixed when the screenshot function is refactored.
+          setTimeout(()=>{
+            self.dispatchEvent(new CustomEvent('model-ready'));
+          }, 300);
```
sun765, line 65: Is 300ms good enough? Do you want it to be higher?

elalish: Yeah, 300ms is the minimum, but you might want to put 500 just to give yourself some headroom. More importantly though, I think this timeout should be in your raf callback instead, as it can take some time to render that first frame and I think we want to wait after that.

sun765: there isn't raf callback in the original script -- there isn't a model-viewer.ts file at all. I guess it's because model-viewer is already a web-component so there's we need to do with it as what we did for filament. Do you want me to add a raf call back for model-viewer?

elalish: No problem; I forgot we do this internally in model-viewer already.

PR #1395, opened 2020-07-22
Fix scene-graph for Safari

### LINKED ISSUE #1379: Glitch shows partial renders outside modelviewer in Safari on MacOS and iOS
Hi, I'm experiencing an issue where I have multiple models on a page in Safari (on both MacOS and iOS) and have a strange glitch (not sure how else to describe) where there seem to be multiple renders of each model happening outside the modelviewer itself. The models are set to autorotate and the glitch appears once rotation starts.

An image attached which hopefully gives the idea. You can see the straight line underneath the blue glow effect which shows where the modelviewer ends, and how the part renders appear outside of this.

Is this a known issue as can't find it in previous logged issues.

Thanks!

### Browser Affected

- [ ] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [x] Safari

### OS

- [ ] Android
- [x] iOS
- [ ] Linux
- [x] MacOS
- [ ] Windows

### Versions
Using this version of modelviewer:
unpkg.com/@google/model-viewer/dist/model-viewer.min.js

Screenshot 2020-07-17 at 17 56 56

elalish:
Fixes #1379

This fix was just a one-liner; apparently `overflow:hidden` is only respected on Safari in combination with `position:relative`. In the process I discovered a larger issue where the scene-graph was not working in Safari, which is also fixed here. Safari doesn't implement EventTarget, so I had to copy a hack from another part of model-viewer. What troubles me most is that our scene-graph-spec should have caught this on our Safari tests. I still have no idea how it passed.

### DIFF
--- a/packages/3dom/src/facade/api.ts
+++ b/packages/3dom/src/facade/api.ts
@@ -131,7 +131,7 @@ export interface Model extends ThreeDOMElement {
  *
  * @see ../context.ts
  */
-export interface ModelGraft extends EventTarget {
+export interface ModelGraft {
   /**
    * A flat list of all unique materials found in this scene graph.
    *
--- a/packages/3dom/src/facade/three-js/model-graft.ts
+++ b/packages/3dom/src/facade/three-js/model-graft.ts
@@ -23,6 +23,7 @@ import {ThreeDOMElement} from './three-dom-element.js';
 const $model = Symbol('model');
 const $correlatedSceneGraph = Symbol('correlatedSceneGraph');
 const $elementsByInternalId = Symbol('elementsByInternalId');
+const $eventDelegate = Symbol('eventDelegate');
 
 /**
  * ModelGraft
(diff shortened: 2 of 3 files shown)


erik-valtechsd:
Hi - I am using cdn to load model-viewer, do you know how I can get this fix via cdn? This is still an issue for me. Thanks 

```
  <script type="module" src="https://unpkg.com/@google/model-viewer/dist/model-viewer.min.js"></script>
  <script nomodule src="https://unpkg.com/@google/model-viewer/dist/model-viewer-legacy.js"></script>
```

elalish:
Don't worry, I'll get a v1.1 release out shortly 😁. If you want to test in the meanwhile, you can pull straight from our docs page, which serves master: `<script type="module" src="https://modelviewer.dev/node_modules/@google/model-viewer/dist/model-viewer.min.js></script>`

PR #1400, opened 2020-07-23
Detect background pixels

sun765:
1. Omit all transparent pixels when taking screenshots of goldens
2. When calculating metrics, ignore transparent(background) pixels.

Since i have made model-viewer's goldens' background transparent, to make it consistent, i also update babylon's goldens to be transparent. Filament's goldens will be transparent after the reactor-filament pr being merged. For gltf, didn't find a way do to it and i think it didn't provide a function for me to do it. 

The difference between metrics looks larger than before, i guess that make sense since we only take model pixels into acount.
image


elalish:
@bsdorra Would you mind enabling `alpha: true` in your WebGLRenderer? The only reason not to is performance, but I don't think frame rate is the biggest concern in your path tracer.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -226,20 +226,26 @@ export class ImageComparator {
           candidateImage.length}, golden: ${goldenImage.length})`);
     }
 
+    let modelPixelCount = 0;
     for (let y = 0; y < height; ++y) {
       for (let x = 0; x < width; ++x) {
         const index = y * width + x;
         const position = index * COMPONENTS_PER_PIXEL;
+
+        if (candidateImage[position + 3] == 0) {
```
elalish, line 235: This is a good candidate for a comment; it may not be obvious to most readers that you are ignoring background pixels by looking for alpha = 0.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -229,7 +229,8 @@ export class ArtifactCreator {
       // Ignored...
     }
 
-    const screenshot = await page.screenshot({path: outputPath});
+    const screenshot =
+        await page.screenshot({path: outputPath, omitBackground: true});
```
elalish, line 233: Good find!

bsdorra:
> @bsdorra Would you mind enabling `alpha: true` in your WebGLRenderer? The only reason not to is performance, but I don't think frame rate is the biggest concern in your path tracer.

Ok, done! I enabled alpha for the WebGLRenderer and added a `disableBackground` function to the PathtracingRenderer. When disabled, the pathtracer will now output transparent pixels for direct background hits. 

Not sure if it's useful, but there's already a function to disable the background rendering in the current version, but this will just set the background to white -> `useBackgroundFromIBL(false)`

elalish:
@bsdorra Thanks! Any idea when you'd like to release the next version on npm?

bsdorra:
@elalish  As soon as I know that it works with the fidelity-tests. I tried to check it today, but ran into issues with update-screenshots. I mean, not with the pathtracer, but it generally refused to render (even with a freshly cloned version). Seems like it had issues with finding the assets. I didn't have much time to look into it, so I'll have another try tomorrow. With some fresh eyes, this will hopefully resolve on its own :) I'll keep you updated.

elalish:
@bsdorra Thanks, feel free to make a small WIP PR updating the version of your library, even to a github branch. If it's not working, @sun765 can take a look.

PR #1404, opened 2020-07-23
kill process when something is wrong in fidelity test

sun765:
Just to test whether fidelity test can fail in github actions or not. Since it's not going to be merged, i didn't care too much about format 

Now the fidelity test did fail immediately when it runs into an error and did give the the error message i want. However, the message is piled in the logs, so i guess i need to make it easier to see and more robust in the next step.
image


.github/workflows/fidelity-tests.yml
```diff
@@ -1,14 +1,14 @@
 name: Fidelity tests
 
-on: [pull_request]
+on: [pull_request, push]
```
sun765, line 3: Do you want fidelity test to run jus on pull request is made , or run every time a new push is made?

.github/workflows/fidelity-tests.yml
```diff
@@ -35,10 +35,10 @@ jobs:
     - name: Bootstrap packages
       run: npm run bootstrap
 
-    - name: Build packages
-      run: npm run build
+    - name: Build packages # may be just build render fidelity packages?
+      run: ./node_modules/.bin/lerna run --scope @google/model-viewer-render-fidelity-tools build --stream
```
sun765, line 39: maybe just build the fidelity test tool pacakge? It saves some time

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
```diff
@@ -1,60 +1,15 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-import HTTPServer from 'http-server';
 import module from 'module';
-import {dirname, join, resolve} from 'path';
-import rimraf from 'rimraf';
 
 const require = module.createRequire(import.meta.url);
```
sun765, line 3: I think this line is trying to make require works in typescript? Because deleting this will make require unrecongnized.

.github/workflows/fidelity-tests.yml
elalish: I think just pull request is fine, unless you have a strong opinion.

.github/workflows/fidelity-tests.yml
elalish: Maybe; what version of model-viewer will be used for the fidelity tools in this case? We needs to ensure it is building the master version of model-viewer and not using the last release on npm.

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
elalish: I have no idea, though I bet we could verify by looking up what rimraf is.

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
```diff
@@ -1,60 +1,15 @@
...
-if (process.argv.length > 3) {
-  scenarioWhitelist = new Set();
-
-  for (let i = 3; i < process.argv.length; i++) {
-    scenarioWhitelist.add(process.argv[i]);
+const core = require('@actions/core');
+
+async function run(): Promise<void> {
+  try {
+    throw new Error('i just want to fail you !');
```
elalish, line 8: Now that you've successfully made it fail, it's probably time to make it fail only when we want it to.

PR #1412, opened 2020-07-25
Update filament goldens

sun765:
Update filament's goldens. Also, i changed the order of scenarios in config, so that more interesting scenarios are in front. If you don't like the order, im happy to change it


packages/render-fidelity-tools/test/renderers/gltfpt/shader/material.glsl
```diff
@@ -0,0 +1,182 @@
+/* @license
```
sun765, line 1: not sure where these shader files come from. It's not in the master branch, but they are generated after i build(not even sure if they comes from building). Should i commit them?

bsdorra: those are left overs from the renaming I did to the pathtracing renderer (gltfpt -> dspbr-pt). the shaders getting copied over from the node module during `prepare`. gltfpt should not exist anymore and the shaders are safe to delete.

packages/render-fidelity-tools/test/config.json
```diff
@@ -157,25 +137,206 @@
       }
```
elalish, line 137: I would put the boom box toward the top, but we can leave the boom box with axes toward the bottom, since they are basically the same.

packages/render-fidelity-tools/test/config.json
```diff
@@ -157,25 +137,206 @@
...
+    },
+    {
+      "name": "khronos-MetalRoughSpheres-LDR",
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/MetalRoughSpheres/glTF/MetalRoughSpheres.gltf",
+      "lighting": "../../../shared-assets/environments/spruit_sunrise_1k_LDR.jpg",
+      "orbit": {
+        "radius": 12
+      },
+      "exclude": [
+        "dspbr-pt",
```
elalish, line 226: Let's organize these so that all the ones with dspbr-pt are above all those without. We should also think about excluding any less-interesting scenarios from dspbr-pt so that we only pay for the rendering where the results are interesting.

packages/render-fidelity-tools/test/config.json
```diff
@@ -157,25 +137,206 @@
...
+      "name": "khronos-Cube",
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/Cube/glTF/Cube.gltf",
+      "orbit": {
+        "theta": 30,
+        "phi": 60,
+        "radius": 5
+      }
+    },
+    {
+      "name": "khronos-Duck",
```
elalish, line 168: It appears from looking at the renders that the tone mapping has in fact changed with the upgrade of Filament. Can you show some examples to demonstrate whether it has gotten closer to or farther from the model-viewer and babylon renders?

packages/render-fidelity-tools/test/renderers/gltfpt/shader/material.glsl
sun765: Thanks for the explanation!

sun765:
i changed the order again and run test in this pr and right before i update filament. (both cases computes metrics in the same way).  There are differences between older and latest version of filament, here's some example.In most "interesting" scenarios, latest filament's absolute value of rms is smaller than older version, which means it got farther from latest model-viewer. Correct me if im wrong, but i feel like the latest filament makes contrast between light area and dark area is a little higher than before

I saved the pdf files of these two fidelity pages in case you may want to see later. (they are too big so i didn't upload them here)

1.5.2:
image

1.8.0
image

1.5.2
image

1.8.0
image

This one differs the most : 
1.5.2
image


1.8.0 
image

sun765:
This pr fails 4 unit tests, do i need to fix the problem?  Besides, this time it didn't trigger fidelity test. Not sure what's the problem. 
image

elalish:
Yeah, I've been seeing more and more flaky tests on mobile safari 12; I think it's time to turn it off, since 14 should be on browser stack soon. Looks like you need to fix some conflicts anyway; if it doesn't pass that time, I'll put in a PR to disable them.

elalish:
I have no idea why it's not running the fidelity tests, but I'd like to get to the bottom of that. Have you seen this happen on any other PRs?

elalish:
Also, whenever you update goldens, please look through the diffs can call out any that have changed significantly. It's easy to review one or two, but github doesn't make it very easy to quickly look through them all. I noticed at some point that Filament's specGlossVsMetalRough got much better, but I'm not actually sure when that was.

sun765:
after pushing again, the fidelity test is triggered. i will ping you whenever the fidelity test is not triggered.

PR #1420, opened 2020-07-28
Update babylonjs and its goldens

sun765:
Did four things in this PR:
1.  update babylonjs to latest version
2. remove 0 radius check  and some babylon scenarios from exclude
3. enable real time filtering for some scenarios 
4. update related goldens

Here's how those high dynamic scenarios looks like now:
image

In older babylonjs:
image


sun765:
PS: all tests are triggered when i keep pushing to this pr this time.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -95,6 +95,7 @@ export interface ScenarioConfig {
   verticalFoV: number;
   exclude?: Array<string>;
   renderSkybox: boolean;
+  realTimeFiltering: boolean;
```
elalish, line 98: Considering that this configuration is specific to babylon, I don't think it should live in the common config file. You'll notice I have a special case in our Filament renderer for the directional light test. I would do something similar here and set this based on name-matching the IBL file.

packages/render-fidelity-tools/test/config.json
```diff
@@ -394,7 +394,8 @@
       "orbit": {
         "radius": 12
       },
-      "renderSkybox": true
+      "renderSkybox": true,
+      "realTimeFiltering": true
```
elalish, line 398: This scenario doesn't need realTimeFiltering because it is using the default IBL, which is not especially HDR.

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -89,6 +89,10 @@ export class BabylonViewer extends LitElement {
     this[$updateSize]();
 
     const {orbit, target, verticalFoV, renderSkybox} = scenario;
+    const lightingBaseName = (scenario.lighting.split('/').pop() as string)
+                                 .split('.')
+                                 .slice(0, -1)
+                                 .join('');
```
sun765, line 95: Learn a lot from how you manipulate string here 👍

packages/render-fidelity-tools/src/common.ts
sun765: Yeah that's much better. I just saw dspbr-pt put some config variables in the scenario config so i thought that's ok. Anyway your way is more organized . Thanks!

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
```diff
@@ -132,11 +133,21 @@ export class BabylonViewer extends LitElement {
...
+    const needRealTimeFilter =
+        (lightingBaseName == 'spruit_sunrise_1k_HDR' ||
+         lightingBaseName == 'spot1Lux');
+
+    // the size of cubemap is 256 for all other renderers
+    const environment = needRealTimeFilter ?
+        new HDRCubeTexture(
+            scenario.lighting, this[$scene], 256, false, false, false, false) :
+        new HDRCubeTexture(
+            scenario.lighting, this[$scene], 256, false, false, false, true);
```
elalish, line 148: I think you could further simplify this by skipping the ternary and putting the `needRealTimeFilter` boolean directly into the last input.

packages/render-fidelity-tools/src/common.ts
elalish: Your point on `pt` is a good one. Do we actually use different numbers of samples for different scenarios? I would prefer to make that a constant inside that renderer for the same reason if possible.

packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
sun765: Good idea!

PR #1422, opened 2020-07-29
Three.js r119

elalish:
Updating from r117 -> r119. Mainly differences in ACESFilmic tone mapping, but it seems to be generally bringing us slightly closer to the other renderers, so good enough. There is also a bug in the new WebGL2Renderer, probably with the RoughnessMipmapper that means I had to switch us to WebGL1Renderer for now. I need to upstream a fix into three.js so that we can allow WebGL2.


elalish:
FYI @mrdoob

packages/model-viewer/src/test/three-components/Renderer-spec.ts
```diff
@@ -102,19 +102,6 @@ suite('Renderer', () => {
       expect(otherScene.renderCount).to.be.equal(1);
     });
 
-    test('marks scenes no longer dirty after rendering', () => {
```
elalish, line 105: This test was flaky on Firefox 77, and its test case is already covered by the one immediately above (since we check that `otherScene` has not rendered a second time).

sun765: i saw unit test fails at this test some times

PR #1423, opened 2020-07-30
Add renderer option when update screenshots

sun765:
Added renderer option in update screenshots command. I didn't add that option in fidelity test because i feel it's not necessary and it requires more refactor.  If you feel that's necessary, im glad to add it to fidelity test command, too.

Also changed the order of scenarios according to today's meeting


packages/render-fidelity-tools/src/workflows/update-screenshots.ts
```diff
@@ -39,12 +39,28 @@ const goldensDirectory = join(rootDirectory, 'goldens');
 const renderersDirectory = join(rootDirectory, 'renderers')
 
 let scenarioWhitelist: Set<string>|null = null;
+let rendererWhitelist: Set<string>|null = null;
+const rendererList = new Set(config.renderers.map((renderer: any) => {
```
sun765, line 43: I feel like this line is a little confusing. If you have more clear and easy to understand way to covert an array's properties to a set, im happy to learn.

packages/render-fidelity-tools/src/workflows/update-screenshots.ts
```diff
@@ -39,12 +39,28 @@ const goldensDirectory = join(rootDirectory, 'goldens');
...
+      if (rendererWhitelist === null) {
+        rendererWhitelist = new Set();
+      }
+      rendererWhitelist.add(argName);
+    } else {
+      if (scenarioWhitelist === null) {
+        scenarioWhitelist = new Set();
+      }
+      scenarioWhitelist.add(argName);
+    }
```
sun765, line 63: Do you think i use too many nested if here?

packages/render-fidelity-tools/test/config.json
```diff
@@ -186,40 +153,72 @@
...
+    },
+    {
+      "name": "khronos-SpecGlossVsMetalRough",
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/SpecGlossVsMetalRough/glTF-Binary/SpecGlossVsMetalRough.glb",
+      "orbit": {
+        "radius": 0.5
+      },
+      "exclude": [
+        "dspbr-pt"
+      ]
```
elalish, line 221: Let's not exclude dspbr-pt from MetalRoughSpheres (just the first one, below).

packages/render-fidelity-tools/test/config.json
```diff
@@ -186,40 +153,72 @@
       }
     },
     {
-      "name": "khronos-NormalTangentMirrorTest",
-      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/NormalTangentMirrorTest/glTF/NormalTangentMirrorTest.gltf",
+      "name": "Astronaut",
```
elalish, line 156: Let's remove the astronaut entirely; it's not even a khronos model and it doesn't test anything novel.

packages/render-fidelity-tools/src/workflows/update-screenshots.ts
elalish: Looks fine to me.

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
```diff
@@ -44,6 +44,8 @@ try {
 
 let scenarioWhitelist: Set<string>|null = null;
 
+// default update screenshots command takes 4 arguments. If there's more than 4,
+// user has specified scenarios to test
```
elalish, line 48: Is there any command-line help shown for `update-screenshots`? If there is, we should add this information there. Also, is this actually true? Based on your logic below I think the whitelist scenarios and renderers can appear in any order, correct?

sun765: No there isn't . I just added an explanation in the ReadMe,  hope it makes things more clear.

By 3 arguments (i said 4 before by mistake), i mean the original command has already taken up 3 arguments.  
 "update-screenshots": "node --experimental-modules ./lib/workflows/update-screenshots.js ./test/config.json",
I just want to put a comment here to explain why it's makes a condition to check whether the lengths of arguments are more than 3 or not (i myself was confused with that at first)

PR #1427, opened 2020-07-31
replace furnace test's hdr file

sun765:
replace furnace test's environment file with the white_furnace.hdr file. Babylon doesn't look very well here:

image

I tried to play with the exposure,  but still doesn't look right (feel like it's not a exposure problem)
exposure = 1.0:
image

exposure = 3.0:
image


elalish:
Actually, that looks great to me. The exposure doesn't matter at all; a furnace test is correct when the spheres are indistinguishable from the background. Babylon is actually doing the best.

elalish:
In fact, I might just cheat and save an image of this with no model at all and call it the dspbr-pt version. That's what we should compare to.

PR #1431, opened 2020-07-31
Auto fidelity test

sun765:
Use a threshold to run automatic fidelity test. When there're scenarios that passes the threshold, logs all those scenarios and fail the test. This is the current test log: (the threshold looks a little too strict here and i should probably change it )
image


packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -83,6 +84,15 @@ export class ArtifactCreator {
       const analysisResults =
           await this.analyze(screenshot, goldens, scenario, dimensions);
 
+      // model-viewer is in 0 index
```
sun765, line 87: In interviews people keep telling me not to use these kind of "magic index" because it's not very readable. Do you feel like this is a problem? Do you have better ideas to make this more readable?

elalish: Yes, the better thing to do is `const modelViewerIndex = 0`, then use that below. If you can remove a comment by adding a name, that's almost always clearer.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -26,6 +26,10 @@ export const MAX_COLOR_DISTANCE: number = 35215;
 
 export const DEVICE_PIXEL_RATIO: number = 2;
 
+// using this threshold to do automatic fidelity test for model-viewer. any
+// scenario whose rms value (in dB) is bigger than the threshold will fail.
+export const FIDELITY_TEST_THRESHOLD: number = -18;
```
sun765, line 31: im just using an arbitrary threshold here.  the highest rms in dB in dev page is -24 (assuming it's right), i guess probably -20 would be a reasonable threshold?

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -69,20 +70,51 @@ export class ArtifactCreator {
...
 
       const analysisResults =
           await this.analyze(screenshot, goldens, scenario, dimensions);
 
+      const modelViewerIndex = 0;
+      const modelViewerRmsInDb =
+          toDecibel(analysisResults[modelViewerIndex].rmsDistanceRatio);
+      // add additional check on whether all renders give bad metrics. Tha
+      // probably means the model-viewer screenshot captured in the fidelity
+      // test maybe completely white or transparent
```
elalish, line 98: This is too approximate. If we're going to have this check, it needs to be part of the analysis where it checks that every pixels is (1,1,1) or (0,0,0).

elalish: If that's looking complicated, we can save it for a future PR.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -26,6 +26,14 @@ export const MAX_COLOR_DISTANCE: number = 35215;
 
 export const DEVICE_PIXEL_RATIO: number = 2;
 
+// use this threshold to do automatic fidelity test for model-viewer. any
+// scenario whose rms value (in dB) is bigger than the threshold will fail.
+export const FIDELITY_TEST_THRESHOLD: number = -24;
+
+// use this threshold to detect whether the screenshot generate in fidlity test
+// is completely white or transparent
+export const WHITE_THRESHOLD: number = -10;
```
elalish, line 35: Let's remove this, as it's too arbitrary to give us certainty.

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
```diff
@@ -58,5 +62,31 @@ screenshotCreator.captureAndAnalyzeScreenshots(scenarioWhitelist)
...
+      const modelViewerErrorPath =
+          join(outputDirectory, 'modelViewerFidelityErrors.json');
+      const modelViewerFidelityErrors = require(modelViewerErrorPath);
+
+      // config contains all scenarios, testConfig contains only scenarios that
+      // the test run on.
+      const testConfigPath = join(outputDirectory, 'config.json');
+      const testConfig = require(testConfigPath);
+      const scenarioCount = testConfig.scenarios.length,
+            failCount = modelViewerFidelityErrors.length,
```
elalish, line 74: This syntax is a little weird to me; let's just put each const on its own line and skip the commas.

.github/workflows/fidelity-tests.yml
```diff
@@ -37,8 +37,13 @@ jobs:
 
     - name: Build packages
```
sun765, line 38: Is it ok if i change this to build only model-viewer package and fidelity test package? As there are more and more packages, this can save quite some time

elalish: Sure, though I'm pretty sure you'll also need to build 3dom, as that's a dependency of model-viewer.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -26,6 +26,14 @@ export const MAX_COLOR_DISTANCE: number = 35215;
...
+
```
sun765, line 36: I think we need to play back and force on these two thresholds.  The fidelity threshold is based on the highest value among all scenarios, the white threshold is based on the crazy metrics i showed you before (every renderers give around -9 dB)

packages/render-fidelity-tools/src/common.ts
```diff
@@ -254,3 +262,7 @@ export class ImageComparator {
     return {analysis: {rmsDistanceRatio}};
   }
 }
+
+export function toDecibel(value: number): number {
+  return 10 * Math.log10(value);
+}
```
sun765, line 268: Since this function is used in more than one script, i just refactored it into common.ts.  I also tried to write the fidelity test result as dB directly, but json file doesn't recognize Infinity and there's some trouble in parsing it, so i just stay with this way. If you have better idea on how to refactor this, please let me know, thanks!

sun765:
@elalish  Making fidelity test timeout if the capture screenshot just hang there is a little more complex than it thought so im still working on that. Is it ok if i implement it in another pr?

.github/workflows/fidelity-tests.yml
```diff
@@ -36,9 +36,17 @@ jobs:
...
-      run: npm run build
+      run: |
+       ./node_modules/.bin/lerna run --scope @google/3dom build --stream
+       ./node_modules/.bin/lerna run --scope @google/model-viewer build --stream
+       ./node_modules/.bin/lerna run --scope @google/model-viewer-render-fidelity-tools build --stream
+    
+    # TODO: without build legacy support, some scenarios give wrong metrics in fidelity test workflow. 
+    # There's some tricky bug here but we don't know how to solve yet. 
+    - name: Generate legacy bundles
+      run: npm run build:legacy-support
```
sun765, line 47: Build legacy-support stops before fidelity test stops it from giving crazy metrics. Model-viewer's renderer's page use the latest version of model-viewer to render golden and should have nothing to do with legacy support (https://github.com/google/model-viewer/blob/master/packages/render-fidelity-tools/test/renderers/model-viewer/index.html#L51), so this is really confusing. Any idea on how this could happen would be super helpful.

packages/render-fidelity-tools/src/artifact-creator.ts
elalish: I think you addressed my feedback here, but the comment looks out of date?

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -55,23 +55,19 @@
...
-          //TODO: this is a temporary workaround for taking screenshots before the model has been loaded
-          //This will be fixed when the screenshot function is refactored.
-          setTimeout(()=>{
-            self.dispatchEvent(new CustomEvent('model-ready'));
-          }, 500);
-        }
-      }, { once: true });
+      modelViewer.addEventListener('poster-dismissed', ()=>{
+        this.dispatchEvent(new CustomEvent('model-ready'));
+      }, {once: true});
```
sun765, line 61: The poster-dismissed event happened after the model-visibility event, (i think the order is announceVisibility(fire model-visibility event) -> wait for transition finish -> fire poster-dismissed event) so i think it's ok to just keep only one event listener for poster-dismissed event.

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -55,23 +55,19 @@
...
       
       modelViewer.style.width = `${scenario.dimensions.width}px`;
       modelViewer.style.height = `${scenario.dimensions.height}px`;
 
       modelViewer.style.backgroundColor = "rgba(255,255,255,0)";
 
       modelViewer.src = scenario.model;
+
+      modelViewer.skyboxImage = null;
+      modelViewer.environmentImage = null;
```
elalish, line 70: I would put these inside the else statement instead; that way when you change scenarios but have the same lighting, it doesn't switch back and forth. Realistically this is probably optimized out by lit-element's batched updates anyway, but it's still cleaner.

packages/model-viewer/src/features/loading.ts
```diff
@@ -432,6 +432,7 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
             // screen readers
             defaultPosterElement.setAttribute('aria-hidden', 'true');
             defaultPosterElement.tabIndex = -1;
+            this.dispatchEvent(new CustomEvent('poster-dismissed'));
```
elalish, line 435: Have you verified this fixes most of the flakey renders? Which ones are still experiencing flakes, and roughly how often (%)?

sun765: i run the fidelity test 3 times so far, all 100% passed. I will keep running it and then tell you the result

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -55,23 +55,19 @@
...
-      modelViewer.addEventListener('model-visibility', (event) => {
-        if (event.detail.visible) {
-          //TODO: this is a temporary workaround for taking screenshots before the model has been loaded
-          //This will be fixed when the screenshot function is refactored.
-          setTimeout(()=>{
-            self.dispatchEvent(new CustomEvent('model-ready'));
-          }, 500);
-        }
-      }, { once: true });
+      modelViewer.addEventListener('poster-dismissed', ()=>{
```
bejado, line 58:
```suggestion
      modelViewer.addEventListener('poster-dismissed', () => {
```

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -66,13 +66,13 @@
 
       modelViewer.src = scenario.model;
 
-      modelViewer.skyboxImage = null;
-      modelViewer.environmentImage = null;
       if (scenario.renderSkybox) {
+        // this set both skyboxImage and environmentImage to scenario.lighting;
```
elalish, line 70: That's not quite true; they will only both be the same if environmentImage is set to null, so you should do that here as well, since you might be switching from a scenario with an environmentImage.

sun765:
Fidelity test run 4 times so far, all 100% passed. (probably just by chance, and i will keep monitoring it)

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -66,14 +66,8 @@
...
-      if (scenario.renderSkybox) {
-        // this set both skyboxImage and environmentImage to scenario.lighting;
-        modelViewer.skyboxImage = scenario.lighting;
-      }
-      else {
-        modelViewer.environmentImage = scenario.lighting;
-        modelViewer.skyboxImage = null;
-      }
+      modelViewer.environmentImage = scenario.lighting;
+      modelViewer.skyboxImage = scenario.renderSkybox? scenario.lighting: null;
```
elalish, line 70: 👍

PR #1435, opened 2020-08-03
Quicklook link event / anchor

### LINKED ISSUE #1030: Feature request: Add support for additional properties to enable Apple Pay Button or a Custom Action in AR Quick Look mode
Would be very nice to see support added for additional properties in the model-viewer markup to allow easily passing in the additional properties/params needed to enable an Apple Pay Button or a Custom Action in AR Quick Look mode. 

Reference documentation here: https://developer.apple.com/documentation/arkit/adding_an_apple_pay_button_or_a_custom_action_in_ar_quick_look

I will take a pass at editing: packages/model-viewer/src/features/ar.ts
https://github.com/google/model-viewer/blob/master/packages/model-viewer/src/features/ar.ts

Off the top of my head, it might just some need a few quick edits to the function "openIOSARQuickLook" in that file, and maybe require adding in the ability to pass in a reference to a custom listener to bind to, in order to support the Custom Action as well...  If I have some time tomorrow and get it working, will submit a PR.

### LINKED ISSUE #1314: addEventListener on iOS
Apple added new capability with buttons and the implementation is via ID - but it doesn’t seem to work. I’ve had a look at the model-viewer.js where <a> is created dynamically and then with rel=ar and even managed to dynamically ad ID but it’s not working when calling it. Any wisdom on this plz? 

document.getElementById("ar-link-3").addEventListener("message", function (event) {   
    if (event.data == "_apple_ar_quicklook_button_tapped") {
        window.location.assign("https://developer.apple.com/documentation/arkit/previewing_a_model_with_ar_quick_look");
    }
}, false);

hybridherbst:
I started adding a fix (or, actually two options) for the issues with iOS QuickLook links that need to be catched from the anchor tags QuickLook has been launched. 
I'm unclear how to proceed, though; 

**Option 1**: adding proper event handling so that one can attach a custom listener to the modelviewer instance (preferred)
Questions: currently it seems events should be in the ARMixin, however openIOSARQuickLook lives outside of the mixin. Not sure what the correct way is to add this event.

**Option 2**: adding the anchor to the DOM.
Questions: if that is done I think it should be added as a child of the AR Button (?), and not just in "body". Also, should it be cached internally to make sure it's not added multiple times?

### Reference Issues
Fixes #1030 
Fixes #1314

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -37,6 +37,23 @@ export const openIOSARQuickLook = (() => {
       modelUrl.hash = 'allowsContentScaling=0';
     }
     anchor.setAttribute('href', modelUrl.toString());
+    // one option: add directly to event listener here,
+    // make a message handler for model-viewer that can be attached to
+    // unclear: where to add handler, openIOSARQuickLook is export const right
+    // now but events should be on the ARMixin?
+    /*
+    anchor.addEventListener("message", (event: Event) => {
+      if ((<MessageEvent>event).data == "_apple_ar_quicklook_button_tapped") {
+        alert("user tapped on link: " + event.target + ", opening page");
+        console.log("user tapped on link: " + event.target + ", opening page");
+      }
+    }, true);
+    */
+    // another option: correctly configure the a tag
+    // (so that it's clear for which modelviewer this was) and add it to the DOM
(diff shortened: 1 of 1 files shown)


hybridherbst:
@googlebot I signed it!

elalish:
Closing in favor #1458

PR #1437, opened 2020-08-03
Add create-react-app example

### LINKED ISSUE #1421: Why doesn't this load anything?
### Description

Downloaded your [Astronaut.glb](https://github.com/google/model-viewer/blob/master/packages/shared-assets/models/Astronaut.glb) and I'm serving it over a CORS-friendly https-friendly proxy (ngrok). I'm trying to load the most basic example I've seen of `<model-viewer>` within a create-react-app project on CodeSandbox.

It seems to load <model-viewer> (I can see its shadow dom elements, etc..)

The network tab indicates the model is successfully being fetched.

But nothing shows up in the viewer...

#### Live Demo

https://codesandbox.io/s/model-viewer-react-t2u1v?file=/src/App.js (have to download the glb, and follow the comment instructions, then replace the ngrok url to replicate)

### Browser Affected
- [x] Chrome 84
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [ ] Safari

### OS
- [ ] Android
- [ ] iOS
- [ ] Linux
- [x] MacOS 10.15.4
- [ ] Windows

### Versions
- model-viewer: v1.1.0
- three.js: 0.119.0

corysimmons:
>Instructions: https://github.com/google/model-viewer/blob/master/packages/model-viewer/CONTRIBUTING.md#code-reviews 👈 404's

### Reference Issue
Fixes #1421 and probably other issues (questions)

---

I don't want to have to put an `Astronaut.glb` in every single example that is ever created, but I can't do relative imports outside of the `src` directory with React.

I tried fetching the Astronaut.glb blob from Github Raw, but hit CORS.

Then tried `yarn add @google/model-viewer-shared-assets` (from https://github.com/google/model-viewer/blob/ec2057513beb40a3c6ea9c2be0ea66d3065f4990/packages/shared-assets/package.json) but it's a private package on npm I suppose?

I was going to try to make something like https://npm.im/shared-astronaut that simply imported and `module.exports` the `Astronaut.glb` file, but seemed like it needed a loader maybe? Maybe I'm dumb and it's simple but the second it gets into loader territory I realize it's gonna be a time sink...

Anyway! It'd be nice to have something like https://npm.im/shared-astronaut so for every example (and if people just wanted to tinker around) they could quickly `yarn add shared-astronaut` then `import astronaut from 'shared-astronaut'`.

I should also mention this repo is **3GB** (with a **G**). It took _forever_ to clone it on my mediocre internet (Github may have been throttling to 1MB/s). If you _do_ want more people to contribute (especially with examples), then it might be worth seriously considering extracting all the shared packages to their own repos so people can quickly `git clone model-viewer && mkdir examples/some-new-example` without it taking 30 mins to download.

For now, in the scheme of a 3GB repo, I don't think the 3MB Astronaut.glb matters and this single example would probably clear up a lot of _"How can this work in React?"_ questions.


corysimmons:
@googlebot I signed it!

corysimmons:
Alternatively, Google could maintain their own CodeSandbox account, and just make this a `EXAMPLES.md` in the root of the project which just lists demos.

elalish:
The rest of our examples are inside of the `modelviewer.dev` package, so they can be served on our docs site. Obviously you wouldn't want to use our example template for this, but could it live there instead? You can see that we also access and serve all the shared-assets from there, so that might also help.

Thanks, I'll fix up the contributors link. Also, sorry about the 3GB download; I need to add cloning instructions to our docs. You want to use `depth=1` to avoid pulling the history. The size is all old golden images from fidelity tests.

corysimmons:
Ah okay, busy chasing a deadline this week but if this starts to go stale, someone ping me next week and I'll get it sussed out. Alternatively feel free to copy/paste the demo to wherever. I'm not worried about contrib credit or anything.

billschnoebelen:
Thanks Cory!

This saved my day!

//  this worked locally, but threw errors on the server
import "@google/model-viewer"; 

// this solved it
import "@google/model-viewer/dist/model-viewer-legacy";

elalish:
@corysimmons I think I might have figured out why the normal import wasn't working: #1470. Anyway, I'd definitely like to get this example merged. Do you have time to move it to the modelviewer.dev package? It would also help immensely if you could put it on its own branch instead of `master`; makes for collisions when I try to prod the tests to run.

aviggiano:
I had the same issue but had to use

```javascript
import '@google/model-viewer/dist/model-viewer'
```

elalish:
I'm now thinking this would actually be easier to ingest not as part of our repo, but as an official example like our Glitch issue template. If anyone knows how to fix #1801, if you'd be willing to post it here I could fork it to make our own codesandbox.io example to put in our documentation. I'm guessing the problem is something simple, but I'm surprised there's no console errors.

corysimmons:
Dunno, I like https://github.com/vercel/next.js/tree/canary/examples + npmignore (so npm package downloaders don't have to download that huge directory) just so contributors can keep the examples fresh (whereas on CodeSandbox they'd probably get stale).

elalish:
Yeah, except the vast majority of our users are not familiar with how to clone or build our repo. I think some kind of Glitch-esque project will be much easier for them to consume, and I won't have to worry about what systems it can build on (Windows is currently a disaster for us). API stability means a simple example like this really shouldn't break, and if it does then users will probably alert me so I can fix it.

softmarshmallow:
Updates on this pr.?

BorisKourt:
This is a nice example but I am wondering if it is ignoring some broader compatibility issues between react and web-components. For example how `<model-viewer>` reacts on re-render and attribute changes in a react component. I feel like that will trip newcomers up once they start building on top of this and expect it to behave like any regular react component. 

I am not sure if this has been done elsewhere (https://github.com/google/model-viewer/issues/1801, seems to similarly avoid the issue.) it would be nice to see a way for this to work with react in a more complete demo. Or maybe some suggestions for best practices. 

---

Is there more information on `import "@google/model-viewer/dist/model-viewer-legacy"` and for how long this will be available/maintained?

elalish:
The trouble is we have no React expertise on this team, so we have to rely on the community to contribute. Feel Free! As for legacy, it was only for supporting IE11, which we no longer do. It hasn't been in any of our latest releases, and hence is not maintained.

corysimmons:
It's really just this

image

Web Components seem to work fine in React https://reactjs.org/docs/web-components.html

And if you're worried about re-renders, `React.memo` can fix that https://reactjs.org/docs/react-api.html#reactmemo

elalish:
Also, since `model-viewer-legacy` is depreacted, I'd recommend using `model-viewer-umd` for anything that doesn't like our default modern bundle.

PR #1455, opened 2020-08-12
Improve auto fidelity test

sun765:
1.  Added two raf when firing the poster-dismissed event, to make sure the event is fired after poster completely disappear
2. Check if model-viewer hang there when rendering golden. If it hang there after some time(maxTime), stop it and report error and keep testing the rest scenarios
3. Check if the golden captured is completely white. If so, report error and keep testing the rest scenarios

A new error emerges. I think 10 seconds is long enough, but i will try longer maxTime later. 
image

Also, not sure what caused the huge chunk of logging. I think the only thing i did related to model-viewer is adding an additional raf. I will hunt it down
image

### DIFF
--- a/packages/render-fidelity-tools/src/artifact-creator.ts
+++ b/packages/render-fidelity-tools/src/artifact-creator.ts
@@ -77,7 +77,7 @@ export class ArtifactCreator {
             scenarioName,
             dimensions,
             join(scenarioOutputDirectory, 'model-viewer.png'),
-            10);
+            60);
       } catch (error) {
         const errorMessage =
             `❌ Failed to capture model-viewer's screenshot of ${
--- a/packages/render-fidelity-tools/src/artifact-creator.ts
+++ b/packages/render-fidelity-tools/src/artifact-creator.ts
@@ -199,7 +199,7 @@ export class ArtifactCreator {
   async captureScreenshot(
       renderer: string, scenarioName: string, dimensions: Dimensions,
       outputPath: string = join(this.outputDirectory, 'model-viewer.png'),
-      useMaxTime: boolean = false, maxTime: number = 10) {
+      useMaxTime: boolean = false, maxTimeInSec: number = 60) {
     const scaledWidth = dimensions.width;
     const scaledHeight = dimensions.height;
     const rendererConfig = this[$configReader].rendererConfig(renderer);
@@ -250,32 +250,33 @@ export class ArtifactCreator {
     // variables are captured in its closure scope. TypeScript compiler
     // currently has no mechanism to detect this and will happily tell you
     // your code is correct when it isn't.
-    const evaluateError = await page.evaluate(async (useMaxTime, maxTime) => {
-      console.log(maxTime);
-      console.log(useMaxTime);
-      const modelBecomesReady = (self as any).modelLoaded ?
-          Promise.resolve() :
-          new Promise((resolve, reject) => {
-            let timeout: NodeJS.Timeout;
-            if (useMaxTime) {
-              timeout = setTimeout(reject, maxTime * 1000);
-            }
+    const evaluateError =
+        await page.evaluate(async (useMaxTime, maxTimeInSec) => {
+          console.log(maxTimeInSec);
+          console.log(useMaxTime);
+          const modelBecomesReady = (self as any).modelLoaded ?
+              Promise.resolve() :
+              new Promise((resolve, reject) => {
+                let timeout: NodeJS.Timeout;
+                if (useMaxTime) {
+                  timeout = setTimeout(reject, maxTimeInSec * 1000);
+                }
 
-            self.addEventListener('model-ready', () => {
-              if (useMaxTime) {
-                clearTimeout(timeout);
-              }
-              resolve();
-            }, {once: true});
-          });
+                self.addEventListener('model-ready', () => {
+                  if (useMaxTime) {
+                    clearTimeout(timeout);
+                  }
+                  resolve();
+                }, {once: true});
+              });
 
-      try {
-        await modelBecomesReady;
-        return null;
-      } catch (error) {
-        return `Stop capturing screenshot after ${maxTime} seconds`;
-      }
-    }, useMaxTime, maxTime);
+          try {
+            await modelBecomesReady;
+            return null;
+          } catch (error) {
+            return `Stop capturing screenshot after ${maxTimeInSec} seconds`;
+          }
+        }, useMaxTime, maxTimeInSec);
 
     if (evaluateError) {
       console.log(evaluateError);
--- a/packages/render-fidelity-tools/src/artifact-creator.ts
+++ b/packages/render-fidelity-tools/src/artifact-creator.ts
@@ -77,7 +77,7 @@ export class ArtifactCreator {
             scenarioName,
             dimensions,
             join(scenarioOutputDirectory, 'model-viewer.png'),
-            true);
+            10);
       } catch (error) {
         const errorMessage =
             `❌ Failed to capture model-viewer's screenshot of ${
@@ -199,7 +199,7 @@ export class ArtifactCreator {
   async captureScreenshot(
       renderer: string, scenarioName: string, dimensions: Dimensions,
       outputPath: string = join(this.outputDirectory, 'model-viewer.png'),
-      useMaxTime: boolean = false, maxTimeInSec: number = 60) {
+      maxTimeInSec: number = -1) {
     const scaledWidth = dimensions.width;
     const scaledHeight = dimensions.height;
     const rendererConfig = this[$configReader].rendererConfig(renderer);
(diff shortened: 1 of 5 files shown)


packages/render-fidelity-tools/src/workflows/test-fidelity.ts
```diff
@@ -70,9 +70,9 @@ screenshotCreator.captureAndAnalyzeScreenshots(scenarioWhitelist)
       // the test run on.
       const testConfigPath = join(outputDirectory, 'config.json');
       const testConfig = require(testConfigPath);
-      const scenarioCount = testConfig.scenarios.length;
       const failCount = modelViewerFidelityErrors.length;
-      const passCount = scenarioCount - failCount;
+      const passCount = testConfig.scenarios.length;
+      const scenarioCount = failCount + passCount;
```
sun765, line 75: Some errors , like hang there when rendering , didn't find goldens for some render, doesn't have a analysisResult and won't be pushed to the testConfig file, so my former way of counting pass and fail is wrong. So i just change to all passed scenarios goest to testConfig, all senarios with errors goes to modelViewerFidelityErrors.  The shortcoming is when deploying, you cant see the failed sceanrios in the dev page because they are not in the conifg file. 

The other option is to calculate the count using the total scenarios count. The shortcoming for this is when running locally and running the test with specific scenarios, the count is not correct

Tell me which method you like better and im happy to change it.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -188,7 +198,8 @@ export class ArtifactCreator {
 
   async captureScreenshot(
       renderer: string, scenarioName: string, dimensions: Dimensions,
-      outputPath: string = join(this.outputDirectory, 'model-viewer.png')) {
+      outputPath: string = join(this.outputDirectory, 'model-viewer.png'),
+      useMaxTime: boolean = false, maxTime: number = 10) {
```
sun765, line 202: im just using an arbitrary maxTime here, so feel free to change

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
elalish: I think I like it the way you've done it here. If any fail, nothing will make it to the dev page anyway since the test will error out. That's good for something facing the public, which this will be shortly.

packages/render-fidelity-tools/src/artifact-creator.ts
elalish: Looks like it wasn't long enough for sponza. Software rendering can be impressively slow, and keep in mind this also includes things like decoding textures, since it's the first frame. Let's go with 30 or 60 seconds.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -239,20 +250,35 @@ export class ArtifactCreator {
     // variables are captured in its closure scope. TypeScript compiler
     // currently has no mechanism to detect this and will happily tell you
     // your code is correct when it isn't.
-    await page.evaluate(async () => {
-      const modelBecomesReady = (self as any).modelLoaded ?
```
sun765, line 243: When console.log self object in puppeteer, it's actually the window object and doesn't have modelLoaded property at all(and i think i shouldn't have).  i removed it and run update-screenshots and test for all scenarios (except for dspbr-pt), no errors found. So i think we should just remove it.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -188,7 +198,8 @@ export class ArtifactCreator {
 
   async captureScreenshot(
       renderer: string, scenarioName: string, dimensions: Dimensions,
-      outputPath: string = join(this.outputDirectory, 'model-viewer.png')) {
+      outputPath: string = join(this.outputDirectory, 'model-viewer.png'),
+      useMaxTime: boolean = false, maxTimeInSec: number = 60) {
```
bejado, line 202: Perhaps instead of adding a `useMaxTime` flag, specifying a `maxTimeInSec` of 0 could denote no timeout.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -253,6 +279,13 @@ export class ImageComparator {
       }
     }
 
+    const imagePixelCount = width * height;
+    if (whitePixelCount === imagePixelCount ||
+        blackPixelCount === imagePixelCount ||
+        transparentPixelCount === imagePixelCount) {
```
bejado, line 285: If an image is rendered correctly but with 50% transparency, then `transparentPixelCount` will still equal `imagePixelCount` because you've defined "transparent pixels" as anything < 100% opaque. Is this what you want? In that case the error "Candidate image is colorless" might be misleading"

packages/render-fidelity-tools/src/common.ts
```diff
@@ -239,11 +242,34 @@ export class ImageComparator {
...
         }
 
+        let isWhitePixel = true;
+        let isBlackPixel = true;
+        for (let i = 0; i < 4; i++) {
+          const colorComponent = candidateImage[position + i];
+          if (colorComponent != 255) {
+            isWhitePixel = false;
+          }
+          if (colorComponent != 255) {
```
bejado, line 261: Should this be `if (colorComponent != 0)`?

packages/render-fidelity-tools/src/common.ts
bejado: My recommendation for this logic would be to pre-multiply the color channels by alpha, and _then_ test for white or black pixels. This way, both [0, 0, 0, 255] and [11, 22, 33, 0] register as "black"

packages/render-fidelity-tools/src/artifact-creator.ts
elalish: Or even -1, to make it even more obvious. It's a good practice to test that inputs are >0 anyway if that's what you expect.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -239,20 +250,35 @@ export class ArtifactCreator {
...
             }, {once: true});
           });
 
-      await modelBecomesReady;
-    });
+          try {
+            await modelBecomesReady;
+            return null;
+          } catch (error) {
+            return `Stop capturing screenshot after ${maxTimeInSec} seconds`;
```
elalish, line 273: I think you can put this error message into your reject(). Then you could display whatever error message you get, because it's possible you got an error that isn't the timeout.

packages/render-fidelity-tools/src/common.ts
elalish: That's a good idea, and simple.

packages/render-fidelity-tools/src/common.ts
```diff
@@ -253,6 +279,13 @@ export class ImageComparator {
       }
     }
 
+    const imagePixelCount = width * height;
+    if (whitePixelCount === imagePixelCount ||
+        blackPixelCount === imagePixelCount ||
```
elalish, line 284: This is subtly different than the logic I was trying to describe. This will fail only if the image is all white or all black or all transparent. We want it to fail none of the pixels are non- white, black, transparent. So you only need to keep count of one category, instead of the three separately.

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -56,7 +56,12 @@
     harness.addEventListener('scenario-change', () => {
       const {scenario} = harness;
       modelViewer.addEventListener('poster-dismissed', () => {
-        this.dispatchEvent(new CustomEvent('model-ready'));
+        // When the poster-dismissed event fired, one raf has already been requrested. 
+        // Use an additional raf here is to make sure model-ready event fires at the frame that 
+        // the poster has completely faded away. 
+        requestAnimationFrame(()=>{
+          this.dispatchEvent(new CustomEvent('model-ready'));
```
elalish, line 63: I think you'll need a raf inside a raf to get past the bug we found. A raf that dispatches an event that requests a raf can still call those two in the same animation frame.

sun765: Thank you so much for pointing this out!

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -250,29 +250,30 @@ export class ArtifactCreator {
...
-        }, useMaxTime, maxTimeInSec);
+          resolve();
+        }, {once: true});
+      });
+
+      try {
+        await modelBecomesReady;
+        return null;
+      } catch (error) {
+        return error;
```
sun765, line 274: Not sure why the error here is a string, not an object. I guess that has something to do with puppeteer?

packages/render-fidelity-tools/src/common.ts
```diff
@@ -231,9 +231,7 @@ export class ImageComparator {
     }
 
     let modelPixelCount = 0;
-    let whitePixelCount = 0;
-    let blackPixelCount = 0;
-    let transparentPixelCount = 0;
+    let colorlessPixelCount = 0;
```
sun765, line 234: Just want to make sure if i understand the logic correctly. I should check whether all pixels are colorless or not, right? And a pixel is colorless if its white or black or transparent.

packages/render-fidelity-tools/src/common.ts
sun765: Yeah it should. Thank you for finding out!

packages/render-fidelity-tools/src/common.ts
sun765: Thank you for the recommendation and it's really much better than mine.  We found an error that model-viewer's golden is captured when the poster hasn't faded away and that's why i thought checking all pixels' alpha this way would be helpful.  
image
But yeah the error message is misleading in this case. Besides, this bug is probably fixed right now, so i removed this check in latest commit.  @elalish Do you still want this check to make sure the poster has faded away when we take screenshots?

packages/render-fidelity-tools/src/artifact-creator.ts
elalish: I think it's because you put a string into `reject()` instead of a new Error: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject

packages/render-fidelity-tools/src/common.ts
```diff
@@ -239,11 +240,27 @@ export class ImageComparator {
         // b, a.  here position is the index for current pixel's r , position+3
         // is index for its alpha
         const position = index * COMPONENTS_PER_PIXEL;
+        const alpha = candidateImage[position + 3];
+
+        let isWhitePixel = true;
+        let isBlackPixel = true;
+        for (let i = 0; i < 4; i++) {
```
elalish, line 247: Maybe `i < 3`?

packages/render-fidelity-tools/src/common.ts
elalish: Correct.

PR #1459, opened 2020-08-14
more raf, transparent check

sun765:
The poster bug still exist even if we have already used two raf. I just add code to test whether the golden is semi transparent. 
image

### DIFF
--- a/packages/render-fidelity-tools/src/common.ts
+++ b/packages/render-fidelity-tools/src/common.ts
@@ -283,7 +283,7 @@ export class ImageComparator {
 
     if (transparentPixelCount === imagePixelCount) {
       throw new Error(
-          'Candidate image is semi transparent, probably model-veiwer\'s screenshot is taken before its poster faded! ')
+          'Candidate image is semi transparent, probably the screenshot is taken before its poster faded! ')
     }
 
     const rmsDistanceRatio =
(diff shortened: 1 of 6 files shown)


packages/render-fidelity-tools/src/common.ts
```diff
@@ -275,6 +281,11 @@ export class ImageComparator {
       throw new Error('Candidate image is colorless!');
     }
 
+    if (transparentPixelCount === imagePixelCount) {
```
sun765, line 284: If the candidate image is completely transparent, it's marked as colorless ,not transparent here.  Do you want it this way or do you want it to be marked as transparent?

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -60,7 +60,9 @@
         // the poster has completely faded away.
```
sun765, line 60: Just added more raf to see if the bug still exist.

bejado: Does this fix the bug? Maybe we can make the number of RAF configurable

packages/render-fidelity-tools/test/renderers/model-viewer/index.html
```diff
@@ -60,7 +60,9 @@
         // the poster has completely faded away. 
         requestAnimationFrame(()=>{
           requestAnimationFrame(()=>{
-            this.dispatchEvent(new CustomEvent('model-ready'));
+            requestAnimationFrame(()=>{
+              this.dispatchEvent(new CustomEvent('model-ready'));
```
elalish, line 64: Might be nice to add a comment about how it seems like three rafs are needed to get the poster to fade out reliably. Nice to document work-arounds.

PR #1461, opened 2020-08-14
Singleton

elalish:
This refactor is the first step toward lazy-loading three.js, and a significant chunk of our own library. I've created the Lazy class, which contains the three.js-related singletons and will become the dynamic import point of the bundle. Remaining work includes removing all references to the DOM elements from the ModelScene, as that will be the per-element entry point to this same bundle and so it must be self-contained. Then the final step will be reexporting all the three and three-component interfaces through Lazy and setting up the dynamic import to trigger on preload.


elalish:
It looks like it's actually quite easy to lazy-load the whole model-viewer library, and you can even get a poster so long as you use a custom div. Therefore it seems unnecessary to bother lazy-loading three.js within model-viewer.

PR #1467, opened 2020-08-19
Multi os fidelity test

sun765:

### DIFF
--- a/packages/model-viewer/src/test/model-viewer-spec.ts
+++ b/packages/model-viewer/src/test/model-viewer-spec.ts
@@ -3,36 +3,96 @@ import {Constructor} from '../utilities';
 
 import {BasicSpecTemplate} from './templates';
 
+
 // const expect = chai.expect;
 
-suite.only('ModelViewerElement', () => {
-  // global variable
-  let nextId: number = 0;
-  let tagName: string;
-  let ModelViewer: Constructor<ModelViewerElement>;
 
-  setup(() => {
-    tagName = `model-viewer-${nextId++}`;
-    ModelViewer = class extends ModelViewerElement {
-      static get is() {
-        return tagName;
-      }
-    }
-    // not sure what does this do
-    customElements.define(tagName, ModelViewer);
-  })
+const DEFAULT_SCENARIO = {
+  name: 'khronos-Cube',
+  model:
+      '../../../shared-assets/models/glTF-Sample-Models/2.0/Cube/glTF/Cube.gltf',
+  lighting: '../../../shared-assets/environments/lightroom_14b.hdr',
+  dimensions: {width: 768, height: 768},
+  target: {x: 0, y: 0, z: 0},
+  orbit: {theta: 30, phi: 60, radius: 5},
+  verticalFoV: 45
+}
 
-  BasicSpecTemplate(() => ModelViewer, () => tagName);
-  // tear down
+const setupModelViewer =
+    async (
+        modelViewer: ModelViewerElement,
+        config: any) => {
+  modelViewer.style.width = `${config.dimensions.width}px`;
+  modelViewer.style.height = `${config.dimensions.height}px`;
 
-  // basic template (optional)
+  modelViewer.style.backgroundColor = 'rgba(255,255,255,0)';
 
-  suite(
-      'Fidelity Test',
-      () => {
-          // setup-fidleity test,  read config file here
+  modelViewer.src = config.model;
 
-          // test each scenarios
+  modelViewer.environmentImage = config.lighting;
+  modelViewer.skyboxImage = config.lighting;
 
-      })
-})
+  const {theta, phi, radius} = config.orbit;
+  modelViewer.minCameraOrbit = `auto auto ${radius}m`;
+  modelViewer.maxCameraOrbit = `auto auto ${radius}m`;
+  modelViewer.cameraOrbit = `${theta}deg ${phi}deg ${radius}m`;
+  const {x, y, z} = config.target;
+  modelViewer.cameraTarget = `${x}m ${y}m ${z}m`;
+  modelViewer.fieldOfView = `${config.verticalFoV}deg`;
+}
+
+
+                        suite.only('ModelViewerElement', () => {
+                          // global variable
+                          let nextId: number = 0;
+                          let tagName: string;
+                          let ModelViewer: Constructor<ModelViewerElement>;
+
+                          setup(() => {
+                            tagName = `model-viewer-${nextId++}`;
+                            ModelViewer = class extends ModelViewerElement {
+                              static get is() {
+                                return tagName;
+                              }
+                            }
+                            // not sure what does this do
+                            customElements.define(tagName, ModelViewer);
+                          })
+
+                          BasicSpecTemplate(() => ModelViewer, () => tagName);
+
+                          suite('Fidelity Test', () => {
+                            let element: ModelViewerElement;
+
+                            setup(() => {
+                              element = new ModelViewerElement();
+                              document.body.insertBefore(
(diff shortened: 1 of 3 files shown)


packages/model-viewer/src/test/helpers.ts
```diff
@@ -225,4 +225,58 @@ export const createFakeThreeGLTF = () => {
     } as unknown as GLTFParser,
     userData: {}
   };
-};
\ No newline at end of file
+};
+
```
sun765, line 229: At first i put all the scenarios config into a json file and tried to fetch that in the suite()'s setup.
```
suite("fidelity test", ()=>{
let scenarios = [];
setup(()=>{
    scenarios = await(await fetch(scenarioPath)).json();
});

for(let scenario of scenarios){
    test(scenario.name, ()=>{
       //do the fidelity test here
    })
}
})
```
But the fact is that setup() only triggers before test(),  and when mocha see the line for(let scenario of scenarios), it didn't see test, so it doesn't call setup and scenarios is just empty.() I have tried some workaround, none work so far. So i just used this ugly inefficient way to read scenario configs. If you have better way to do this, im happy to learn

packages/model-viewer/src/test/helpers.ts
```diff
@@ -225,4 +225,58 @@ export const createFakeThreeGLTF = () => {
...
+}
+;
+
+export const SCENARIOS: Array<scenarioConfig> = [
+  {
+    name: 'khronos-MetalRoughSpheres',
+    model:
+        'models/glTF-Sample-Models/2.0/MetalRoughSpheres/glTF/MetalRoughSpheres.gltf',
+    lighting: 'environments/lightroom_14b.hdr',
+    dimensions: {width: 100, height: 100},
```
sun765, line 254: the origin dimension is 768, 768, im just using an arbitrary small dimension here to make the test faster, so feel free to change it.

packages/model-viewer/src/test/model-viewer-spec.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  return Promise.resolve(fiddleCanvas);
+};
+
+// TODO(sun765): this only test whether the screenshot
+// is colorless or not. Replace this with more robust
+// test in later pr.
+function testFidelity(screenshotCanvas: HTMLCanvasElement): FidelityResult {
+  let colorlessPixelCount = 0;
+  let transparentPixelCount = 0;
```
sun765, line 58: Since i have used a different way to capture screenshot, the poster bug probably wont exist, so do i still need to check whether the background is semi-transparent?

packages/model-viewer/src/test/model-viewer-spec.ts
```diff
@@ -0,0 +1,136 @@
...
+        teardown(() => {
+          if (element.parentNode != null) {
+            element.parentNode.removeChild(element);
+          }
+        });
+
+        test(scenario.name, async () => {
+          const blob = await element.toBlob();
+          const screenshotCanvas = await captureScreenshot(blob);
+          const result = testFidelity(screenshotCanvas);
```
elalish, line 125: Is there a reason you can't just say `screenshotCanvas = element[$renderer].displayCanvas(element[$scene])`? I don't think you should need to do all the round-trips through blobs and images, right?

packages/model-viewer/src/test/model-viewer-spec.ts
```diff
@@ -0,0 +1,136 @@
...
+        return tagName;
+      }
+    }
+    customElements.define(tagName, ModelViewer);
+  });
+
+  BasicSpecTemplate(() => ModelViewer, () => tagName);
+
+  suite('Fidelity Test', () => {
+    SCENARIOS.forEach((scenario) => {
```
elalish, line 106: For tests, it's nicer to have them be clear and separate. I'd skip this and remove the configs entirely. These three scenarios use the same configuration anyway, so you can just hard-code it in `setupModelViewer()`. Then just have three separate tests that just set the src and the environment, so it's easy to read through and see what each one does.

packages/model-viewer/src/test/model-viewer-spec.ts
elalish: Yeah, that makes sense; let's skip the transparent count.

packages/model-viewer/src/test/model-viewer-spec.ts
```diff
@@ -0,0 +1,169 @@
...
+      screenshotContext.UNSIGNED_BYTE,
+      pixels);
+
+  let colorlessPixelCountCanvas = 0;
+  for (let row = 0; row < height; row++) {
+    for (let col = 0; col < width; col++) {
+      let isWhite = true;
+      let isBlack = true;
+
+      // read pixel data from top left corner
```
sun765, line 64: I feel like reading from top-down direction is more intuitive. If you feel this is not necessary, im happy to change. Besides, i double checked the pixels data and compared them with the data from the canvas i previously used, 100% match.

packages/model-viewer/src/test/model-viewer-spec.ts
```diff
@@ -0,0 +1,169 @@
...
+// is colorless or not. Replace this with more robust
+// test in later pr.
+function testFidelity(screenshotContext: WebGLRenderingContext) {
+  const width = screenshotContext.drawingBufferWidth;
+  const height = screenshotContext.drawingBufferHeight;
+
+  const pixels = new Uint8Array(width * height * COMPONENTS_PER_PIXEL);
+  // this function reads in the bottom-up direction from the coordinate
+  // specified ((0,0) is the bottom-left corner).
+  screenshotContext.readPixels(
```
sun765, line 49: Although im not using alpha value at all, i think it may be useful for later work, so i read the pixels in rgba format.

packages/model-viewer/src/test/model-viewer-spec.ts
```diff
@@ -0,0 +1,169 @@
...
+          isWhite = false;
+        }
+        if (colorComponent != 0) {
+          isBlack = false;
+        }
+      }
+
+      if (isWhite || isBlack) {
+        colorlessPixelCountCanvas++;
+      }
```
bejado, line 80: This logic is still a bit strange to me. If you rendered a black and white checkerboard, the test would fail. I think what you're going for is checking if _all_ the pixels are black or if _all_ the pixels are white. Maybe it would be better to keep track of two sums separately, one for black pixels and one for white.

Also, a pixel with color [0, 0, 0, 255] doesn't get counted as "black" but I think it should?

elalish: I think this is okay, since we aren't rendering any checkerboards. A common type of error is a black model on a white/transparent background and we want to ensure we catch that.

packages/model-viewer/src/test/model-viewer-spec.ts
```diff
@@ -0,0 +1,169 @@
...
+  BasicSpecTemplate(() => ModelViewer, () => tagName);
+
+  suite('Fidelity Test', () => {
+    suite('Metal roughness spheres', () => {
+      let element: ModelViewerElement;
+
+      setup(async () => {
+        element = new ModelViewerElement();
+        document.body.insertBefore(element, document.body.firstChild);
+        await setupModelViewer(element, LIGHTROOM_PATH);
```
elalish, line 112: Let's remove lighting from `setupModelViewer` so that all the tests can share the same setup function. Then just apply the individual lighting in each test. The whole point of these setup functions is to avoid code duplication.

elalish: In fact you can probably make a separate function that sets the lighting and awaits the poster and call that in each test.

PR #1468, opened 2020-08-19
Ar ux buttons to move objects

Wulff2099:
Can somebody make some buttons to move objects foward backward and turn left and right.
Thanks


elalish:
Did you really mean to open a PR here? Also, you can implement buttons like this on your website with a little bit of HTML; just use the onClick method to change the `camera-orbit` and `camera-target` attributes as desired.

PR #1473, opened 2020-08-20
Fidelity docs

elalish:
This adds description to our fidelity page and reorders the scenarios a little.

### DIFF
--- a/packages/render-fidelity-tools/test/results-viewer.html
+++ b/packages/render-fidelity-tools/test/results-viewer.html
@@ -26,13 +26,79 @@
   <!-- Web Components polyfill is required to support Edge and Firefox < 63: -->
   <script src="../node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 <style>
-body {
-  margin: 1em;
-  background-color: #f7f7f7;
-}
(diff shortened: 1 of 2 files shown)


packages/render-fidelity-tools/test/results-viewer.html
```diff
@@ -26,13 +26,79 @@
...
+    (represented by &lt;model-viewer&gt;), <a
+    href="https://google.github.io/filament/webgl/"
+    target="_blank">filament.js</a>, and <a href="https://www.babylonjs.com/"
+    target="_blank">babylon.js</a>. If any other renderers would like to be
+    included, please open a PR adding them to the <a
+    href="https://github.com/google/model-viewer/tree/master/packages/render-fidelity-tools/src/components/renderers"
+    target="_blank">render-fidelity-tools</a>
+    package.</p>
+
+    <p>In addition, we are showing a ground-truth render provided by <a
```
sun765, line 69: maybe renderer instead of render? but im bad in in English , if they mean the same thing then just ignore it .

elalish: In this case it is a render (image) provided by a renderer (dspbr-pt).

PR #1477, opened 2020-08-24
Hide model viewer

sun765:
Refactor the fidelity test in this way: for every scenario,  compare model-viewer's golden with other renderers' goldens first (compareRenderers), then compare model-viewer's golden with its screenshot generate from fidelity test(autoTest). 

This is how the dev page looks like now. (metrics is the same as in dev page)
image

Here's how the workflow logs looks like
image


packages/render-fidelity-tools/src/workflows/test-fidelity.ts
```diff
@@ -57,27 +57,17 @@ if (process.argv.length > 3) {
   }
 }
 
-screenshotCreator.captureAndAnalyzeScreenshots(scenarioWhitelist)
+screenshotCreator.fidelityTest(scenarioWhitelist)
```
sun765, line 60: if you don't like the name changes i can change it back

packages/render-fidelity-tools/src/common.ts
```diff
@@ -237,10 +243,12 @@ export class ImageComparator {
     // when the golden image's background (use top left pixel to represent) is
     // transparent while the candidate(mode-viewer) image is not, it's
     // definitely that case
-    const candidateTopLeftAlpha = candidateImage[3];
-    const goldenTopLeftAlpha = goldenImage[3];
-    if (goldenTopLeftAlpha != candidateTopLeftAlpha) {
-      throw new Error(WARNING_MESSAGE);
+    if (checkSemiTransparent) {
```
sun765, line 246: Some filament's goldens' topleft pixel's alpha is not 0 (probably something i did wrong?), causing a lot of trouble. I just disable this check when comparing model-viewers' goldens with other renderers' golden for now, and i think we should discuss a better solution

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -48,168 +48,186 @@ export class ArtifactCreator {
...
+            (10 * Math.log10(rmsDistanceRatio)).toFixed(2)}`);
 
-    return analysisResults;
+    return analysis;
   }
 
   async captureScreenshot(
       renderer: string, scenarioName: string, dimensions: Dimensions,
-      outputPath: string = join(this.outputDirectory, 'model-viewer.png'),
+      outputPath: string|null = join(this.outputDirectory, 'model-viewer.png'),
```
sun765, line 230: just dont want to write the model-viewer's screenshot into result folder anymore because it's not used in the dev page

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -48,168 +48,186 @@ export class ArtifactCreator {
...
-      console.log(
-          `\n🔍 Comparing <model-viewer> to ${goldenConfig.description}`);
+      console.log(`\n🎨 Scenario: ${scenarioName}`);
 
-      const goldenPath =
-          join(rootDirectory, 'goldens', scenarioName, goldenConfig.file)
+      const scenarioOutputDirectory = join(outputDirectory, scenarioName);
+      mkdirp.sync(scenarioOutputDirectory);
+
+      await this.compareRenderers(scenario);
```
sun765, line 177: im not using try catch block for comapre renderers because i feel like auto test has already most possible errors. But compareRenderers probably may have its own errors as well, such as it can't find some renderers' goldens. If you want me to try and catch these errors as well im happy to change.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -48,18 +48,124 @@ export class ArtifactCreator {
...
+    try {
+      modelViewerGolden = await fs.readFile(modelViewerGoldenPath);
+    } catch (error) {
+      throw new Error(`❌ Failed to read model-viewer's ${
+          scenarioName} golden! Error message: ${error.message}`);
+    }
+
+    await fs.writeFile(
+        join(outputDirectory, scenarioName, goldens[modelViewerIndex].file),
+        modelViewerGolden);
```
elalish, line 72: What file are we writing here? What's it for?

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -48,18 +48,124 @@ export class ArtifactCreator {
...
+
+    const modelViewerGoldenImage = pngjs.PNG.sync.read(modelViewerGolden).data;
+
+    for (let golden of goldens) {
```
elalish, line 76: I think you can do `const golden of goldens` if you're not actually modifying the elements.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -48,18 +48,124 @@ export class ArtifactCreator {
...
+      let candidateGolden;
+      try {
+        candidateGolden = await fs.readFile(candidateGoldenPath);
+      } catch (error) {
+        throw new Error(`❌ Failed to read ${golden.name}'s ${
+            scenarioName} golden! Error message: ${error.message}`);
+      }
+
+      await fs.writeFile(
+          join(outputDirectory, scenarioName, golden.file), candidateGolden);
```
elalish, line 95: Same question as above.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -48,18 +48,124 @@ export class ArtifactCreator {
...
+    }
+    const scenarioRecord = {analysisResults, scenario};
+
+    console.log(`\n💾 Recording analysis`);
+    await fs.writeFile(
+        join(outputDirectory, scenarioName, 'analysis.json'),
+        JSON.stringify(scenarioRecord));
+  }
+
+  async autoTest(scenario: ScenarioConfig): Promise<ImageComparisonAnalysis> {
```
elalish, line 110: autoTest is a pretty generic name. Let's try to be specific about what we're doing. Maybe this is the function that should be called `captureAndAnalyzeScreenshot`?

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -48,18 +48,124 @@ export class ArtifactCreator {
...
-    const {goldens, outputDirectory} = this;
-    const modelViewerFidelityErrors: Array<string> = [];
-    const modelViewerFidelityWarnings: Array<string> = [];
+    const autoTestResults:
+        AutoTestResults = {results: [], errors: [], warnings: []};
+
+    const compareRenderersErrors: Array<string> = [];
 
-    for (const scenarioBase of scenarios) {
+    for (let scenarioBase of scenarios) {
```
elalish, line 166: Why was this change necessary?

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -68,148 +174,71 @@ export class ArtifactCreator {
...
-        join(outputDirectory, 'modelViewerFidelityErrors.json'),
-        JSON.stringify(modelViewerFidelityErrors));
-
-    await fs.writeFile(
-        join(outputDirectory, 'modelViewerFidelityWarnings.json'),
-        JSON.stringify(modelViewerFidelityWarnings));
-
-    return scenarios;
+        join(outputDirectory, 'autoTestResults.json'),
+        JSON.stringify(autoTestResults));
```
elalish, line 217: Another possible name: fidelityRegressionResults

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -301,8 +330,9 @@ export class ArtifactCreator {
       // Ignored...
     }
 
-    const screenshot =
-        await page.screenshot({path: outputPath, omitBackground: true});
+    const screenshot = outputPath ?
+        await page.screenshot({path: outputPath, omitBackground: true}) :
+        await page.screenshot({omitBackground: true});
```
elalish, line 335: Can this be combined using `path: outputPath?` Also, how do you use/override the default value of outputPath? There might be a simpler way to write that function input.

packages/render-fidelity-tools/src/workflows/test-fidelity.ts
```diff
@@ -57,53 +57,68 @@ if (process.argv.length > 3) {
...
-        for (const error of modelViewerFidelityErrors) {
+        for (const error of compareRendererErrors) {
           console.log(error);
         }
+      }
 
+      if (autoTestErrorCount > 0 || compareRendererErrorCount > 0) {
         throw new Error(
-            'Model Viewer failed the fidelity test! Please fix the fidelity error before merging to master!');
+            ' ❌ Fidelity test failed! Please fix the errors list above before mering this pr!');
```
elalish, line 121: errors listed above

packages/render-fidelity-tools/src/artifact-creator.ts
sun765: These golden images are saved to the result folder which are use to show in the result-viewer page. I guess the reason is sometimes you only run fidelity test on specified scenarios, so saving to another folder instead of just reading it from the golden folder can help distinguish which ones to show in the local result-viewer page. (probably there are some other reasons i did get). I have just added a comment here, but do you want me to refactor it so that the result page find the golden images from the golden folder directly?

elalish: Thank you, I'd completely forgotten about that. No, let's leave it as is; thanks for the comment.

packages/render-fidelity-tools/src/artifact-creator.ts
```diff
@@ -330,9 +339,8 @@ export class ArtifactCreator {
       // Ignored...
     }
 
-    const screenshot = outputPath ?
-        await page.screenshot({path: outputPath, omitBackground: true}) :
-        await page.screenshot({omitBackground: true});
+    const screenshot =
+        await page.screenshot({path: outputPath, omitBackground: true});
```
elalish: 👍

PR #1506, opened 2020-09-03
round canvas size

### LINKED ISSUE #1503: Model-Viewer sometimes renders blurry (wrong res?) on some devices when some sort of zoom is used
### Description

We're receiving reports from some customers that content is rendering blurry. We seem to have it tracked down to zoom settings (in the browser/in Windows/not sure).

200903-185002-ApplicationFrameHost
_Note the super blurry image on the left – backgrounds are actually different_

#### Live Demo
https://prefrontalcortex.de/labs/siemensstadt-ar-v2/
https://www.siemensstadt.siemens.com/en/ar-map

We were able to reproduce the issue (both on 0.9 and 1.1 actually, so not 100% sure about the "regression" remark) by going to the Chrome dev tools, enabling "mobile mode", and then setting the zoom settings to 150%:
image

Note that when you refresh the page in that 150% mode, everything is crisp and clear and nice, until you set it to 100%, go back to desktop mode, repeat the process. So the assumption would be that under specific circumstances (e.g. page is scaled, Windows is scaled, mobile is "zoomed in", somehow the resulting canvas resolution is wrong.

Actually just found another way to reproduce:
- Open https://prefrontalcortex.de/labs/siemensstadt-ar-v2/
- Hover over a non-model-viewer part (e.g. some button)
- use Ctrl + Mousewheel to change page zoom
- Note that at 100%, everything is sharp
- 90% and 110% are the worst with very noticeable blur
- at 80% and 125% it's slightly noticeable
- at both lower and higher zooms it's not noticeable at all, looks sharp

### Browser Affected

- [X] Chrome
- [ ] Edge
- [X] Firefox
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [X] Windows

elalish:
Fixes #1503 

Well, I can't repro the issue so I can't be sure, but this seems a likely fix. @hybridherbst Can you verify this fix before I merge?

### DIFF
--- a/packages/model-viewer/src/three-components/Renderer.ts
+++ b/packages/model-viewer/src/three-components/Renderer.ts
@@ -205,8 +205,8 @@ export class Renderer extends EventDispatcher {
     // and only the portion that is shown is copied over.
     for (const scene of this.scenes) {
       const {canvas} = scene;
-      canvas.width = width * dpr;
-      canvas.height = height * dpr;
+      canvas.width = Math.round(width * dpr);
+      canvas.height = Math.round(height * dpr);
       canvas.style.width = `${widthCSS}px`;
       canvas.style.height = `${heightCSS}px`;
       scene.isDirty = true;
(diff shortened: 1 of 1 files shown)


hybridherbst:
Sorry I was too slow with testing 😔

hybridherbst:
@elalish I finally got to test this; looks like it's nearly fixed!
From zooming around I can see that now all resolutions stay at exactly the same pixel values but 90% and 110% still jump by a pixel, reducing clarity. This means that almost every zoom looks super crisp and sharp but the common values of 90% and 110% are still somewhat blurry.


(the relevant part is that in the lower right, the `<canvas width=518 height=1121>` jumps to `<canvas width=519 height=1122>` when zooming at 90% and 110%.

elalish:
@hybridherbst Thanks for verifying. Let's see if this second PR fixes it, as Three wasn't using the same kind of rounding I was.

PR #1542, opened 2020-09-16
Fix further canvas rounding issues

elalish:
Follow on to #1506

### DIFF
--- a/packages/model-viewer/src/test/three-components/Renderer-spec.ts
+++ b/packages/model-viewer/src/test/three-components/Renderer-spec.ts
@@ -27,6 +27,9 @@ const ModelViewerElement = class extends ModelViewerElementBase {
   }
 };
 
+// Ensure tests are not rescaling
+ModelViewerElement.minimumRenderScale = 1;
+
 interface TestScene {
   renderCount?: number;
 }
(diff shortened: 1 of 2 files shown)


hybridherbst:
Hmm, unfortunately I see a regression here. **Now 100% is blurry!** And it seems that while height stays constant, width is jumping by a pixel. So weird!

Comparing with https://modelviewer.dev/examples/lighting-and-environment.html shows me that when I change zoom,
- on master, canvas width stays constant at 623 and height toggles between 968 (sharp) and 967 (blurry)
  - with 100% → 623x968 (sharp)
- on canvasRounding, canvas width toggles between 624 (blurry) and 623 (sharp), height is constant at 968
  - with 100% → 624x968 (blurry)

ezgif com-optimize (8)

hybridherbst:
@elalish please note my comment here — you might not want to merge that in as-is, it introduces blurriness to the default 100% zoom for some reason

elalish:
Okay, reverting. This is looking like browser bugs of some kind. Remind me, are you testing chrome on Windows? Of course this all works perfectly on my mac. This was all the ideas I had, so please feel free to play with it yourself since you can repro.

PR #1558, opened 2020-09-22
Progress bar updated

chrismgeorge:
Example progress bar based on: https://twitter.com/modelviewer/status/1306671314749464578

### DIFF
--- a/packages/modelviewer.dev/examples/lighthouse.html
+++ b/packages/modelviewer.dev/examples/lighthouse.html
@@ -287,7 +287,7 @@
     const onProgress = (event) => {
       const progressBar = event.target.querySelector('.progress-bar');
       const updatingBar = event.target.querySelector('.update-bar');
-      updatingBar.style.width=`${event.detail.totalProgress*100}%`;
+      updatingBar.style.width = `${event.detail.totalProgress*100}%`;
       if (event.detail.totalProgress == 1) {
         progressBar.classList.add('hide');
       }
--- a/packages/modelviewer.dev/examples/lighthouse.html
+++ b/packages/modelviewer.dev/examples/lighthouse.html
@@ -92,7 +92,7 @@
     }
  
     .update-bar {
-      background-color: rgba(255, 255, 255, 0.9);
+      background-color: rgba(255, 0, 0, 0.9);
       width: 0%;
       height: 100%;
       border-radius: 25px;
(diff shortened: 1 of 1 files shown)


packages/modelviewer.dev/examples/lighthouse.html
```diff
@@ -268,11 +285,13 @@ <h2>Setting up <code>&lt;model-viewer&gt;</code> for good Lighthouse scores</h2>
...
-        firstBar.classList.add('hide');
-      }else{
-        firstBar.style.width=`${event.detail.totalProgress*50}%`;
+      const progressBar = event.target.querySelector('.progress-bar');
+      const updatingBar = event.target.querySelector('.update-bar');
+      if (event.detail.totalProgress == 1) {
+        updatingBar.style.width=`${event.detail.totalProgress*100}%`;
+        progressBar.classList.add('hide');
+      } else {
+        updatingBar.style.width=`${event.detail.totalProgress*100}%`;
```
elalish, line 294: Since this is happening on both sides of the conditional, let's move it out and simplify.

packages/modelviewer.dev/examples/lighthouse.html
```diff
@@ -74,19 +74,32 @@
 
     .progress-bar {
       display: block;
-      width: 0%;
+      width: 33%;
       height: 10%;
+      max-height: 2%;
```
elalish, line 79: Does height 10% mean much when max-height is 2%?

elalish (changes requested):
Looks great, just a couple minor tweaks.

packages/modelviewer.dev/examples/lighthouse.html
chrismgeorge: The height has to be specified, given the setup. If you play with [this](https://www.bitdegree.org/learn/best-code-editor/css-height-and-width-dimensions-example-5) - Comment out min-height and height, then the max-height doesn't do anything. As long as height is set, though, max-height will override it.

chrismgeorge: 10% was an arbitrary number

packages/modelviewer.dev/examples/lighthouse.html
chrismgeorge: done

elalish (approved):
Great, thanks!

PR #1564, opened 2020-09-24
Improve lighthouse example prompt

elalish:
This adds the interaction prompt manually before the element has loaded (or before the model is loaded in the second version). Possible breaking UX to consider for the future: make this the default behavior of the interaction prompt before the model has been revealed.


packages/modelviewer.dev/examples/lighthouse2.html
```diff
@@ -62,6 +62,69 @@
       display: none;
     }
 
+    .center {
```
chrismgeorge, line 65: possibly want a better name, what are you centering? i.e. center-pre-prompt

packages/modelviewer.dev/examples/lighthouse2.html
```diff
@@ -62,6 +62,69 @@
...
+      display: flex;
+      justify-content: center;
+      align-items: center; 
+      height: 100%;
+      left: 0;
+      right: 0;
+      top: 0;
+      bottom: 0;
```
chrismgeorge, line 73: According to this: https://css-tricks.com/almanac/properties/t/top-right-bottom-left/#setting-opposite-sides

if you have (top and left) set (right and bottom) are usually ignored.

chrismgeorge (commented):
Other than the comments, looks good!

chrismgeorge (dismissed):

PR #1570, opened 2020-09-28
added lighting control code

dhruvadhia-cela:
Adding a way to control environment-image by pressing a key and drag rotate with mouse. This will help users to check how the environment light casts onto an object from different angles.


dhruvadhia-cela:
@googlebot I signed it!

packages/modelviewer.dev/examples/staging-and-camera-control.html
```diff
@@ -185,11 +185,13 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
 <script>
 (() => {
+  
   const modelViewer = document.querySelector('#pan-demo');
+  modelViewer.addEventListener('contextmenu', event => event.preventDefault());
   const tapDistance = 2;
   let panning = false;
   let panX, panY;
-  let startX, startY;
+  let firstX, firstY;
```
elalish, line 194: Looks like you need to pull master; these are recent changes your PR is trying to undo.

packages/modelviewer.dev/examples/staging-and-camera-control.html
```diff
@@ -302,7 +302,60 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
+  // Add the event listeners for mousedown, mousemove, and mouseup
+  modelViewer.addEventListener('mousedown', (event) => {
+    isLightChanging = true;
+  }, true);
+
+  modelViewer.addEventListener('mousemove', (event) => {
+	  
+  if (isLightChanging === true) {
+    if (event.shiftKey && event.which == 1) {
+	  modelViewer.cameraOrbit = `${orbit.theta + 0.01}rad ${orbit.phi}rad ${orbit.radius}m`;
```
elalish, line 338: I think it might make more sense to connect the rotation to the amount of mouse motion; see the panning example.

packages/modelviewer.dev/examples/staging-and-camera-control.html
```diff
@@ -302,7 +302,60 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
+	  modelViewer.resetTurntableRotation(-1 * orbit.theta);
```
elalish, line 339: If you want the object to stay steady while the environment moves, you'll want to store your turntable rotation and then subtract the orbit.theta change you made from it.

packages/modelviewer.dev/examples/staging-and-camera-control.html
```diff
@@ -302,7 +302,60 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
+  const modelViewer = document.querySelector("#envlight-demo");
+	  
+  const updateOrbit = () => {
+  const orbit = modelViewer.getCameraOrbit(); 
+  modelViewer.jumpCameraToGoal();
+  requestAnimationFrame(() => updateOrbit());
+  var isLightChanging = false;
+  // Add the event listeners for mousedown, mousemove, and mouseup
+  modelViewer.addEventListener('mousedown', (event) => {
+    isLightChanging = true;
```
elalish, line 331: I don't think you really need this mode switch or the extra event listeners.

packages/modelviewer.dev/examples/staging-and-camera-control.html
```diff
@@ -302,7 +302,60 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
+  }, true);
+
+  modelViewer.addEventListener('mousemove', (event) => {
+	  
+  if (isLightChanging === true) {
+    if (event.shiftKey && event.which == 1) {
```
elalish, line 337: You'll probably want to call `event.preventDefault()` in this `if` statement so that you don't get the normal camera motion.

elalish (changes requested):
Thanks! Have you tried this out with both mouse and touch?

packages/modelviewer.dev/examples/staging-and-camera-control.html
dhruvadhia-cela: > Thanks! Have you tried this out with both mouse and touch?

Not yet, just doing it with mouse currently.

dhruvadhia-cela:
Hi @elalish,

I have so far resolved all the comments you mentioned. However, I was facing permission issues as I was using different GitHub user on my local machine. It seems like you will have to consent in order to verify the author. 

Apart from this, I got the desired effect with the mouse, however, you will notice it is not perfect in the sense I want to rotate the environment-image while a shift + left click-drag. When I let go of the mouse, I can only rotate the model along the y axis. Also After some drag rotation in both the direction, there is this weird effect that the model loses its quality and takes a lot of CPU usage. Your input at this stage will be appreciated as I think we are pretty close to achieving the desired effect of environment map rotation.

dhruvadhia-cela:
@googlebot I consent

dhruvadhia1:
@googlebot I consent

dhruvadhia-cela:
> Hi @elalish,
> 
> I have so far resolved all the comments you mentioned. However, I was facing permission issues as I was using different GitHub user on my local machine. It seems like you will have to consent in order to verify the author.
> 

ok nevermind, this is resolved. I just had to consent from another account. Kindly review the code as commented above.

dhruvadhia-cela:
> Thanks! Have you tried this out with both mouse and touch?
@elalish,

Are you waiting for me to add touch inputs as well? I was wondering in the meantime you can have a look at the rest of the code.

packages/modelviewer.dev/examples/staging-and-camera-control.html
```diff
@@ -302,7 +307,66 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
+  var speedX = event.movementX;
+	  
+  if (event.shiftKey && event.which == 1) {
+			
+    const updateOrbit = () => {
+    const orbit = modelViewer.getCameraOrbit(); 
+    console.log(speedX);
+    modelViewer.cameraOrbit = `${orbit.theta + 0.001*(-speedX)}rad ${orbit.phi}rad`;
+    modelViewer.jumpCameraToGoal();
+    modelViewer.resetTurntableRotation(orbit.theta - 0.01*(-speedX));
```
elalish, line 347: Did you test that this works? I see you multiplying by 0.01 here and 0.001 above. Probably better to replace `speedX` by `deltaTheta` and only do the multiplication once.

packages/modelviewer.dev/examples/staging-and-camera-control.html
```diff
@@ -302,7 +307,66 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
+   
+  // Speed and direction on X axis   
+  var speedX = event.movementX;
+	  
+  if (event.shiftKey && event.which == 1) {
+			
+    const updateOrbit = () => {
+    const orbit = modelViewer.getCameraOrbit(); 
+    console.log(speedX);
+    modelViewer.cameraOrbit = `${orbit.theta + 0.001*(-speedX)}rad ${orbit.phi}rad`;
```
elalish, line 345: you should probably add `auto` to the end of this string for the radius, to make it more clear.

packages/modelviewer.dev/examples/staging-and-camera-control.html
```diff
@@ -302,7 +307,66 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
+  // Add the event listeners for mousedown, mousemove, and mouseup
+  modelViewer.addEventListener('mousedown', (event) => {
 
+  	console.log("mouse down");
+  }, true);
+  
+  modelViewer.addEventListener('mousemove', (event) => {
+   
+  // Speed and direction on X axis   
+  var speedX = event.movementX;
```
elalish, line 338: This isn't supported in IE11 and there's no equivalent with touch, so probably better to avoid.

packages/modelviewer.dev/examples/staging-and-camera-control.html
```diff
@@ -302,7 +307,66 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
+    modelViewer.jumpCameraToGoal();
+    modelViewer.resetTurntableRotation(orbit.theta - 0.01*(-speedX));
+    requestAnimationFrame(() => updateOrbit()); 
+   };
+   updateOrbit();
+   }
+  }, true);
+
+  modelViewer.addEventListener('mouseup', (event) => {
+    const updateOrbit = () => {
```
elalish, line 355: Why is this function needed?

elalish (changes requested):
Thanks, have you tried `npm run serve` to test out your example? You might consider using PointerEvents instead in order to handle touch and mouse at the same time.

packages/modelviewer.dev/examples/staging-and-camera-control.html
dhruvadhia-cela: Ok, will replace that with next check in

packages/modelviewer.dev/examples/staging-and-camera-control.html
dhruvadhia-cela: I wanted to reset everything on mouse up event. Like reinstating to initial state of stoping environment image to rotate and only rotate object on mouse drag.

packages/modelviewer.dev/examples/staging-and-camera-control.html
dhruvadhia-cela: Ok will get it resolved in next commit.

dhruvadhia-cela:
> Thanks, have you tried `npm run serve` to test out your example? You might consider using PointerEvents instead in order to handle touch and mouse at the same time.

Yes, I am testing everything with npm run serve. Will consider PointerEvents in next commit.

packages/modelviewer.dev/examples/staging-and-camera-control.html
dhruvadhia-cela: oh yes thats a typo. I think i have fixed that locally already. I don't get on what you mean by doing multiplication once. You mean outside update call?

dhruvadhia1:
@elalish So I have finally made progress and refined the code. It's very close to the final experience I am trying to have. Usage goes as follow:

First press shift and then mouse drag move along horizontal/X axis to move the background. The only problem now is that every time I let go of environment rotating control and let's say I simply rotate the model and leave it at a certain orientation. And then I try to rotate the environment map again, it will reset the model. 

This line causes the reset, and if I don't use this line then obviously the model does not stay in its own place while rotating the environment map. 

 modelViewer.resetTurntableRotation(orbit.theta);

Any leads on resolving this issue would be perfect and I am sure we can then check into the examples section. 

Cheers
Dhruv

dhruvadhia-cela:
@elalish did you get a chance to look at the commit?

PR #1571, opened 2020-09-29
Render fidelity: upgrade Filament.

prideout:
This upgrade affects AlphaModeBlendTest as well as memory usage when using the test framework interactively.


packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -234,10 +244,12 @@ export class FilamentViewer extends LitElement {
...
     });
 
-    loader.delete();
-
     this[$boundingBox] = asset.getBoundingBox();
-    this[$scene].addEntities(asset.getEntities());
+    const entities = asset.getEntities();
+    this[$scene].addEntities(entities);
+    for (const entity of entities) {
+      entity.delete();
```
elalish, line 251: Why do we delete the entities right after adding them?

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
```diff
@@ -149,37 +152,40 @@ export class FilamentViewer extends LitElement {
...
+      this[$assetLoader]!.destroyAsset(existingAsset);
       this[$currentAsset] = null;
+      for (const entity of entities) {
+        entity.delete();
+      }
     }
 
     if (this[$ibl] != null) {
       this[$scene].setIndirectLight(null);
+      this[$engine].destroyTexture(this[$ibl]!.getReflectionsTexture());
```
elalish, line 168: We can remove all these `!` by doing `const ibl = this[$ibl]` and then operating no `ibl` instead. Actually, I'd really rather just strip all the Symbols out of this file, but I'm not going to make you do that. :)

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
prideout: This is an unfortunate emscripten-ism.  The emscripten entity wrapper is being deleted, not the entity itself.

elalish: ah, good to know. Is it still necessary to also delete them above then, or will they already be gone?

elalish: As in they are also deleted around line 160.

prideout: Added a commit with a comment.

packages/render-fidelity-tools/src/components/renderers/filament-viewer.ts
prideout: Added a commit to clean this up.

elalish (approved):
Thanks!

PR #1578, opened 2020-09-30
update deploy

chrismgeorge:
update the deploy script to deploy /docs and /data


elalish (approved):
Let's see if we've forgotten anything else...

PR #1580, opened 2020-10-01
Add Google Analytics tracking code

rkochman:


elalish (approved):
LGTM, thanks! This isn't going to put those "I accept cookies" popups on our site, is it?

staus:
> LGTM, thanks! This isn't going to put those "I accept cookies" popups on our site, is it?

This looks quite concerning to me. I really hope it gets removed!
Who owns this analytics? Is it everyone at Google? Is it this team? Is it an individual? Why?

elalish:
@staus Thanks for raising the concern. We want to get an idea of how much traffic our dev pages are getting, but I failed to realize that full analytics was doing so much more than that. We've reverted and will look for a lighter-weight solution that doesn't involve any cookies.

PR #1588, opened 2020-10-01
Update TextureUtils.ts

mqg734:
Allow HDR image files with an optional '.js' extension.


elalish (approved):
LGTM! Any idea how long we'll need this hack for?

PR #1590, opened 2020-10-01
Support multiple categories and example flow

chrismgeorge:
* Added legal switching between examples / documentation such that being inside one category takes you to the corresponding category.


packages/modelviewer.dev/examples/loading/index.html
```diff
@@ -0,0 +1,275 @@
...
+  </div>
+      
+</div>
+
+</div>
+
+<script type="module" src="../../examples/built/docs-and-examples.js">
+</script>
+<script type="module">
+  (() => { init('examples/loading'); })();
```
elalish, line 256: Presumably this addition is why these new file copies are here? And once we do the final docs merge, we'll replace the existing examples with these versions?

packages/modelviewer.dev/src/docs-and-examples/create-html.ts
```diff
@@ -33,6 +33,43 @@ const CategoryConstant: Category = {
...
+    <div class="subCategory">
+      ${isActive ? getExamples(category) : ''}
+    </div>
+  </div>
+    `;
+}
+
+export function createExamplesSidebar(json: any[]) {
+  for (const category of json) {
+    createExampleSidebarCategory(category);
```
elalish, line 69: This function is trivial enough you might just combine them (assuming the inner one isn't used elsewhere).

packages/modelviewer.dev/src/docs-and-examples/create-html.ts
```diff
@@ -209,19 +243,18 @@ function createSubcategory(
   }
 }
 
-export function convertJSONToHTML(
-    json: any[], docsOrExamples: 'docs'|'examples') {
```
elalish, line 213: 👍

packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
```diff
@@ -112,6 +112,7 @@ function updateSidebarView(
 
 /*
  * Hide all of the entries not within the current subcategory
+ * entries should be every entry on the page when this is called
```
elalish, line 115: Try to make comments complete sentences with punctuation and such.

packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
```diff
@@ -208,7 +209,7 @@ function handlePageJump(entries: IntersectionObserverEntry[]) {
 /*
  * Update the table of contents based on how the page is viewed.
  */
-export function sidebarObserver(docsOrExamples: 'docs'|'examples') {
+export function sidebarObserver(docsOrExamples: string) {
```
elalish, line 212: Maybe `docsOrExample` since we're choosing a specific one now?

elalish (approved):
LGTM.

packages/modelviewer.dev/examples/loading/index.html
chrismgeorge: yep! keeping two copies right now so nothing breaks -- necessary for testing

packages/modelviewer.dev/src/docs-and-examples/create-html.ts
chrismgeorge: what is the thumbs up for?

packages/modelviewer.dev/src/docs-and-examples/create-html.ts
chrismgeorge: hmmm, I like the little bit of extra readability this does, and it is consistent with how I iterate through jsons in other places.

packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
chrismgeorge: will do in next pr!

packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
chrismgeorge: hmm, yeah, I think you're right, I'll update that in the next one

packages/modelviewer.dev/src/docs-and-examples/create-html.ts
elalish: I like that you managed to remove the `docsOrExamples` argument.

packages/modelviewer.dev/src/docs-and-examples/create-html.ts
elalish: My only trouble with the readability is that I can't easily tell which function is which because they only differ by one letter in the middle. But I agree, it's not a big deal.

packages/modelviewer.dev/src/docs-and-examples/create-html.ts
chrismgeorge: cool cool

packages/modelviewer.dev/src/docs-and-examples/create-html.ts
chrismgeorge: createExample**s**Sidebar vs createExampleSidebar**Category**?

PR #1605, opened 2020-10-05
Sidebar clicking on small section bug fix

chrismgeorge:
Sidebar bug fixes and CSS updates to header and layout on mobile.

What's not in this PR:
* Examples header on mobile to flip between sidebar/ no sidebar
* Update toggle button to be it's own column and only on mobile instead of desktop

-- looking for general code review

### DIFF
--- a/packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
+++ b/packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
@@ -16,7 +16,8 @@
 let globalCurrentView: string[] = [];
 let previouslyActive: string = '';
 let toRemove = '';
-let order: any[] = [];  // TODO, switch to dictionary
+let order = new Map<string, number>();
+let isSideBarClick = false;
 
 let isFirstOpen = true;      // is true on the first observation of all entries
 let everyEntry: any[] = [];  // a list of all attributes/properties etc.
@@ -153,8 +154,7 @@ function handleHTMLEntry(htmlEntry: IntersectionObserverEntry) {
       activateSidebar(sidebarIds);
       previouslyActive = sidebarIds.name;
       globalCurrentView.push(sidebarIds.name);
-    } else if (
-        order.indexOf(previouslyActive) > order.indexOf(sidebarIds.name)) {
+    } else if (order.get(previouslyActive)! > order.get(sidebarIds.name)!) {
       // scrolling up
       updateFromOldToNew(globalCurrentView[0], sidebarIds);
       globalCurrentView.unshift(sidebarIds.name);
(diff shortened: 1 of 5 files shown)


packages/modelviewer.dev/src/docs-and-examples/utils.ts
```diff
@@ -86,36 +86,31 @@ function jumpToSection() {
...
-  const stringZero = zero.toString().concat('px');
-  const threeHundred = 300;
-  const stringThreeHundred = threeHundred.toString().concat('px');
-
   if (nav.offsetWidth > 150) {
-    root.style.setProperty('--sidebar-width', stringZero);
+    root.style.setProperty('--sidebar-width', '0px');
   } else {
-    root.style.setProperty('--sidebar-width', stringThreeHundred);
+    root.style.setProperty('--sidebar-width', '300px');
```
elalish, line 113: 👍

packages/modelviewer.dev/styles/docs.css
```diff
@@ -81,6 +81,7 @@ hr.solid {
 .tab {
   display: flex;
   align-items: center;
+  padding-right: 10px;
```
elalish, line 84: The style is looking good. Only thing I noticed is when you click on a sidebar item on mobile it turns gray (and stays gray) like it's somehow getting the hover state applied.

elalish (approved):
A few minor things; up to you if you want to address them in this PR or not.

packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
```diff
@@ -194,27 +194,43 @@ function handleHTMLEntry(htmlEntry: IntersectionObserverEntry) {
...
+      globalCurrentView.push(sidebarIds.name);
     }
   }
+
+  // sort current view
+  globalCurrentView.sort(function(nameA, nameB) {
+    if (order.get(nameA)! < order.get(nameB)!) {
+      return -1;
+    } else if (order.get(nameA)! > order.get(nameB)!) {
+      return 1;
```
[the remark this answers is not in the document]
chrismgeorge: I believe the custom sorting expects a number returned (at least typescript does), and I wouldn't want to run into any edge cases with the alternative of casting a boolean to a number

packages/modelviewer.dev/styles/docs.css
chrismgeorge: I think this is a consistent mobile feature. I could have the sidebar close after the click, which would avoid that.

packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
```diff
@@ -256,26 +272,29 @@ function handleExamples(entries: IntersectionObserverEntry[], _observer: any) {
...
       }
       updateHeader();
     });
     // i.e. attributes, properties, events, methods, slots, custom css.
+    let orderIndex = 0;
     document.querySelectorAll('div[id*="docs"]').forEach((section) => {
       const idSplitList = section.getAttribute('id')!.split('-');
       if (idSplitList.length === 4) {
-        order.push(idSplitList.slice(1, 10).join('-'));
+        const id = idSplitList.slice(1, 10).join('-');
```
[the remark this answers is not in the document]
chrismgeorge: Yeah, this is just my way of finding all of the entries, so I'll give that a global

packages/modelviewer.dev/styles/docs.css
elalish: Yes, I think closing makes sense, especially because you can't really see the content has changed now.

packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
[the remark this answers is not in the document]
elalish: Ah, you're right, but there's still a simple solution: `return order.get(nameA)! - order.get(nameB)!;`, from here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

packages/modelviewer.dev/styles/docs.css
chrismgeorge: done in new pr

PR #1613, opened 2020-10-08
Enable viewport scaling, fix view assumptions

### LINKED ISSUE #1612: add WebXR dynamic viewport scaling support
### Description

Feature request: add support for dynamic viewport scaling as recently added to the WebXR spec. Since this isn't widely supported in user agents yet, make it conditional on support being present.

The WebXR explainer section https://github.com/immersive-web/webxr/blob/master/explainer.md#dynamic-viewport-scaling has documentation for the feature.

klausw:
Use dynamic viewport scaling using the recommended viewport scale
if supported.

For correct rendering, loop over the returned view array instead of just 
hardcoding the first view, and use each view's corresponding viewport.

### Reference Issue
Fixes #1612


packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -457,6 +460,15 @@ export class ARRenderer extends EventDispatcher {
       this.dispatchEvent({type: 'status', status: ARStatus.SESSION_STARTED});
     }
 
+    // Use automatic dynamic viewport scaling if supported.
+    if (view.requestViewportScale && view.recommendedViewportScale) {
+      const scale = view.recommendedViewportScale;
+      view.requestViewportScale(Math.max(scale, MIN_VIEWPORT_SCALE));
+    }
+    const layer = this[$currentSession]!.renderState.baseLayer;
+    const viewport = layer!.getViewport(view);
```
elalish, line 469: Our tests are failing on this line in Chrome 85. Any idea why?

elalish (changes requested):

packages/model-viewer/src/three-components/ARRenderer.ts
klausw: The mock XRSession data was missing the getViewport method on the layer.

Is there a way to run tests individually? My laptop seems to have trouble running the full suite, so the fix is a bit speculative.

elalish: Yes, on any of the `test()` or `suite()` calls, just add a `.only`, like `test.only()`.

elalish: Anyway, it passed our CI, so looks like it's all good. Thanks!

elalish (approved):
For some reason my Chrome Canary refuses to update, even though Play knows it's out of date (still on 85 for some reason). This is working fine on 85, so I'll just merge it once the tests pass and let's test it on the new thing.

PR #1615, opened 2020-10-08
add redux devtools

chrismgeorge:
adding redux devtools

--general review of content


elalish (approved):
I checked and the extension is working fine for me. Good find, this seems like it'll make debug much easier!

PR #1617, opened 2020-10-08
Docs updates

chrismgeorge:
- update docs with new information
- remove table "type"
- add css for attributes

### DIFF
--- a/packages/modelviewer.dev/data/docs.json
+++ b/packages/modelviewer.dev/data/docs.json
@@ -203,24 +203,19 @@
       {
         "name": "progress",
         "htmlName": "progress",
-        "description": "TODO"
-      },
-      {
-        "name": "poster-dismissed",
-        "htmlName": "posterDismissed",
-        "description": "TODO"
+        "description": "This event fires while a model, environment, and/or skybox is loading, during both download and processing. The progress of all of these concurrent tasks will be given by a value between 0 and 1: <span class=\"attribute\">event.detail.totalProgress</span>."
       }
     ],
     "Slots": [
       {
         "name": "poster",
         "htmlName": "poster",
-        "description": "TODO"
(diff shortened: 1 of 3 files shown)


packages/modelviewer.dev/data/docs.json
```diff
@@ -395,77 +367,69 @@
...
         }
       },
       {
         "name": "camera-target",
         "htmlName": "cameraTarget",
         "description": "Set the starting and/or subsequent point the camera orbits around. Accepts values of the form \"$X $Y $Z\", like \"0m 1.5m -0.5m\". Also supports units in centimeters (\"cm\") or millimeters (\"mm\"). A special value \"auto\" can be used, which sets the target to the center of the model's bounding box in that dimension. Any time this value changes from its initially configured value, the camera will interpolate from its current position to the new value. Defaults to \"auto auto auto\".",
         "default": {
-          "type": "$X $Y $Z",
           "default": "auto auto auto",
-          "options": "Multiple forms described in description"
```
elalish, line 410: Instead of "multiple forms", for ones like this I think what we had in the type was more descriptive and we can just move it to options.

packages/modelviewer.dev/styles/examples.css
```diff
@@ -629,7 +629,8 @@ paper-button {
 }
 
 .attribute {
-  white-space:nowrap;
+  white-space: pre-wrap !important;
+  font-family: var(--font-monospace);
```
elalish, line 633: 👍

elalish (changes requested):
Looks like you made your branch of my branch instead of off master? Generally speaking you always want to branch of the branch that you're going to merge directly into, otherwise git tends to get merge conflicts. Anyway, they shouldn't be hard to resolve. These two branches wouldn't have conflicts normally anyway, since they didn't operate on the same lines.

packages/modelviewer.dev/data/docs.json
chrismgeorge: done

chrismgeorge: good call--

elalish (approved):

PR #1623, opened 2020-10-09
remove parenthesis from sidebar methods

chrismgeorge:

### DIFF
--- a/packages/modelviewer.dev/data/docs.json
+++ b/packages/modelviewer.dev/data/docs.json
@@ -135,7 +135,7 @@
       {
         "name": "modelCacheSize",
         "htmlName": "modelCacheSize",
-        "description": "This static, writable property sets <span class='attribute'>&lt;model-viewer&gt;</span>'s internal glTF model cache size, controlling number of individual models that should be cached. Note that any number of models may be cached at a given time. This number just sets the maximum number of models that can be cached even if they are not being used by a <span class='attribute'>&lt;model-viewer&gt;</span> element in the document. Note also that the cache size is measured in number of glTF models, not bytes! This is important to keep in mind, since any two models may have wildly different byte sizes.",
+        "description": "This static, writable property sets <span class='attribute'>&lt;model-viewer&gt;</span>'s internal glTF model cache size, controlling number of individual models that should be cached. Note that any number of models may be cached at a given time. This number sets the maximum number of models that can be cached even if they are not being used by a <span class='attribute'>&lt;model-viewer&gt;</span> element in the document. Note also that the cache size is measured in number of glTF models, not bytes! This is important to keep in mind, since any two models may have wildly different byte sizes.",
         "default": {
           "default": "5",
           "options": "any number >= 0"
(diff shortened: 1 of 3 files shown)


packages/modelviewer.dev/data/docs.json
```diff
@@ -548,7 +548,7 @@
       {
         "name": "hotspot-*",
         "htmlName": "hotspot",
-        "description": "Any child element under <span class=\"attribute\">&lt;model-viewer&gt;</span> with a slot name starting with \"hotspot\" will be aligned with the 3D model using its <span class=\"attribute\">data-position</span> and <span class=\"attribute\">data-normal</span> attributes. See the <a href=\"../examples/annotations\">annotations example</a> for details."
```
elalish, line 551: Why did you take these spans away?

packages/modelviewer.dev/data/docs.json
```diff
@@ -146,19 +146,19 @@
...
         "htmlName": "dismissPoster",
-        "description": "Dismisses the poster, causing the model to load and render if necessary. This is currently effectively the same as interacting with the poster via user input."
+        "description": "Dismisses the <span class='attribute'>poster</span>, causing the model to load and render if necessary. This is currently effectively the same as interacting with the <span class='attribute'>poster</span> via user input."
       }
     ],
     "Static Methods": [
       {
         "name": "mapURLs(callback)",
         "htmlName": "mapURLs",
-        "description": "If provided, the callback will be passed each resource URL before a request is sent. The callback may return the original URL, or a new URL to override loading behavior. This behavior can be used to load assets from .ZIP files, drag-and-drop APIs, and Data URIs. This is a static method."
```
elalish, line 156: 👍

packages/modelviewer.dev/data/docs.json
```diff
@@ -114,7 +114,7 @@
       {
         "name": "modelIsVisible",
         "htmlName": "modelIsVisible",
-        "description": "This property is read-only. It returns true if the element is visible on the page (assuming there is an IntersectionObserver) and the poster has been dismissed. Related to the model-visibility event.",
```
elalish, line 117: 👍

elalish (approved):
Thanks for the editing!

packages/modelviewer.dev/data/docs.json
chrismgeorge: they aren't direct model-viewer attributes, I can add them back if you think it makes more sense to keep them?

elalish: yeah, probably so; they are attributes defined by model-viewer. And besides, that class name is a bit misleading; it's really anything we want to look like code.

chrismgeorge: yeah... but also, if everything looks like code, then the helpfulness also goes down and it can get confusing when viewing it, i.e. it would be nice if the code on the site signified something we created

elalish: Good point; I don't have a strong opinion on it. But these are at least attributes we created.

chrismgeorge: gotcha, and yeah maybe there should be quotation marks or a different style for other code that isn't ours -- just something to think about

PR #1631, opened 2020-10-13
fix expandable content

chrismgeorge:
fix expandable content:
* eases
* properties transition correctly
* content stays hidden when opening


elalish (approved):
I haven't actually pulled your branch, but I trust you. Maybe let's just do a bigger review when the new styling is ready. You're welcome to do bigger PRs on the CSS, since it's more UI review than code review.

chrismgeorge:
This was actually one of the bugs from the feedback we received from that review

PR #1636, opened 2020-10-14
fix testing for editor

chrismgeorge:
- added karma-spec-reporter for visualizing the test cases
- fixed broken model_viewer_preview_test
- commented out expandable_contents test for content (because of change to UI) - will update later if really needed

PR #1638, opened 2020-10-14
Fix space-opera tests

elalish:
Reverts #1631 
Reverts #1381 

There's still one error, but not in a specific test, so we need to figure that out too.


packages/space-opera/src/components/shared/expandable_content/expandable_section.css.ts
```diff
@@ -25,16 +25,13 @@ export const styles: CSSResult = css`:host {
...
   overflow: hidden;
   padding: 0 20px;
-  transition-duration: 300ms;
-  transition-timing-function: ease-out;
 }
 
 .SectionContent[open] {
   background-color: var(--me-theme-container-background);
-  overflow: hidden;
+  overflow: unset;
```
stevesan, line 32: These changes don't seem related. Are they in the original PR?

stevesan (approved):

packages/space-opera/src/components/shared/expandable_content/expandable_section.css.ts
elalish: No, that's from reverting chris' recent PR, which broke some tests because they weren't being run.

elalish:
Closing in favor of #1636

PR #1642, opened 2020-10-15
quick-fix for color picker

chrismgeorge:
quick-fix for color picker


elalish (approved):
Would be nice to have slightly more description for these UI changes; maybe a before and after screenshot? Or reference to an issue.

PR #1643, opened 2020-10-15
begin moving reducers into logically correct place

chrismgeorge:
Refactoring reducers to a file/feature
-no logical changes, just moving dispatch reducers for camera_settings into its own file
-will make this adjustment for all components, after this first one is merged


packages/space-opera/src/components/camera_settings/reducer.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+import {registerStateMutator, State} from '../../redux/space_opera_base.js';
+import {SphericalPositionDeg, Vector3D} from '../../redux/state_types.js';
+
+/*
+ * Register state mutators and get corresponding dispatchers.
+ */
+
+const SET_CAMERA_CONTROLS_ENABLED = 'SET_CAMERA_CONTROLS_ENABLED';
```
elalish, line 25: I feel like these lines aren't adding much value. I'd either put them back how they were, or export a union type so TS can check these for errors. You can take a look at `ARStatus` in ARRenderer.ts for an example.

packages/space-opera/src/components/camera_settings/reducer.ts
```diff
@@ -0,0 +1,72 @@
...
+export const dispatchCameraControlsEnabled = registerStateMutator(
+    SET_CAMERA_CONTROLS_ENABLED, (state: State, enabled?: boolean) => {
+      state.config = {...state.config, cameraControls: !!enabled};
+    });
+
+// Orbit
+const SAVE_CAMERA_ORBIT = 'SAVE_CAMERA_ORBIT';
+export const dispatchSaveCameraOrbit =
+    registerStateMutator(SAVE_CAMERA_ORBIT, (state: State) => {
```
elalish, line 34: I'm going to trust this a copy and paste from above, since github is not helping with the diff.

elalish (approved):
Not quite sure how this will move all the reducers to a single folder, when your new reducer.ts is still inside the camera_settings folder? Or are you going to move it again?

packages/space-opera/src/components/camera_settings/reducer.ts
chrismgeorge: this is me setting up for an actual reducer

packages/space-opera/src/components/camera_settings/reducer.ts
chrismgeorge: yep, sans the consts

packages/space-opera/src/components/camera_settings/reducer.ts
chrismgeorge: these are usually exported so they can be used in multiple places, at the moment, I am just moving everything into the respective location

chrismgeorge:
> Not quite sure how this will move all the reducers to a single folder, when your new reducer.ts is still inside the camera_settings folder? Or are you going to move it again?

Single reducer / feature. No other application has the reducers inside of the actual component, it just isn't good style. The alternate approaches are having a reducer/ folder with all the reducers or have a reducer/feature. The reducer/feature is the widely accepted best practice for redux at the moment.

packages/space-opera/src/components/camera_settings/reducer.ts
elalish: In that case, I'd recommend going with the `ARStatus` approach so you can also get type checking. But you can do that down the road in your refactor.

PR #1645, opened 2020-10-15
Pointer events

elalish:
Switching to pointerEvents to get touch-action: pan-y to work properly. 

UPDATE: Argh! I really prefer this method, but though iOS Safari claims to support pointerEvents and touch-action, it appears to be unusably buggy. Basically, with `touch-action: auto;`, you still get horizontal swipes coming through as pointer-events, but with `touch-action: none;` it works maybe 80% of the time, but the other 20% it will randomly get stuck back in `auto` mode, where pinch zooms the page and slight vertical motion starts scrolling. It seems to be non-deterministic. 

Looks like I'll have to do it myself after all with #1566, as touchEvents and preventDefault seems to be quite reliable. It's a shame not to use standard web features though.


elalish:
closing in favor of #3404.

PR #1648, opened 2020-10-16
move states into corresponding module

chrismgeorge:
moved lighting, camera, and material state into corresponding module


elalish (approved):
LGTM

PR #1650, opened 2020-10-16
move testing and update base reducer for readability

chrismgeorge:
- moved all testing files into a separate folder within src/test/
- TODO: Update base reducer to actually act as a reducer


packages/space-opera/src/components/model_viewer_snippet/reducer.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ */
+
+import {ModelViewerConfig} from '@google/model-viewer-editing-adapter/lib/main.js';
+
+import {registerStateMutator} from '../../space_opera_base.js';
+import {State} from '../../space_opera_base.js';
+import {INITIAL_CAMERA} from '../camera_settings/camera_state.js';
+
+// TODO: BUGGY! This is mutating the state!
```
elalish, line 24: Was this here before or did you add it? Also, isn't a state mutator supposed to mutate the state?

packages/space-opera/src/components/reset/reducer.ts
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import {reduxStore} from '../../space_opera_base.js';
+
+/** Mostly for unit tests. */
+export const RESET_STATE_ACTION_TYPE = 'RESET_SPACE_OPERA_STATE';
+export function dispatchResetState() {
```
elalish, line 22: Does this really deserve its own file? Seems more like a helper.

packages/space-opera/src/reducers.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// TODO: Refactor reducers to follow best practices, then combine them all here.
```
elalish, line 18: Oh, perhaps this addresses my comments above? Will a bunch of those short files end up combined here instead?

elalish (approved):

packages/space-opera/src/components/model_viewer_snippet/reducer.ts
chrismgeorge: I added this. It is supposed to mutate the state, but it is supposed to mutate a copy of the original state. So a good number of the reducers are in fact "broken" by redux standards.

packages/space-opera/src/components/reset/reducer.ts
chrismgeorge: for consistencies sake, I gave it it's own folder but will probably move it into shared at some point

packages/space-opera/src/reducers.ts
chrismgeorge: This file is for importing and combining the reducers into one using the combineReducers method from redux. In general, the top level isn't supposed to contain any reducers.

PR #1658, opened 2020-10-21
Head ex

chrismgeorge:
removed header from examples on desktop, add extra scrolling to entries and subheaders of docs

### DIFF
--- a/packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
+++ b/packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
@@ -337,9 +337,6 @@ export function sidebarClick() {
     const root = document.documentElement;
     root.style.setProperty('--sidebar-width', '0px');
   }
-  // if (extraScroll) {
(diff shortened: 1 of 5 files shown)


packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
```diff
@@ -337,6 +337,9 @@ export function sidebarClick() {
     const root = document.documentElement;
     root.style.setProperty('--sidebar-width', '0px');
   }
+  // if (extraScroll) {
+  //   window.scrollTo(window.scrollX, window.scrollY - 100);
+  // }
```
elalish, line 342: Do you want this?

elalish (approved):

packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
chrismgeorge: nope, forgot to push that removal!

elalish (approved):

PR #1669, opened 2020-10-27
force examples to push to new examples

chrismgeorge:
replace examples with redirect links


elalish (approved):
Thanks! Let's give it a shot.

PR #1670, opened 2020-10-27
replace landing page

chrismgeorge:
don't know if you were planning on doing this, just went ahead and made this if you wanted it

I deleted the original index.html, and replaced it with what was previously home.html


elalish (approved):
Ah yes, completely missed this wasn't in the other PR. Thanks!

PR #1692, opened 2020-11-04
Refactor scene-graph

elalish:
This removes the 3dom package and pulls in a minimal portion of it to model-viewer in order to keep the scene-graph API working. There are some minor breaking changes in this update, but I'm not too concerned since we labeled scene-graph as unstable. The biggest one is probably that elements without names now return an empty string instead of null. Also, all the methods are now sync instead of async, except `setURI()`, which needs to download a texture. 

This removes a lot of code and will ease both maintenance and feature additions. Feedback welcome. The vast majority of Github's idea of lines removed is just package-lock garbage; it actually only removed about 4,000 to 5,000 LOC.


chrismgeorge:
> This removes the 3dom package and pulls in a minimal portion of it to model-viewer in order to keep the scene-graph API working. There are some minor breaking changes in this update, but I'm not too concerned since we labeled scene-graph as unstable. The biggest one is probably that elements without names now return an empty string instead of null. Also, all the methods are now sync instead of async, except `setURI()`, which needs to download a texture.
> 
> This removes a lot of code and will ease both maintenance and feature additions. Feedback welcome. The vast majority of Github's idea of lines removed is just package-lock garbage; it actually only removed about 4,000 to 5,000 LOC.

What kind of effects does this/ could this have on the editor?

elalish:
Hopefully none; I'll give it a try to see if anything seems broken. This code review is largely FYI for you. You may want to take a look at the changes to api.ts, since that's your interface.

PR #1714, opened 2020-11-19
Vertical ar

### LINKED ISSUE #687: Ability to place object on vertical plane in AR
### Description

Hi, i'm trying to use model viewer with ar, however i would like to display pictures, so i would need to match vertical surfaces. Is there a way to achieve that?

so far i've seen this code:
```

// NOTE: Currently rays will be cast from the middle of the screen.
// Eventually we might use input coordinates for this.
    this.raycaster.setFromCamera({x: 0, y: 0}, this.camera);
    const ray = this.raycaster.ray;
    originArray.set(ray.origin.toArray());
    directionArray.set(ray.direction.toArray());

```

However i can't find a way to actually switch plane detection.

### Browser Affected

All

### OS

- [x] Android
- [x] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### Versions

latest

elalish:
Fixes #687 

I'm taking the simple approach for now, outlined helpfully in this comment: https://github.com/google/model-viewer/issues/687#issuecomment-683315689, where we now have an attribute `ar-placement` which can be either `"wall"` or the default `"floor"`. This also configures SceneViewer to use vertical placement. QuickLook is not configurable on the intent side as far as I'm aware.

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -318,6 +318,9 @@ configuration or device capabilities');
       if (this.arScale === 'fixed') {
         intentParams += `&resizable=false`;
       }
+      if (this.arPlacement === 'wall') {
+        intentParams += `&enable_vertical_placement=true`;
+      }
 
       const intent = `intent://arvr.google.com/scene-viewer/1.0${
           intentParams}#Intent;scheme=https;package=com.google.ar.core;action=android.intent.action.VIEW;S.browser_fallback_url=${
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -83,6 +83,9 @@ export const ARMixin = <T extends Constructor<ModelViewerElementBase>>(
 
     @property({type: String, attribute: 'ar-scale'}) arScale: string = 'auto';
 
+    @property({type: String, attribute: 'ar-placement'})
+    arPlacement: string = 'floor';
+
     @property({type: String, attribute: 'ar-modes'})
     arModes: string = DEFAULT_AR_MODES;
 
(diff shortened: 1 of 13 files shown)


packages/model-viewer/src/features/ar.ts
```diff
@@ -308,6 +318,9 @@ configuration or device capabilities');
       if (this.arScale === 'fixed') {
         intentParams += `&resizable=false`;
       }
+      if (this.arPlacement === 'wall') {
+        intentParams += `&enable_vertical_placement=true`;
```
aviggiano, line 322: @elalish sorry to ask, but where did you find the documentation about this intent parameter? We couldn't find it anywhere on the ARCore SDK docs

elalish: I have an inside contact 😁. I believe they're working on updating their docs.

p0wl:
awesome!!! Thanks for your work! Will test this hopefully next week!

Aubi0ne:
Hello ! I tried to test the new wall ar-placement feature but i see no difference between the floor and the wall mode. Is it normal ? Or maybe i am doing something wrong... ^^"

elalish:
@Aubi0ne How are you testing it? If you're using a release version, this hasn't been released yet, so that's the most likely problem. It should work on our dev pages, and keep in mind it only applies to WebXR and SceneViewer, so you won't notice a difference on iOS.

Aubi0ne:
Oh ok i thought you would publish after the merge, i misunderstood sorry. And so if there is no difference on iOS, do you know a way to improve wall detection on iOS ?

elalish:
I don't know much about quick-look, but you can check their docs. You may be able to specify something in the usdz or .reality file.

packages/model-viewer/src/features/ar.ts
FelipeR2U: @elalish , do you know if there is any other option related to vertical placement? Something like "vertical_only" 😅

vartist:
Hi everyone! I have tried the **wall** placing mode but it’s not working. (iPhone). Even the original model does not work. (https://kaboomlaser.com/products/test)
Reading through some posts I thought it’s fixed ? Or am I doing something wrong?

aviggiano:
> Hi everyone! I have tried the **wall** placing mode but it’s not working. (iPhone). Even the original model does not work. (https://kaboomlaser.com/products/test)
> 
> Reading through some posts I thought it’s fixed ? Or am I doing something wrong?

You have a problem with your USDZ model. How did you export/convert it?

PR #1717, opened 2020-11-22
Update ZIP

chrismgeorge:
## Changes to user experience:
* When a user adds a model to the editor, the original name of the file will be displayed in the snippet. And the relative file path of the model will be included in the snippet on export. 
* Similarly, the environment image (skybox, as well, if checked) has the same effect. I.e. relative path will be displayed. 
* Same for poster, except the exported poster name is always, "poster.png"

### Side effects
* Some side effects of this are that the way we handle the import snippet has changed slightly. Importantly, the environment image is retained if the **relative name** in the import snippet isn't changed.

### DIFF
--- a/packages/space-opera/src/components/ibl_selector/ibl_selector.ts
+++ b/packages/space-opera/src/components/ibl_selector/ibl_selector.ts
@@ -87,6 +87,7 @@ export class IblSelector extends ConnectedLitElement {
             this.config.environmentImage) {
       reduxStore.dispatch(dispatchEnvrionmentImage(
           dropdownElement.selectedItem.getAttribute('value') || undefined));
+      // dropdown value equals null when "Default" is selected
       if (dropdownElement.selectedItem.getAttribute('value') === null) {
         reduxStore.dispatch(dispatchSetEnvironmentName(undefined));
       } else {
--- a/packages/space-opera/src/components/ibl_selector/ibl_selector.ts
+++ b/packages/space-opera/src/components/ibl_selector/ibl_selector.ts
@@ -33,6 +33,7 @@ import {State} from '../../types.js';
 import {dispatchEnvrionmentImage, dispatchExposure, dispatchShadowIntensity, dispatchShadowSoftness, dispatchUseEnvAsSkybox, getConfig} from '../config/reducer.js';
 import {ConnectedLitElement} from '../connected_lit_element/connected_lit_element.js';
 import {FileModalElement} from '../file_modal/file_modal.js';
+import {dispatchSetEnvironmentName} from '../relative_file_paths/reducer.js';
 import {CheckboxElement} from '../shared/checkbox/checkbox.js';
 import {Dropdown} from '../shared/dropdown/dropdown.js';
 import {SliderWithInputElement} from '../shared/slider_with_input/slider_with_input.js';
@@ -86,6 +87,14 @@ export class IblSelector extends ConnectedLitElement {
             this.config.environmentImage) {
       reduxStore.dispatch(dispatchEnvrionmentImage(
           dropdownElement.selectedItem.getAttribute('value') || undefined));
+      if (dropdownElement.selectedItem.getAttribute('value') === null) {
+        reduxStore.dispatch(dispatchSetEnvironmentName(undefined));
+      } else {
+        const envImageList =
+            dropdownElement.selectedItem.getAttribute('value')?.split('/');
+        const envImageName = envImageList![envImageList!.length - 1];
+        reduxStore.dispatch(dispatchSetEnvironmentName(envImageName));
+      }
     }
   }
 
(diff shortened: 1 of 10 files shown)


chrismgeorge:
newstate

The ModelViewerSnippet state was updated to include a section for relativeFilePaths, these are used when displaying the snippet and when exporting the snippet such that the paths for the src, poster, and environment image are relative to the snippet output.

chrismgeorge:

packages/space-opera/src/components/ibl_selector/ibl_selector.ts
```diff
@@ -86,6 +87,15 @@ export class IblSelector extends ConnectedLitElement {
             this.config.environmentImage) {
       reduxStore.dispatch(dispatchEnvrionmentImage(
           dropdownElement.selectedItem.getAttribute('value') || undefined));
+      // dropdown value equals null when "Default" is selected
+      if (dropdownElement.selectedItem.getAttribute('value') === null) {
+        reduxStore.dispatch(dispatchSetEnvironmentName(undefined));
```
elalish, line 92: Not sure where this error is, but everything works great when you choose one of our supplied models and environments. However, if I load my own, the zip no longer contains a GLB at all, and the HDR comes down as "undefined", though it is the correct file.

packages/space-opera/src/components/ibl_selector/ibl_selector.ts
```diff
@@ -86,6 +87,15 @@ export class IblSelector extends ConnectedLitElement {
...
+      } else {
+        const envImageList =
+            dropdownElement.selectedItem.getAttribute('value')?.split('/');
```
elalish, line 95: Also, possibly unrelated, but when I select BoomBox, and then load my own model, it gets loaded with the BoomBox textures and appears to just remove the incoming textures completely.

elalish (changes requested):
PR looks good, but I found some bugs.

packages/space-opera/src/components/ibl_selector/ibl_selector.ts
chrismgeorge: fixed

packages/space-opera/src/components/ibl_selector/ibl_selector.ts
chrismgeorge: fixed

packages/space-opera/src/components/model_viewer_snippet/components/open_button.ts
```diff
@@ -197,15 +197,13 @@ export class ImportCard extends LitElement {
   @internalProperty() selectedDefaultOption: number = 0;
 
   async onUploadGLB() {
-    const files = await this.fileModal.open();
+    const files: any = await this.fileModal.open();
```
elalish, line 200: What actual TS errors are you getting around with `any`? That's a pretty blunt tool, so it's better if we can be more delicate and retain more of it's type checking power.

packages/space-opera/src/components/model_viewer_snippet/components/open_button.ts
```diff
@@ -222,13 +220,13 @@ export class ImportCard extends LitElement {
     const dropdown = event.target as Dropdown;
     const value = dropdown.selectedItem?.getAttribute('value') || undefined;
     let snippet = '';
-    const simpleMap = {
+    const simpleMap: any = {
```
elalish, line 223: In this modern era it's better to use `?? undefined` than `|| undefined`, or whatever other default value you want. Might want to do a quick code search to see where else `||` crops up.

packages/space-opera/src/components/model_viewer_snippet/components/open_button.ts
```diff
@@ -244,14 +242,12 @@ export class ImportCard extends LitElement {
         this.selectedDefaultOption = 0;
         return;
       } else if (value in simpleMap) {
```
elalish, line 244: Your maps are key:value pairs, but your keys are called values and I'm not sure what the values (numbers) are used for. Any chance this can be simplified? I haven't read it thoroughly enough to be sure, so feel free to correct me.

elalish (approved):

packages/space-opera/src/components/model_viewer_snippet/components/open_button.ts
chrismgeorge: It just isn't recognizing the key "name", i,.e. file[0] is an object that has a key, so that should probably be cast to the correct type, but I'm not sure what that is

packages/space-opera/src/components/model_viewer_snippet/components/open_button.ts
chrismgeorge: the dropdown needs an index of what is selected, this was the most straightforward way of doing it. I can look into other options moving forward. I'll change the "value" variable to "key", you're right, that is confusing.

elalish (approved):
Zip export is look great now, thanks!

PR #1721, opened 2020-11-23
Adjust byteLength after deleting texture

mqg734:
Fixed issue with packGlb creating an invalid GLB after deleting unused texture(s). The byteLength in the header still has the length of the buffer prior to deleting the texture(s). The buffer views were being adjusted, but the corresponding buffer's byteLength wasn't.


elalish (approved):
Thanks!

PR #1723, opened 2020-11-23
delay load event until jumpCameraToGoal() can execute

elalish:
@chrismgeorge Can you check if this fixes your Glitch?


chrismgeorge:
> @chrismgeorge Can you check if this fixes your Glitch?

Will do,

PR #1735, opened 2020-11-30
added setEmissiveFactor to scene-graph API

elalish:
This was one of the gaps between scene-graph and editing-adapter and it was blocking some new UI in the editor, so it got higher priority.


chrismgeorge:
looking good

PR #1738, opened 2020-12-01
update EnvironmentScene.ts for even lighting that is color balanced w…

vanlicht:
When using the Model viewer with default exposure of 1.0, the 3D assets color visualization look bit too hot in generate. I prepared a color chart glb with Khronos unlit extension color chart as the baseline, with the medium gray (122,122,122), in the Model viewer it reads as (132,132,132) due to tone mapping I assume. 

I then use the same color chart glb with regular PBR as the candidate to tweak the lighting environment to match the target model color, to achieve the same effect.

modeViewerCompare_20201130


vanlicht:
will resubmit

PR #1741, opened 2020-12-02
Rotate environment example

### LINKED ISSUE #1727: Reset on rotating environment image
I havent got any reply on pull request https://github.com/google/model-viewer/pull/1570 for many days now so I thought ill create an issue if anyone can answer. 

Rotating environment-image can now be achieved in the live demo link by pressing shift.

First press shift and then mouse drag move along horizontal/X axis to move the background. The only problem now is that every time I let go of environment rotating control and let's say I simply rotate the model and leave it at a certain orientation. And then I try to rotate the environment image again, it will reset the model.

This line causes the reset, and if I don't use this line then obviously the model does not stay in its own place while rotating the environment map.

modelViewer.resetTurntableRotation(orbit.theta);

#### Live Demo
https://thecela.com/playground/configurator/experiment/rotateEnvironmentMap.html

#####Version
latest cloud based version

elalish:
Fixes #1727

This is a new example for rotating the skybox instead of the camera. Thanks to @dhruvadhia-cela as this builds off of his #1570. I also added the missing `toString()` method to `getCameraOrbit` to further simplify this.

### DIFF
--- a/packages/model-viewer/src/features/controls.ts
+++ b/packages/model-viewer/src/features/controls.ts
@@ -73,6 +73,7 @@ export interface SphericalPosition {
   theta: number;  // equator angle around the y (up) axis.
   phi: number;    // polar angle from the y (up) axis.
   radius: number;
+  toString(): string;
 }
 
 export type InteractionPromptStrategy = 'auto'|'when-focused'|'none';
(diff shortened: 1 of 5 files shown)


dhruvadhia1:
Awesome! tested and it works perfect. The reset is no more an issue.

dhruvadhia-cela (dismissed):
Awesome! tested and it works perfect. The reset is no more an issue.

PR #1761, opened 2020-12-10
change EnvironmentScene.ts

thomastgt:
When using the Model viewer with default exposure of 1.0, the 3D assets color visualization looks a bit too hot with over exposure. I prepared a color chart glb with Khronos unlit extension color chart as the baseline, with the medium gray (122,122,122), in the Model viewer it reads as (132,132,132) due to tone mapping I assume.
I then use the same color chart glb with regular PBR as the candidate to tweak the lighting environment to match the target model color, to achieve the same effect.
modeViewerCompare_20201130


packages/model-viewer/src/three-components/EnvironmentScene.ts
```diff
@@ -29,12 +29,12 @@ export default class EnvironmentScene extends Scene {
...
     const mainLight = new PointLight(0xffffff, 500.0, 28, 2);
-    mainLight.position.set(0.418, 16.199, 0.300);
+    mainLight.position.set(0.5, 18.0, 0.5);
     this.add(mainLight);
 
     const room = new Mesh(geometry, roomMaterial);
-    room.position.set(-0.757, 13.219, 0.717);
-    room.scale.set(31.713, 28.305, 28.591);
+    room.position.set(0.0, 13.0, 0.0);
+    room.scale.set(30.0, 30.0, 30.0);
```
elalish, line 37: 👍

elalish (changes requested):
Thanks, this is really helpful!

packages/model-viewer/src/three-components/EnvironmentScene.ts
```diff
@@ -74,40 +74,34 @@ export default class EnvironmentScene extends Scene {
...
-    light5.position.set(3.235, 11.486, -12.541);
-    light5.scale.set(2.5, 2.0, 0.1);
+    const light5 = new Mesh(geometry, this.createAreaLightMaterial(9.2));
+    light5.position.set(0, 10.0, -14.0);
+    light5.scale.set(5.0, 10.0, 0.1);
     this.add(light5);
 
     // +y
-    const light6 = new Mesh(geometry, this.createAreaLightMaterial(100));
+    const light6 = new Mesh(geometry, this.createAreaLightMaterial(4));
```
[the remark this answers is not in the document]
thomastgt: I pushed a new one, now the all the four sides + top all matches the current model viewer's front RGB value, when using the PBR color chart model. 

I can crank down the exposure value down to 0.46 to match closely of the RGB color values of the chart from the PBR mode one the new environment, to unlit_extension color chart asset displayed on the current Model Viewer.

modeViewerCompare_20201211

modeViewerCompare_20201211b

packages/model-viewer/src/three-components/EnvironmentScene.ts
```diff
@@ -29,12 +29,12 @@ export default class EnvironmentScene extends Scene {
...
     this.add(mainLight);
 
     const room = new Mesh(geometry, roomMaterial);
-    room.position.set(-0.757, 13.219, 0.717);
-    room.scale.set(31.713, 28.305, 28.591);
+    room.position.set(0.0, 13.0, 0.0);
+    room.scale.set(30.0, 30.0, 30.0);
     this.add(room);
 
     const box1 = new Mesh(geometry, boxMaterial);
```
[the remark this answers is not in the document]
thomastgt: New commit has the boxes position (box 1 and 3) adjusted.

packages/model-viewer/src/three-components/EnvironmentScene.ts
```diff
@@ -29,16 +29,16 @@ export default class EnvironmentScene extends Scene {
     const boxMaterial = new MeshStandardMaterial({metalness: 0});
 
     const mainLight = new PointLight(0xffffff, 500.0, 28, 2);
-    mainLight.position.set(0.418, 16.199, 0.300);
+    mainLight.position.set(0.5, 18.0, 0.5);
```
elalish, line 32: I'm trying to figure out why the floor looks darker in these reflections than it did. My only guess is it's because the main light moved up by 2? Can you take a look at these height dimensions (of the room as well) and see if we can get the reflection to match the original a bit closer? None of this should have any noticeable effect on the lighting of your model.

elalish (changes requested):
Thanks, this is looking really good, just one further change requested.

packages/model-viewer/src/three-components/EnvironmentScene.ts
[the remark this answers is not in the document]
elalish: Thank you, these comparison images are really helpful! I'll make sure to put that 0.46 factor into our docs

packages/model-viewer/src/three-components/EnvironmentScene.ts
thomastgt: Hi, by adjust the room scale and tweak the point light height, I brought the reflection image of the room closer to the current version. Please take a look.

elalish: Thanks! I don't suppose I could trouble you to post updated renders here? The ones you did before were very helpful.

thomastgt: Definitely. Here's the comparison. Left is the original, Right is the new environment.

modeViewerCompare_20201215

elalish: Thanks, that looks great! I'm going to hold off on merging this for a bit as there are some other's who want to take a look at it. I'll make sure it gets in for the next release, which is looking early January now.

elalish:
I'm sorry for waffling, but after looking at this more closely I'm a bit concerned about the amount of difference (breaking change) for our existing users. Consider the before/after of our homepage: 
imageimage

Not only is the result now effectively brighter (I think the front face is probably close, but everything else has brightened a lot), but there's also much less contrast to give the 3D shape a sense of depth. Unfortunately I have the feeling that this is directly at odds with the concept of even lighting, so even with more work there may always be a tradeoff. 

I'm thinking of avoiding the breaking change by adding an attribute like `default-lighting` with options "legacy" and "even", defaulting to the old lighting. And maybe drop the strength on "even" down to where you get RGB parity. How. does that sound? I'm working on a release for next week, so I'd like to close on this quickly.

thomastgt:
I like the idea of option to switch lightings. Say from the even lighting it can be for analytical purpose. Therefore it could be named as analytical-lighting or even-lighting.

As discussed, with the even-lighting it is aimed to match the front light of the current default one, and when lower the expected sure to 0.46 seems to match the white balance, which is true for the currently default lighting except the dark would look too dark...which may be for future improvement if the we / users continue to see the need. Happy to have the new lighting included for content creation purpose.

Thomas

Q
Sent from Workspace ONE Boxer

On Jan 12, 2021 7:06 PM, Emmett Lalish <notifications@github.com> wrote:

I'm sorry for waffling, but after looking at this more closely I'm a bit concerned about the amount of difference (breaking change) for our existing users. Consider the before/after of our homepage:
[image]<https://user-images.githubusercontent.com/1649964/104392527-17547b80-54f7-11eb-988c-5a10743cd2c8.png>[image]<https://user-images.githubusercontent.com/1649964/104392513-0efc4080-54f7-11eb-970c-835a881e4027.png>

Not only is the result now effectively brighter (I think the front face is probably close, but everything else has brightened a lot), but there's also much less contrast to give the 3D shape a sense of depth. Unfortunately I have the feeling that this is directly at odds with the concept of even lighting, so even with more work there may always be a tradeoff.

I'm thinking of avoiding the breaking change by adding an attribute like default-lighting with options "legacy" and "even", defaulting to the old lighting. And maybe drop the strength on "even" down to where you get RGB parity. How. does that sound? I'm working on a release for next week, so I'd like to close on this quickly.

mikkoh:
Hey @elalish I'm going to get @bldelgado to take a look at this

bldelgado:
This same issue came up when we were initially looking at the lighting. I think _dropping the strength to where you have RGB parity_ might be the best move forward, or at least closer. From our side, merchants had used that darker shadowing to help showcase the metal details of their products. But that came at the cost of the colours being slightly off. Fixing that colour issue would at least help to solve one of the problems while we can look at the potential of adding other ways to showcase different materials.

mikkoh:
I rendered a few mp4s of old vs new. Click to view videos:

Old "hot lighting":
https://user-images.githubusercontent.com/496903/104600866-8e008e80-5647-11eb-8ded-7c104a6e4d02.mp4

New "neutral lighting":
https://user-images.githubusercontent.com/496903/104601156-e2a40980-5647-11eb-930a-927311f071c4.mp4

Old "hot lighting":
https://user-images.githubusercontent.com/496903/104601291-0d8e5d80-5648-11eb-82da-d8561cc4a4ce.mp4

New "neutral lighting":
https://user-images.githubusercontent.com/496903/104601436-3f072900-5648-11eb-86b6-51e712697c2a.mp4

thomastgt:
Thanks for rendering out the video comparison test!
Some observation and thought about the "old-hotter-lighting" and "new neutral lighting":
1. It depends on what viewing angles we are looking at and what's the purpose
2. For the original Cube test I posted, the goal is to create even lighting from all 4 sides + top, matching the value from the front side of the  "old-hotter-lighting"
3. For astronaut: it interprets good description of multi-faceted angle, instead of cube mesh... from the front view astronaut looks overall slightly darker I assume it's because the angle is not 90 degrees but a combination of varied degrees.
3. For the Mixer example, it looks drastically different at first, but really the fair comparison is the front side of the asset, which is the thin part of the model, where the "V" mark is. If particularly looking at that angle (limited viewing area however), that the color matches, and not surprising having more highlighted spectacular on the new lighting environment.

My thought is, if the new lighting setup is technically sounding and has value for analytical color checking purpose to check the assets from all angles, that's where I see it as beneficial, by adding the new neutral lighting to the repertoire; it is not aiming for replacing the current "old-hotter-lighting" if the merchants already are used to this aesthetics. Side note: a separate topic of improving the current "old-hotter-lighting" as the darker side is too dark...left for debate when it comes to aesthetics however.

bldelgado:
**Neutral Lit**
image
**Older Lighting**
image

@thomastgt As a test for comparing colour I think we may be overlooking some key factors. When you ask the question of "What is the purpose?" I think more often than not we're talking about 3D models that are meant to be digital representations of their real life counter part. In the two images I posted above I would assume the materials for the mixer to be different. 

For the Neutral Lit Mixer I would assume the material is more of a plastic type material. The fact that there's less depth in the colours would make me infer that the material is more akin to a PVC pipe or something. When looking at the grey metal pieces, I would assume that they are more hard plastic in versus brushed stainless steel. 

In the case of the Older lighting, I think we fall prey to a problem in the intensity of lighting. The dark parts are too dark and the hot spots are too strong.

In both cases the issue is having lighting feel more _accurate_ to the materials. It's the combination of the material and the colour that help make it feel real.

elalish:
Okay, I've updated this PR to be non-breaking; you can now get the @thomastgt's new lighting by passing "neutral" to `environment-image`. I've reverted to baking the 0.46 exposure into the lights, since we no longer need to match anything. I like this quite a lot, and note the new example where you can compare the default to this new neutral. It would be great if someone could verify that my changes still have the RGB color swatches rendering as desired.

PR #1767, opened 2020-12-16
Fix delete texture bugs

mqg734:
Fixed two delete texture bugs that resulted in an invalid GLB file:
1. The byteOffset is optional for a BufferView. We weren't handling the case where a byteOffset was missing, resulting in invalid GLBs.
2. Deleting a texture file of an odd file size results in an invalid GLB. This happened because the place holder image we replace deleted textures with is 68 bytes, resulting in odd byteOffsets when we adjust the byteOffset for subsequent bufferViews. Accessors have specific data alignment requirements for the byteOffset.


elalish (approved):
Thanks!

PR #1784, opened 2020-12-29
Disable scene-viewer occlusion

### LINKED ISSUE #1733: How to disable blending with scene viewer?
By default, object blending is enabled in scene-viewer. Is there a way we can have it disabled as default?

elalish:
Fixes #1733 

Occlusion in scene-viewer is not a great experience with the current sensors, so we're disabling it by default now. It can be enabled by putting a query param on the `src` URL: `https://somewhere.org/model.glb?disable_occlusion=false`. You can also set any other scene-viewer query params in this way, like title, link, etc. This PR has also removed the default values for title and link that used to be set from alt and the source website, as that is rarely what a user would want anyway. Alt text is not a title, and the link was doing the same thing as the exit button, except sucking you into a WebView instead of taking you back to your browser.

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -281,11 +281,8 @@ configuration or device capabilities');
       // modelUrl can contain title/link/sound etc.
       // These are already URL-encoded, so we shouldn't do that again here.
       let intentParams = `?file=${modelUrl.toString()}&mode=ar_only`;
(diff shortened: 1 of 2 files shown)


dungnguyen234:
how can i add this value "disable_occlusion=false" in html , because i want to show my customer blending effect

elalish:
Did you try the directions above? If it doesn't work, please feel free to start a discussion with a link to your page that's not working so we can find the error.

PR #1792, opened 2020-12-30
Add orientation and scale attributes

### LINKED ISSUE #426: User-configurable transform for model when placed in AR
We scale and translate a model in order to frame it properly in inline scenarios. However, when the user goes to place the model in WebXR-based AR, the natural scale is used. And, the translation may not look right depending on geometry of the model in question.

This issue proposes that we enable authors to configure the relative scale, translation and rotation of the model for the purposes of presenting in AR.

### LINKED ISSUE #1118: Ability to position model external to glb/gltf content
It would be helpful to be able to translate and rotate content external to the model file so that it loads in a desired position. Perhaps through additional tag attributes.

A good use case for this is when loading content from a 3rd party that you don't have edit control over but want to load in a specific orientation.

### LINKED ISSUE #1724: Rotate model on X axis
### Description

Is there an option to rotate the model on an X axis? Here's an example:
image
If not, it'd be nice to have this feature implemented.

#### Live Demo
I've found the example model viewer here -> https://www.creators3d.com/online-viewer

elalish:
Fixes #426
Fixes #1118 
Fixes #1724 

This adds an orientation attribute specified as Euler angles and a 3-element scale attribute.

### DIFF
--- a/packages/model-viewer/src/features/controls.ts
+++ b/packages/model-viewer/src/features/controls.ts
@@ -80,6 +80,7 @@ export type InteractionPromptStrategy = 'auto'|'when-focused'|'none';
 export type InteractionPromptStyle = 'basic'|'wiggle';
 export type InteractionPolicy = 'always-allow'|'allow-when-focused';
 export type TouchAction = 'pan-y'|'pan-x'|'none';
+export type Bounds = 'tight'|'legacy';
 
 export const InteractionPromptStrategy:
     {[index: string]: InteractionPromptStrategy} = {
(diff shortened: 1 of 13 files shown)


vortice3D:
Hi there, @elalish:

Any news about including the position attributes (X, Y, Z), besides the already implemented on orientation and scale?

Best regards.

elalish:
That's what `camera-target` is for.

vortice3D:
Good morning, @elalish:

I've tried that way (camera-target), but sadly, talking about the Y-axis (up) it seems not work, at least when passing to the AR experience. I mean that only the other two axes are updated, as they are the ones relatives to the floor plane.

This way, is my only option to have a different model for each of the seven (7) elevations over the floor that the user can choose for the product?

The solution of having a specific animation for the elevation, in order to choose from that the specific elevation is not applicable in my case, as I need to have my model animated and model-viewer doesn't support additive animations at the moment.

The other way I think I could try here is to gain access to the internal ThreeJS scenegraph, in order to manipulate the product model directly from there. Any advice on this?

Thanks for your time. Really, appreciate it.

elalish:
That's outside our public API, so you're welcome to hack, but you're on your own :) If you want proper support, I'd recommend helping with #4442.

PR #1798, opened 2021-01-06
Change broken link in docs.json

devbridie:
The current link, https://github.com/google/model-viewer#supported-formats, refers to a non-existent anchor. I think this is supposed to refer to GLTFLoader, but I can't find any details on supported formats aside from the GLTF extensions list.

### DIFF
--- a/packages/modelviewer.dev/data/docs.json
+++ b/packages/modelviewer.dev/data/docs.json
@@ -5,7 +5,7 @@
       {
         "name": "src",
         "htmlName": "src",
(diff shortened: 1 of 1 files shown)


elalish:
Thanks for noticing, I have no idea where that link was supposed to point to either. Honestly, considering we already have links to the formats we support, I think we could just leave off that sentence fragment entirely. What do you think?

devbridie:
I've changed the commit to remove the link entirely.

elalish (approved):
Thanks!

PR #1804, opened 2021-01-08
Ios ar

chrismgeorge:
* Added try catch statement to keep recursive loops running.
* Fixed iOS POSTing to send a USDZ url, and have an iOS device post the url when the ar button is clicked.


chrismgeorge:
New management of multiple mobile sessions with the editor:

Screen Shot 2021-01-11 at 10 16 41 AM

chrismgeorge:

packages/space-opera/src/components/mobile_view/mobile_view.ts
```diff
@@ -146,15 +149,19 @@ export class MobileView extends LitElement {
...
+      }, 7000);
       await this.waitForData(json);
     } else {
       console.error('Error:', response);
     }
   }
 
   async triggerFetchLoop() {
-    await this.fetchLoop();
+    try {
```
chrismgeorge, line 160: fixes issue where the fetch loop timing out would crash the loop causing the mobile device to close its "port"

packages/space-opera/src/components/mobile_view/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+export function envToSession(
+    pipeId: number|string, sessionID: number, envIsHdr: boolean): string {
+  const addOn = envIsHdr ? '#.hdr' : '';
+  return `${DOMAIN}${pipeId}-${sessionID}-env${addOn}`;
+}
+
+// ex: 'https://piping.nwtgck.repl.co/123-456-789'
+export function usdzToSession(
+    pipeId: number|string, sessionID: number, modelId: number): string {
+  return `${DOMAIN}${pipeId}-${sessionID}-${modelId}.usdz`;
```
chrismgeorge, line 53: fixes iOS by adding usdz file type to url

packages/space-opera/src/components/mobile_view/open_mobile_view.ts
```diff
@@ -195,19 +199,38 @@ export class OpenMobileView extends ConnectedLitElement {
...
-  // with knowing when isSendingData should be set to false.
   async postInfo() {
+    if (this.isSendingData) {
+      return;
+    }
     this.isSendingData = true;
+    const sessionList = [...this.sessionList];
+    setTimeout(() => {
+      this.isSendingData = false;
+    }, 20000);
```
elalish, line 233: Might be nice to replace times like these with `ALL_CAPS_NAMES` that are defined at the top of the file or imported from a parameters file, so that it's easy to know what they are and change them if necessary.

packages/space-opera/src/components/mobile_view/utils.ts
elalish: Have we checked this works with a .reality file too?

packages/space-opera/src/components/mobile_view/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+/**
+ * Determine the mobile operating system.
+ * This function returns one of 'iOS', 'Android', 'Windows Phone', or
+ * 'unknown'.
+ * https://stackoverflow.com/questions/21741841/detecting-ios-android-operating-system
+ */
+export function getMobileOperatingSystem(): string {
+  // @ts-ignore
```
elalish, line 88: let's get rid of all the @ts-ignore. You can always do `(window as any).opera` if you need to.

elalish (approved):
Nice, this looks ready for testing.

PR #1840, opened 2021-01-19
Unflag mobile section

chrismgeorge:
Removes id in URL to display mobile section


packages/space-opera/src/components/mobile_view/open_mobile_view.ts
```diff
@@ -360,8 +360,9 @@ export class OpenMobileView extends ConnectedLitElement {
   // Initialize AR values and start deploy loop
   async onInitialDeploy() {
     this.isDeployed = true;
-    reduxStore.dispatch(dispatchAr(true));
```
chrismgeorge, line 363: Don't need these here (they were just here when these checkboxes weren't initially exposed)

elalish (approved):
Ship it!

PR #1846, opened 2021-01-20
Add modelviewer link to editor

chrismgeorge:
* Per title


chrismgeorge:
Screen Shot 2021-01-20 at 10 53 41 AM

chrismgeorge:

elalish:
Would it be hard to move it above the tabs? Then you could also make it black text on white, which would help separate it from the controls.

chrismgeorge:
> Would it be hard to move it above the tabs? Then you could also make it black text on white, which would help separate it from the controls.

Would you want a border around the box that it, that would align with the editor's edge next to the model-viewer preview?

elalish:
I think that's unnecessary, but let's see how it looks.

chrismgeorge:
Screen Shot 2021-01-20 at 12 17 35 PM

elalish:
Fair enough. How about a light grey background instead of a box? Also, let's up the font size to match the icon.

chrismgeorge:
Screen Shot 2021-01-20 at 12 29 53 PM

??

elalish (approved):
Thanks!

PR #1850, opened 2021-01-20
deploy svg for editor

chrismgeorge:


elalish:
Should we just use the icon that's in shared-assets in the first place? Would that have avoided this bug? I'm thinking we should refactor our repo to put all assets in shared-assets for this reason, because it's a bit confusing as is.

chrismgeorge:
> Should we just use the icon that's in shared-assets in the first place? Would that have avoided this bug? I'm thinking we should refactor our repo to put all assets in shared-assets for this reason, because it's a bit confusing as is.

The icon that's in shared assets is a png. I could put the svg I need into shared assets, and then use it?

elalish:
> The icon that's in shared assets is a png. I could put the svg I need into shared assets, and then use it?

Yes, sounds good. Is that SVG used anywhere else? Would be nice to consolidate.

elalish (approved):

PR #1858, opened 2021-01-21
Adjusting pannel height because of modelviewer header

chrismgeorge:


elalish (approved):
Ah, I totally should have noticed that! Thanks.

PR #1859, opened 2021-01-21
bug fix: horse flickers when loaded after model

chrismgeorge:

### DIFF
--- a/packages/space-opera/src/components/materials_panel/materials_panel.ts
+++ b/packages/space-opera/src/components/materials_panel/materials_panel.ts
@@ -239,7 +239,8 @@ export class MaterialPanel extends ConnectedLitElement {
         >${
         this.materials.map(
             (material, id) => html`<paper-item value="${id}">(${id}) ${
-                material.name}</paper-item>`)}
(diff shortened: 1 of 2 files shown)


packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -235,7 +239,8 @@ export class MaterialPanel extends ConnectedLitElement {
         >${
         this.materials.map(
             (material, id) => html`<paper-item value="${id}">(${id}) ${
-                material.name}</paper-item>`)}
+                material.name ? material.name :
```
chrismgeorge, line 242: There is an issue in the editor (which can be seen on the current version) where if you start with the default Robot GLB, and then upload the horse GLB, the dropdown menu would glitch out and constantly "re-select" an element from the list. The dropdown menu seemed to dislike a material's name being undefined if a specific material isn't named.

The reselecting is an issue because we have a feature where if a material is selected it blinks red for a brief time.

chrismgeorge:

elalish (approved):
Thanks!

PR #1972, opened 2021-01-26
Add basic error handling in space-opera viewer preview when loading malformed glb.

hjeldin:
Trying to load a "malformed" glb into the editor yields a white screen with no information to the end user.
The issue with the attached glb file is that while the model is fine, some materials reference textures from './'. 
three.js tries to `fetch` those files, the requests fail and the "error" custom event is triggered but not listened to from space-opera.

Malformed glb: [Ring.glb.zip](https://github.com/google/model-viewer/files/5874330/Ring.glb.zip)

### DIFF
--- a/packages/space-opera/src/components/model_viewer_preview/model_viewer_preview.ts
+++ b/packages/space-opera/src/components/model_viewer_preview/model_viewer_preview.ts
@@ -232,6 +232,9 @@ export class ModelViewerPreview extends ConnectedLitElement {
                 if (this.addHotspotMode) {
                   this.addHotspot(event);
                 }
+              },
+              error: (error: CustomEvent) => {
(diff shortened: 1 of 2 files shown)


hjeldin:
@googlebot I fixed it.

hjeldin:
I added the email I used to commit to my GitHub account, is it enough? Let's try:

@googlebot I fixed it.

chrismgeorge:
Currently looking to using validation from https://github.com/donmccurdy/three-gltf-viewer , which would solve this problem I believe.

chrismgeorge (approved):
LGTM, this functionality wasn't added in the gltf-validator merge, so let's merge this as well now!

elalish (approved):
Thanks for the contribution!

PR #1979, opened 2021-01-27
extend material graph-api

matios13:
with alphaMode , doubleSded and cutoff. 

Based on threejs implementation and [ModelViewerGLTFInstance](model-viewer/packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts )


matios13:
@googlebot I signed it!

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -123,4 +135,55 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     (this[$sourceObject] as GLTFMaterial).emissiveFactor = rgb;
     this[$onUpdate]();
   }
+
+  setAlphaCutoff(value: number) {
+    (this[$sourceObject] as GLTFMaterial)
+    for (const material of this[$correlatedObjects] as
+         Set<MeshStandardMaterial>) {
+      // This little hack ignores alpha for opaque materials, in order to comply
+      // with the glTF spec.
```
elalish, line 144: 👍

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -123,4 +135,55 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+      if (alphaMode === 'MASK') {
+        material.transparent = false;
+        material.depthWrite = true;  // default
+        this.setAlphaCutoff(
+            this.alphaCutoff !== undefined ? this.alphaCutoff : 0.5)
+      }
+    }
+  }
+
+  setDoubleSided(doubleSided: boolean) {
```
elalish, line 178: The trouble here is that we do some special stuff for materials that are transparent and double-sided in order to get the triangle sorting working mostly right: https://github.com/google/model-viewer/blob/b5b86b3f681dc724fe9499c192041d2aa4e19fd8/packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts#L77-L113

The trouble is that the special stuff is changes to the mesh objects that use this material, so I think you'll have to do a quick traversal to find meshes you may need to modify.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -123,4 +135,55 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+  setAlphaCutoff(value: number) {
+    (this[$sourceObject] as GLTFMaterial)
+    for (const material of this[$correlatedObjects] as
+         Set<MeshStandardMaterial>) {
+      // This little hack ignores alpha for opaque materials, in order to comply
+      // with the glTF spec.
+      if (!material.alphaTest && !material.transparent) {
+        material.alphaTest = -0.5;
+      } else if (this.alphaMode === 'MASK') {
+        material.alphaTest = value;
```
elalish, line 148: It might be nice to make functions of things like this so that we can reduce duplication between here and `ModelViewerGLTFInstance.ts`.

elalish (changes requested):
Thanks for taking the initiative here! Would you mind adding a few tests?

elalish:
Thanks! Closing if favor #2616

PR #1842, opened 2021-01-19
Best Practices

chrismgeorge:
[go/modelviewer-best-practices](https://docs.google.com/document/d/1JU8fAWCS4hZIMNcB8GhfTzUHqphIHpivpNamy7_m01g/edit?usp=sharing)

- [x] Allow users to enable the custom loading bar
- [x] Add a outside elements section for modelviewer scripts such as adding the progress bar event listener (and others in the future)
- [x] Allow users to enable the custom ar button
- [x] Change the output of the model viewer snippet from a txt file to an HTML & CSS file.
- [x] Update mobile to simply use and send "modelviewersnippet" state & edits
- [x] Fix issue where progress bar will re-render after a model is loaded / before a model is loaded if it is checked again (conditional rendering)
- [x] Turn js files into a single file for all js. 
- [x] Clean up html output file
- Feedback Todos
- [x] Add AR Prompt
- [x] Clean up progress bar JS to make it more universal (removing ?)
- [x] Decide on whether or not to stringify the progress bar code
- [x] Use single replacement for modelviewer & script instead of individual ones.
- [x] Fetch png to export for ar button
- [x] Fetch hand to export hand prompt
- [x] Comment beatify

PR #1983, opened 2021-01-28
editor-mobile fixes w/ best practices

chrismgeorge:
Fixed bugs from Best Practices #1842 ::
* Hand prompt wasn't animated in WebXR
* The progress bar didn't render on mobile session from the editor
* The hand ar prompt would render on the editor main screen, removed it


elalish (approved):
Thanks for noticing and the quick turnaround!

PR #1984, opened 2021-01-28
Fix variant normals

### LINKED ISSUE #1763: Models displayed diffrently on PC and Mobile
### Description

Model is displaying incorrectly on mobile devices . Here is screenshot from Chrome on Windows PC:
image
And here is screenshot from Chrome on Android device:
image

Material looks totally diffrent, it is too metallic. What can be a cause for such behaviour? 
Here is model affected by this behavoiur:
[model.zip](https://github.com/google/model-viewer/files/5687866/model.zip)

### Version

- model-viewer: v1.3.0

### Browser Affected

- [x] Chrome
- [x] Edge
- [x] Firefox
- [ ] IE
- [ ] Safari

### OS

- [x] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

elalish:
This updates us to three.js r125 and pulls in the equivalent of this change: https://github.com/mrdoob/three.js/pull/21148 as we suffered from the same problem with variants that don't include vertex tangents. I believe r125 also fixes #1763, however it made a regression on the Khronos NormalTangentTest (which is less severe, but should still be addressed down the road).


elalish:
FYI @mrdoob

PR #1962, opened 2021-01-25
Add gltf sample viewer

UX3D-haertl:
This PR integrates the current alpha version (published to npm) of the gltf-sample-viewer from Khronos into the render fidelity tools.

I did not include the golden screenshots, since on my machine the scrollbar is visible on the right side. This also happens for me with the babylon and filament renderers. There is already an issue for filament: https://github.com/google/model-viewer/issues/720

Furthermore, I suppressed a typescript error during importing the npm package (error TS7016: Could not find a declaration file), which seems to appear if someone wants to import js into ts. I did not see how this was solved with the other renderers.

PR #1990, opened 2021-01-29
Add gltf sample viewer

elalish:
Shadowing #1962, adding golden images.


elalish:
Doesn't look like we'll need this branch.

PR #1992, opened 2021-02-01
GLTF Validator in Editor

chrismgeorge:
Merging Validation as is done in Don's https://github.com/donmccurdy/three-gltf-viewer/ into space-opera


chrismgeorge:

chrismgeorge:
Screen Shot 2021-02-04 at 6 14 48 AM
Screen Shot 2021-02-04 at 6 14 55 AM

packages/space-opera/src/components/model_viewer_snippet/components/validation_utils.ts
```diff
@@ -0,0 +1,211 @@
...
+  return await validate(url, gltf);
+}
+
+/**
+ * Loads the gltf using the GLTF Loader.
+ *
+ * @param url The gltf's url in state.entities.gltf.gltfUrl, which is set
+ *     when the model is initially loaded.
+ */
+async function loadGltf(url: string) {
```
elalish, line 44: Is this actually necessary? It seems odd since we already have this code inside model-viewer. If I'm reading this correctly, it seems like all this gltf does is get passed down through functions until it displays the file's metadata, which doesn't seem super necessary. Can we just cut a bunch of code out by only validating?

elalish: If we think this is important, it would probably be better to add a `getMetadata` method on model-viewer and call that.

packages/modelviewer.dev/src/docs-and-examples/sidebar.ts
```diff
@@ -328,17 +328,7 @@ export function sidebarObserver(docsOrExample: string) {
   }
 }
 
-export function sidebarClick(titleName: string = '') {
-  // if a title was supplied on the docs page as being clicked on, we want to
-  // decrease the --scroll-top css so that the top of the title is flush
-  // with the top of the window
```
elalish, line 334: 👍

packages/space-opera/src/components/model_viewer_snippet/components/validation_utils.ts
chrismgeorge: 👍

elalish (approved):
Thanks, looks great!

PR #2018, opened 2021-02-16
updated gltf-sample-viewer goldens

elalish:
Getting rid of the scrollbars. 

@UX3D-haertl Looking at https://modelviewer.dev/fidelity/, it appears you still have a problem with tone-mapping or exposure. It's especially obvious on Cube, Duck, and the simple grey tests toward the bottom. Also the furnace test is wrong, as it doesn't have the proper grey background and the spheres look over-exposed. Would you mind fixing these issues before I merge this PR?


UX3D-nopper:
@elalish We have implemented two version of ACES tonemapping. Actually, we need to discuss in 3D Formats / 3D Commerce, which one should be used and/or is preferred.
I try to raise this topic in this evening call.
In any case, we will revisit, if we have a bug somewhere.

UX3D-haertl:
I updated the gltf-sample-viewer on our branch. Now the issues with the screenshots should be fixed. VC is still brighter than in the other renderers, but the other screenshots look good.

https://github.com/ux3d/model-viewer/tree/addGltfSampleViewer

elalish:
Great, can you make a PR? I'll update this one after we merge yours. 

VC is weird; Filament had a bunch of trouble with it too. Something to do with how ambient occlusion is handled iirc.

elalish:
Closing in favor of #2128.

PR #2043, opened 2021-02-23
update to WebGL2 where available

elalish:
Three.js updated for WebGL2, but there was a Chrome bug preventing us. Here we'll test progress.


elalish:
Nope, still blocked by https://bugs.chromium.org/p/chromium/issues/detail?id=1120100

elalish:
This happened as part of our upgrade to r129, since three.js put in a work-around for the Chrome bug.

PR #2058, opened 2021-02-26
removed fancy default for pan-y

elalish:
There was built into the default `touch-action: "pan-y"` a special case to effectively turn it into `touch-action: "none"` when the modelviewer element was fullscreen. However, this doesn't work well at all for iframes and it's honestly a little confusing anyway. From now on, if you know your page doesn't need to scroll, you'll have to remember to set `touch-action: "none"` manually.


elalish:
FYI @mqg734

PR #2078, opened 2021-03-02
starting animationMixer

shamsail:
### Reference Issue

Hi, I wanted to work/contribute on the animation side so that we can provide animationMixer feasibility as three.js does in the modelviewer's API.

### DIFF
--- a/packages/model-viewer/src/features/animation.ts
+++ b/packages/model-viewer/src/features/animation.ts
@@ -12,11 +12,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/*
+animationMixer
+*/
 
-import {property} from 'lit-element';
+import { property } from 'lit-element';
 
-import ModelViewerElementBase, {$hasTransitioned, $needsRender, $onModelLoad, $renderer, $scene, $tick, $updateSource} from '../model-viewer-base.js';
-import {Constructor} from '../utilities.js';
+import ModelViewerElementBase, { $hasTransitioned, $needsRender, $onModelLoad, $renderer, $scene, $tick, $updateSource } from '../model-viewer-base.js';
+import { Constructor } from '../utilities.js';
 
 const MILLISECONDS_PER_SECOND = 1000.0
 
@@ -25,7 +28,7 @@ const $paused = Symbol('paused');
 
 export declare interface AnimationInterface {
   autoplay: boolean;
-  animationName: string|void;
+  animationName: string | void;
   animationCrossfadeDuration: number;
   readonly availableAnimations: Array<string>;
   readonly paused: boolean;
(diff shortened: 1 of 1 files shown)


elalish:
Great, thank you! Can you give a description of a use case you'd like to solve and take a stab at an API you'd like to expose to do that? Feel free to reference the three.js documentation, but we want to keep our API agnostic of theirs.

elalish:
Also, if you're developing on VSCode, please install the clang-format extension so that you get consistent formatting with our repo.

shamsail:
Hi, Ok, let me explain some scenarios .. like 

1) what if user wants to play some chosen animations 3 out of 8.. (forward and backward directions)
2) what if user wants to play animation only one time (not in loop/continous)(currently there is only play and pause for specific animation)
3) what if user wants to combine the available animations or putting some weights (not to make new animations)
4) export/import animation clips from/in the file/scene   


https://threejs.org/docs/#api/en/animation/AnimationClip
https://threejs.org/docs/#api/en/animation/AnimationAction
https://threejs.org/docs/#api/en/animation/AnimationMixer

Also, I have tried with clang-format but it's not working , may be i'm setting wrong path for this settings.json, but i'll see it again. anyhow, i have disabled my previous formatter for this project.

shamsail:
Screenshot from 2021-03-03 13-28-20
Also, I'm doing "npm run build", it's giving me errors for "lerna". Although I have installed dependencies. I'm attaching screenshot of the error.
I'm newbie in this so pardon me.

elalish:
I was getting those error too when I upgraded to npm 7; try downgrading to 6.14.

elalish:
> 1. what if user wants to play some chosen animations 3 out of 8.. (forward and backward directions)
> 2. what if user wants to play animation only one time (not in loop/continous)(currently there is only play and pause for specific animation)
I believe these are possible with our current API, though you might need a RAF callback to set the animation time.
> 3. what if user wants to combine the available animations or putting some weights (not to make new animations)
This seems to be the core of the new work. What are the use cases for this in model-viewer? Why not say new animations should be exported from the animation software like any other significant model edit?
> 4. export/import animation clips from/in the file/scene
Animations are already imported and exported with glTF. Is something not working?

KartikSoneji:
I found this PR while searching for a way to allow multiple chosen animations to play simultaneously.
Is this currently supported by the library?
If not, would a patch to add this feature be accepted?

I was thinking of a simple extension like `animation-name="animation1 animation2"` for selecting animations and `autoplay="all"` to play all animations.

elalish:
> I was thinking of a simple extension like `animation-name="animation1 animation2"` for selecting animations and `autoplay="all"` to play all animations.

@KartikSoneji We'd absolutely accept a PR like that. That API shape sounds reasonable. These animation APIs tend to have a lot of edge cases, so please try to think beyond your  own use case when testing it. Note that we now have `play({repetitions: n})`, which if called with two animations of different length will make them play out of sync. I think that's a reasonable behavior, but these are the kinds of things I'd like you to think through.

hybridherbst:
Just stumbled upon this issue; for reference, Gestaltor has a concept of "non-conflicting animations" and actually prevents playback of conflicting ones (e.g. multiple animations target the same thing, thus behaviour would be undefined):
image.

Don't think that would be necessary here for playing multiple ones but I thought that's interesting for a potential mixer :)

PR #2092, opened 2021-03-05
Fix CacheEvictionPolicy retainerCounts  

### LINKED ISSUE #1674: Changing src reloads/crashes the page
### Description

To test this behaviour I've used the example code from modelviewer.dev along with models from Smithsonians open gallery (https://3d.si.edu/explore).
When changing the src property 3-4 times, the page reloads/crashes. I've downloaded models from modelviewer.dev to make sure it's not the models, but the page reloaded/crashed after 10-15 times of chaning the src.
Keep chaning the src until it reloads. With other models (larger perhaps?) it happens very quickly (3-4 times).
It doesn't happen on desktop, only on Android and iOS.

Edit:
Watching RAM on PC makes me realize that Memory is going up every time I change the src. Which probably leads to out of memory on mobile.

#### Live Demo
https://www.deltareality.com/webar/smithsonian-gallery/
Open this on mobile device or Chrome inspection toolbar for mobile

### Version
- model-viewer: v1.3.0

### Browser Affected

- [x ] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [ x] Safari

### OS

- [ x] Android
- [ x] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

mqg734:
When a model is in the viewport but hasn't finished loading yet, the intersection observer triggers updateSource multiple times with the same source, resulting in extra retainerCounts for that source's URL in CacheEvictionPolicy. The url check at the top of the function is supposed to avoid that, but wasn't working because this.url is not set until after some async calls.

Fixes #1674 
It's possible there are other leaks still, but I think this was the big one.

### DIFF
--- a/packages/model-viewer/src/three-components/ModelScene.ts
+++ b/packages/model-viewer/src/three-components/ModelScene.ts
@@ -166,6 +166,7 @@ export class ModelScene extends Scene {
       }
       return;
     }
+    this.url = url;
 
     // If we have pending work due to a previous source change in progress,
     // cancel it so that we do not incur a race condition:
@@ -198,7 +199,6 @@ export class ModelScene extends Scene {
     }
 
     this.reset();
-    this.url = url;
     this._currentGLTF = gltf;
 
     if (gltf != null) {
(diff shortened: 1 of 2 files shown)


elalish (changes requested):
This is great, thanks! Can you also go ahead and remove the three.js cache? I've looked into that more and I'm pretty sure that's just a memory leak from old code. I think that'll fix the issue I linked.

elalish (approved):
Thank you!

PR #2096, opened 2021-03-05
Fix SceneViewer intent params

elalish:
Fixes #2093 and some related problems. Also cleans up the code since we don't have to worry about IE11 support anymore.

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -268,31 +268,37 @@ configuration or device capabilities');
      * the current device.
      */
     [$openSceneViewer]() {
-      // This is necessary because the original URL might have query
-      // parameters. Since we're appending the whole URL as query parameter,
-      // ? needs to be turned into & to not lose any of them.
-      const gltfSrc = this.src!.replace('?', '&');
       const location = self.location.toString();
       const locationUrl = new URL(location);
-      const modelUrl = new URL(gltfSrc, location);
+      const modelUrl = new URL(this.src!, location);
+      const params = new URLSearchParams(modelUrl.search);
 
       locationUrl.hash = noArViewerSigil;
 
       // modelUrl can contain title/link/sound etc.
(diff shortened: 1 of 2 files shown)


elalish:
FYI @tpsiaki

packages/model-viewer/src/features/ar.ts
```diff
@@ -268,31 +268,37 @@ configuration or device capabilities');
...
+        params.set('enable_vertical_placement', 'true');
+      }
+      if (params.has('sound')) {
+        const soundUrl = new URL(params.get('sound')!, location);
+        params.set('sound', soundUrl.toString());
+      }
+      if (params.has('link')) {
+        const linkUrl = new URL(params.get('link')!, location);
+        params.set('link', linkUrl.toString());
       }
```
tpsiaki, line 297: it seems not the best that users pass a model url w/ sound & link get params that are used for the actual model request in addition to sending to Scene Viewer for sound/link. I don't think there's a great alternative solution though without adding Scene Viewer specific model-viewer attributes which might be more trouble than they're worth.

packages/model-viewer/src/features/ar.ts
```diff
@@ -268,31 +268,37 @@ configuration or device capabilities');
      * the current device.
      */
     [$openSceneViewer]() {
-      // This is necessary because the original URL might have query
-      // parameters. Since we're appending the whole URL as query parameter,
-      // ? needs to be turned into & to not lose any of them.
-      const gltfSrc = this.src!.replace('?', '&');
```
tpsiaki, line 274: removing this makes me happy :)

tpsiaki (approved):

packages/model-viewer/src/features/ar.ts
elalish: Agreed, it's not the best, but no one has complained so far. At least it's fairly unlikely their server is looking for the same query params as SceneViewer, but an intersection would be a problem.

PR #2128, opened 2021-03-15
update gltf-sample-viewer goldens

elalish:
Replacing #2018.


elalish:
@UX3D-haertl @UX3D-nopper; thanks for the updates! These renders are looking much better; not just the tone-mapping but the diffuse reflections and transparency as well.

PR #2286, opened 2021-04-12
Adds 'interpolation-decay' attribute

### LINKED ISSUE #580: Make camera controls physics configurable
In #288 we introduced a new camera control scheme that features many potentially configurable parameters.

In https://github.com/GoogleWebComponents/model-viewer/issues/498#issuecomment-486709830 we discussed making the physics of camera interpolation configurable. This includes qualities such as velocity and dampening. We should add sufficient configurability to `<model-viewer>` to ensure that camera interpolation looks and feels the way a user expects.

timmmeh:
* Setting the attribute affects camera smoothing and model position interpolation.
* Adds 'decayMilliseconds' to Damper.update() method call.
* Updates to docs - adds interpolation-decay
* Updates to the Interpolation example - adds interpolation-decay attribute example

### Reference Issue

Fixes #580

### DIFF
--- a/packages/model-viewer/src/features/controls.ts
+++ b/packages/model-viewer/src/features/controls.ts
@@ -15,7 +15,6 @@
 
 import {property} from 'lit-element';
 import {Event, PerspectiveCamera, Spherical, Vector3} from 'three';
-
 import {style} from '../decorators.js';
 import ModelViewerElementBase, {$ariaLabel, $container, $hasTransitioned, $loadedTime, $needsRender, $onModelLoad, $onResize, $renderer, $scene, $tick, $userInputElement, toVector3D, Vector3D} from '../model-viewer-base.js';
 import {degreesToRadians, normalizeUnit} from '../styles/conversions.js';
@@ -33,6 +32,7 @@ import {timeline} from '../utilities/animation.js';
 // would cause the interaction prompt to glitch unexpectedly
 // @see https://github.com/google/model-viewer/issues/839
 const PROMPT_ANIMATION_TIME = 5000;
+export const DEFAULT_DECAY_MILLISECONDS = 50;
 
 // For timing purposes, a "frame" is a timing agnostic relative unit of time
 // and a "value" is a target value for the keyframe.
@@ -243,6 +243,7 @@ export declare interface ControlsInterface {
   orbitSensitivity: number;
   touchAction: TouchAction;
   bounds: Bounds;
+  interpolationDecayMilliseconds: number;
   getCameraOrbit(): SphericalPosition;
   getCameraTarget(): Vector3D;
   getFieldOfView(): number;
@@ -339,6 +340,9 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
 
     @property({type: Boolean, attribute: 'disable-zoom'})
     disableZoom: boolean = false;
+
+    @property({type: Number, attribute: 'interpolation-decay-milliseconds'})
+    interpolationDecayMilliseconds: number = DEFAULT_DECAY_MILLISECONDS;
 
     @property({type: String, attribute: 'bounds'}) bounds: Bounds = 'legacy';
 
@@ -495,6 +499,11 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
 
       if (changedProperties.has('orbitSensitivity')) {
         controls.sensitivity = this.orbitSensitivity;
+      }
+
+      if (changedProperties.has('interpolationDecayMilliseconds')) {
+        controls.interpolationDecayMilliseconds =
+            this.interpolationDecayMilliseconds;
       }
 
       if (this[$jumpCamera] === true) {
--- a/packages/model-viewer/src/features/staging.ts
+++ b/packages/model-viewer/src/features/staging.ts
@@ -79,7 +79,6 @@ export const StagingMixin = <T extends Constructor<ModelViewerElementBase>>(
       this[$autoRotateStartTime] = performance.now();
     }
 
-    // tslint:disable-next-line:no-any
     updated(changedProperties: Map<string, any>) {
       super.updated(changedProperties);
 
--- a/packages/model-viewer/src/features/staging.ts
+++ b/packages/model-viewer/src/features/staging.ts
@@ -15,15 +15,15 @@
  */
 
 import {property} from 'lit-element';
-
 import {style} from '../decorators.js';
 import ModelViewerElementBase, {$hasTransitioned, $renderer, $scene, $tick} from '../model-viewer-base.js';
 import {degreesToRadians} from '../styles/conversions.js';
 import {EvaluatedStyle, Intrinsics} from '../styles/evaluators.js';
 import {numberNode, NumberNode} from '../styles/parsers.js';
 import {Constructor} from '../utilities.js';
+import {CameraChangeDetails, DEFAULT_DECAY_MILLISECONDS} from './controls.js';
 
-import {CameraChangeDetails} from './controls.js';
+
 
 // How much the model will rotate per
 // second in radians:
@@ -45,6 +45,7 @@ const $onCameraChange = Symbol('onCameraChange');
 export declare interface StagingInterface {
   autoRotate: boolean;
   autoRotateDelay: number;
+  interpolationDecayMilliseconds: number;
   readonly turntableRotation: number;
   resetTurntableRotation(theta?: number): void;
 }
@@ -63,6 +64,9 @@ export const StagingMixin = <T extends Constructor<ModelViewerElementBase>>(
     @property({type: String, attribute: 'rotation-per-second'})
     rotationPerSecond: string = 'auto';
 
+    @property({type: Number, attribute: 'interpolation-decay-milliseconds'})
+    interpolationDecayMilliseconds: number = DEFAULT_DECAY_MILLISECONDS;
+
     private[$autoRotateStartTime] = performance.now();
     private[$radiansPerSecond] = 0;
 
@@ -80,11 +84,16 @@ export const StagingMixin = <T extends Constructor<ModelViewerElementBase>>(
       this[$autoRotateStartTime] = performance.now();
     }
 
+    // tslint:disable-next-line:no-any
     updated(changedProperties: Map<string, any>) {
       super.updated(changedProperties);
 
       if (changedProperties.has('autoRotate')) {
         this[$autoRotateStartTime] = performance.now();
+      }
+      if (changedProperties.has('interpolationDecayMilliseconds')) {
+        this[$scene].interpolationDecayMilliseconds =
+            this.interpolationDecayMilliseconds;
       }
     }
 
--- a/packages/model-viewer/src/three-components/Damper.ts
+++ b/packages/model-viewer/src/three-components/Damper.ts
@@ -14,10 +14,7 @@
  */
 
 export const SETTLING_TIME = 10000;  // plenty long enough
-const DECAY_MILLISECONDS = 50;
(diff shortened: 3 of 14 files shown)


packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -356,12 +358,11 @@ export class SmoothControls extends EventDispatcher {
...
-    const deltaRatio = deltaZoom === 0 ?
-        0 :
-        deltaZoom > 0 ? (maximumRadius! - radius) /
-                (Math.log(maximumFieldOfView!) - this.goalLogFov) :
+    const deltaRatio = deltaZoom === 0 ? 0 :
+        deltaZoom > 0                  ? (maximumRadius! - radius) /
+            (Math.log(maximumFieldOfView!) - this.goalLogFov) :
                         (radius - minimumRadius!) /
-                (this.goalLogFov - Math.log(minimumFieldOfView!));
+            (this.goalLogFov - Math.log(minimumFieldOfView!));
```
elalish, line 365: This formatting looks weird; do you have clang-format set up?

packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -496,6 +508,7 @@ export class SmoothControls extends EventDispatcher {
     if (TOUCH_EVENT_RE.test(event.type)) {
       const {touches} = event as TouchEvent;
 
+      // tslint:disable-next-line:switch-default
```
elalish, line 511: I try to avoid these where possible; what is this about?

packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -536,7 +549,7 @@ export class SmoothControls extends EventDispatcher {
 
     if (event.cancelable) {
       event.preventDefault();
-    };
+    }
```
elalish, line 552: 👍

packages/model-viewer/src/three-components/Damper.ts
```diff
@@ -34,7 +31,11 @@ export class Damper {
 
   update(
       x: number, xGoal: number, timeStepMilliseconds: number,
-      xNormalization: number): number {
+      xNormalization: number, decayMilliseconds: number = 50): number {
+    const naturalFrequency =
+        1 / Math.max(MIN_DECAY_MILLISECONDS, decayMilliseconds);
```
elalish, line 36: This is probably a good idea, but I'd still like to hear your reasoning on making 50 the minimum value. Also, this reminds me we should add documentation before we merge this, as details like this are important to communicate.

elalish: You might consider parameterizing the damper with naturalFrequency directly, so that the max and divide can happen once instead of inside the loop.

packages/model-viewer/src/features/staging.ts
```diff
@@ -63,6 +64,9 @@ export const StagingMixin = <T extends Constructor<ModelViewerElementBase>>(
     @property({type: String, attribute: 'rotation-per-second'})
     rotationPerSecond: string = 'auto';
 
+    @property({type: Number, attribute: 'interpolation-decay-milliseconds'})
+    interpolationDecayMilliseconds: number = DEFAULT_DECAY_MILLISECONDS;
```
elalish, line 68: I don't think this will work, since the attributes from all of the mixins are surfaced on the element and this will create a name collision. Also, I don't see the point of specifying them separately anyway; let's just set both controls and scene from the controls mixin.

packages/model-viewer/src/features/controls.ts
```diff
@@ -340,6 +341,9 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
     @property({type: Boolean, attribute: 'disable-zoom'})
     disableZoom: boolean = false;
 
+    @property({type: Number, attribute: 'interpolation-decay-milliseconds'})
```
elalish, line 344: I don't think we need milliseconds in the name here, since that's the JS default time unit anyway (we don't bother for `auto-rotate-delay` for instance). How about `interpolation-decay-time`?

elalish (changes requested):

elalish:
Also, please mark `Fixes ...` so that Github will automatically link and close the issue when this is merged.

packages/model-viewer/src/three-components/SmoothControls.ts
timmmeh: tslint is complaining that there is no 'default:' case here. It looks like if no case label is matched the behavior is to fall out of the switch statement and execute the line 'this.lastTouches = touches;'

I will just add a default statement that breaks out of the switch, a bit of future proofing.

packages/model-viewer/src/three-components/SmoothControls.ts
elalish: funny, it's not doing it for me. Well, they both seem equally weird, so no big deal. Honestly when the formatting gets this ugly it probably means we should just make some named `const`s and split this into multiple lines.

packages/model-viewer/src/three-components/Damper.ts
timmmeh: I made the default value 50 because this was the previous value, I added a default to not break any code that might be using the Damper. For the minimum I just picked an arbitrary small value to avoid a divide by 0 situation, since the change allows anyone to set this value, and this is one way it could go wrong.

Added setting the naturalFrequency to avoid doing the divide every frame.

packages/model-viewer/src/features/staging.ts
timmmeh: Done.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -107,8 +107,7 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
       const duplicateMaterial = Array.isArray(mesh.material) ?
           duplicateMaterials :
           duplicateMaterials[0];
-      const meshBack = mesh.clone() as Mesh;
-      meshBack.material = duplicateMaterial;
+      const meshBack = new Mesh(mesh.geometry, duplicateMaterial);
```
elalish, line 110: This is reverting a change in master. Maybe a bad pull? It wasn't in your first commit.

packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -132,6 +132,8 @@ export class SmoothControls extends EventDispatcher {
   private thetaDamper = new Damper();
   private phiDamper = new Damper();
   private radiusDamper = new Damper();
+  // Controls the various damper interpolation decays
+  interpolationDecayMilliseconds: number = 50;
```
elalish, line 136: Is this supposed to be here?

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -93,6 +93,8 @@ export class ModelScene extends Scene {
   public canScale = true;
   public tightBounds = false;
 
+  public interpolationDecayMilliseconds = 50;
```
elalish, line 96: Is this vestigial?

packages/model-viewer/src/three-components/Damper.ts
```diff
@@ -14,27 +14,40 @@
...
- * NIL_SPEED clamping also happens at roughly the same convergence for all
- * states.
+ * DECAY_MILLISECONDS. This common parameter makes all states converge at the
+ * same rate regardless of scale. xNormalization is a number to provide the
+ * rough scale of x, such that NIL_SPEED clamping also happens at roughly the
+ * same convergence for all states.
  */
 export class Damper {
   private[$velocity]: number = 0;
+  private[$naturalFrequency]: number = 0;
```
elalish, line 32: You followed our style nicely here, but it turns out I just hadn't gotten around to fixing this yet. I've been removing symbols from all the classes that are not publicly visible, since they're pretty annoying to read (I changed this style rule when I took over the project, so there's still some left). Basically whenever I edit a file that still has them, I remove them (except for the mixins).

packages/model-viewer/src/three-components/Damper.ts
elalish: Oh, now I see. I missed that `MIN_DECAY_MILLISECONDS` was a different constant. Thanks.

elalish (changes requested):

packages/model-viewer/src/three-components/SmoothControls.ts
timmmeh: nope, done.

packages/model-viewer/src/three-components/ModelScene.ts
timmmeh: Removed

packages/model-viewer/src/three-components/Damper.ts
timmmeh: lol, ok, done.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
timmmeh: I see, this popped out of the merge, putting it back.

packages/model-viewer/src/features/staging.ts
```diff
@@ -80,6 +79,7 @@ export const StagingMixin = <T extends Constructor<ModelViewerElementBase>>(
       this[$autoRotateStartTime] = performance.now();
     }
 
+    // tslint:disable-next-line:no-any
```
elalish, line 82: Let's skip this. In fact, you could probably just revert this file, since there aren't any real changes here.

elalish (changes requested):
This looks great, thanks! Let's add the documentation and then we can merge it.

packages/model-viewer/src/features/staging.ts
timmmeh: Done.

packages/modelviewer.dev/data/docs.json
```diff
@@ -552,6 +552,15 @@
         "links": [
           "<a href=\"../examples/scenegraph/#transforms\">Related example</a>"
         ]
+      },
+      {
+        "name": "interpolation-decay",
+        "htmlName": "interpolationDecay",
+        "description": "Controls model interaction sensitivity, the smaller the value the more sensitive the interaction will be.",
```
elalish, line 559: How about: "Controls the rate of interpolation when the camera or model moves, due to either user interaction or attribute changes. The decay is asymptotic and the value is in milliseconds, where the majority of the movement will occur within this value's time. Doubling this value will cut the speed in half."

timmmeh: Sure sounds great

packages/modelviewer.dev/data/docs.json
```diff
@@ -552,6 +552,15 @@
...
         ]
+      },
+      {
+        "name": "interpolation-decay",
+        "htmlName": "interpolationDecay",
+        "description": "Controls the rate of interpolation when the camera or model moves, due to either user interaction or attribute changes. The decay is asymptotic and the value is in milliseconds, where the majority of the movement will occur within this value's time. Doubling this value will cut the speed in half.",
+        "default": {
+          "default": "50",
+          "options": "values greater than 0"
+        }
```
elalish, line 563: Can you also add the link section like above to point to your example?

timmmeh: No problem

elalish (dismissed):
Thanks, this looks great!

elalish (approved):
Thanks!

PR #2328, opened 2021-04-27
add external renderer support

elalish:
This is experimental, for using model-viewer to control an external renderer instead of using Three.js. This removes any dependence on glTF and even allows non-model-based experiences (e.g. generative systems, etc). Of course the model-dependent parts of our API will also cease to function in this situation.


mikkoh:
❤️

PR #2379, opened 2021-05-13
Set texture api

### LINKED ISSUE #1678: Material SetURI Bug
### Description

Having two <model-viewer /> with the same `src` and trying to "setURI" in one of them changes the same material/texture on the other.

#### Live Demo
Accessing [Staging and Cameras Page](https://modelviewer.dev/examples/stagingandcameras/) and running the following code on devTools console:
`document.getElementsByTagName('model-viewer')[0].model.materials[0].pbrMetallicRoughness.baseColorTexture.texture.source.setURI('https://picsum.photos/20')`

(Works with other models as well and other URLs as well)

### Version

- model-viewer: v1.3.0

### Browser Affected

- [x] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [x] Windows

timmmeh:
### Reference Issue
Resolves issue with setURI where updating a texture on one object updates all like objects.

Fixes #1678

### DIFF
--- a/packages/model-viewer/src/features/scene-graph/material.ts
+++ b/packages/model-viewer/src/features/scene-graph/material.ts
@@ -116,18 +116,6 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     return (this[$sourceObject] as GLTFMaterial).emissiveFactor!;
   }
 
-  setBaseColorMap(uri: string) {
-    if (this[$pbrMetallicRoughness] !== null) {
-      this[$pbrMetallicRoughness].setBaseColorMap(uri);
-    }
-  }
-
-  setMetallicRoughnessMap(uri: string) {
-    if (this[$pbrMetallicRoughness] !== null) {
-      this[$pbrMetallicRoughness].setMetallicRoughnessMap(uri);
-    }
-  }
-
   setNormalMap(uri: string) {
     if (this[$normalTexture] !== null) {
       this[$normalTexture]!.texture.source.setURI(uri);
--- a/packages/model-viewer/src/features/scene-graph/material.ts
+++ b/packages/model-viewer/src/features/scene-graph/material.ts
@@ -23,7 +23,6 @@ import {TextureInfo} from './texture-info.js';
 import {$correlatedObjects, $onUpdate, $sourceObject, ThreeDOMElement} from './three-dom-element.js';
 
 
-
 const $pbrMetallicRoughness = Symbol('pbrMetallicRoughness');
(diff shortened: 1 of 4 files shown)


packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -177,6 +179,24 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
     }
 
     const clone = material.clone() as MeshStandardMaterial;
+    if (material.map != null) {
+      material.map = material.map.clone();
```
elalish, line 183: This is how I tried to fix this the first time too, but it doesn't work. Cloning does not do the right thing and it ended up breaking things, see #1711 and #2279.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -115,6 +116,36 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     return (this[$sourceObject] as GLTFMaterial).emissiveFactor!;
   }
 
+  setBaseColorMap(uri: string) {
+    if (this[$pbrMetallicRoughness] !== null) {
+      this[$pbrMetallicRoughness].setBaseColorMap(uri);
+    }
+  }
+
+  setMetallicRoughnessMap(uri: string) {
```
elalish, line 125: I wonder if we should have these duplicated here or not. I can see both sides, but I tend toward keeping the API minimal. Do you have a strong opinion?

elalish (changes requested):

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: removed

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
timmmeh: Per conversation, shininess bug is not reproducible

elalish (approved):
I just removed all the helper methods for now because I'd like to keep this minimal for the moment. I'm really glad you found the simple fix for this that somehow eluded me. This makes me feel much better about my understanding of three.js. Let's come back to the API shape discussion as we look to build it out further.

PR #2380, opened 2021-05-14
simplified USDZ and added docs

elalish:
@kolodi A slight simplification of #2374 in order to release the USDZ memory earlier. Also added info about `ios-src` and how auto-generation interacts with the scene-graph API.


Aubi0ne:
Hello, thanks for the feature ! Do you know when it will be available in the release version ?

elalish:
Now: https://github.com/google/model-viewer/releases/tag/v1.7.0

PR #2399, opened 2021-05-18
workaround for scene-viewer bug

### LINKED ISSUE #2381: [SceneViewer] AR not working on new installations/after new update.
### Description

Hi,
after update of Google App from 14th May 2021 (v. 12.18.11.23.arm) SceneViewer stopped working in "ar_only" mode which is used in model-viewer.

### To reproduce
1. Uninstall updates of [Google App](https://play.google.com/store/apps/details?id=com.google.android.googlequicksearchbox)
2. Install them again
3. Go to https://modelviewer.dev/examples/augmentedreality/#sceneViewer and try to place model in AR.
It results in error such as "File could not be loaded". It seems it happens because SceneViewer doesn't have permissions to camera etc. and app fails and exits before user can grant those permissions. If I run SceneViewer app in "3d_preferred" mode which is used in Google AR examples (like duck or other animals) and then switch to ar mode, I can grant those permissions and model loads in AR. After I grant those permissions opening any model in "ar_only" mode (via model-viewer) also works correctly.

It is really a **major** bug that just breaks usage of SceneViewer.

### OS
- [x] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### AR
- [ ] WebXR
- [x] SceneViewer
- [ ] QuickLook

elalish:
Fixes #2381 

Thanks @60days for finding the quick work-around. Scene Viewer has also fixed the underlying issue, but this will keep us from having to wait a couple weeks for that to roll out.

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -303,7 +303,7 @@ configuration or device capabilities');
       locationUrl.hash = noArViewerSigil;
 
       // modelUrl can contain title/link/sound etc.
-      params.set('mode', 'ar_only');
+      params.set('mode', 'ar_preferred');
       if (!params.has('disable_occlusion')) {
         params.set('disable_occlusion', 'true');
       }


MicDesign1:
There is no way to deploy this workaround for those of us using model viewer via web app, correct?

elalish:
It's been release in v1.7.1; does that work for you?

PR #2411, opened 2021-05-21
Possible fix for textures that appear fully glossy or black

timmmeh:

### DIFF
--- a/packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
+++ b/packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
@@ -185,12 +185,18 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
       clone.emissiveMap.needsUpdate = true;
     }
 
+    // Clones the roughnessMap if it exists.
     const roughnessMap: Texture|null = material.roughnessMap!.clone();
 
+    // Assigns the roughnessMap to the cloned material and generates mipmaps.
     if (roughnessMap !== null) {
       roughnessMap.needsUpdate = true;
       clone.roughnessMap = roughnessMap;
 
+      // Generates mipmaps from the clone of the roughnessMap
+      // TODO (b/188906197) Render artifacts when generating mipmaps after
+      // cloning (texture). Move back to 'scene.traverse()' when bug is
+      // resolved.
       const {threeRenderer, roughnessMipmapper} = Renderer.singleton;
       // XR must be disabled while doing offscreen rendering or it will
       // clobber the camera.
@@ -200,6 +206,8 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
       threeRenderer.xr.enabled = enabled;
     }
 
+    // Checks if roughnessMap and metalnessMap share the same texture and
+    // either clones or assigns.
     if (material.roughnessMap === material.metalnessMap) {
       clone.metalnessMap = roughnessMap;
     } else if (material.metalnessMap != null) {
@@ -207,6 +215,8 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
       clone.metalnessMap.needsUpdate = true;
     }
 
+    // Checks if roughnessMap and aoMap share the same texture and
+    // either clones or assigns.
     if (material.roughnessMap === material.aoMap) {
       clone.aoMap = roughnessMap;
     } else if (material.aoMap != null) {
--- a/packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
+++ b/packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import {BackSide, DoubleSide, FrontSide, Material, Mesh, MeshStandardMaterial, Object3D, Shader} from 'three';
+import {BackSide, DoubleSide, FrontSide, Material, Mesh, MeshStandardMaterial, Object3D, Shader, Texture} from 'three';
 import {GLTF} from 'three/examples/jsm/loaders/GLTFLoader.js';
 
 import {$clone, $prepare, $preparedGLTF, GLTFInstance, PreparedGLTF} from '../GLTFInstance.js';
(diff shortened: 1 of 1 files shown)


packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -185,21 +178,38 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
...
+    } else if (material.metalnessMap != null) {
       clone.metalnessMap = material.metalnessMap.clone();
       clone.metalnessMap.needsUpdate = true;
     }
-    if (material.roughnessMap != null) {
-      clone.roughnessMap = material.roughnessMap.clone();
-      clone.roughnessMap.needsUpdate = true;
+
+    if (material.roughnessMap === material.aoMap) {
+      clone.metalnessMap = roughnessMap;
```
elalish, line 209: aoMap?

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -185,21 +178,38 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
...
+    if (roughnessMap !== null) {
+      roughnessMap.needsUpdate = true;
+      clone.roughnessMap = roughnessMap;
+
+      const {threeRenderer, roughnessMipmapper} = Renderer.singleton;
+      // XR must be disabled while doing offscreen rendering or it will
+      // clobber the camera.
+      const {enabled} = threeRenderer.xr;
+      threeRenderer.xr.enabled = false;
+      roughnessMipmapper.generateMipmaps(clone as MeshStandardMaterial);
```
elalish, line 197: Okay, so this looks like what I missed in my first version. This shouldn't be necessary, so we should try to figure out why this is broken in three. Still, good enough for now.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -185,21 +178,38 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
...
-      clone.aoMap = material.aoMap.clone();
-      clone.aoMap.needsUpdate = true;
-    }
     if (material.emissiveMap != null) {
       clone.emissiveMap = material.emissiveMap.clone();
       clone.emissiveMap.needsUpdate = true;
     }
-    if (material.metalnessMap != null) {
+
+    const roughnessMap: Texture|null = material!.roughnessMap!.clone();
```
elalish, line 186: nit: don't think you need the first "!".

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
timmmeh: doh - good catch.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
timmmeh: done.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
timmmeh: Agreed, I will setup a test case to dig further, if nothing else it will provide a repro case to give to threejs (assuming it does repro)

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -186,7 +186,10 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
     }
 
     // Clones the roughnessMap if it exists.
-    const roughnessMap: Texture|null = material.roughnessMap!.clone();
```
elalish: Ah yes, the danger of the ! - I should have noticed that.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -185,21 +180,51 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
...
     if (material.emissiveMap != null) {
       clone.emissiveMap = material.emissiveMap.clone();
       clone.emissiveMap.needsUpdate = true;
     }
-    if (material.metalnessMap != null) {
+
+    // Clones the roughnessMap if it exists.
+    let roughnessMap: Texture|null = null;
+    if (material.roughnessMap) {
+      roughnessMap = material.roughnessMap!.clone();
```
elalish, line 191: TS is smart enough that you don't need the `!` here. Also, we generally prefer to be more explicit with `material.roughnessMap != null`.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -185,21 +180,51 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
...
+    }
+
+    // Assigns the roughnessMap to the cloned material and generates mipmaps.
+    if (roughnessMap !== null) {
```
elalish, line 195: This is the one time where we prefer `==` and `!=`, since they catch both null and undefined. There are lots of opinions about it, but that's model-viewer style.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -185,21 +180,51 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
...
+      roughnessMap.needsUpdate = true;
+      clone.roughnessMap = roughnessMap;
+
+      // Generates mipmaps from the clone of the roughnessMap
+      // TODO (b/188906197) Render artifacts when generating mipmaps after
```
elalish, line 200: Let's avoid internal references since this code is all public. Let's put a three.js bug link here instead if we have one.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
timmmeh: Done.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
timmmeh: Done.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
timmmeh: I'm not sure what you're suggesting. Do you want a != check to see if it can be coerced into null?
Added != null check.

elalish: no, instead of, not in addition to. We just don't use `=== null` or `!== null`.

elalish (approved):

PR #2414, opened 2021-05-21
simplified progress bar CSS for easier styling

elalish:
Fixes #2393 

The simplifies the progress bar CSS so that it is more ergonomic to style using `part`. It is slightly different than it used to be, fading out instead of sliding up, but given its small default height, I'd say most people will be unlikely to notice the difference. @maciekglowka does this look like a reasonable change?  I verified that your initial CSS works as expected now.

### DIFF
--- a/packages/model-viewer/src/template.ts
+++ b/packages/model-viewer/src/template.ts
@@ -171,30 +171,16 @@ canvas.show {
   left: 0;
   width: 100%;
   height: var(--progress-bar-height, 5px);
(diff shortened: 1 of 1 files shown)


maciekglowka:
It's hard to say definitely by only looking at the code, but if we have to use only one CSS selector than for sure it is the way to go.
I am not bothered either by the opacity fade instead of sliding up. (and I actually use much thicker bar)

PR #2424, opened 2021-05-25
removed bad example path

elalish:
Fixes the AR Custom Button example on iOS. Apparently "needs a newer version of iOS" also means "couldn't find that URL".


packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -187,12 +187,12 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
...
     // Clones the roughnessMap if it exists.
     let roughnessMap: Texture|null = null;
-    if (material.roughnessMap) {
+    if (material.roughnessMap != null) {
       roughnessMap = material.roughnessMap.clone();
     }
 
     // Assigns the roughnessMap to the cloned material and generates mipmaps.
-    if (roughnessMap) {
+    if (roughnessMap != null) {
```
elalish, line 195: This is a total nit, but is what I was failing to communicate on your (@timmmeh) last PR review.

PR #2463, opened 2021-06-08
Fixes issue with framing a quantized GLTF

### LINKED ISSUE #2450: Bounding box for quantized models computed incorrectly?
### Description

In three.js r128 a fix was added to compute the correct bounding boxes for quantized models (https://github.com/mrdoob/three.js/pull/21554). That fix is working in my viewer now (https://gltf-viewer.donmccurdy.com/) but for some reason the latest version of model-viewer still does not center models using quantized position attributes.

The effect of the incorrect bounding box is that the model is not visible at all.

#### Live Demo

[export-v1.glb.zip](https://github.com/google/model-viewer/files/6585763/export-v1.glb.zip)

Compare:

- https://modelviewer.dev/editor/ (nothing shown)
- https://gltf-viewer.donmccurdy.com/ (correct)
- https://sandbox.babylonjs.com/ (correct)

### Version
- model-viewer: v1.7.2

ghost:
Resolve framing/rendering issue with quantized models.

fixes #2450

### DIFF
--- a/packages/model-viewer/src/three-components/ModelUtils.ts
+++ b/packages/model-viewer/src/three-components/ModelUtils.ts
@@ -85,9 +85,7 @@ export const reduceVertices = <T>(
 
               for (i = 0, l = position.count; i < l; i++) {
                 vertex.fromBufferAttribute(position, i);
-                if (position.normalized) {
-                  vertex.multiplyScalar(scale);
-                }
+                vertex.multiplyScalar(scale);
                 vertex.applyMatrix4(object.matrixWorld);
 
                 value = func(value, vertex);


packages/model-viewer/src/three-components/ModelUtils.ts
```diff
@@ -49,17 +74,21 @@ export const reduceVertices = <T>(
...
             const {position} = geometry.attributes;
 
             if (position !== undefined) {
+              const scale = getNormalizedComponentScale(position);
+
               for (i = 0, l = position.count; i < l; i++) {
-                vertex.fromBufferAttribute(position, i)
-                    .applyMatrix4(object.matrixWorld);
+                vertex.fromBufferAttribute(position, i);
+                if (position.normalized) {
```
elalish, line 88: Is this check necessary? Won't scale always be 1 if position is not normalized? Or could we make it that way?

timmmeh: This was cribbed from Three's update but we return 1 to cover the Int32 case, Three seems to ignore this case but we are a bit more strict because of typescript. Also this depends on the flag being set _and_ the buffer type being Int16 (for instance) but it _should_ be common that you have a Int16 buffer that is _not_ normalized so we definitely need the check somewhere.

elalish (approved):
Thanks!

timmmeh:
No problem, I can add this check. Do you per-chance know of a model that
falls into this category of having UInt8Array of position data (for
instance) but is not normalized?

donmccurdy:
Sure, gltfpack produces models like that — here's an aggressively quantized duck for example:

[DuckQuantized.glb.zip](https://github.com/google/model-viewer/files/6665320/DuckQuantized.glb.zip)

PR #2469, opened 2021-06-12
Remove editing adapter

elalish:
The goal is to remove all state from Redux that is contained in an exported glTF. The idea is that the Redux state will contain only the snippet setup for model-viewer, and between that and the exported files, we can transfer all necessary state to the mobile view. This will mean that we won't be able to enable undo functionality for material edits, but I'm okay with that, especially considering we don't have undo at all right now. 

20,000 of those lost LOC are just from the stupid package-lock files, but I'm still feeling pretty good about a net loss of ~6,000 lines of actual TS.


packages/space-opera/src/components/utils/create_object_url.ts
```diff
@@ -0,0 +1,81 @@
+/**
```
chrismgeorge, line 1: I assume these files are just copied over from the editing adapter?

packages/space-opera/src/components/animation_controls/animation_controls.ts
```diff
@@ -42,10 +37,10 @@ interface AnimationControlsInterface {
 export class AnimationControls extends ConnectedLitElement {
   @query('me-checkbox#animation-autoplay') autoplayCheckbox?: CheckboxElement;
   @internalProperty() animationNames: string[] = [];
-  @internalProperty() config: AnimationControlsInterface = {};
+  @internalProperty() config: ModelViewerConfig = {};
 
   stateChanged(state: State) {
-    this.animationNames = getAnimationNames(state);
+    this.animationNames = getModelViewer()?.availableAnimations ?? [];
     this.config = getConfig(state);
```
chrismgeorge, line 44: This may be leftover, but if the only thing this.config is used for in this component is animationName, you'd probably want to have a getAnimationName(state) method that returns this.config.animationName

Using getConfig() will cause you to re-render this component every time state.config changes, whereas you only want this component to re-render every time this.config.animationName changes (or any other values for this component)

chrismgeorge (commented):
You said this is just the WIP? What other changes are you thinking about making?

In terms of removing state, I think you did everything you'd need to. Replacing the getters, updating the combineReducer and types. 

Something to look out for is if the updates to the model you are replacing are tied to the redux state or the model currently. I.e. if we're adding more stuff to the config and you update the substate that is updated many things might get re-rendered, which could hurt performance. It probably won't be an issue, but it's just something to look out for.

packages/space-opera/src/components/utils/create_object_url.ts
elalish: Yes indeed. I think I have most of what I need to keep now.

packages/space-opera/src/components/animation_controls/animation_controls.ts
elalish: Ah, that's a very helpful insight! Will do.

packages/space-opera/src/components/animation_controls/animation_controls.ts
```diff
@@ -42,10 +37,10 @@ interface AnimationControlsInterface {
 export class AnimationControls extends ConnectedLitElement {
   @query('me-checkbox#animation-autoplay') autoplayCheckbox?: CheckboxElement;
   @internalProperty() animationNames: string[] = [];
-  @internalProperty() config: AnimationControlsInterface = {};
+  @internalProperty() config: ModelViewerConfig = {};
 
   stateChanged(state: State) {
-    this.animationNames = getAnimationNames(state);
+    this.animationNames = getModelViewer()?.availableAnimations ?? [];
```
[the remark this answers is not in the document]
elalish: So, regarding what else I am going do to, there's actually a lot. I'm going to remove editing-adapter completely, along with all state related to the glTF. My basic thought is that the snippet state lives in Redux and in combination with the glTF model we push over the wire is enough to reconstitute everything for the mobile view. I'm going to throw out any hope for doing incremental undos to material edits and just let model-viewer handle the state for all that itself. The material state will be serialized by `exportScene()` rather than Redux. I'm going to write a little design doc for this.

Coming back to your question above, remove animation will be another glTF edit, so possible, but not tracked by Redux or directly undoable (I'm going to stick with the Revert feature we have now instead).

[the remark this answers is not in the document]
chrismgeorge: Gotcha, will you be keeping the "original" materials in redux to do an overall revert back to?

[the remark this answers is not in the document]
elalish: I'll believe they're already buried in MV somewhere, but I may need to add a hook to access them. They're read-only (for a given model-load), so probably no need to include them in the state.

packages/space-opera/src/components/model_viewer_preview/types.ts
```diff
@@ -41,3 +42,13 @@ export interface GltfState {
   gltf?: GltfModel;
   gltfJsonString: string;
 }
+
+export const INITIAL_MODEL_STATE: ModelState = {
+  thumbnailsById: new Map<string, string>(),
```
chrismgeorge, line 47: I notice the previous texturesById was a Map<string, Texture> , why is this one a Map<string, string>?

chrismgeorge: Nvm, read texturesById as thumbnailsById

packages/space-opera/src/components/inspector/inspector.ts
```diff
@@ -19,71 +19,35 @@
...
-    await this.requestUpdate();
-  }
-
-  protected updated(changedProperties: PropertyValues) {
-    if (changedProperties.has($texturesById)) {
-      // Only call on change. Otherwise we could infinite-async-loop.
-      this.updateTexturesComplete = this.updateTextures(this[$texturesById]);
-    }
+    this.thumbnails = [...state.entities.model.thumbnailsById.values()];
+    this.gltfJsonstring = state.entities.model.originalGltfJson;
```
chrismgeorge, line 37: I always like to use getters here (it is obviously just a style choice though)

packages/space-opera/src/components/animation_controls/animation_controls.ts
elalish: Fixed.

packages/space-opera/src/components/model_viewer_preview/reducer.ts
```diff
@@ -91,6 +128,12 @@ export function dispatchGltfJsonString(gltfJsonString?: string) {
   return {type: SET_GLTF_JSON_STRING, payload: gltfJsonString};
 }
 
+const SET_THUMBNAILS = 'SET_THUMBNAILS'
+export async function dispatchThumbnails() {
+  const thumbnailsById = await createThumbnails();
```
chrismgeorge, line 133: Are the thumbnails ever updated during an editor session? Or is this just called once per model?

elalish: This function is per model load, but I'll be adding a function to append a newly-opened texture.

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -75,16 +60,7 @@ export class Image extends ThreeDOMElement implements ImageInterface {
     });
 
     for (const texture of this[$threeTextures]) {
-      // If the URI is set to null but the Image had an associated buffer view
-      // (this would happen if it started out as embedded), then fall back to
-      // the cached object URL created by GLTFLoader:
-      if (image == null &&
-          (this[$sourceObject] as GLTFImage).bufferView != null) {
-        texture.image = this[$bufferViewImages].get(texture);
```
elalish, line 81: @timmmeh As I've been going through this API, I realized that this complexity was confusing and uncalled-for. We never even specified that the input could be null, and it makes very little sense to go back to an original image, but only in certain circumstances. Anyway, hopefully this will simplify things for you as well.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -65,6 +66,18 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+      gltfMaterial.doubleSided = false;
+    }
+
+    if (gltfMaterial.alphaMode == null) {
+      gltfMaterial.alphaMode = 'OPAQUE';
+    }
+
+    if (gltfMaterial.alphaCutoff == null) {
+      gltfMaterial.alphaCutoff = 0.5;
+    }
```
elalish, line 79: @timmmeh I should have noticed this in code review; we fill in all the defaults at load time, which simplifies logic down the road. I may end up creating a new set of types to represent this, as I'm tired of all the `!`s.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -156,20 +169,31 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     this[$onUpdate]();
   }
 
-  setAlphaCutoff(cutoff: number): void {
-    cutoff = Math.min(1.0, cutoff);
+  [$applyAlphaCutoff]() {
```
elalish, line 172: @timmmeh I found some bugs as I was using these APIs in the editor. My fixes are here, so please give them a solid review.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
```diff
@@ -142,7 +143,10 @@ suite('scene-graph/material', () => {
       await loadModel(HELMET_GLB_PATH);
       // Disabled cutoff value hack see:
       // https://github.com/google/model-viewer/blob/e3a000111980f5cf018c61dec8695463e0d843a0/packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts#L210
-      expect(element.model!.materials[0].getAlphaCutoff())
+      expect(element.model!.materials[0]![$correlatedObjects]
+                 ?.values()
+                 .next()
+                 .value.alphaTest)
```
elalish, line 149: @timmmeh I changed this test because I only wanted the -0.5 in three.js and not reflected in the glTF.

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -47,7 +47,7 @@ export class Image extends ThreeDOMElement implements ImageInterface {
       gltfImage: GLTFImage|null) {
     gltfImage = gltfImage ?? {
       name: 'adhoc_image',
-      uri: (texture && texture.image) ? texture.image.uri : 'adhoc_image'
+      uri: (texture && texture.image) ? texture.image.src : 'adhoc_image'
```
elalish, line 50: @timmmeh Another small bug I found (TS should have caught this, but it apparently thinks `image` is `any`.

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -27,40 +27,39 @@ import '@polymer/paper-item';
...
 import '@material/mwc-icon-button';
 
-import {ALPHA_BLEND_MODES, checkFinite, DEFAULT_EMISSIVE_FACTOR} from '@google/model-viewer-editing-adapter/lib/main.js'
-import {createSafeObjectUrlFromUnsafe, SafeObjectUrl} from '@google/model-viewer-editing-adapter/lib/util/create_object_url.js'
 import {RGB, RGBA} from '@google/model-viewer/lib/model-viewer';
-import {customElement, html, internalProperty, property, query} from 'lit-element';
+import {customElement, html, internalProperty, query} from 'lit-element';
 import * as color from 'ts-closure-library/lib/color/color';  // from //third_party/javascript/closure/color
 
+import {TextureInfo} from '../../../../model-viewer/lib/features/scene-graph/texture-info.js';
```
elalish, line 34: @timmmeh Probably not worth looking at the diff since I pretty much rewrote this file. Still, it'd be great if you can look over this whole file a bit, since this is where my entire usage of your API lives.

elalish:
FYI @asemanh Here is this PR for removing editing-adapter and replacing it with @timmmeh's new APIs.

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -295,358 +309,257 @@ export class MaterialPanel extends ConnectedLitElement {
...
-    reduxStore.dispatch(
-        dispatchSetAlphaCutoff(getEditsMaterials(reduxStore.getState()), {
-          id: this.selectedMaterialId,
-          alphaCutoff: this.selectedAlphaCutoff,
-        }));
+    this.getMaterial().setAlphaCutoff(this.selectedAlphaCutoff);
+    reduxStore.dispatch(dispatchModelDirty());
   }
 
   revertMetallicRoughnessTexture() {
```
timmmeh, line 479: All these revert functions look awful similar, seems like a good candidate for refactoring into one helper method with a parameter to indicating the texture to revert. Then we could delete a whole lot more code.

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -295,358 +309,257 @@ export class MaterialPanel extends ConnectedLitElement {
...
 
   onMetallicRoughnessTextureChange() {
-    if (this.selectedMaterialId === undefined) {
-      throw new Error('No material selected');
-    }
-    const id = this.selectedMaterialId;
-    const textureId = this.selectedMetallicRoughnessTextureId;
-    reduxStore.dispatch(dispatchMetallicRoughnessTexture(
-        getEditsMaterials(reduxStore.getState()), {id, textureId}));
+    this.onTextureChange(
```
timmmeh, line 420: This change is nice, is there a way to get rid of all the bespoke OnThisTextureChange() methods entirely and just use this instead?

timmmeh (commented):
Lots of changes here, looked over materials_panel.ts - some high level comments, but I have some overall thoughts on the editor as I've been exploring this in my side projects.

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -69,80 +68,83 @@ export class MaterialPanel extends ConnectedLitElement {
...
+  }
+
+  rgbToHex(rgba: RGBA|RGB): string {
+    const selectedColorRgb =
+        rgba.slice(0, 3).map((color: number) => Math.round(color * 255));
+    return color.rgbArrayToHex(selectedColorRgb);
   }
 
   /* Interpolate base color as curr approaches duration */
   getInterpolatedColor(original: RGBA, curr: number, duration: number): RGBA {
```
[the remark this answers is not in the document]
elalish: It just flashes the material so you can see which one you just selected. Probably not worth optimizing the visual too much.

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -69,80 +68,83 @@ export class MaterialPanel extends ConnectedLitElement {
...
     const INTERP_COLOR = [0, 0, 0];
     // determine how much of interp color to use
-    const interpRatio = (duration - curr) / duration;
+    const interpRatio = Math.max(0, Math.min(1, (duration - curr) / duration));
```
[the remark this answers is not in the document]
elalish: Honestly, we probably have those functions kicking around somewhere already; I'll take a look.

packages/space-opera/src/components/materials_panel/materials_panel.ts
elalish: Yeah, I was thinking the same thing. I'll give it a shot.

elalish: Hmm, it looks like because of the way the HTML is interpreted, that's not so easy. Probably why they did it this way in the first place.

packages/space-opera/src/components/materials_panel/materials_panel.ts
elalish: I think I'm going to leave this for now in the interest of getting this merged. There's plenty more cleanup to do in space-opera when someone is excited about it.

packages/space-opera/src/components/materials_panel/materials_panel.ts
timmmeh: Yeah I thought maybe something like that, overall this is much better

PR #2471, opened 2021-06-13
Add WebXR Light estimation support

### LINKED ISSUE #2479: AR Environment Estimation Map
### Description

Follow up on #167.
This has landed 🎉 in Chrome so it would be great to get support for choosing if WebXR should use the pre-set env map or the estimated user environment.

Demo / Feature:
https://www.chromestatus.com/feature/5704707957850112

Applies to WebXR only

takahirox:
Fixes #2479

This PR adds [WebXR AR Light estimation](https://immersive-web.github.io/lighting-estimation/) support.


hybridherbst:
Question after reading the code: it looks like this isn't configurable. I think there should be an option on model-viewer to either keep or replace the environment (e.g. an additional "xr-environment" attribute). It's not _always_ desired to have the real-world environment affect the AR scene as that can change the perception of brand colors etc.

maciekglowka:
> 
> 
> Question after reading the code: it looks like this isn't configurable. I think there should be an option on model-viewer to either keep or replace the environment (e.g. an additional "xr-environment" attribute). It's not _always_ desired to have the real-world environment affect the AR scene as that can change the perception of brand colors etc.

I would like to support the idea of light estimation being an option through a flag setting or smth.
I tried the estimation in scene viewer and it is not always producing pleasing effects, so often it is safer to set a neutral lighting or neutral hdri.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -122,6 +125,22 @@ export class ARRenderer extends EventDispatcher {
...
+
+    this.xrLight = new XREstimatedLight(this.threeRenderer);
+
+    this.xrLight.addEventListener('estimationstart', () => {
+      if (!this.isPresenting) {
+        return;
+      }
+
+      const scene = this.presentedScene!;
+      scene.add(this.xrLight!);
```
elalish, line 137: nit: can we put the xrLight null check up with the return statement so we can get rid of all these bangs?

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -122,6 +125,22 @@ export class ARRenderer extends EventDispatcher {
...
+
+      if (this.xrLight!.environment) {
+        this.oldEnvironment = scene.environment;
+        scene.environment = this.xrLight!.environment;
```
elalish, line 141: Have you tested this with a shiny object? A big complaint we got in Scene Viewer is that shiny objects look matte because the ARCore environment image is so low-res (16x16 I believe) that it cannot produce a sharp reflection. I was thinking it would be better to either ignore its environment entirely or generate an LDR version of our PMREM and modulate it with the estimated environment (so that the high res one does most of the reflections while the low-res one takes care of the HDR lighting).

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -312,6 +331,14 @@ export class ARRenderer extends EventDispatcher {
     if (scene != null) {
       const {element} = scene;
 
+      if (this.xrLight!.parent != null) {
```
elalish, line 334: if you put `this.xrLight != null &&`, then you won't need the bangs.

elalish (commented):
Thanks for looking at this; it's been on my ToDo list for months now.

elalish:
I agree with the other commenters that this should be configurable, because ARCore's lighting estimation is often a disaster. I agree a boolean `xr-environment` attribute that defaults to false so as not to be a breaking change is a good way to go.

hybridherbst:
One advantage of this being a separate attribute would be that there could be (future) improvements and new modes, e.g. "xr-environment: none", "xr-environment: estimated" (?), "xr-environment: mixed" to combine low-res env + high-res sharp reflections from another (the existing?) env map (good idea!).

packages/model-viewer/src/three-components/ARRenderer.ts
takahirox: Thanks. I'm not familiar with TypeScript yet. Your comment is very helpful. Updated.

takahirox:
Thanks for the comments.

Regarding low resolution environment image (on ARCore), I didn't know that, interesting. I only tested with [the space suit model](space suit model). Maybe we need to test more to determine the best solution.

I added boolean `xr-environment` attribute (default is `false`) so far. If it's `false` estimated environment will be ignored. I think it's an easy and good start. Feel free to make follow up PRs for the "mixed" (or other) approach.

I'm not sure the boolean `xr-environment` attribute will be the final solution, so I added `xr-environment` attribute to only the top page example of `modelviewer.dev` and I didn't update the document yet. We can update the document when we determine the solution.

packages/model-viewer/src/features/ar.ts
```diff
@@ -71,7 +72,7 @@ export declare interface ARInterface {
   arModes: string;
   arScale: string;
   iosSrc: string|null;
-  readonly canActivateAR: boolean;
+  xrEnvironment: boolean, readonly canActivateAR: boolean;
```
elalish, line 75: I think you're looking for a `;` here instead of a `,`

packages/model-viewer/src/features/ar.ts
```diff
@@ -106,6 +110,7 @@ export const ARMixin = <T extends Constructor<ModelViewerElementBase>>(
     protected[$arModes]: Set<ARMode> = new Set();
     protected[$arMode]: ARMode = ARMode.NONE;
     protected[$preload] = false;
+    private[$xrEnvironment] = false;
```
elalish, line 113: No need for this; the `@property` above automatically creates a `this.xrEnvironment` member.

packages/modelviewer.dev/index.html
```diff
@@ -56,7 +56,7 @@ <h3 class="grouping-title grouping-title-new quick-start">Quick Start</h3>
 <script type="module" src="https://unpkg.com/@google/model-viewer/dist/model-viewer.js"></script>
 
 <!-- Use it like any other HTML element -->
-<model-viewer src="shared-assets/models/Astronaut.glb" alt="A 3D model of an astronaut" ar ar-modes="webxr scene-viewer quick-look" environment-image="neutral" auto-rotate camera-controls></model-viewer>
+<model-viewer src="shared-assets/models/Astronaut.glb" alt="A 3D model of an astronaut" ar ar-modes="webxr scene-viewer quick-look" environment-image="neutral" xr-environment auto-rotate camera-controls></model-viewer>
```
elalish, line 59: Let's put this in a slightly less prominent location. How about here: https://modelviewer.dev/examples/augmentedreality/#ar

elalish (changes requested):
Almost there. Agreed we can work on more advanced combinations as a follow-on, since this is opt-in.

packages/model-viewer/src/features/ar.ts
takahirox: Updated, thanks!

packages/model-viewer/src/features/ar.ts
takahirox: Updated, thanks!

packages/modelviewer.dev/index.html
takahirox: Sounds good to me. Updated, thanks!

hybridherbst:
One question to those possible future improvements, if xrEnvironment starts as boolean, what will happen once/if it becomes an enum later? Will there be a graceful transition (e.g. xrEnvironment is then equivalent to xrEnvironment="estimated") or will this be a breaking change?

elalish (approved):
Thanks! Regarding @hybridherbst, I'm not convinced this will become more than a boolean attribute. I consider the current deficiencies of xr-environment to be bugs, so I won't be shy about changing the behavior if I can improve it. I don't want to complicate our API with a bunch of options that are hard to explain.

PR #2483, opened 2021-06-16
adds buffer normalization check

timmmeh:
Adds buffer normalization check before de-quantizing

follow up to #2450


elalish (approved):
Thanks!

Issue #1621, opened 2020-10-09
Javascript API to create textures


civa86:
**Description**
I load a gltf model into the viewer with a material that has only a `baseColorFactor` defined inside its `pbrMetallicRoughness` properties

**Problem**
I want to change material properties in order to load a `baseColorTexture`.
According with the [examples](https://modelviewer.dev/examples/scene-graph.html) there is an API to handle this:

```javascript
material.pbrMetallicRoughness.baseColorTexture.texture.source.setURI
```
However, In my scenario, `baseColorTexture` is null so I'll get an error if I try to manipulate its source. 
Is it possibile to create textures with javascript code? Is there some API that I'm missing?

elalish:
This API is pretty new and is definitely missing some key features. I believe we expose the constructors so that you can call `material.pbrMetallicRoughness.baseColorTexture = new modelviewer.Texture();`, though I won't claim it's well-tested. You might also try doing this in our [editor](https://modelviewer.dev/editor/) which makes use of this API. Keep in mind that you won't be able to apply a texture unless you already have UV coordinates, which is beyond our ability to generate.

civa86:
I can't see the Texture constructor exposed inside the modelViewer instance that I can retrieve with the `document.querySelector`.
I will investigate the editor more and I will try to figure out how the example manages this part.
No problem for the UV, I know they must be in place in the gltf model and it's not part of the viewer taking care of them.

adamcal:
@civa86 did you get this API working? We're having the same issue. Thanks!

civa86:
@adamcal Not yet! 
I have to say that I was just making some test on this new component while we are using other realtime viewers for our real projects right now.

I didn't found the constructor I need anywhere, did you look for the Texture constructor too? It doesn't seem to be exposed.

Eventually I didn't have some time to investigate on the [editor](https://modelviewer.dev/editor/) example to try to figure out how they managed the same scenarios. Did you take a look on that?

adamcal:
@civa86 

I did indeed look at the editor. It can swap out the textures perfectly on the models we're using but I'm honestly not skilled enough in JS to figure out how it works. 

I'll try to dig into it more but please let me know if you ever find anything.

Thanks!

JamesBotterill:
I'm looking for the same thing, was hoping there'd be more about it in the documentation update but sadly not.

niwasmala:
@elalish when I try to set texture like

[...]

PR #2505, opened 2021-06-23
Adds empty images if the loaded glTF does not implement them.

timmmeh:
For instance TexturedBox has no normal map, the code would create an 'empty normal map' this would allow the texture to be added at runtime.

WIP an empty material.aoMap causes objects to be rendered black. Next commit will add a 'texture setter' that will update the texture material but leave it null otherwise, this should avoid the rendering issues in three.

### Reference Issue

Addresses #1621


packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -64,14 +88,26 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
 
       if (normalTexture != null && normalMap != null) {
         normalTextures.add(normalMap);
+      } else {
+        normalTexture = new EmptyNormalTexture();
+        material.normalMap = new ThreeTexture();
```
elalish, line 93: I don't think you want to do this. We want to create a new texture wrapper, not an actual empty Three.js texture. That could have all sorts of strange side effects, especially for performance (and potentially running out of samplers, if we take this too far). We should add the logic to `setURI()` that will create a texture if one is not present on the three.js material yet. That probably means you'll need to store a reference to the material and not just the correlated textures.

timmmeh: Update the PR as per our conversation, externally the API is nice to work with. Internally I'm not so happy with the weird contortions and indirections, it works though, and it does highlight (painfully so) some areas to address in later scene-graph updates.

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -87,4 +110,12 @@ export class Image extends ThreeDOMElement implements ImageInterface {
     }
     this[$onUpdate]();
   }
+
+  applyTexture(applicator: ((texture: ThreeTexture) => void)|undefined) {
```
elalish, line 114: Can we make this private? Meaning naming it with a symbol.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -64,31 +88,61 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
 
       if (normalTexture != null && normalMap != null) {
         normalTextures.add(normalMap);
+      } else {
+        normalTexture = new EmptyNormalTexture();
```
elalish, line 92: Can you just write `normalTexture = {index: -1};`? And remove all the `EmptyTexture`s?

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -15,17 +15,40 @@
...
 
+
+
 const loader = new ImageLoader();
 
 const $threeTextures = Symbol('threeTextures');
 const $uri = Symbol('uri');
 const $bufferViewImages = Symbol('bufferViewImages');
 
+export class GLTFImageDefinition implements ExternalImage {
```
elalish, line 31: Again, I think you can remove all of this and just construct an object directly like `{name: 'null_image', uri: 'null_image'}`.

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -88,6 +93,25 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+        const texture = await new TextureLoader().load(uri);
+        const textures = new Set<Texture>();
+        const textureDef = new GLTFTextureDefinition();
+        const imageDef = new GLTFImageDefinition(uri);
+        textures.add(texture);
+
+        const {gltf} = currentGLTF.correlatedSceneGraph;
+
+        return new ModelViewerTexture(
+            () => {}, gltf, textureDef, textures, imageDef);
```
elalish, line 108: I think we need our `onUpdate()` function here or else if any setting is changed on a descendent of this texture (say a sampler) it won't cause our render to redraw.

elalish (changes requested):
Thanks, looking good! Let's update the example in this PR so we can test that it's working. A few basic automated tests would be nice too.

packages/model-viewer/src/features/scene-graph/image.ts
timmmeh: Unfortunately I need to call it from Texture-Info, the overall architecture here is fighting against keeping this internal and believe me I tried. There is no concept of a friend class in typescript but I should be able to mimic it and thus hide this method, but it will just add more code. IMO with a little re-design we can have our cake and eat it too, but that's outside of this CL.

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: Oh, looks like you can, learning what typescript lets you get away with is a WIP

packages/model-viewer/src/features/scene-graph.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/image.ts
timmmeh: Looks like it's possible, but typescript question, is that good practice long term though? Seems like well defined types could provide some safety and readability.

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -100,20 +101,32 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
       return onUpdate
     }
 
+    // Load texture helper waits for the texture and image to finish loading
+    // before returning.
+    async[$loadTexture](uri: string) {
+      let result: Texture|null = null;
+      await new Promise(resolve => {
+        return new TextureLoader().load(uri, texture => {
+          resolve(texture);
```
elalish, line 110: This is very close to what `TextureUtils.load()` does. Think you can use that instead, perhaps with some light modification?

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -100,20 +101,32 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+        });
+      }).then(completedTexture => {
```
elalish, line 112: `await` is the modern version of `.then()`. There should be no need to use both, so we don't have `.then()` in our code base.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -28,6 +28,58 @@ const $pbrMetallicRoughness = Symbol('pbrMetallicRoughness');
...
+      onUpdate: () => void,
+      correlatedMaterials: Set<MeshStandardMaterial>|undefined,
+      usage: TextureUsage, encodding: TextureEncoding, format: PixelFormat) {
+    return (texture: ThreeTexture) => {
+      if (correlatedMaterials) {
+        for (const material of correlatedMaterials) {
+          texture.encoding = encodding;
+          texture.format = format;
+          texture.wrapS = RepeatWrapping;
+          texture.wrapT = RepeatWrapping;
```
elalish, line 56: Okay, so this is our default values for a new texture. That's fine, but it seems odd that this creation all happens during `setTexture()`; shouldn't it be in `createTexture()` instead? Also, when creating a texture, you'll also need to create its children, like `sampler` which must have corresponding defaults. Is that all getting built consistently?

packages/model-viewer/src/features/scene-graph.ts
elalish: It'll save you from creating and destroying `TextureLoader`s all the time, which I think is discouraged.

timmmeh: Ok updated to do something similar.

packages/model-viewer/src/features/scene-graph.ts
timmmeh: Done, this worked
'''
await new Promise<Texture>(
          (resolve) => this[$textureLoader].load(uri, resolve));
'''
 I was hitting an issue where texture.img was undefined, not fully loaded

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: Samplers are getting built - default samplers, I didn't have to change anything at least to make it work.
As for the other issue, these are not the default values these are values that get set/changed depending on the type of texture you want to set, thats why it's happing in setTexture. I can remove the wrap settings, I think those might be default values, left over from testing. Of-course we might want to provide some sort of override setting so you can specify your own settings, but these are following the glTF spec.

timmmeh: Turns out clamp is the Three default, so I do need to keep repeat, repeat is the glTF default.

timmmeh: moved setting the wrap and flipY to createTexture, and added a default encoding and format.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -64,31 +114,61 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+        gltf,
+        normalTexture!,
+        normalTextures,
+        // Applicator provides method for applying a texture to a material.
+        TextureApplicator[$provideApplicator](
+            onUpdate,
+            correlatedMaterials,
+            TextureUsage.Normal,
+            LinearEncoding,
+            RGBFormat));
```
elalish, line 145: Considering the encoding default is directly related to the texture usage, which is already specified here, how about we remove this input and handle the logic directly inside the applicator? Also, the occlusion and MR textures are usually the same (occlusion simply uses the red channel). I believe you have to set the format based on the incoming buffer, which I think the loader already does.

packages/model-viewer/src/features/scene-graph/texture.ts
```diff
@@ -45,12 +50,17 @@ export class Texture extends ThreeDOMElement implements TextureInterface {
...
     this[$sampler] = new Sampler(onUpdate, sampler, correlatedTextures);
 
-    if (gltf.images != null && imageIndex != null) {
+    if (gltf.images != null && imageIndex != null && imageIndex !== -1) {
       const image = gltf.images[imageIndex];
 
       if (image != null) {
         this[$source] = new Image(onUpdate, image, correlatedTextures);
       }
+    } else if (imageIndex === -1) {
```
elalish, line 59: If you move this block ahead, you can just put an early return and skip and extra check.

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -88,6 +95,38 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
     static Texture: Constructor<Texture>;
     static Image: Constructor<Image>;
 
+    private[$provideOnUpdate](): () => void {
```
elalish, line 98: Why do we need to wrap this function in a provider? A function is already an object you can pass if you simply pass `onUpdate` instead of `onUpdate()`. Or is there some problem with the meaning of `this`?

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: I can move the encoding logic, no problem.
I see for the occlusion it's set to RGB after loading, I guess that's to be expected if it's sharing with the MR map.

packages/model-viewer/src/features/scene-graph.ts
timmmeh: So this was to prevent repeating the same code in different methods. Since I needed to create the `onUpdate` function the same way that was being done in the `updated` method I just pulled it out and wrapped to prevent it from having to be maintained in multiple places. I could just make it method that lives on `scene-graph` is that what you mean?

packages/model-viewer/src/features/scene-graph/texture.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph.ts
timmmeh: I just turned it into a private method, this seems to work fine, not seeing any problems with the `this` reference

elalish: Actually, looking at it now, I see even the `$onUpdate` wrapper is unnecessary; we can just pass `$needsRender` directly.

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -400,24 +400,85 @@ <h2 class="demo-title">Swap textures - Normals/Occlusion/Emission</h2>
...
+<script type="module">
+const modelViewerTexture = document.querySelector("model-viewer#duck");
+
+modelViewerTexture.addEventListener("load", () => {
+
+  let material = modelViewerTexture.model.materials[0];
+
+  let createAndApplyTexture = async (channel, event) => {
+    if(event.target.value == "None"){
+      // Clears the texture.
```
elalish, line 463: Nice, I look forward to trying this out!

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -88,6 +95,32 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+        return null;
+      }
+      // Applies default settings.
+      texture.encoding = sRGBEncoding;
+      texture.wrapS = RepeatWrapping;
+      texture.wrapT = RepeatWrapping;
+      texture.flipY = false;
+
+      const textureDef = {source: -1} as GLTFTexture;
+      const imageDef = {name: 'null_image', uri: 'null_image'} as GLTFImage;
```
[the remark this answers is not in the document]
timmmeh: Well intellisense couldn't figure out, so I added them for readability

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -13,38 +13,85 @@
...
     }
   }
 
-  get texture(): Texture {
+  get texture(): Texture|null {
     return this[$texture];
   }
+
+  setTexture(texture: Texture|null): void {
+    let threeTexture: ThreeTexture|null = null;
```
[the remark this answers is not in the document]
timmmeh: Done.

packages/modelviewer.dev/examples/scenegraph/index.html
timmmeh: :)

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -13,38 +13,85 @@
...
+    let encoding: TextureEncoding = sRGBEncoding;
+
+    // Assigns the passed in textures threeTexture to the material. Otherwise
+    // the texture is cleared by removing setting the material texture and
+    // the model-viewer texture to null.
+    if (texture) {
```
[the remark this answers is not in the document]
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -13,38 +13,85 @@
...
+      threeTexture = texture.source[$threeTexture];
+    } else {
+      // Removes texture from the texture-info.
+      this[$texture] = null;
```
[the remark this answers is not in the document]
timmmeh: whoops 🙃

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -15,51 +15,53 @@
...
+import {$gltf, $gltfImage, $gltfTexture, $threeTexture, TextureContext} from './material.js';
+import {$onUpdate, $sourceObject, ThreeDOMElement} from './three-dom-element.js';
 
 
 
 const loader = new ImageLoader();
 
-const $threeTextures = Symbol('threeTextures');
-export const $threeTexture = Symbol('threeTextures');
+// const $threeTextures = Symbol('threeTextures');
```
timmmeh, line 28: delete

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -91,17 +94,17 @@ export class Image extends ThreeDOMElement implements ImageInterface {
...
   }
 
-  [$applyTexture](applicator: ((texture: ThreeTexture) => void)|undefined) {
-    if (applicator) {
-      for (const texture of this[$threeTextures]) {
-        applicator(texture);
-      }
-    }
-  }
+  // [$applyTexture](applicator: ((texture: ThreeTexture) => void)|undefined) {
```
timmmeh, line 102: delete

packages/model-viewer/src/features/scene-graph/pbr-metallic-roughness.ts
```diff
@@ -82,31 +82,21 @@ export class PBRMetallicRoughness extends ThreeDOMElement implements
...
-            firstValue(correlatedMaterials),
-            firstValue(baseColorTextures),
-            TextureUsage.Occlusion,
-            sRGBEncoding,
-            RGBAFormat));
-
-    this[$metallicRoughnessTexture] = new TextureInfo(
+        firstValue(correlatedMaterials),
+        firstValue(baseColorTextures),
+        TextureUsage.Occlusion,
```
timmmeh, line 90: should be base

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -37,58 +133,87 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
-      material.pbrMetallicRoughness = {};
+    if (gltfMaterial.pbrMetallicRoughness == null) {
+      gltfMaterial.pbrMetallicRoughness = {};
     }
     this[$pbrMetallicRoughness] = new PBRMetallicRoughness(
-        onUpdate, gltf, material.pbrMetallicRoughness, correlatedMaterials);
+        onUpdate, gltf, gltfMaterial.pbrMetallicRoughness, correlatedMaterials);
 
-    const {normalTexture, occlusionTexture, emissiveTexture} = material;
+    let {normalTexture: normalTextureDef, occlusionTexture, emissiveTexture} =
```
elalish, line 160: Is this rename necessary?

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -13,20 +13,116 @@
...
+  [$gltf]: GLTF;
+  [$material]: MeshStandardMaterial|null;
+  [$threeTexture]: ThreeTexture|null = null;
+  [$gltfTextureInfo]: GLTFTextureInfo;
+  [$gltfTexture]: GLTFTexture;
+  [$gltfSampler]: GLTFSampler;
+  [$gltfImage]: GLTFImage;
+  [$usage]: TextureUsage;
+  [$encoding]: TextureEncoding;
+  [$format]: PixelFormat;
```
elalish, line 67: Is this used? I believe we aren't allowed to change format from what the file specifies.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -13,20 +13,116 @@
...
+ * a texture in one place.
+ */
+export class TextureContext {
+  [$gltf]: GLTF;
+  [$material]: MeshStandardMaterial|null;
+  [$threeTexture]: ThreeTexture|null = null;
+  [$gltfTextureInfo]: GLTFTextureInfo;
+  [$gltfTexture]: GLTFTexture;
+  [$gltfSampler]: GLTFSampler;
+  [$gltfImage]: GLTFImage;
```
elalish, line 64: A comment about why these are needed and how they're used could be nice here.

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: nope, done.

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: oh yeah, good catch

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: Resolved

packages/model-viewer/src/test/features/scene-graph/texture-spec.ts
```diff
@@ -0,0 +1,153 @@
...
+      expect(uuidFromTextureInfoObject).to.be.equal(newUUID);
+
+      const uuidFromTextureObject: string =
+          (element.model!.materials[0]
+               .pbrMetallicRoughness.baseColorTexture?.[$correlatedObjects]
+               ?.values()
+               .next()
+               .value as ThreeTexture)
+              .uuid;
+      expect(uuidFromTextureObject).to.be.equal(newUUID);
```
elalish, line 101: I think this is identical to the one above?

packages/model-viewer/src/test/features/scene-graph/texture-spec.ts
```diff
@@ -0,0 +1,153 @@
...
+      const element = new ModelViewerElement();
+      element.src = ASTRONAUT_GLB_PATH;
+      document.body.insertBefore(element, document.body.firstChild);
+      await waitForEvent(element, 'load');
+
+      const texture = await element.createTexture(assetPath(
+          'models/glTF-Sample-Models/2.0/BoxTextured/glTF/CesiumLogoFlat.png'));
+
+      element.model!.materials[0]
+          .pbrMetallicRoughness.baseColorTexture!.setTexture(texture);
```
elalish, line 134: Looks like the code to here is shared in all these tests. Let's pack it into a `beforeEach()` method to reduce duplication.

elalish (commented):
The tests look good. I don't know that we need a whole matrix, but maybe pick one of these tests and copy it for each of the texture types to get some breadth of coverage?

elalish:
Also, you added Default_Abedo.jpg; are you using it for anything?

packages/model-viewer/src/test/features/scene-graph/texture-spec.ts
timmmeh: Done.

packages/model-viewer/src/test/features/scene-graph/texture-spec.ts
timmmeh: Done.

timmmeh: onBefore was not recognized, so I followed the pattern where it was being used elsewhere now getting things like:
```
Chrome Headless 91.0.4472.114 (Mac OS 10.15.7) ERROR
  Uncaught ReferenceError: describe is not defined
  at lib/test/features/scene-graph/texture-spec.js:22:1
 ```

Comment (timmmeh) [reply packages/model-viewer/src/test/features/scene-graph/texture-spec.ts:134]: resolved with a helper method for now.

Comment (elalish) [reply packages/model-viewer/src/test/features/scene-graph/texture-spec.ts]: Oh, sorry, I used the jasmine term instead of the chai term. Really wish we didn't have two different testing frameworks... You're looking for `setup` and `teardown`. If you look at some of the other MV tests you'll see examples. Remember to remove elements in `teardown` to keep the tests hermetic.

File: packages/model-viewer/src/test/features/scene-graph/material-spec.ts [hunk 25]
```diff
@@ -34,18 +34,23 @@ suite('scene-graph/material', () => {
...
       element.src = HELMET_GLB_PATH;
       document.body.insertBefore(element, document.body.firstChild);
       await waitForEvent(element, 'load');
 
       texture = await element.createTexture(REPLACEMENT_TEXTURE_PATH);
-    };
 
-    test('Set a new base map', async () => {
-      await init();
+      await timePasses();
```
elalish, line 45: Were the tests not passing without this extra await? Any idea why this would be needed?

elalish (approved):
Can you prod this to rerun the tests and see if they pass? Looks like it's just our flaky iOS 13 runner.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
timmmeh: I pulled this over from example code and never removed it, it seems to run fine without it.

elalish (dismissed):
Thanks!

elalish:
Oh, and can you switch it from Draft to a regular PR?

elalish (approved):

shamsail:
Hi @elalish , when we can expect this thing in a new release?  It will be helpful to assign the textures which value are Null already.

timmmeh:
This should be out in the next release

PR #2563, opened 2021-07-07
Three.js r130

elalish:
Also removes our hack to get transparent materials working properly now that three.js does this for us.


timmmeh (approved):
LGTM

PR #2566, opened 2021-07-09
Add createThumbnails

elalish:
Fixes #2531

I changed the scene-graph API a little to support the work I'm doing in the editor. It includes a method to get CPU images out of our GPU textures, which was just requested as a feature. I had some conflicts merging master today, so I'm pulling the model-viewer changes out of the editor branch into this so they can be merged and stop having conflicts.


packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -84,6 +86,10 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
       return this[$variants];
     }
 
+    get gltfJson() {
+      return this[$currentGLTF]?.correlatedSceneGraph.gltf;
```
elalish, line 90: I'm using this in the editor as well. Perhaps would be better to clone? As I'm looking at it, I'm wondering if someone could do something odd by editing this...

packages/model-viewer/src/features/scene-graph/api.ts
```diff
@@ -228,11 +228,23 @@ export declare interface Image {
    */
   readonly uri?: string;
 
+  /**
+   * The bufferView of the image, if it is embedded.
+   */
+  readonly bufferView?: number
```
elalish, line 234: This is so I can make a cache identifier; I should either have a uri or a bufferView, and either way it should be unique per image.

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -52,37 +51,64 @@ export class Image extends ThreeDOMElement implements ImageInterface {
...
     });
 
-    const texture = this[$textureInfo][$threeTexture];
-    if (texture) {
-      // If the URI is set to null but the Image had an associated buffer view
-      // (this would happen if it started out as embedded), then fall back to
-      // the cached object URL created by GLTFLoader:
-      if (image == null &&
-          (this[$sourceObject] as GLTFEmbeddedImage).bufferView != null) {
-        texture.image = this[$bufferViewImages].get(texture);
```
elalish, line 80: I removed the bufferViewImages because they don't really make any sense to me. It's like you can change the image and then *sometimes* you have a previous image to go back to? Nah, and the simplification will probably be necessary as we expand this material API.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -60,6 +60,10 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     this[$pbrMetallicRoughness] = new PBRMetallicRoughness(
         onUpdate, gltf, gltfMaterial.pbrMetallicRoughness, correlatedMaterials);
 
+    if (gltfMaterial.emissiveFactor == null) {
+      gltfMaterial.emissiveFactor = [0, 0, 0];
```
elalish, line 64: This was just an existing bug I noticed.

packages/model-viewer/src/features/scene-graph/pbr-metallic-roughness.ts
```diff
@@ -50,10 +50,10 @@ export class PBRMetallicRoughness extends ThreeDOMElement implements
       pbrMetallicRoughness.baseColorFactor = [1, 1, 1, 1];
     }
     if (pbrMetallicRoughness.roughnessFactor == null) {
-      pbrMetallicRoughness.roughnessFactor = 0;
+      pbrMetallicRoughness.roughnessFactor = 1;
     }
     if (pbrMetallicRoughness.metallicFactor == null) {
-      pbrMetallicRoughness.metallicFactor = 0;
+      pbrMetallicRoughness.metallicFactor = 1;
```
elalish, line 56: These were also existing bugs relative to the glTF spec.

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -146,15 +146,15 @@ export class TextureInfo extends ThreeDOMElement implements
...
+  get texture(): Texture {
     return this[$texture];
   }
 
   setTexture(texture: Texture|null): void {
     const threeTexture: ThreeTexture|null =
         texture != null ? texture.source[$underlyingTexture] : null;
     let encoding: TextureEncoding = sRGBEncoding;
-    this[$texture] = texture;
+    this[$texture] = texture ?? new Texture(this);
```
elalish, line 157: @timmmeh I added this because I wanted to make `.texture` always defined. Looking at the constructor, this seemed to be what you did when nothing was specified. Does this seem reasonable?

packages/model-viewer/src/three-components/gltf-instance/gltf-2.0.ts
```diff
@@ -188,23 +188,15 @@ export interface Node {
   extras?: Extras;
 }
 
-export interface ExternalImage {
```
elalish, line 191: Combining the two glTF image types because I didn't feel the type complexity was doing anything to help us.

packages/model-viewer/src/features/scene-graph/texture-info.ts
timmmeh: So I added nullness support so you could remove a texture, basically allows for an empty texture-info.  I think this change would break that, but you can check it with the "Create Textures" example you should be able to add a normal map to the duck and remove it.

packages/model-viewer/src/features/scene-graph.ts
timmmeh: yeah cloning wouldn't hurt, changing the gltf object directly could probably interfere with internal representation, although it would make sense to edit and then reload the scene with the changes, but that's a different can of worms.

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -52,37 +51,64 @@ export class Image extends ThreeDOMElement implements ImageInterface {
...
+    const camera = new OrthographicCamera(-1, 1, 1, -1, 0, 1);
+
+    const {threeRenderer} = Renderer.singleton;
+    const renderTarget = new WebGLRenderTarget(width, height);
+    threeRenderer.setRenderTarget(renderTarget);
+    threeRenderer.render(scene, camera);
+    threeRenderer.setRenderTarget(null);
+
+    const buffer = new Uint8Array(width * height * 4);
+    threeRenderer.readRenderTargetPixels(
```
timmmeh, line 96: Can you not just read pixels from the source buffer with glReadPixels(). It's been a while since I've done this in GL but in Vulkan you can take a texture in video memory and transition it to an RGBA uncompressed format and write it out, I would think readPixels() basically does that.

elalish: Honestly `canvas.toBlob()` is basically just a wrapper around `readPixels` anyway. The convenience it affords is that it'll resample to a new resolution for me automatically. If this were something that was on the performance critical path, I'd look into it more, but meh.

packages/model-viewer/src/features/scene-graph/texture-info.ts
elalish: Hmm, I guess I was hoping this would still null it out. I'll check. If it breaks that example, we should probably add a test for this.

elalish: Yeah, the example seems to work fine switching back to none. Can you review again and make sure you're happy with the logic here?

packages/model-viewer/src/features/scene-graph/image.ts
timmmeh: Yeah I guess I wasn't thinking about the fact that this is for thumbnails, this is the best way to take advantage of the hardware linear interpolation to quickly downsample.

packages/model-viewer/src/features/scene-graph/texture-info.ts
timmmeh: Is there a new commit for this? I don't see it yet.
So the other issue that could crop up here is on the API side, if the texture is not null or at least indicated to the user to be invalid than you could easily call setURI() (like in the Swap Textures example)
```
 let applyPBRTexture = (channel, event) => {
    material.pbrMetallicRoughness[channel].texture.source.setURI(event.target.value);
  }
```
If texture were null you could check use that to at least check the if the texture is valid, if not I think it would succeed but just not do anything, not 100% sure it might still produce an error somewhere, but the point is there is no check to see if it's actually a valid texture or not.

elalish: Interesting; so how is this different than how your constructor works when there is no texture in the glTF? I thought I was following that same logic as that, but I may be mistaken.

timmmeh: I may have not completed that idea thoroughly in the code, although that was my intent at one point and I think it makes sense for the API, taking a look...

timmmeh: So taking a closer look, your right the constructor is creating a valid texture, in the initial design of things this was how it was working but once we switched to the setTexture() API I was going to change this but forgot. I'd still like to do that though, so I'd say I can make that change now, I think it will be a small change

timmmeh: Ok fixed the API to create a null Texture object when the backing three-texture is null or undefined. Also added test to exercise this code.
https://github.com/google/model-viewer/pull/2568

elalish: Thanks!

elalish: Fixed.

timmmeh (approved):
lgtm

PR #2569, opened 2021-07-12
Added AttenuationTest to render-fidelity-tools tests

augmentedsky:
Added a test for AttenuationTest under render-fidelity-tools/tests


packages/render-fidelity-tools/test/renderers/filament/filament/docs/filamesh.md
```diff
@@ -0,0 +1,264 @@
+# Filamesh
```
timmmeh, line 1: I see a bunch of filament files here, just wondering why, is this intended?

augmentedsky: I didn't realize I shouldn't add those. Will revert and send a new iteration

PR #2570, opened 2021-07-12
remove ios13 testing while BrowserStack's test runner is down

elalish:
BrowserStack is investigating why iOS13 has gone from flaky to not working at all. Meanwhile, I'd like our automated tests to pass.


timmmeh (approved):
👍

PR #2571, opened 2021-07-13
Added AttenuationTest to render-fidelity-tools tests

augmentedsky:
Added AttenuationTest model and its golden files to render-fidelity-tools tests


elalish:
I just disabled the broken test runner. Please pull master and let's see if we can get these tests to pass.

packages/render-fidelity-tools/test/config.json
```diff
@@ -75,6 +75,23 @@
...
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/AttenuationTest/glTF/AttenuationTest.gltf",
+      "dimensions": {
+        "width": 512,
+        "height": 1024
+      },
+      "target": {
+        "y": 3.6
+      },
+      "orbit": {
+        "radius": 11.2
```
elalish, line 89: I would start with the default values (just removing these listings), then add back in the ones you need until the model is nicely framed. Only use custom dimensions if the model has a really odd aspect ratio.

elalish (changes requested):
Good start!

packages/render-fidelity-tools/test/config.json
augmentedsky: Sorry I just saw this. Let me update them in the next commit

augmentedsky: Just sent another commit let me know what you think

elalish (approved):
I'm not sure why Github's rich diff tool seems to be not working. Would be great if you could paste one of your goldens into a comment so we can see.

elalish:
@googlebot I signed it!

augmentedsky:
Here are the attenuationtest screenshots. How do they look?

babylon-golden
filament-golden
gltf-sample-viewer-golden
model-viewer-golden

timmmeh:
Based on the Don McCurdy viewer https://gltf-viewer.donmccurdy.com/ this looks reasonable

Issue #2419, opened 2021-05-24
TypeError: Cannot read property 'forEach' of undefined


zmnv:
Hello there. General Kenobi.

I created a PNG  Thumbnails Batch Renderer with your awesome project and sometimes it stops just because source model file has something like 'problems with materials' i think.. BUT. I can see the model.. 
Idk how to organize try catch with model-viewer just with vanilla js... sorry))

Can you handle this case without type error please? 

```
model-viewer.js:56233 Uncaught (in promise) TypeError: Cannot read property 'forEach' of undefined
    at new Sm (model-viewer.js:56233)
    at HTMLElement.[onModelLoad] (model-viewer.js:56342)
    at vc.<anonymous> (model-viewer.js:48752)
    at vc.dispatchEvent (model-viewer.js:1689)
    at vc.setupScene (model-viewer.js:44589)
    at vc.setSource (model-viewer.js:44577)
    at async HTMLElement.[updateSource] (model-viewer.js:49041)
    at async HTMLElement.[updateSource] (model-viewer.js:53334)
```

Screens:

![image](https://user-images.githubusercontent.com/25253501/119393144-3e0e9400-bcd9-11eb-9b46-3bd98bf4c166.png)

![image](https://user-images.githubusercontent.com/25253501/119392672-ba54a780-bcd8-11eb-815b-ed6639347aae.png)

From:

```
class Model {
    constructor(correlatedSceneGraph, onUpdate = () => { }) {
        this[_a] = [];
        const { gltf, gltfElementMap } = correlatedSceneGraph;
        gltf.materials.forEach(material => {
            this[$materials].push(new Material(onUpdate, gltf, material, gltfElementMap.get(material)));
        });
    }
```

source glb (14mb) (not mine)
https://cloudflare-ipfs.com/ipfs/QmeqKjz7KLigLis1AJGbuu22uXdgGdNaVneFF1MiDkoaPJ

from object 6137 at hic et nunc gallery.

I've checked source file and don't see any "materials" property.

elalish:
This looks like a regression; we have code specifically to handle this case where the glTF only has the default material.

zmnv:
@elalish Sorry for the question. Is any way to update `onerror` event to provide error handle in this situation:

> connect google model viewer as <script> html
> work with model viewer as `const viewer =  document.getElementById("viewer");`
> viewer.onerror 

or `onerror` event ONLY for http requests?

Now sometimes I get:
```
[...]

PR #2575, opened 2021-07-14
* Resolves extraneous materials being created.

timmmeh:
Default material was being created even in cases where an object node did have a material (such as a group node)

This is a follow up to #2419 - tested with the same asset to verify that this PR does not cause a regression with the case where the glTF only has the default material.

### DIFF
--- a/packages/model-viewer/src/test/three-components/gltf-instance/correlated-scene-graph-spec.ts
+++ b/packages/model-viewer/src/test/three-components/gltf-instance/correlated-scene-graph-spec.ts
@@ -21,6 +21,8 @@ import {CorrelatedSceneGraph} from '../../../three-components/gltf-instance/corr
 import {Material, PBRMetallicRoughness, Texture, TextureInfo} from '../../../three-components/gltf-instance/gltf-2.0.js';
 import {assetPath, loadThreeGLTF} from '../../helpers.js';
 
+
+
 const expect = chai.expect;
 
 const HORSE_GLB_PATH = assetPath('models/Horse.glb');
@@ -115,6 +117,23 @@ suite('correlated-scene-graph', () => {
               }
             });
         expect(name).to.be.eq('Default');
+      });
+
+      test('Only one default material after cloning', async () => {
+        const threeGLTF = await loadThreeGLTF(KHRONOS_TRIANGLE_GLB_PATH);
+        const correlatedSceneGraph = CorrelatedSceneGraph.from(threeGLTF);
+
+        const scene = SkeletonUtils.clone(threeGLTF.scene) as Group;
+        const scenes: Group[] = [scene];
+
+        const cloneThreeGLTF: GLTF = {...threeGLTF, scene, scenes};
+
+        const cloneCorrelatedSceneGraph =
+            CorrelatedSceneGraph.from(cloneThreeGLTF, correlatedSceneGraph);
+
+        expect(cloneCorrelatedSceneGraph.gltf.materials!.length).to.be.eq(1);
+        expect(cloneCorrelatedSceneGraph.gltf.materials![0].name)
+            .to.be.eq('Default');
       });
     });
   });
--- a/packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
+++ b/packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
@@ -135,7 +135,7 @@ export class CorrelatedSceneGraph {
             if (((object as Mesh).isMesh || (object as Material).isMaterial) &&
                 elementReference == null) {
               // Checks if default material was allready addded to the gltf.
-              if (cloneGLTF.materials) {
+              if (cloneGLTF.materials && cloneGLTF.materials.length) {
                 const material =
                     cloneGLTF.materials[cloneGLTF.materials.length - 1];
                 if (material.name === 'Default') {
@@ -151,9 +151,10 @@ export class CorrelatedSceneGraph {
                 defaultReference.index = cloneGLTF.materials.length;
                 cloneGLTF.materials.push(defaultMaterial);
               }
+
+              elementReference = defaultReference;
             }
 
-            elementReference = defaultReference;
 
             if (elementReference == null) {
               return;
--- a/packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
+++ b/packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
@@ -123,19 +123,37 @@ export class CorrelatedSceneGraph {
 
     const defaultMaterial = {name: 'Default'} as Material;
     const defaultReference = {type: 'materials', index: -1} as GLTFReference;
-    if (cloneGLTF.materials == null) {
-      cloneGLTF.materials = [];
-    }
-    defaultReference.index = cloneGLTF.materials.length;
-    cloneGLTF.materials.push(defaultMaterial);
 
     for (let i = 0; i < originalThreeGLTF.scenes.length; i++) {
       this[$parallelTraverseThreeScene](
           originalThreeGLTF.scenes[i],
           cloneThreeGLTF.scenes[i],
           (object: ThreeSceneObject, cloneObject: ThreeSceneObject) => {
-            const elementReference =
+            let elementReference =
                 upstreamCorrelatedSceneGraph.threeObjectMap.get(object);
+
+            if (((object as Mesh).isMesh || (object as Material).isMaterial) &&
+                elementReference == null) {
+              // Checks if default material was allready addded to the gltf.
+              if (cloneGLTF.materials) {
+                const material =
+                    cloneGLTF.materials[cloneGLTF.materials.length - 1];
+                if (material.name === 'Default') {
+                  defaultReference.index = cloneGLTF.materials.length - 1;
+                }
+              }
+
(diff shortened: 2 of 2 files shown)


timmmeh:
Found why there were extraneous empty materials and came up with a possible fix, it turns out we needed some more testing to narrow down the default-material cases to Meshes that are only using the default material. It seems maybe all scene objects where getting a default material element created.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -122,35 +149,46 @@ export class CorrelatedSceneGraph {
...
                 }
                 defaultReference.index = cloneGLTF.materials.length;
                 cloneGLTF.materials.push(defaultMaterial);
               }
-              elementReference = defaultReference;
             }
 
-            const {type, index} = elementReference;
-            const cloneElement = cloneGLTF[type]![index];
+            if (elementReference != null) {
```
elalish, line 180: How about an early return instead to keep the indenting down?

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -122,35 +149,46 @@ export class CorrelatedSceneGraph {
...
                 upstreamCorrelatedSceneGraph.threeObjectMap.get(object);
 
-            if (elementReference == null) {
+            const getMaterial = (mesh: Mesh) => {
+              return mesh.material as MeshStandardMaterial;
+            };
+            // Only sets up default material if the mesh references it.
+            if ((object as Mesh).isMesh &&
+                getMaterial(object as Mesh).uuid ===
+                    parserCache.defaultMaterialUUID) {
```
elalish, line 170: Aside: it's nice to avoid putting top-level comments in the PR because github doesn't let you reply to them. I prefer to pick even a random line of code to attach them to instead, so that they can become a thread.

elalish: > Found why there were extraneous empty materials and came up with a possible fix, it turns out we needed some more testing to narrow down the default-material cases to Meshes that are only using the default material. It seems maybe all scene objects where getting a default material element created.

I was just noticing this bug myself (and I introduced it) so thanks for fixing it! I suppose the UUID thing will work, but it's a tad complex. I wonder if the `isMesh` check is all you really need? Could you fix this by reverting all of this and just inserting that as a ternary where `elementReference` is defined?

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -1,8 +1,10 @@
-import {Group, Material, Mesh, Object3D, Texture} from 'three';
+import {Group, Material, Mesh, MeshStandardMaterial, Object3D, Texture} from 'three';
 import {GLTF as ThreeGLTF, GLTFReference} from 'three/examples/jsm/loaders/GLTFLoader.js';
 
 import {GLTF, GLTFElement} from '../../three-components/gltf-instance/gltf-2.0.js';
 
+
+
```
elalish, line 7: With your tests failing iOS 13 it looks like you may have forgotten to pull master before you branched this? Anyway, a pull now should be a quick fix.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
timmmeh: Done.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
timmmeh: I guess that works too, actually, in that case why not just insert the default material once and be done, like so...

elalish: hmm, isn't that just creating a default material all the time even if it's not needed again? Can we add a simple test while we're at it?

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
timmmeh: Done.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
timmmeh: Well originally it was generating one for every object not it just does it once, I had the check (to see if there is actually DefaultMaterial being used) to prevent this so it's only ever added in the legit cases. If that check is removed than, one will always be created so long as the gltf has a mesh in it, which is practically every gltf. But not sure if having a default representation there is a bad thing though.

elalish: Fair, but I think it's still confusing to show a material that's unused. I think the even smaller change I suggested will fix both?

timmmeh: New updates added, found another issue, which is that the Default material was being added by the original gltf correlation and then added again when cloned.

timmmeh: Adding tests to catch these other issues such as multiple defaults, and extraneous materials, but this bit of code seems bug-prone, probably better to do this fix upstream by patching the incoming scene and gltf before this correlation step.

elalish: Oh wow, I see how it's happening both in the original and in the cloning step now. I can't remember why I did that, but I agree it sounds like a terrible idea. Go for it.

elalish (commented):
Just clearing the review status so you can request it again after your tests are added.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -123,20 +123,38 @@ export class CorrelatedSceneGraph {
...
           (object: ThreeSceneObject, cloneObject: ThreeSceneObject) => {
-            const elementReference =
+            let elementReference =
                 upstreamCorrelatedSceneGraph.threeObjectMap.get(object);
 
+            if (((object as Mesh).isMesh || (object as Material).isMaterial) &&
+                elementReference == null) {
+              // Checks if default material was allready addded to the gltf.
+              if (cloneGLTF.materials) {
+                const material =
```
timmmeh, line 139: Could do an exhaustive check for default material here, not sure if the default could ever end up _not_ being the last material, but as the code ages and things check 🤷  who knows

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -123,20 +123,38 @@ export class CorrelatedSceneGraph {
...
     for (let i = 0; i < originalThreeGLTF.scenes.length; i++) {
       this[$parallelTraverseThreeScene](
           originalThreeGLTF.scenes[i],
           cloneThreeGLTF.scenes[i],
           (object: ThreeSceneObject, cloneObject: ThreeSceneObject) => {
-            const elementReference =
+            let elementReference =
                 upstreamCorrelatedSceneGraph.threeObjectMap.get(object);
 
+            if (((object as Mesh).isMesh || (object as Material).isMaterial) &&
```
timmmeh, line 135: Found the a mesh with default material or a Material itself (with no reference) can be traversed here, I find that odd in and of itself, but thats a battle for another day.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -123,20 +123,38 @@ export class CorrelatedSceneGraph {
...
+                elementReference == null) {
+              // Checks if default material was allready addded to the gltf.
+              if (cloneGLTF.materials) {
+                const material =
+                    cloneGLTF.materials[cloneGLTF.materials.length - 1];
+                if (material.name === 'Default') {
```
timmmeh, line 141: This prevents a Default material from being added twice, the first time it's added is during the original GLTF correlation step.

packages/model-viewer/src/test/three-components/gltf-instance/correlated-scene-graph-spec.ts
```diff
@@ -116,6 +118,23 @@ suite('correlated-scene-graph', () => {
...
+
+        const scene = SkeletonUtils.clone(threeGLTF.scene) as Group;
+        const scenes: Group[] = [scene];
+
+        const cloneThreeGLTF: GLTF = {...threeGLTF, scene, scenes};
+
+        const cloneCorrelatedSceneGraph =
+            CorrelatedSceneGraph.from(cloneThreeGLTF, correlatedSceneGraph);
+
+        expect(cloneCorrelatedSceneGraph.gltf.materials!.length).to.be.eq(1);
```
elalish, line 134: 👍

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
elalish: Good enough for now, thanks! At least we have a test to save us.

elalish (approved):
Thanks for fixing my bug!

PR #2577, opened 2021-07-14
Sorted the scenarios in config.json to improve readability and ease of…

augmentedsky:

### DIFF
--- a/packages/render-fidelity-tools/test/config.json
+++ b/packages/render-fidelity-tools/test/config.json
@@ -26,193 +26,19 @@
   ],
   "scenarios": [
     {
-      "name": "khronos-BoomBox",
-      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/BoomBox/glTF/BoomBox.gltf",
+      "name": "AlphaBlendLitmusTest",
+      "model": "../../../shared-assets/models/alpha-blend-litmus.glb",
       "orbit": {
-        "radius": 0.03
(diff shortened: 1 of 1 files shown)


packages/render-fidelity-tools/test/config.json
```diff
@@ -26,40 +26,126 @@
   ],
   "scenarios": [
     {
-      "name": "khronos-BoomBox",
-      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/BoomBox/glTF/BoomBox.gltf",
+      "name": "AlphaBlendLitmusTest",
```
elalish, line 29: So, the trouble is this order is actually sort of intentional. The reason being that the order in the json currently determines the order in our public comparisons: https://modelviewer.dev/fidelity/

There are a bunch of really boring conformance models that we sort to the bottom and we keep the pretty ones near the top for visibility. However, I agree that sorting internally is convenient. Perhaps we could add an index property or something?

elalish (changes requested):

packages/render-fidelity-tools/test/config.json
augmentedsky: Ah I see. My bad, I should have checked before making the changes. I'm not sure how adding an index may help. I can then just append any new models to the end unless we want them to be high up

augmentedsky:
Intended to be unsorted

PR #2579, opened 2021-07-15
Added DragonAttenuation to fidelity tests

asemanh:
Added the images for reference

babylon-golden
filament-golden
gltf-sample-viewer-golden
model-viewer-golden


elalish:
This is a good example of one we could reduce the height of. More than half the image is just background.

packages/render-fidelity-tools/test/config.json
```diff
@@ -374,6 +374,16 @@
         "dspbr-pt"
       ]
     },
+    {
+      "name": "khronos-DragonAttenuation",
```
elalish, line 378: Also, this model is pretty (in my opinion), so let's put it toward the top.

elalish (changes requested):

packages/render-fidelity-tools/test/config.json
asemanh: Agreeeed! I love it myself!

asemanh:
babylon-golden
filament-golden
gltf-sample-viewer-golden
model-viewer-golden

packages/render-fidelity-tools/test/config.json
```diff
@@ -25,6 +25,20 @@
     }
   ],
   "scenarios": [
+    {
+      "name": "khronos-DragonAttenuation",
```
elalish, line 29: Okay, maybe not the *first* model :) This is a bit complicated (in terms of glTF extensions) and not all the renderers are even showing it properly yet. So, we don't need to showcase it right at the top, but at least in the upper section of pretty models.

packages/render-fidelity-tools/test/config.json
```diff
@@ -25,6 +25,20 @@
...
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/DragonAttenuation/glTF/DragonAttenuation.gltf",
+      "dimensions": {
+        "height": 120,
+        "width": 130
```
elalish, line 33: Sorry, this wasn't what I meant. Let's keep the resolution around what we usually use; I just don't want lots of blank pixels. For this one, I'm realizing it'll probably make more sense to zoom in a bit as we don't need to see the whole drape. Let's just focus on the dragon and frame that.

elalish (changes requested):

packages/render-fidelity-tools/test/config.json
asemanh: sorry I missed this

elalish (approved):

PR #2580, opened 2021-07-15
Added Fox model to fidelity tests

asemanh:
Added screenshots here for reference
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Fox

babylon-golden
filament-golden
gltf-sample-viewer-golden
model-viewer-golden


packages/render-fidelity-tools/test/config.json
```diff
@@ -385,6 +385,18 @@
         "dspbr-pt"
       ]
     },
+    {
+      "name": "khronos-Fox",
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/Fox/glTF/Fox.gltf",
+      "orbit": {
```
elalish, line 391: Let's adjust the target to center the model (Y is up, so that may be all you need to change). And probably reduce the height, since it's a wide model.

elalish (changes requested):

asemanh:
babylon-golden
filament-golden
gltf-sample-viewer-golden
model-viewer-golden

packages/render-fidelity-tools/test/config.json
asemanh: How is the latest fox?

elalish:
Looks like you could zoom in by about a factor of 2. The vast majority of the pixels are blank.

packages/render-fidelity-tools/test/config.json
elalish: Looks like you could zoom in by about a factor of 2. The vast majority of the pixels are blank.

PR #2582, opened 2021-07-15
Added MaterialsVariantShoe to the fidelity tests

asemanh:
model : https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/MaterialsVariantsShoe

Screenshots for reference:
babylon-golden
filament-golden
gltf-sample-viewer-golden
model-viewer-golden


packages/render-fidelity-tools/test/config.json
```diff
@@ -404,6 +404,17 @@
         "dspbr-pt"
       ]
     },
+    {
+      "name": "khronos-MaterialsVariantsShoe",
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/MaterialsVariantsShoe/glTF/MaterialsVariantsShoe.gltf",
+      "orbit": {
+        "theta":35,
+        "phi": 49
```
elalish, line 412: Let's reduce radius here from default so that it fills the viewport better. No need for perfection, I just want to make use of the pixels we have.

elalish (changes requested):

PR #2584, opened 2021-07-16
Puppeteer upgrade

asemanh:
I haven't figured out why bootstrap fails in the CI test, while it succeeds in my local machine.


packages/render-fidelity-tools/package.json
```diff
@@ -43,7 +43,7 @@
     "lit-element": "^2.5.1",
     "mkdirp": "^1.0.4",
     "pngjs": "^6.0.0",
-    "puppeteer": "^5.2.1",
+    "puppeteer": "^10.1.0",
```
elalish, line 46: It's strange that this single change is causing massive changes to our package-lock files in all the packages. Sounds to me like upgrade:package-lock is not behaving right on your machine. What happens if you run that on a clean branch off master without changing package.json at all? It should not change package-lock much if at all. If it is, we'll need to debug that first.

elalish (changes requested):

PR #2585, opened 2021-07-16
Restore 1 to many material relationship

timmmeh:


packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -106,8 +106,10 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
     static Texture: Constructor<Texture>;
     static Image: Constructor<Image>;
 
-    private[$onUpdate]() {
-      this[$needsRender]();
+    private[$getOnUpdateMethod]() {
+      return () => {
+        this[$needsRender]();
```
timmmeh, line 111: Discovered that change variant example was breaking because 'this' was undefined. I don't think it was ever being captured in a lambda, this fixes it.

packages/model-viewer/src/features/scene-graph/api.ts
```diff
@@ -127,6 +127,12 @@ export declare interface TextureInfo {
    * The Texture being referenced by this TextureInfo
    */
   readonly texture: Texture|null;
+
+  /**
+   * Sets a texture on the texture info, or removes the texture if argument is
+   * null.
+   */
+  setTexture(texture: Texture|null): void;
```
timmmeh, line 135: This api should probably be here now.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -66,57 +66,69 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+        normalTextures,
         TextureUsage.Normal,
         normalTexture!);
 
     this[$occlusionTexture] = new TextureInfo(
         onUpdate,
         gltf,
-        this[$backingThreeMaterial],
-        firstValue(occlusionTextures),
+        correlatedMaterials,
```
timmmeh, line 122: The main thing that changes here is that all correlated materials are now being being updated when editing a texture. However, I simplified things a bit be removing, what ends up being, duplicate references to the same textures. So when a material is branched in Three.js, because of some state change, both branches still  point to the same texture, any request to change the texture affects all materials but there is no need to represent the same texture many times.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -66,57 +66,69 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
-    const firstValue = (set: Set<ThreeTexture>): ThreeTexture => {
-      return set.values().next().value;
-    };
+    for (const gltfMaterial of correlatedMaterials) {
+      const {
+        normalMap: verifyNormalMap,
+        aoMap: verifyAoMap,
+        emissiveMap: verifyEmissiveMap
+      } = gltfMaterial;
+      if (verifyNormalMap !== normalMap) {
```
timmmeh, line 100: Just check to make sure texture references have not diverged.

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -56,7 +56,7 @@ export class TextureInfo extends ThreeDOMElement implements
   // Holds a reference to the glTF file data.
   [$gltf]: GLTF;
   // Holds a reference to the Three data that backs the material object.
-  [$material]: MeshStandardMaterial|null;
+  [$material]: Set<MeshStandardMaterial>|null;
```
elalish, line 59: Does this need `|null`? Should we just initialize it to an empty set?

elalish: Also, should the name be plural: `$materials`?

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -66,57 +66,69 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
 
     let {normalTexture, occlusionTexture, emissiveTexture} = gltfMaterial;
 
-    const normalTextures = new Set<ThreeTexture>();
-    const occlusionTextures = new Set<ThreeTexture>();
-    const emissiveTextures = new Set<ThreeTexture>();
+    let normalTextures: ThreeTexture|null = null;
+    let occlusionTextures: ThreeTexture|null = null;
+    let emissiveTextures: ThreeTexture|null = null;
```
elalish, line 71: Since these are no longer sets, perhaps they should be singular?

packages/model-viewer/src/features/scene-graph/pbr-metallic-roughness.ts
```diff
@@ -56,44 +56,47 @@ export class PBRMetallicRoughness extends ThreeDOMElement implements
...
+        correlatedMaterials.values().next().value ?
+        correlatedMaterials.values().next().value.map :
+        null;
+
+    // NOTE: GLTFLoader users the same texture for metalnessMap and
+    // roughnessMap in this case
+    // @see https://github.com/mrdoob/three.js/blob/b4473c25816df4a09405c7d887d5c418ef47ee76/examples/js/loaders/GLTFLoader.js#L2173-L2174
+    const metallicRoughnessTextures: ThreeTexture|null =
+        correlatedMaterials.values().next().value ?
+        correlatedMaterials.values().next().value.metalnessMap :
```
elalish, line 75: For consistency, might be nice to use the same destructing method you used above for the textures on material.

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -82,13 +82,13 @@ export class TextureInfo extends ThreeDOMElement implements
...
-      texture: ThreeTexture|null, textureUsage: TextureUsage,
-      gltfTextureInfo: GLTFTextureInfo) {
+      onUpdate: () => void, gltf: GLTF,
+      material: Set<MeshStandardMaterial>|null, texture: ThreeTexture|null,
+      textureUsage: TextureUsage, gltfTextureInfo: GLTFTextureInfo) {
     super(
         onUpdate,
         gltfTextureInfo ? gltfTextureInfo : {index: -1},
-        new Set<ThreeTexture>([texture!]));
+        new Set<ThreeTexture>(texture ? [texture] : []));
```
elalish, line 91: `$threeTexture` is duplicating `$source`; let's just use one; probably `$source` for consistency.

elalish (changes requested):
Looking good, thanks! Can you think of any simple tests for this?

packages/model-viewer/src/features/scene-graph/texture-info.ts
timmmeh: Unfortunately it does need the null, this is to support the ad-hoc creation of a Texture. The Texture needs a TextureInfo at construction however in the ad-hoc case there is no material that the texture is being applied to.

Rename, done.

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -192,7 +199,7 @@ export class TextureInfo extends ThreeDOMElement implements
       // Applies the existing context to the new texture.
       this[$texture]!.applyNewTextureInfo(this);
     }
-    this[$material]!.needsUpdate = true;
+    // this[$material]!.needsUpdate = true;
```
timmmeh, line 202: remove

packages/model-viewer/src/features/scene-graph/pbr-metallic-roughness.ts
timmmeh: Sorry, I'm not sure what you mean by destructing method.

elalish: With emissiveTexture, normalTexture, etc, you pulled them all out at once. Pretty sure this code does exactly the same thing, it just looks different. It's a tad easier to read when the same operation looks the same. Just a nit.

timmmeh: Ah I see

packages/model-viewer/src/features/scene-graph/texture-info.ts
timmmeh: done.

packages/model-viewer/src/features/scene-graph/pbr-metallic-roughness.ts
```diff
@@ -56,49 +56,58 @@ export class PBRMetallicRoughness extends ThreeDOMElement implements
...
     }
 
-    const baseColorTexture =
-        pbrMetallicRoughness.baseColorTexture ?? {index: -1};
+    let {
+      baseColorTexture: gltfBaseColorTexture,
+      metallicRoughnessTexture: gltfMetallicRoughnessTexture
+    } = pbrMetallicRoughness;
 
-    const metallicRoughnessTexture =
```
timmmeh: refactored this code to make it look consistent with the way things are done in material

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -81,6 +79,11 @@ export class TextureInfo extends ThreeDOMElement implements
   [$encoding]: TextureEncoding;
   onUpdate: () => void;
 
+  // Returns a reference to the Three data that backs the texture object.
+  get[$threeTexture](): ThreeTexture|null {
+    return this[$correlatedObjects]?.values().next().value;
+  }
```
elalish, line 85: 👍

elalish (approved):
Thanks!

PR #2596, opened 2021-07-21
Texture info refactor

timmmeh:
Refactor greatly simplifies texture-info, texture, image, and sampler setup. Allows objects to more independent than previous, making it easier to change and update textures.


packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -20,34 +20,34 @@ import {Image as GLTFImage} from '../../three-components/gltf-instance/gltf-2.0.
...
 export const $applyTexture = Symbol('applyTexture');
 
 /**
  * Image facade implementation for Three.js textures
  */
 export class Image extends ThreeDOMElement implements ImageInterface {
-  get[$underlyingTexture]() {
-    return this[$textureInfo][$threeTexture];
+  get[$sourceTexture]() {
+    console.assert(
```
timmmeh, line 39: correlatedObjects are allowed to be null, this ensures that operations can't be applied to potential temporary objects without any backing data. I don't think this can happen in the current state of things, but overtime this might save some headaches.

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -103,12 +103,13 @@ export class Image extends ThreeDOMElement implements ImageInterface {
...
-          }
-          resolve(URL.createObjectURL(blob));
-        }, 'image/png')});
+    return new Promise<string>(async (resolve, reject) => {
+      blobCanvas.toBlob(blob => {
+        if (!blob) {
+          return reject('Failed to capture thumbnail.');
+        }
+        resolve(URL.createObjectURL(blob));
+      }, 'image/png');
```
timmmeh, line 112: missing a semicolon

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -64,36 +65,38 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
-    let occlusionTexture: ThreeTexture|null = null;
-    let emissiveTexture: ThreeTexture|null = null;
-
     const {normalMap, aoMap, emissiveMap} =
         correlatedMaterials.values().next().value;
 
-    if (gltfNormalTexture != null && normalMap != null) {
-      normalTexture = normalMap;
-    } else {
-      gltfNormalTexture = {index: -1};
```
timmmeh, line 83: The big simplification here was to allow glTF data to be represented as null in the cases where we would want to conjure things up out of thin air. This drastically simplifies all the bookkeeping we need to do to support adding for write operations such as adding a new texture.

packages/model-viewer/src/features/scene-graph/texture-factory.ts
```diff
@@ -0,0 +1,66 @@
...
+  // A helper method for creating TextureInfo objects from gltf data.
+  static[$createTextureInfo](
+      onUpdate: () => void, usage: TextureUsage, gltf: GLTF,
+      threeTexture: ThreeTexture|null, gltfTextureInfo: GLTFTextureInfo|null,
+      correlatedMaterials: Set<MeshStandardMaterial>): TextureInfo {
+    let gltfTexture: GLTFTexture|null = null;
+
+    let texture: Texture|null = null;
+    // Creates image, sampler and texture if valid texture info is provided.
+    if (gltfTextureInfo) {
```
timmmeh, line 42: Being verbose about creating the images and sampler needed to create a texture. This could be hidden inside texture constructor, but I like simplicity of doing it this way where the constructor defines what objects you need and we construct them before hand and pass them in rather than hiding their creation inside the constructor.

packages/model-viewer/src/features/scene-graph/texture-factory.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Helper class for creating Textures and TextureInfos.
+ */
+export class TextureFactory {
+  // Creates context from Texture. This creats a texture with a valid
+  static[$createTexture](threeTexture: ThreeTexture): Texture {
+    // Creates an empty glTF data set, allows for creating a texture that's
+    // not bound to a material.
+    const emptyOnUpdate = () => {};
+    return new Texture(
```
timmmeh, line 25: creates a texture from a ThreeJS texture, notice the glTF data is just null instead of being spoofed. Much simpler to write, use, and read this code.

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -53,99 +44,24 @@ export class TextureInfo extends ThreeDOMElement implements
...
     this.onUpdate = onUpdate;
-    this[$gltf] = gltf;
     this[$materials] = material;
     this[$usage] = textureUsage;
-
-    // Gathers glTF texture info data.
-    this[$gltfTextureInfo] = this[$sourceObject] as GLTFTextureInfo;
-
-    // Gathers glTF texture data.
-    if (gltf.textures && this[$gltfTextureInfo]!.index !== -1) {
```
timmmeh, line 105: No need to spoof glTF data so all the nonsense can be deleted.

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -154,15 +70,9 @@ export class TextureInfo extends ThreeDOMElement implements
 
   setTexture(texture: Texture|null): void {
     const threeTexture: ThreeTexture|null =
-        texture != null ? texture.source[$underlyingTexture] : null;
+        texture != null ? texture.source[$sourceTexture] : null;
     let encoding: TextureEncoding = sRGBEncoding;
     this[$texture] = texture;
-    // Ensures correlatedObjects is up to date.
-    const correlatedObjects = (this[$correlatedObjects] as Set<ThreeTexture>);
```
timmmeh, line 161: This is an interesting change, I removed the TextureInfo correlatedObject. I think this better for a couple of reasons, 1) the TextureInfo has no real correlation in Three.js, previously the it was correlated to a three-texture, now Texture is correlated to a three-texture 2) this reduces the number of way you can access and change the underlying object and should help reduce changing things from the wrong place.

packages/model-viewer/src/features/scene-graph/texture.ts
```diff
@@ -15,42 +15,36 @@
...
+  ) {
     super(
-        textureInfo.onUpdate,
-        textureInfo[$gltfTexture],
-        new Set<ThreeTexture>([textureInfo[$threeTexture]!]));
-    this[$sampler] = new Sampler(textureInfo);
-    this[$source] = new Image(textureInfo);
-  }
-
-  applyNewTextureInfo(textureInfo: TextureInfo): void {
```
timmmeh, line 47: no need to do this now

packages/model-viewer/src/features/scene-graph/three-dom-element.ts
```diff
@@ -32,12 +34,12 @@ type CorrelatedObjects = Set<Object3D>|Set<Material>|Set<Texture>;
...
   readonly[$onUpdate]: () => void;
   // The canonical GLTF or GLTFElement represented by this facade.
-  readonly[$sourceObject]: GLTFElement;
+  readonly[$sourceObject]: GLTFElement|null;
   // The backing Three.js scene graph construct for this element.
   readonly[$correlatedObjects]: CorrelatedObjects|null;
 
   constructor(
-      onUpdate: () => void, element: GLTFElement,
+      onUpdate: () => void, element: GLTFElement|null,
```
timmmeh, line 42: All the simplifications hinge on this, this seems like an appropriate change given that we want create new objects that will not have a backing glTF file.

packages/model-viewer/src/features/scene-graph/texture.ts
```diff
@@ -15,42 +15,36 @@
...
-    (this[$onUpdate] as () => void) = textureInfo.onUpdate;
-    (this[$sourceObject] as GLTFElement) = textureInfo[$gltfTexture];
-    (this[$correlatedObjects] as Set<ThreeTexture>) =
-        new Set<ThreeTexture>([textureInfo[$threeTexture]!]);
-    this[$sampler] = new Sampler(textureInfo);
-    this[$source] = new Image(textureInfo);
+        onUpdate, gltfTexture, new Set<ThreeTexture>(texture ? [texture] : []));
```
timmmeh, line 45: fixes a source object problem, glTF TextureInfo was being supplied as the source data but should have been glTF Texture

timmmeh, line 45: fixes a source object problem, glTF TextureInfo was being supplied as the source data but should have been glTF Texture

packages/model-viewer/src/features/scene-graph/texture-factory.ts
```diff
@@ -0,0 +1,66 @@
...
+        new Image(emptyOnUpdate, threeTexture, null),
+        threeTexture,
+        null);
+  }
+
+  // A helper method for creating TextureInfo objects from gltf data.
+  static[$createTextureInfo](
+      onUpdate: () => void, usage: TextureUsage, gltf: GLTF,
+      threeTexture: ThreeTexture|null, gltfTextureInfo: GLTFTextureInfo|null,
+      correlatedMaterials: Set<MeshStandardMaterial>): TextureInfo {
```
elalish, line 37: This seems like the same argument list as the TextureInfo constructor; can we simply move this logic in there so we don't need the extra static function?

packages/model-viewer/src/features/scene-graph/texture-factory.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const $createTexture = Symbol('createTexture');
+export const $createTextureInfo = Symbol('createTextureInfo');
+
+/**
+ * Helper class for creating Textures and TextureInfos.
+ */
+export class TextureFactory {
+  // Creates context from Texture. This creats a texture with a valid
+  static[$createTexture](threeTexture: ThreeTexture): Texture {
```
elalish, line 21: Can we move this static constructor into the Texture class to keep everything together and so we can avoid making this texture-factory.ts file?

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -20,34 +20,34 @@ import {Image as GLTFImage} from '../../three-components/gltf-instance/gltf-2.0.
...
-    super(
-        textureInfo.onUpdate,
-        textureInfo[$gltfImage],
-        new Set<ThreeTexture>([textureInfo[$threeTexture]!]));
-    this[$textureInfo] = textureInfo;
+  constructor(
+      onUpdate: () => void, texture: ThreeTexture|null,
+      gltfImage: GLTFImage|null) {
+    gltfImage =
+        gltfImage ?? {name: 'adhoc_image', uri: 'adhoc_image'} as GLTFImage;
```
elalish, line 49: If we have a ThreeTexture, let's set the uri here to its URI.

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -20,34 +20,34 @@ import {Image as GLTFImage} from '../../three-components/gltf-instance/gltf-2.0.
...
+export const $sourceTexture = Symbol('sourceTexture');
 export const $applyTexture = Symbol('applyTexture');
 
 /**
  * Image facade implementation for Three.js textures
  */
 export class Image extends ThreeDOMElement implements ImageInterface {
-  get[$underlyingTexture]() {
-    return this[$textureInfo][$threeTexture];
+  get[$sourceTexture]() {
```
elalish, line 38: Maybe $threeTexture? We use $sourceObject to mean the glTF, vs $correlatedObject to refer to three.js stuff, so I'd like to not confuse the term source any more than we have to.

packages/model-viewer/src/features/scene-graph/texture-factory.ts
elalish: I see your point, but this seems inconsistent with the rest of the style here. Each other class's constructor here constructs the next classes that are members of it. Let's stick with that convention for now for consistency.

packages/model-viewer/src/features/scene-graph/three-dom-element.ts
elalish: Correct me if I'm wrong, but it looks to me like the important change you made is making each inheriting constructor take `GLTFElement|null`; as far as I can tell you do in fact pass a real source object to the super call in each one. The one exception perhaps being TextureInfo, but I'm wondering (based on your other comments) if that class simply should not inherit from ThreeDOMElement at all.

elalish (changes requested):
This is looking like a vast improvement!

packages/model-viewer/src/features/scene-graph/texture-factory.ts
```diff
@@ -0,0 +1,66 @@
...
+    // Creates an empty glTF data set, allows for creating a texture that's
+    // not bound to a material.
+    const emptyOnUpdate = () => {};
+    return new Texture(
+        emptyOnUpdate,
+        new Sampler(emptyOnUpdate, threeTexture, null),
```
elalish, line 27: When you set a null sampler, you're getting glTF default sampler values, which don't necessarily equal the three.js texture's sampler values. I think we need to set them explicitly.

packages/model-viewer/src/features/scene-graph/three-dom-element.ts
timmmeh: Actually when you create a texture using the createTexture() API null is provided for all gltf data, which simplifies a lot of things. Elsewhere in the code null is 'allowed' but not directly specified, this makes setting up TextureInfos with empty textures much more simple because you don't have to have branching logic based on whether the gltf data is real or not. But your right TextureInfo doesn't need to be a ThreeDomElement, but that makes me wonder what material difference is ThreeDomElement making at this stage and why have it all, if I can remove something in the middle of the Dom and it makes no difference than.... 🤷  why have it all? I should  look deeper into that existential question.

elalish: That's a fair point, but let's not go too far down the rabbit hole. I'm suggesting TextureInfo is a special case because as you mentioned, it doesn't really have a correlation to three.js. I think of this base element as representing something that's related to three.js (based on its members).

timmmeh: Done.

packages/model-viewer/src/features/scene-graph/texture-factory.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/texture-factory.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/image.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/image.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/texture-factory.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/texture-factory.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/texture.ts
```diff
@@ -36,15 +37,25 @@ export class Texture extends ThreeDOMElement implements TextureInterface {
...
+
+    this[$sampler] = new Sampler(onUpdate, threeTexture, gltfSampler);
+    this[$image] = new Image(onUpdate, threeTexture, gltfImage);
+  }
+
+  // Helpter method creates a Texture from a ThreeTexture that is not bound to
+  // any material.
+  static[$createTexture](threeTexture: ThreeTexture, onUpdate: () => void):
+      Texture {
+    return new Texture(onUpdate, threeTexture, null, null, null);
```
elalish, line 58: Well, that's kind of trivial now. Should we just remove this and call the constructor directly?

packages/model-viewer/src/features/scene-graph/three-dom-element.ts
elalish: You said done, but this still says `|null`. Am I correct that it's no longer required after your other changes? If so, it may also render some of your console checks unnecessary.

elalish (commented):
We're close!

packages/model-viewer/src/features/scene-graph/texture-info.ts
```diff
@@ -53,14 +51,28 @@ export class TextureInfo extends ThreeDOMElement implements
...
 
   constructor(
-      onUpdate: () => void, textureUsage: TextureUsage = TextureUsage.Base,
-      texture: Texture|null, material: Set<MeshStandardMaterial>|null,
-      gltfTextureInfo: GLTFTextureInfo|null) {
-    super(onUpdate, gltfTextureInfo, new Set<ThreeTexture>([]));
+      onUpdate: () => void, usage: TextureUsage,
+      threeTexture: ThreeTexture|null, material: Set<MeshStandardMaterial>,
+      gltf: GLTF, gltfTextureInfo: GLTFTextureInfo|null) {
+    let texture: Texture|null = null;
```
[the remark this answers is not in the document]
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/three-dom-element.ts
timmmeh: Right it's still needed in order to represent image, sampler, and texture (and more to come I think) that _don't_ have glTF data.

packages/model-viewer/src/features/scene-graph/texture.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/three-dom-element.ts
timmmeh: Went back and looked at this, yeah, I think I can put this back.

elalish (dismissed):
Great, thank you! More than 100 lines of net code loss; that's what I call improvement!

packages/model-viewer/src/three-components/gltf-instance/gltf-2.0.ts
```diff
@@ -1,3 +1,5 @@
+import {TextureFilter, Wrapping} from 'three';
```
elalish, line 1: Can we remove this? I don't like the idea of glTF being dependent on three.js. I believe we're already making the assumption that these filter constants are consistent between glTF and three, which I don't love, but is probably fine since they both come from WebGL.

packages/model-viewer/src/three-components/gltf-instance/gltf-2.0.ts
```diff
@@ -78,23 +80,38 @@ export interface OrthographicCamera {
...
+  NearestMipmapLinear = 9986,
+  LinearMipmapLinear = 9987,
+}
+
+export type NearestFilter = Filter.Nearest;
+export type LinearFilter = Filter.Linear;
+export type NearestMipmapNearestFilter = Filter.NearestMipmapNearest;
+export type LinearMipmapNearestFilter = Filter.LinearMipmapNearest;
+export type NearestMipmapLinearFilter = Filter.NearestMipmapLinear;
+export type LinearMipmapLinearFilter = Filter.LinearMipmapLinear;
```
elalish, line 97: Do we really need both an enum and the individual constants?

packages/model-viewer/src/features/scene-graph/sampler.ts
```diff
@@ -15,28 +15,35 @@
...
-  const minFilterValues: Array<MinFilter> =
-      [9728, 9729, 9984, 9985, 9986, 9987];
+  const minFilterValues: Array<MinFilter> = [
+    Filter.Nearest,
+    Filter.Linear,
+    Filter.NearestMipmapNearest,
+    Filter.LinearMipmapLinear,
+    Filter.NearestMipmapLinear,
+    Filter.LinearMipmapLinear
+  ];
```
elalish, line 33: Would `= [...Filter]` work here?

packages/model-viewer/src/three-components/gltf-instance/gltf-2.0.ts
timmmeh: I removed them and it created several type conversion issues elsewhere, it started to creep out of scope so I backed off. But happy to lay waste to that area of code perhaps in a different CL?

packages/model-viewer/src/features/scene-graph/sampler.ts
timmmeh: Interesting, I tried but get
`Type 'typeof Filter' must have a '[Symbol.iterator]()' method that returns an iterator.ts(2488)`
I looked around for enum expansion in typescript but I'm not finding anything off the cuff.

packages/model-viewer/src/three-components/gltf-instance/gltf-2.0.ts
timmmeh: That was my thinking as well, but given my change it needed that to be able to understand the type conversions between TextureFilter and MagFilter for instance.

elalish (approved):
Thanks!

Issue #2139, opened 2021-03-19
Memory leak after repeated adding and removing model-viewer component to the DOM


ukito-pl:
### Description

Repeated adding and removing of model-viewer component to the DOM will eventually lead to memory overflow and loss of GL Context. Use live demo to test this behaviour. In my case, after around 70 refreshes, memory starts to rise up rapidly and after 120 views opengl crashes. It can vary based on user's machine. When using model-viewer on Single Page Application memory overflow can happen very quickly (imagine online shop where each item has its own model view).

#### Live Demo
https://plnkr.co/edit/PGOfoqXN8iXuhAzk

### Version

- model-viewer: v1.6.0

### Browser Affected

- [x] Chrome, version: 89.0.4389.90
- [ ] Edge
- [x] Firefox, version 86.0
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [x] Windows 10

elalish:
@timmmeh FYI here's that memory leak.

vrboobies:
Confirmed issue on android. Crashes the chrome browser after about 10 load/unloads.

fgillmann:
Also confirmed for Edge (Chromium) with the plunker in the first post.
Can also confirm it for our own usage of model-viewer.

We have model-viewer running in a React website with files up to 400MB which work fine within desktop browsers and even with mobile browsers its ok for our use case, if the device is not too old.

Once you click on an image representing the object, one instance of model-viewer is created within a fullscreen modal/overlay and once the user closes the overlay, everything is discarded. 

Doing so multiple times, with the same or different objects, regardless of modelCacheSize set to 0, considerably increases the memory consumption.

Also regarding the plunker: 
I had it run 24 times in Edge, finished with model-viewer not visible any more and then just went away for 20min, and coming back, the memory usage was dropped from 1600MB down to 180MB

PR #2600, opened 2021-07-21
dispose material textures when disposing GLTFInstance

ukito-pl:
Resolves problem with memory overflow when GLTFInstance object is being repeatedly created and disposed.

### Reference Issue
Fixes: [#2139](https://github.com/google/model-viewer/issues/2139)

### DIFF
--- a/packages/model-viewer/src/three-components/GLTFInstance.ts
+++ b/packages/model-viewer/src/three-components/GLTFInstance.ts
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import {Group, Mesh, Object3D} from 'three';
+import {Group, Mesh, Object3D, Texture} from 'three';
 import {GLTF} from 'three/examples/jsm/loaders/GLTFLoader.js';
 import {SkeletonUtils} from 'three/examples/jsm/utils/SkeletonUtils.js';
 import {Constructor} from '../utilities.js';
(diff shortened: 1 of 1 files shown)


packages/model-viewer/src/three-components/GLTFInstance.ts
```diff
@@ -136,6 +136,13 @@ export class GLTFInstance implements GLTF {
         const materials =
             Array.isArray(mesh.material) ? mesh.material : [mesh.material];
         materials.forEach(material => {
+          // Explicitly dispose any textures assigned to this material
+          for (const propertyName in material) {
+            const propertyValue = (material as any)[propertyName];
+            if (propertyValue instanceof Texture) {
+              propertyValue.dispose();
```
elalish, line 143: Good call! Not sure how we went this long without noticing our textures needed disposal. Did you verify this fixes the leak?

elalish (approved):

timmmeh (approved):

packages/model-viewer/src/three-components/GLTFInstance.ts
ukito-pl: Yes, the example I supplied in [#2139](https://github.com/google/model-viewer/issues/2139) doesn't create memory overflow when tested on my local build with proposed fix.

PR #2601, opened 2021-07-21
Upgrading to npm7 and latest version of puppeteer

asemanh:


packages/shared-assets/package-lock.json
```diff
@@ -1,5 +1,13 @@
 {
   "name": "@google/model-viewer-shared-assets",
   "version": "0.0.1",
-  "lockfileVersion": 1
+  "lockfileVersion": 2,
```
elalish, line 4: hmm, when I pull this branch everything seems to work fine, but when I run `npm run update:package-lock` it claims my package-lock files are from an old version of npm and it changes them all. I'm on npm 7.19.1; can you repro that?

elalish:
@googlebot I signed it!

asemanh:
@googlebot I signed it!

asemanh:
> @googlebot I signed it!

@googlebot I signed it!

packages/shared-assets/package-lock.json
asemanh: I do see the same thing.
npm run update:package-lock == rm -f ./package-lock.json && npm i && ./node_modules/.bin/lerna run update:package-lock --stream

so I'm guessing it still looks at the package-lock that is on the master?

asemanh: And I'm on the same version of npm 7.19.1

asemanh: Actually let me get back to you on this.

asemanh: Ok summary of all the commits.

1) I removed every package-lock.json [the one at the root and the ones per each package]
and I run `npm i` from root and from each package. This way all the package-lock s are version 2.

2) I updated puppeteer version and I ran `npm run update:package-lock` from render-fidelity-tools package. This updated package.json and package-lock.json for this package.

3) After Emmett's comment, I ran `npm run update:package-lock` from the root, and that updated the package-lock for all the packages under model-viewer/packages. If I re-run the same command after this step, I don't get any more changes.

elalish: Strange, I'm still getting the same warnings when I run update:package-lock at the root: `npm WARN old lockfile The package-lock.json file was created with an old version of npm`. One time I tried and it failed complaining about not finding shared-assets, and another time it succeeded. Almost seemed like the run ordering was non-deterministic (or maybe in parallel)? Can you try update:package-lock on your different computers from this branch and verify they are all giving consistent results?

If they are, maybe I need to run npm install? I'm not really sure what that would change, but that's the only difference I see in your steps.

PR #2603, opened 2021-07-22
Upgrade to NPM 7

asemanh:
I tried to update our script. Something I'm not sure is if it's ok to delete the package-lock.json for shared-asset


package.json
```diff
@@ -12,7 +12,8 @@
     "clean": "./node_modules/.bin/lerna run clean --stream",
     "build": "./node_modules/.bin/lerna run build --stream",
     "serve": "./node_modules/.bin/http-server -a 127.0.0.1 -o /packages/modelviewer.dev -c-1",
-    "update:package-lock": "rm -f ./package-lock.json && npm i && ./node_modules/.bin/lerna run update:package-lock --stream"
+    "update:package-lock": "rm -f ./package-lock.json && npm i && ./node_modules/.bin/lerna run update:package-lock --stream",
```
elalish, line 15: I have no problem with removing the shared-assets package-lock, but we should also remove its package.json.

package.json
```diff
@@ -12,7 +12,8 @@
...
+    "upgrade": "rm -rf node_modules package-lock.json && npm install && lerna exec 'rm -rf node_modules package-lock.json' && lerna bootstrap && npm run clean"
```
elalish, line 16: I am still getting `npm WARN old lockfile The package-lock.json file was created with an old version of npm` when I run update:package-lock from your branch. Do you get that if you pull it to your other machine?

package.json
asemanh: I think we still need the package.json for the script

package.json
asemanh: Sounds like you're not seeing that if you run upgrade?

package.json
elalish: It seems to work, so we can skip it for now, but let's do a follow-on to clean up shared-assets.

package.json
elalish: Indeed, let's simplify so we only include the working script, under the old name.

elalish (approved):

package.json
asemanh: Yeah, when we run "npm install", it does call "prepare" script [aligns with npm documentation] which ends up fetching the khornos assets. And I agree that's outside of the scope of this task

package.json
asemanh: Just sent an updated version

PR #2606, opened 2021-07-23
Move before place

elalish:
Addressing what I hope is the biggest part of #2413, this changes the WebXR UX to allow user interactions all the time, instead of blocking them until the model is placed on the floor/wall. I think this improves the experience quite a bit and makes it less annoying/visible to the user when ARCore is struggling. However, there is not much we can do about ARCore's failures to find especially walls. We could potentially implement something like Scene Viewer's UI here, but that's pretty complicated and has it's own user education issues. We'll go with simple for now, but please add your thoughts to the above-linked discussion. 

Also fixes an existing bug where the zoom was sometimes not where you left it when returning from a WebXR session.

### DIFF
--- a/packages/model-viewer/src/test/three-components/ARRenderer-spec.ts
+++ b/packages/model-viewer/src/test/three-components/ARRenderer-spec.ts
@@ -18,7 +18,6 @@ import {Matrix4, PerspectiveCamera, Vector2, Vector3} from 'three';
 import {ControlsInterface, ControlsMixin} from '../../features/controls.js';
 import ModelViewerElementBase, {$scene} from '../../model-viewer-base.js';
 import {ARRenderer} from '../../three-components/ARRenderer.js';
-import {SETTLING_TIME} from '../../three-components/Damper.js';
 import {ModelScene} from '../../three-components/ModelScene.js';
 import {Renderer} from '../../three-components/Renderer.js';
 import {waitForEvent} from '../../utilities.js';
@@ -266,22 +265,7 @@ suite('ARRenderer', () => {
         expect(forward.y).to.be.equal(0);
         expect(cameraToHit.cross(forwardProjection)).to.be.closeTo(0, epsilon);
         expect(cameraToHit.dot(forwardProjection)).to.be.lessThan(0);
-      });
-
-      suite('after hit placement', () => {
-        let hitPosition: Vector3;
-
-        setup(async () => {
-          hitPosition = new Vector3(5, -1, 1);
-          await arRenderer.placeModel(hitPosition);
-          // Long enough time to settle at new position.
-          arRenderer.onWebXRFrame(
-              SETTLING_TIME, new MockXRFrame(arRenderer.currentSession!));
-        });
-
-        test('scene has the same orientation', () => {
-          expect(modelScene.yaw).to.be.equal(yaw);
-        });
+        expect(modelScene.yaw).to.be.equal(yaw);
       });
     });
   });
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -43,10 +43,11 @@ const HIT_ANGLE_DEG = 20;
 const INTRO_DECAY = 120;
 const SCALE_SNAP_HIGH = 1.2;
 const SCALE_SNAP_LOW = 1 / SCALE_SNAP_HIGH;
-
 // For automatic dynamic viewport scaling, don't let the scale drop below this
 // limit.
 const MIN_VIEWPORT_SCALE = 0.25;
+// Furthest away you can move an object (meters).
+const MAX_DISTANCE = 10;
 
 export type ARStatus =
     'not-presenting'|'session-started'|'object-placed'|'failed';
@@ -307,11 +308,12 @@ export class ARRenderer extends EventDispatcher {
       if (this.placeOnWall) {
         // Move the scene's target to the center of the back of the model's
         // bounding box.
-        scene.setTarget(target.x, target.y, scene.boundingBox.min.z);
+        target.z = scene.boundingBox.min.z;
       } else {
         // Move the scene's target to the model's floor height.
-        scene.setTarget(target.x, scene.boundingBox.min.y, target.z);
+        target.y = scene.boundingBox.min.y;
       }
+      scene.setTarget(target.x, target.y, target.z);
     }
   }
 
@@ -419,6 +421,7 @@ export class ARRenderer extends EventDispatcher {
     (this.threeRenderer.xr as any).updateCamera(camera);
     //@ts-ignore
     const xrCamera = this.threeRenderer.xr.getCamera();
+    scene.xrCamera = xrCamera;
 
     xrCamera.position.copy(camera.position);
     xrCamera.quaternion.copy(camera.quaternion);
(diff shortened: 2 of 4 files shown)


elalish:
@timmmeh @asemanh This is mostly practice with Github PR review process and a chance to start learning more of the MV code base.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -495,6 +477,18 @@ export class ARRenderer extends EventDispatcher {
...
   }
 
+  private getTouchLocation(): Vector3|null {
+    const {axes} = this.inputSource!.gamepad;
+    let location = this.placementBox!.getExpandedHit(
+        this.presentedScene!, axes[0], axes[1]);
+    if (location != null) {
+      vector3.copy(location).sub(this.presentedScene!.getCamera().position);
+      if (vector3.length() > MAX_DISTANCE)
+        location = null;
```
asemanh, line 487: here we could just return null

packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -370,10 +370,10 @@ export class SmoothControls extends EventDispatcher {
...
     const deltaRatio = deltaZoom === 0 ?
         0 :
-        deltaZoom > 0 ? (maximumRadius! - radius) /
-                (Math.log(maximumFieldOfView!) - this.goalLogFov) :
-                        (radius - minimumRadius!) /
-                (this.goalLogFov - Math.log(minimumFieldOfView!));
+        ((deltaZoom > 0 ? maximumRadius! : minimumRadius!) - radius) /
+            (Math.log(
+                 deltaZoom > 0 ? maximumFieldOfView! : minimumFieldOfView!) -
+             this.goalLogFov);
```
elalish, line 376: This isn't actually a change: the line wrapping was getting very confused and it was too long anyway, so I tried to simplify it a bit. Still not great.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -415,40 +415,15 @@ export class ARRenderer extends EventDispatcher {
...
     const scene = this.presentedScene!;
+    const xr = this.threeRenderer.xr as any;
 
-    (this.threeRenderer.xr as any).updateCamera(camera);
-    //@ts-ignore
-    const xrCamera = this.threeRenderer.xr.getCamera();
-
-    xrCamera.position.copy(camera.position);
-    xrCamera.quaternion.copy(camera.quaternion);
-    xrCamera.scale.copy(camera.scale);
```
elalish, line 425: These are no longer necessary thanks to an upstream change in three.js.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -39,14 +39,13 @@ const ROTATION_RATE = 1.5;
 // assumption for the start of the session and UI will lack landscape mode to
 // encourage upright use.
 const HIT_ANGLE_DEG = 20;
-// Slow down the dampers for initial placement.
-const INTRO_DECAY = 120;
-const SCALE_SNAP_HIGH = 1.2;
+const SCALE_SNAP_HIGH = 1.3;
```
elalish, line 42: I also increased the snap value a bit based on feedback from Chikezie.

packages/model-viewer/src/test/three-components/ARRenderer-spec.ts
```diff
@@ -266,22 +265,7 @@ suite('ARRenderer', () => {
...
         expect(cameraToHit.cross(forwardProjection)).to.be.closeTo(0, epsilon);
         expect(cameraToHit.dot(forwardProjection)).to.be.lessThan(0);
-      });
-
-      suite('after hit placement', () => {
-        let hitPosition: Vector3;
-
-        setup(async () => {
-          hitPosition = new Vector3(5, -1, 1);
-          await arRenderer.placeModel(hitPosition);
```
elalish, line 276: I'm not thrilled with our testing here anyway, and I removed this method, so I just combined these tests.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -463,29 +446,36 @@ export class ARRenderer extends EventDispatcher {
...
 
-    const hit = hitTestResults[0];
-    const hitPoint = this.getHitPoint(hit);
-    if (hitPoint == null) {
-      return;
-    }
+    xrCamera.projectionMatrixInverse.copy(xrCamera.projectionMatrix).invert();
+
+    const {theta, radius} =
+        (element as ModelViewerElementBase & ControlsInterface)
```
timmmeh, line 452: 🤯  ModelViewerElementBase & ControlsInterface

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -415,40 +415,15 @@ export class ARRenderer extends EventDispatcher {
 
   private updateView(view: XRView) {
     const scene = this.presentedScene!;
+    const xr = this.threeRenderer.xr as any;
```
elalish, line 418: three.js types are out of date.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -495,6 +485,41 @@ export class ARRenderer extends EventDispatcher {
         });
   }
 
+  // private placeInitially(frame: XRFrame) {
```
timmmeh: meant to remove this?

timmmeh (dismissed):

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: Looks like you were looking at an earlier commit?

asemanh (dismissed):

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -638,14 +652,20 @@ export class ARRenderer extends EventDispatcher {
       this.lastAngle = angle;
     } else if (this.isTranslating) {
       fingers.forEach(finger => {
-        if (finger.inputSource !== this.inputSource ||
-            finger.results.length < 1) {
+        if (finger.inputSource !== this.inputSource) {
           return;
         }
 
-        const hit = this.getHitPoint(finger.results[0]);
```
timmmeh, line 646: I think this would work
```
        const hit = finger.results.length >= 1 ? this.getHitPoint(finger.results[0]) : this.getTouchLocation();
        if (hit == null) {
          return;
        }
```

packages/model-viewer/src/three-components/SmoothControls.ts
timmmeh: Actually the parentheses help a lot, still takes a few cycles to process though.

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: Ooh, I like that.

packages/model-viewer/src/three-components/ARRenderer.ts
FelipeR2U: My current setup, to avoid `any` 😢 :
```ts
type ExtendedRenderer = WebGLRenderer & {
  cameraAutoUpdate: boolean
  xr: {
    updateCamera: (camera: Camera) => void
    getCamera: () => Camera
  }
}
...
const renderer = new WebGLRenderer({
  canvas,
  alpha: true,
  antialias: true,
  powerPreference: 'high-performance' as WebGLPowerPreference,
  preserveDrawingBuffer: true,
}) as ExtendedRenderer
```

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -515,17 +509,37 @@ export class ARRenderer extends EventDispatcher {
         null;
   }
 
-  public placeModel(hit: Vector3) {
-    this.placementBox!.show = true;
+  public moveToFloor(frame: XRFrame) {
```
FelipeR2U, line 512: Maybe `moveToPlane` since there's `this.placeOnWall` in this method

elalish: Good call; thanks for the drive-by review 😁

PR #2607, opened 2021-07-23
Khornous test for Fox

asemanh:
Not sure what to do about the scroll bar
babylon-golden
filament-golden
gltf-sample-viewer-golden
model-viewer-golden


elalish:
That looks fine. We can go ahead and get these models in with scroll bars for now. Then you can update the other renderers to the latest versions and when that all works, let me know and I'll run update-screenshots on my mac and push them into your PR. The scrollbars only happen on Linux, for what reason I don't know. If you have an idea on how to fix it, that would be awesome, but don't spend much time on it.

elalish:
Hey, and it looks like Github fixed their rich diff tool, so no need to paste the images in anymore. Once these are uploaded I'll approve.

elalish (approved):
LGTM!

PR #2608, opened 2021-07-23
fix comparison generating false-positive

timmmeh:
false-positive check generating erroneous logs


packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -110,13 +110,13 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
         aoMap: verifyAoMap,
         emissiveMap: verifyEmissiveMap
       } = gltfMaterial;
-      if (verifyNormalMap !== normalMap) {
+      if (verifyNormalMap != normalMap) {
```
elalish, line 113: Ah, I see why you had `!==` here. I don't love this, but I think it's fine for now. Let's see if we can get rid of these warnings entirely down the road a little.

elalish (approved):

PR #2616, opened 2021-07-26
Adding doubleSided and alphaCutoff API

### LINKED ISSUE #1661: Alpha blend
A quick question on how to access alpha blend modes as executed in https://modelviewer.dev/editor/

In the materials - > other section. 

Also alpha cutoff in mask mode and double-sided properties. 

Thanks

timmmeh:
Adds API for getting/setting alphaCutoff, doubleSided, and alphaMode.
Adresses: #2558 
Fixes #1661

### DIFF
--- a/packages/model-viewer/src/features/scene-graph/api.ts
+++ b/packages/model-viewer/src/features/scene-graph/api.ts
@@ -12,7 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import {MagFilter, MinFilter, WrapMode} from '../../three-components/gltf-instance/gltf-2.0.js';
+import {AlphaMode, MagFilter, MinFilter, WrapMode} from '../../three-components/gltf-instance/gltf-2.0.js';
 
 /**
  * All constructs in a 3DOM scene graph have a corresponding string name.
@@ -64,6 +64,9 @@ export declare interface Material {
   getAlphaCutoff(): number;
   setDoubleSided(doubleSided: boolean): void;
   getDoubleSided(): boolean;
+  setAlphaMode(alphaMode: AlphaMode): void;
+  getAlphaMode(): AlphaMode;
+
   /**
    * The PBRMetallicRoughness configuration of the material.
    */
--- a/packages/model-viewer/src/features/scene-graph/api.ts
+++ b/packages/model-viewer/src/features/scene-graph/api.ts
@@ -60,7 +60,10 @@ export declare interface Material {
 
   readonly emissiveFactor: Readonly<RGB>;
   setEmissiveFactor(rgb: RGB): void;
-
+  setAlphaCutoff(cutoff: number): void;
+  getAlphaCutoff(): number;
+  setDoubleSided(doubleSided: boolean): void;
+  getDoubleSided(): boolean;
   /**
    * The PBRMetallicRoughness configuration of the material.
    */
--- a/packages/model-viewer/src/features/scene-graph/material.ts
+++ b/packages/model-viewer/src/features/scene-graph/material.ts
@@ -16,7 +16,7 @@
 import {DoubleSide, FrontSide, MeshStandardMaterial, NoBlending, NormalBlending} from 'three';
 
 import {AlphaMode, GLTF, Material as GLTFMaterial} from '../../three-components/gltf-instance/gltf-2.0.js';
-import {ALPHA_CUTOFF_DISABLED, ALPHA_CUTOFF_GLTF_DEFAULT} from '../../three-components/gltf-instance/ModelViewerGLTFInstance.js';
+import {ALPHA_CUTOFF_BLEND, ALPHA_CUTOFF_GLTF_DEFAULT, ALPHA_CUTOFF_OPAQUE} from '../../three-components/gltf-instance/ModelViewerGLTFInstance.js';
 
 import {Material as MaterialInterface, RGB} from './api.js';
 import {PBRMetallicRoughness} from './pbr-metallic-roughness.js';
@@ -194,15 +194,13 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
   setAlphaMode(alphaMode: AlphaMode): void {
     for (const material of this[$correlatedObjects] as
          Set<MeshStandardMaterial>) {
+      material.blending = NormalBlending;
       if (alphaMode === 'OPAQUE') {
-        material.blending = NoBlending;
-        this.setAlphaCutoff(ALPHA_CUTOFF_DISABLED);
+        this.setAlphaCutoff(ALPHA_CUTOFF_OPAQUE);
       } else if (alphaMode === `BLEND`) {
-        material.blending = NormalBlending;
-        this.setAlphaCutoff(ALPHA_CUTOFF_DISABLED);
+        this.setAlphaCutoff(ALPHA_CUTOFF_BLEND);
       } else {
         // MASK mode.
-        material.blending = NormalBlending;
         if (this.getAlphaCutoff() < 0) {
           this.setAlphaCutoff(ALPHA_CUTOFF_GLTF_DEFAULT);
         }
@@ -221,7 +219,7 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
 
     const blendMode = this[$correlatedObjects]!.values().next().value.blending;
 
-    if (blendMode === NormalBlending) {
+    if (blendMode === NormalBlending && this.getAlphaCutoff() >= 0) {
       return 'BLEND';
     } else if (blendMode === NoBlending && this.getAlphaCutoff() > 0) {
       // Checks if masking is in use, given that three.js implements masking in
--- a/packages/model-viewer/src/features/scene-graph/material.ts
+++ b/packages/model-viewer/src/features/scene-graph/material.ts
@@ -13,10 +13,10 @@
  * limitations under the License.
  */
 
-import {DoubleSide, FrontSide, MeshStandardMaterial} from 'three';
+import {DoubleSide, FrontSide, MeshStandardMaterial, NoBlending, NormalBlending} from 'three';
 
-import {GLTF, Material as GLTFMaterial} from '../../three-components/gltf-instance/gltf-2.0.js';
-import {ALPHA_TEST_DEFAULT} from '../../three-components/gltf-instance/ModelViewerGLTFInstance.js';
+import {AlphaMode, GLTF, Material as GLTFMaterial} from '../../three-components/gltf-instance/gltf-2.0.js';
+import {ALPHA_CUTOFF_DISABLED, ALPHA_CUTOFF_GLTF_DEFAULT} from '../../three-components/gltf-instance/ModelViewerGLTFInstance.js';
 
 import {Material as MaterialInterface, RGB} from './api.js';
 import {PBRMetallicRoughness} from './pbr-metallic-roughness.js';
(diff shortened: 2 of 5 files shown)


packages/model-viewer/src/test/features/scene-graph/material-spec.ts
```diff
@@ -117,4 +120,58 @@ suite('scene-graph/material', () => {
...
+      await waitForEvent(element, 'load');
+    };
+
+    test('test alpha cutoff expect default', async () => {
+      await loadModel(HELMET_GLB_PATH);
+      // Default value hack see:
+      // https://github.com/google/model-viewer/blob/e3a000111980f5cf018c61dec8695463e0d843a0/packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts#L210
+      expect(element.model!.materials[0].getAlphaCutoff())
+          .to.be.equal(ALPHA_TEST_DEFAULT);
+      element.model!.materials[0].setAlphaCutoff(0.25);
```
elalish, line 147: Actually, I don't think this is right. It's fine to set the cutoff to zero; if you look at the code for that hack, this negative value is actually what's set when blend mode equals opaque.

timmmeh: Sure that makes sense. But this edge case could be exposed given that we can now change the blend mode. Not that it makes sense in the first place, working around weird little worts is not ideal, I can just leave it out.

timmmeh: Wait so this must mean three.js is applying masking even when blend-mode is opaque. I guess that makes sense given that masking is a shader snippet, seems like maybe it should be compiled out when updating the material though.

elalish: Actually it's a little weirder. I believe the problem is that for three's opaque materials, if there's an alpha channel value, that value ends up in the output pixel color, which is weird on a transparent rendering. I think alpha cutoff = 0 is used by three to turn off the masking code in the shader. Instead, we use a negative value to turn it on and replace that shader chunk with our own: https://github.com/google/model-viewer/blob/master/packages/model-viewer/src/three-components/shader-chunk/alphatest_fragment.glsl.ts

That is where the alpha gets set back to fully opaque for the output, in order to be glTF spec-compliant.

timmmeh: Interesting, curious about the alpha channel value that ends up in the output color, is it just random garbage? There used to be a Unity bug like this.
This explains why the default blend-mode is always set to blending enabled.
So really negative alpha-mask is how we enable Opaque rendering.

packages/model-viewer/src/features/scene-graph/pbr-metallic-roughness.ts
```diff
@@ -61,7 +61,7 @@ export class PBRMetallicRoughness extends ThreeDOMElement implements
       metallicRoughnessTexture: gltfMetallicRoughnessTexture
     } = pbrMetallicRoughness;
 
-    const {map, metalnessMap} = correlatedMaterials.values().next().value;
+    let {map, metalnessMap} = correlatedMaterials.values().next().value;
```
elalish, line 64: was this change intentional?

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
```diff
@@ -117,4 +120,92 @@ suite('scene-graph/material', () => {
...
+
+    test('test alpha cutoff expect valid value as default', async () => {
+      await loadModel(ALPHA_BLEND_MODE_TEST);
+      expect(element.model!.materials[2].getAlphaCutoff()).to.be.equal(0.25);
+    });
+
+    test('test alpha cutoff test setting and getting', async () => {
+      await loadModel(ALPHA_BLEND_MODE_TEST);
+      element.model!.materials[2].setAlphaCutoff(0.0);
+      // Setting Zero-Alpha should revent to default settings.
```
elalish, line 157: revert (does anyone know of a good VSCode spellchecker? It must exist...)

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
```diff
@@ -117,4 +120,92 @@ suite('scene-graph/material', () => {
...
+    });
+
+    test('test alpha cutoff expect valid value as default', async () => {
+      await loadModel(ALPHA_BLEND_MODE_TEST);
+      expect(element.model!.materials[2].getAlphaCutoff()).to.be.equal(0.25);
+    });
+
+    test('test alpha cutoff test setting and getting', async () => {
+      await loadModel(ALPHA_BLEND_MODE_TEST);
+      element.model!.materials[2].setAlphaCutoff(0.0);
```
elalish, line 156: This is kind of a funny test because if you set a MASK mode material to a cutoff of 0.0, you end up getting a BLEND material, right? I don't think there's much we can do about that until three fixes things upstream, but maybe test a non-zero value?

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -154,4 +155,78 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+
+  getDoubleSided(): boolean {
+    return (this[$sourceObject] as GLTFMaterial).doubleSided ??
+        this[$correlatedObjects]!.values().next().value.side === DoubleSide;
+  }
+
+  setAlphaMode(alphaMode: AlphaMode): void {
+    for (const material of this[$correlatedObjects] as
+         Set<MeshStandardMaterial>) {
+      material.blending = NormalBlending;
```
elalish, line 197: Is this never set to another value?

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -154,4 +155,78 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+  getAlphaMode(): AlphaMode {
+    if ((this[$sourceObject] as GLTFMaterial).alphaMode !== undefined) {
+      return (this[$sourceObject] as GLTFMaterial).alphaMode!;
+    }
+
+    const blendMode = this[$correlatedObjects]!.values().next().value.blending;
+
+    if (blendMode === NormalBlending && this.getAlphaCutoff() >= 0) {
+      return 'BLEND';
+    } else if (blendMode === NoBlending && this.getAlphaCutoff() > 0) {
```
elalish, line 224: There's something funny here about checking `NoBlending`, but not setting it. Might mean we need another test too.

packages/model-viewer/src/features/scene-graph/api.ts
```diff
@@ -60,7 +60,10 @@ export declare interface Material {
 
   readonly emissiveFactor: Readonly<RGB>;
   setEmissiveFactor(rgb: RGB): void;
-
+  setAlphaCutoff(cutoff: number): void;
+  getAlphaCutoff(): number;
+  setDoubleSided(doubleSided: boolean): void;
+  getDoubleSided(): boolean;
```
elalish, line 66: The new APIs should be here too.

elalish (changes requested):

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: No, I had noticed that materials that should be 'opaque' (according to the file spec) are actually set to 'blend' on import, my assumption was that this was intentional, because 'blend' In combination with cutoff hack (set to -0.5) results in an opaque material. However, no that you mention it, this would probably interfere with render performance, on large scenes anyway, where sorting front-to-back actually makes a difference.

packages/model-viewer/src/features/scene-graph/pbr-metallic-roughness.ts
timmmeh: weird

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
timmmeh: Honestly, why is that not there by default.
I installed [this one](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker)
And _apparently_ 'revent' is not a word.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
timmmeh: Really it just needs to read the same value that was set, but in the crazy world, I see your point.

packages/model-viewer/src/features/scene-graph/api.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: Hmm, I realize now that we can just return 'OPAQUE' if `alphaMode` is not defined. This is because the spec says `alphaMode` default is OPAQUE, so everything else has to have a defined value.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -154,4 +155,70 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+  getDoubleSided(): boolean {
+    return (this[$sourceObject] as GLTFMaterial).doubleSided ??
+        this[$correlatedObjects]!.values().next().value.side === DoubleSide;
+  }
+
+  setAlphaMode(alphaMode: AlphaMode): void {
+    for (const material of this[$correlatedObjects] as
+         Set<MeshStandardMaterial>) {
+      if (alphaMode === 'OPAQUE') {
+        material.blending = NoBlending;
```
elalish, line 198: I still don't buy this. Have you looked at how three.js sets up materials based on glTFLoader? They do not seem to change `blending`, but they do change `transparent` and `depthWrite`: https://github.com/mrdoob/three.js/blob/dev/examples/jsm/loaders/GLTFLoader.js#L3092-L3111

elalish (changes requested):

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: Hmm, I see. I guess there are multiple ways to set this, changing the blend mode does work in practice, I've been verifying it against the AlphaBlendTest model. But it looks like three wants us to use the transparent setting.

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: From the docs it sounds as though Three uses the `transparent` flag to do sorting.
https://threejs.org/docs/#api/en/materials/Material.transparent

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: Setting depth write to false for transparent objects seems like a hammer,  that's why they are sorted back-to-front, but I guess it's the best way to ensure minimal artifacts on tricky models.

elalish: Yeah, three is pretty complicated (and has duplicate features that get deprecated over time), but we'd better stick very close to what GLTFLoader does, or we're likely to hit strange edges cases. And yes, there was a long debate about depthWrite false.

elalish (approved):
Thanks!

dhruvadhia1:
It would be great to get new examples for some of these features. Thanks

elalish:
@dhruvadhia1 Would you like to contribute some? These are pretty advanced features, probably only useful to others writing their own web editors. They would likely get more from looking through our own editor code, which is being updated to use all of this: #2469

dhruvadhia1:
@elalish Sure will look into that.

PR #2619, opened 2021-07-27
Separate scaling from rotating on WebXR

felipe-chamas:
// one can happen at a time, but we can switch during a gesture.

With two finger interaction, scaling and rotating are not really mutually exclusive.

#### Description
 I've created a threshold to detect if it's a real rotation or a small movement when pulling fingers apart. `ROTATION_THRESHOLD` was chosen based on some experimentation locally, but I would like to open this value for discussion.
If the threshold is crossed the rotation is active, if not then in the first frame we set the new `firstRatio` value for scaling.
@elalish , I've also changed the function name I pinpointed yesterday 😅

### DIFF
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -79,9 +79,9 @@ const hitPosition = new Vector3();
 const camera = new PerspectiveCamera(45, 1, 0.1, 100);
 
 export class ARRenderer extends EventDispatcher {
-  threeRenderer: WebGLRenderer;
-  currentSession: XRSession|null = null;
-  placeOnWall = false;
+  public threeRenderer: WebGLRenderer;
+  public currentSession: XRSession|null = null;
+  public placeOnWall = false;
 
   private placementBox: PlacementBox|null = null;
   private lastTick: number|null = null;
@@ -474,20 +474,19 @@ export class ARRenderer extends EventDispatcher {
     session.addEventListener('selectend', this.onSelectEnd);
     session
         .requestHitTestSourceForTransientInput({profile: 'generic-touchscreen'})
-        .then((hitTestSource) => {
+        .then(hitTestSource => {
           this.transientHitTestSource = hitTestSource;
         });
   }
 
   private getTouchLocation(): Vector3|null {
     const {axes} = this.inputSource!.gamepad;
-    const location = this.placementBox!.getExpandedHit(
+    let location = this.placementBox!.getExpandedHit(
         this.presentedScene!, axes[0], axes[1]);
     if (location != null) {
       vector3.copy(location).sub(this.presentedScene!.getCamera().position);
-      if (vector3.length() > MAX_DISTANCE) {
+      if (vector3.length() > MAX_DISTANCE)
         return null;
-      }
     }
     return location;
   }
@@ -512,7 +511,7 @@ export class ARRenderer extends EventDispatcher {
         null;
   }
 
-  moveToFloor(frame: XRFrame) {
+  public moveToPlane(frame: XRFrame) {
     const hitSource = this.initialHitSource;
     if (hitSource == null) {
       return;
@@ -734,7 +733,7 @@ export class ARRenderer extends EventDispatcher {
   /**
    * Only public to make it testable.
    */
-  onWebXRFrame(time: number, frame: XRFrame) {
+  public onWebXRFrame(time: number, frame: XRFrame) {
     this.frame = frame;
     ++this.frames;
     const refSpace = this.threeRenderer.xr.getReferenceSpace()!;
@@ -760,12 +759,12 @@ export class ARRenderer extends EventDispatcher {
     // isn't really supported at this point, but make a best-effort
     // attempt to render other views also, using the first view
     // as the main viewpoint.
-    let isFirstView = true;
+    let isFirstView: boolean = true;
     for (const view of pose.views) {
       this.updateView(view);
 
       if (isFirstView) {
-        this.moveToFloor(frame);
+        this.moveToPlane(frame);
 
         this.processInput(frame);
 
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -628,8 +628,10 @@ export class ARRenderer extends EventDispatcher {
         this.isTwoFingering = false;
       } else {
         const {separation, deltaYaw} = this.fingerPolar(fingers);
-        this.isRotating = Math.abs(deltaYaw) > ROTATION_THRESHOLD;
-        if (!this.isRotating && !this.isScaling) {
+        if (Math.abs(deltaYaw) > ROTATION_THRESHOLD) {
+          this.isScaling = false;
+          this.isRotating = true;
+        } else if (!this.isScaling) {
           this.isScaling = true;
           this.firstRatio = separation / scene.scale.x;
         }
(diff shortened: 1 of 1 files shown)


FelipeR2U:
@googlebot I signed it!

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -33,6 +33,7 @@ const INIT_FRAMES = 30;
 // estimation, which will be used once available in WebXR.
 const AR_SHADOW_INTENSITY = 0.3;
 const ROTATION_RATE = 1.5;
+const ROTATION_THRESHOLD = 0.05;
```
elalish, line 36: Thanks for the PR! Can you describe why you think this UX is an improvement? I was following more the Google Maps style where both actions happen together. There is already a 30% snap on scaling to make it sticky near 100%. Also I'd always recommend setting `ar-scale: "fixed"` to disable scaling entirely for products that are meant to be seen at 100%.

FelipeR2U: We do use `ar-scale: "fixed"` whenever we need the product scale precision!
I'm checking a lot on the WebXR code to make our migration and when I read the comment of actions being independent it felt awkward trying to rotate a model without scaling, mostly because it's hard to maintain a constant distance while rotating two fingers. I've also realized I personally use fingers from both hands in opposite directions to rotate, which always make the model shrink then expand. The same movement has always worked nicely on scene-viewer where movements are not interchangeable during a two-finger interaction.
On the other hand, a little bit of rotation while scaling feels acceptable so it's mainly for the reasons above.
Maybe I'm being picky on something that's actually a good feature for most users, since there's already a precise 1 finger rotation, so I don't mind leaving this PR for some extra research😁

elalish: Thanks for the explanation! I'll leave this for some further discussion for now.

FelipeR2U: > I'll leave this for some further discussion for now.
Awesome, thanks for opening it in the thread!

maciekglowka: Well, I do agree that slight rotation while scaling is not an issue, but accidentally scaling while trying to rotate is not desired. So if I understood correctly it should be definitely an improvement :) (when rotating beyond threshold scaling is disabled, right ?)

FelipeR2U: > when rotating beyond threshold scaling is disabled, right ?
That's it!

maciekglowka: For me it is clear now and I support the change :) But maybe for the others :) ?

FelipeR2U: https://demos.r2u.io/model-viewer/index.html

FelipeR2U: Online simplified [demo](https://demos.r2u.io/model-viewer/index.html) for testing

elalish:
@lucadalli I'd be interested to get your opinion on this. It only affects WebXR mode, so just Android for now, so at least you can test it.

lucadalli:
I don't use the WebXR viewer, nor have I have ever worked with WebXR input events, so I'm assuming the same capabilities as touch events.

I can agree with the sentiment to make rotation and scaling distinct. In fact I would be inclined to have 1 finger for rotation and 2 fingers exclusively for scaling, similarly to `camera-controls`.
However, when interacting with the AR model, I find myself hovering my fingers directly over the placement box and performing all interactions there. With mutually exclusive interactions, since a single touch on the placement box moves the model, to rotate the model I would be forced to put my finger outside the placement box. I find this to be awkward not only because of the uncomfortable finger position but also because of the additional cognitive load associated with recognizing where the placement box is so that I can press outside of it.

This brings me to the changes proposed in this PR. They make rotation and scaling somewhat distinct, without compromising the ability to use 2 fingers within the placement box to rotate the model. I do consider it an improved UX.
Alternative/supplementary behavior comes to mind, which I'd like to share.

To allow for scale adjustments without impacting rotation we can make the interactions which start **outside** the placement box mutually exclusive.
The user is already introduced to the pattern that the same interaction outside and inside the placement box produces different behaviors. A single touch in the placement box moves the model, while a single touch outside rotates it. 

Similarly, 2 fingers inside the placement box both rotate and scale, while 2 fingers outside would only scale.

On top of all this, I think we should also introduce the ability to alternate between the 1 finger interaction and the 2 finger interaction without having to reset by lifting all fingers, like we just did with `camera-controls` (merged but unreleased, observable by interacting with any of the model-viewer elements in the docs).
In other words, for an interaction which started outside the placement box, lifting one finger after scaling the model would resume rotation and adding a finger during rotation would initiate scaling.
For an interaction which started inside the placement box, lifting one finger after scaling/rotating would resume model placement and adding a finger during placement would initiate scaling/rotating.

This leaves us with 4 possible behaviors. Behaviors in the same column can be alternated between by adding/removing the second finger. (For simplicity's sake, inside or outside of the placement box would be determined by the first touch. It is not recalculated until all fingers are lifted and a new interaction is started.)
To help the user distinguish between the two sets of behaviors, we could highlight the placement box a different color (blue?) when the touches are inside the placement box.

Touches | Inside Placement Box | Outside Placement Box |
| --- | --- | --- |
| 1 | move | rotate |
| 2 | rotate/scale (possibly with threshold) | scale |

I am not strongly against or in favor of this PR, nor do I feel strongly about the alternative I suggested. I'm sharing my thoughts to further the discussion.

lucadalli:
I'm not sure if this is some intended behavior which is being erratic but it seems like a bug to me.
When testing the [demo linked above](https://demos.r2u.io/model-viewer/index.html) by @FelipeR2U, sometimes I am able to double tap box corners which are diagonally opposite and make the model do a 180. I am able to reproduce on `master` but it's far less common.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -569,19 +571,18 @@ export class ARRenderer extends EventDispatcher {
     } else if (fingers.length === 2) {
       box.show = true;
       this.isTwoFingering = true;
-      const {separation} = this.fingerPolar(fingers);
-      this.firstRatio = separation / scene.scale.x;
```
FelipeR2U, line 573: > 
> 
> I'm not sure if this is some intended behavior which is being erratic but it seems like a bug to me.
> When testing the [demo linked above](https://demos.r2u.io/model-viewer/index.html) by @FelipeR2U, sometimes I am able to double tap box corners which are diagonally opposite and make the model do a 180. I am able to reproduce on `master` but it's far less common.

Well noticed, I had the same issue sometimes and realized I should probably rollback this change so that `this.lastAngle` is updated.

FelipeR2U:
@lucadalli - About your first Comment
Thanks for the inputs, I really appreciate all the study you proposed and I mostly agree! I'm just concerned about the difficulty for a user to put both fingers inside the placement box, since it's not the whole model bounding-box, but only an extension of it's base. Also keeping in mind that the box purpose is to detect a `XRTransientInputHitTestResult` (finger touch hits the model-base?) and that `scale/rotate` keep the model in place, maybe those actions should not be attached to it.
If the proposed table is in fact accepted, I'm a little divided about which action should occur inside/outside the box for 2-touch.
Finally, I'd also like to highlight the current UX of Scene-Viewer: Once the 2-touch action is decided, we can only switch to the other if we start another interaction, which is in my opinion a better experience in fine tuning the model rotation/size. I didn't make the PR like this because of the same comment:
_// Rotating, translating and scaling are mutually exclusive operations; only one can happen at a time, **but we can switch during a gesture.**_

elalish:
Sorry I forgot about this one! I'm in the midst of switching to pointer events, and still open to this once that lands.

felipe-chamas:
> Sorry I forgot about this one! I'm in the midst of switching to pointer events, and still open to this once that lands.

Sorry, I was organizing my GH account and closed some of my PRs, I'm gonna reopen it for further discussion

PR #2620, opened 2021-07-28
Fix iOS 13 automated testing

elalish:
Based on some info from BrowserStack support. It looks like their bugs are fixed so we can finally remove the karma hacks (yay!). They also recommended unique identifiers which is supposed to help somehow. Anyway, we'll know it worked if iOS 13 (and the rest) are no longer flaky.


elalish:
FYI @timmmeh @asemanh

elalish:
Well, for some bizarre reason this is now failing on the editing adapter tests (outside of BrowserStack entirely). I may just wait and merge this after my editor refactor which removes that package. Should be landing fairly soon.

asemanh (approved):

PR #2621, opened 2021-07-28
Minor fix for singleton pattern

asemanh:


elalish (approved):
Yeah, good idea. We tend to be a bit lazy sometimes with the TS private field since it gets stripped in the JS build anyway, but it's still a good practice.

PR #2632, opened 2021-08-02
three.js r131

elalish:
@mrdoob Looking good! Unfortunately I can't yet remove PMREMGenerator from our code base, since iOS Safari 14 doesn't support FloatType and without it the RGBE backgrounds become nearest interpolated (which they shouldn't; we have this solved in model-viewer). Also we don't yet have internal support for `fromScene()` for lighting.


elalish:
@asemanh I've updated the model-viewer goldens here for the new version of three.js; figured you might want to take a look at how rendering has improved.

PR #2633, opened 2021-08-02
Add powerPreference option

asemanh:
This PR only adds the powerPreference option, we default to high-performance to keep the behavior compatible as before.

https://github.com/google/model-viewer/issues/2541

I can either send a PR or another commit once I have the test working


packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -55,15 +58,31 @@ const DEFAULT_LAST_STEP = 3;
...
+  }
+
+  static getDefaultPowerPreference() : string {
+    return DEFAULT_POWER_PREFERENCE;
+  }
 
   static get singleton() {
+    if (this._singleton == null) {
+      this._singleton=  new Renderer({debug: isDebugMode(), powerPreference: this.powerPreference});
+    }
```
elalish, line 79: Why is this needed?

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -55,15 +58,31 @@ const DEFAULT_LAST_STEP = 3;
...
+    this._powerPreference = powerPreferenceOption || DEFAULT_POWER_PREFERENCE;
+  }
+  
+  static get powerPreference() {
+    return this._powerPreference;
+  }
+
+  static getDefaultPowerPreference() : string {
+    return DEFAULT_POWER_PREFERENCE;
+  }
```
elalish, line 74: you can just `export const DEFAULT_POWER_PREFERNCE` instead of making a getter.

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -55,15 +58,31 @@ const DEFAULT_LAST_STEP = 3;
...
+  private static _powerPreference = DEFAULT_POWER_PREFERENCE;
+  private static _singleton : Renderer;
+
+  static set powerPreference(powerPreferenceOption:string) {
+    this._powerPreference = powerPreferenceOption || DEFAULT_POWER_PREFERENCE;
+  }
+  
+  static get powerPreference() {
+    return this._powerPreference;
+  }
```
elalish, line 70: If we have both a getter and a setter that don't have logic in them, we might as well just have a public member and skip the boiler plate.

elalish: Also, is it actually possible to set the powerPreference outside of the constructor and have it affect anything? If not, then maybe it shouldn't be a member at all.

elalish (changes requested):

packages/model-viewer/src/three-components/Renderer.ts
asemanh: I want to force setting powerPreference before singleton gets initialed.
But also, that's the pattern recommended and wont' change the existing behavior.
https://basarat.gitbook.io/typescript/main-1/singleton

packages/model-viewer/src/three-components/Renderer.ts
asemanh: ah ok, that makes sense, I was wondering how to access it.

packages/model-viewer/src/three-components/Renderer.ts
asemanh: So is it only recommended to have setters/getters only if they have logic in them?

asemanh: I'm not quiet sure what you mean by your other comment (having powerPreference outside of the constructor)

elalish: I would say use setters and getters when they are necessary. Likewise, rather than having this state in two separate static variables, perhaps just use the ModelViewer one directly so you don't need it here too.

I see now what you did with the singleton, though it does seem to have changed some behavior since a test is failing now. Perhaps it needs to be prodded into being instantiated?

packages/model-viewer/src/features/loading.ts
```diff
@@ -310,6 +311,8 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
       const ModelViewerElement: ModelViewerGlobalConfig =
           (self as any).ModelViewerElement || {};
 
+      Renderer.powerPreference = ModelViewerElement.powerPreference || DEFAULT_POWER_PREFERENCE;
```
elalish, line 314: I think you still need to define the `powerPreference` property somewhere, and probably better in `model-viewer-base` than `loading`.

packages/modelviewer.dev/examples/loading/index.html
```diff
@@ -330,7 +330,37 @@ <h4>
...
+              the models.
+            </h4>
+          </div>
+          <example-snippet stamp-to="powerPrefSupport" highlight-as="html">
+            <template>
+<model-viewer camera-controls alt="A 3D model of a fish" src="../../shared-assets/models/BarramundiFish.mixed.glb">
+</model-viewer>
+<script>
+ModelViewerElement = self.ModelViewerElement || {};
+ModelViewerElement.powerPreference = "low-power";
```
elalish, line 356: Do you have any thoughts on how to make an automated test?

packages/model-viewer/src/three-components/Renderer.ts
asemanh: I updated the setters/getters, but I'm not convinced why my change should break the test

packages/model-viewer/src/features/loading.ts
asemanh: It seems to be working when I check in debugger for the case I introduced in examples.

elalish: How did you check?

elalish: I would recommend checking this: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getContextAttributes

PR #2639, opened 2021-08-03
don't show AR button when there's no src

elalish:
Fixes a bug where the AR button showed up on android even if there was no `src`.

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -217,28 +217,26 @@ configuration or device capabilities');
     async[$selectARMode]() {
       this[$arMode] = ARMode.NONE;
       if (this.ar) {
-        const arModes: ARMode[] = [];
-        this[$arModes].forEach((value) => {
-          arModes.push(value);
-        });
-
-        for (const value of arModes) {
-          if (value === 'webxr' && IS_WEBXR_AR_CANDIDATE && !isWebXRBlocked &&
-              await this[$renderer].arRenderer.supportsPresentation()) {
-            this[$arMode] = ARMode.WEBXR;
-            break;
-          } else if (
-              value === 'scene-viewer' && IS_SCENEVIEWER_CANDIDATE &&
-              !isSceneViewerBlocked) {
-            this[$arMode] = ARMode.SCENE_VIEWER;
-            break;
-          } else if (value === 'quick-look' && IS_AR_QUICKLOOK_CANDIDATE) {
-            this[$arMode] = ARMode.QUICK_LOOK;
-            break;
+        if (this.src != null) {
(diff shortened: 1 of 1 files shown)


packages/model-viewer/src/features/ar.ts
```diff
@@ -217,28 +217,26 @@ configuration or device capabilities');
...
-              value === 'scene-viewer' && IS_SCENEVIEWER_CANDIDATE &&
-              !isSceneViewerBlocked) {
-            this[$arMode] = ARMode.SCENE_VIEWER;
-            break;
-          } else if (value === 'quick-look' && IS_AR_QUICKLOOK_CANDIDATE) {
-            this[$arMode] = ARMode.QUICK_LOOK;
-            break;
+        if (this.src != null) {
+          for (const value of this[$arModes]) {
+            if (value === 'webxr' && IS_WEBXR_AR_CANDIDATE && !isWebXRBlocked &&
```
asemanh, line 222: not related to this PR, but this is hard to read and understand. Maybe as i read the code more, it makes more sense, but it's strange that the code is designed such that we need to have the check for value and then the other global variables.

packages/model-viewer/src/features/ar.ts
```diff
@@ -217,28 +217,26 @@ configuration or device capabilities');
...
+                await this[$renderer].arRenderer.supportsPresentation()) {
+              this[$arMode] = ARMode.WEBXR;
+              break;
+            } else if (
+                value === 'scene-viewer' && IS_SCENEVIEWER_CANDIDATE &&
+                !isSceneViewerBlocked) {
```
asemanh, line 228: Nit. Since we're touching this code, a better approach could be as below. Having the else part only makes the code harder to read and based on the logic we can simply have:
`
for (const value of arModes) {
          if (value === 'webxr' && IS_WEBXR_AR_CANDIDATE && !isWebXRBlocked &&
              await this[$renderer].arRenderer.supportsPresentation()) {
            this[$arMode] = ARMode.WEBXR;
            break;
          } 
          if (
              value === 'scene-viewer' && IS_SCENEVIEWER_CANDIDATE &&
              !isSceneViewerBlocked) {
            this[$arMode] = ARMode.SCENE_VIEWER;
            break;
          } 
          if (value === 'quick-look' && IS_AR_QUICKLOOK_CANDIDATE) {
            this[$arMode] = ARMode.QUICK_LOOK;
            break;
          }
        }`

asemanh (approved):

packages/model-viewer/src/features/ar.ts
elalish: Good call, done.

asemanh (approved):

PR #2640, opened 2021-08-03
Editor fixes

elalish:
This fixes a couple of small bugs I noticed after my refactor, namely that the texture icons weren't showing up at first and the autoplay button was broken.

### DIFF
--- a/packages/space-opera/src/components/animation_controls/animation_controls.ts
+++ b/packages/space-opera/src/components/animation_controls/animation_controls.ts
@@ -47,13 +47,6 @@ export class AnimationControls extends ConnectedLitElement {
     this.autoplay = !!config.autoplay;
   }
 
-  // Specifically overriding a super class method.
-  // tslint:disable-next-line:enforce-name-casing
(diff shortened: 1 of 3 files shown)


packages/space-opera/package.json
```diff
@@ -36,13 +36,13 @@
...
-    "@material/mwc-tab": "^0.21.0",
-    "@material/mwc-tab-bar": "^0.21.0",
-    "@material/mwc-textfield": "^0.21.0",
+    "@material/mwc-button": "^0.18.0",
+    "@material/mwc-checkbox": "^0.18.0",
+    "@material/mwc-icon-button": "^0.18.0",
+    "@material/mwc-switch": "^0.18.0",
+    "@material/mwc-tab": "^0.18.0",
+    "@material/mwc-tab-bar": "^0.18.0",
+    "@material/mwc-textfield": "^0.18.0",
```
elalish, line 45: I had to downgrade mwc to get our tabs to stop scrolling. There should be a CSS way to fix it in the new version, but I couldn't figure it out.

PR #2641, opened 2021-08-03
different approach for setting powerPreference. 

asemanh:
This currently doesn't work because Render is already initalized/created then we read the script global variables and I'm not sure how else to approach it.


packages/modelviewer.dev/examples/loading/index.html
```diff
@@ -41,386 +41,415 @@
 
 <body>
 
-<div class="examples-page">  
-  <div class="sidebar" id="sidenav"></div>
-  <div id="toggle"></div>
+  <div class="examples-page">
```
elalish, line 44: I forgot to mention; we need to turn off HTML formatting in VSCode because it'll mess up how the examples look. I think you can just revert this since I don't think we're planning to add anything here, right?

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -110,6 +110,8 @@ export interface RendererInterface {
 export default class ModelViewerElementBase extends UpdatingElement {
   protected static[$template]: HTMLTemplateElement|void;
 
+  public static powerPreference:string = 'high-perfromance';
```
elalish, line 113: Regarding your question; perhaps a static set method that calls `resetSingleton()` if the value has changed? If they set it before the element is loaded, it should still start up correctly without a reset (though I agree our automated test won't be covering that situation).

packages/model-viewer/src/utilities.ts
```diff
@@ -198,6 +198,15 @@ export const isDebugMode = (() => {
        self.location.search.match(debugQueryParameter));
 })();
 
+/**
+ * @returns high-perofmance by default. However, developer has the option to set this.
+ * In that case we return developer's choice
+ */
+export const getPowerPreference = ():string => {
```
elalish, line 205: I think you can remove this method since you're setting it to the default value as it's constructed. If it's really not defined when you need it, we should probably understand why.

elalish (changes requested):

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -308,6 +326,12 @@ export default class ModelViewerElementBase extends UpdatingElement {
       // at all times:
       this[$isElementInViewport] = true;
     }
+    const ModelViewerGlobalConfig: ModelViewerGlobalConfig =
+        (self as any).ModelViewerElement || {};
+    if (ModelViewerGlobalConfig.powerPreference != null) {
+      ModelViewerElementBase.powerPreference =
+          ModelViewerGlobalConfig.powerPreference;
```
elalish, line 333: `ModelViewerGlobalConfig` is an interface, which means it's just a type, not a real object. It's a decorator to help TS know that `ModelViewerElementBase.powerPreference` exists, but I don't think this logic is doing what you expect.

packages/model-viewer/src/test/model-viewer-base-spec.ts
```diff
@@ -99,6 +99,30 @@ suite('ModelViewerElementBase', () => {
...
+          element.parentNode.removeChild(element);
+        }
+      });
+
+      test(
+          'validate threeRenderer is using the given powerPreference',
+          async () => {
+            const {threeRenderer} = Renderer.singleton;
+            expect(threeRenderer.getContextAttributes().powerPreference)
+                .to.be.equal('low-power');
```
elalish, line 122: 👍

packages/model-viewer/src/test/model-viewer-base-spec.ts
```diff
@@ -99,6 +99,30 @@ suite('ModelViewerElementBase', () => {
...
+
+      setup(() => {
+        element = new ModelViewerElement();
+        document.body.insertBefore(element, document.body.firstChild);
+      });
+
+      teardown(() => {
+        if (element.parentNode != null) {
+          element.parentNode.removeChild(element);
+        }
```
elalish, line 114: This is fine, but especially in tests I'll often use the shorthand `element.parentNode!.removeChild(element);` since that's just telling TS I'm sure it exists. If it doesn't, the test will fail anyway.

elalish (changes requested):

packages/model-viewer/src/model-viewer-base.ts
asemanh: It's true that ModelViewerGlobalConfig is an interface. What it does here is that I say hay in the window/global space I'm going to have a modelviewerelement. Sometimes this object may have initialized this static member. If that's the case, read it. This is literally the pattern used for the other two static properties. Otherwise, how do you explain how this works in loading.ts for draco.

elalish: Oh, good call.  I just read over the similar code in loading.ts, which apparently I haven't looked at in a long time. Thanks for the explanation!

packages/model-viewer/src/test/model-viewer-base-spec.ts
asemanh: thanks for sharing.

packages/modelviewer.dev/examples/loading/index.html
```diff
@@ -30,7 +30,7 @@
   <script src="../../node_modules/focus-visible/dist/focus-visible.js" defer></script>
 
   <script>
-    window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
```
asemanh, line 33: I tried reverting this, but not sure why I still see the changes :/ 
git checkout origin/master ./packages/modelviewer.dev/index.html

elalish: I think you're looking for `git checkout origin/master -- ./packages/modelviewer.dev/index.html` a subtle but important distinction.

asemanh: I tried that too, didn't work :/

PR #2643, opened 2021-08-04
cleanup and fixed an onUpdate variants bug

elalish:
@timmmeh I know you're out this week, but I'd like you to review this when you get back so the changes don't surprise you.

### DIFF
--- a/packages/model-viewer/src/features/scene-graph.ts
+++ b/packages/model-viewer/src/features/scene-graph.ts
@@ -127,8 +127,7 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
       return new ModelViewerTexture(this[$getOnUpdateMethod](), texture);
     }
 
-
-    updated(changedProperties: Map<string, any>) {
+    async updated(changedProperties: Map<string, any>) {
       super.updated(changedProperties);
 
       if (changedProperties.has('variantName')) {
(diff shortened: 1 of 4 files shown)


packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -139,21 +138,19 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
         }
 
-        const updatedMaterialsPromise =
-            threeGLTF.correlatedSceneGraph.loadVariant(
-                variantName!, this[$getOnUpdateMethod]);
+        const updatedMaterials =
+            await threeGLTF.correlatedSceneGraph.loadVariant(variantName!);
         const {gltf, gltfElementMap} = threeGLTF.correlatedSceneGraph;
 
-        updatedMaterialsPromise.then(updatedMaterials => {
```
elalish, line 147: Switching from `.then` to async/await since it's a bit easier to read and it matches our style.

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -139,21 +138,19 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-        });
+        for (const index of updatedMaterials) {
+          const material = gltf.materials![index];
+          this[$model]!.materials[index] = new Material(
+              this[$getOnUpdateMethod](),
+              gltf,
+              material,
+              gltfElementMap.get(material) as Set<MeshStandardMaterial>);
+        }
+        this[$needsRender]();
```
elalish, line 153: I made this PR to fix a bug, and this line was the actual fix. The render was not being updated after a variant was loaded.

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -195,9 +192,8 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
 
         if (correlatedSceneGraph != null &&
             currentGLTF !== this[$currentGLTF]) {
-          this[$model] = new Model(correlatedSceneGraph, () => {
-            this[$needsRender]();
-          });
+          this[$model] =
+              new Model(correlatedSceneGraph, this[$getOnUpdateMethod]());
```
elalish, line 196: for consistency

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -123,27 +123,6 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
         gltf,
         gltfEmissiveTexture ? gltfEmissiveTexture : null,
     );
-
-    const message = (textureType: string) => {
-      console.info(`A group of three.js materials are represented as a
-        single material but share different ${textureType} textures.`);
```
elalish, line 129: These logs were being recorded for the variant example, which indicates to me this is not really an error. I went ahead and removed them for now to keep from polluting the console.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -284,7 +283,6 @@ export class CorrelatedSceneGraph {
             this.gltfElementMap.set(gltfElement, threeObjects);
           }
           threeObjects.add((object as Mesh).material as Material);
-          onUpdate();
```
elalish, line 287: I removed this `onUpdate` function, because we don't need to request a new render on every visit to a scene node, plus is wasn't doing what we wanted anyway.

PR #2645, opened 2021-08-05
Editor UX improvements

### LINKED ISSUE #1508: [Editor] Animations stop playing when dropping a model with animations for the 2nd time
### Description

When dropping an animated GLB into modelviewer.dev/editor, the animation plays in a loop. However, dropping that model for the 2nd time (or, as in my case, a new file with some fixes) seems to break animation playback with no way to recover (only page reload).

#### Live Demo
[AR-Teapot.zip](https://github.com/google/model-viewer/files/5173895/AR-Teapot.zip)
Drop this into modelviewer.dev/editor, note anim plays
Drop it again, note anim does not play
Toggle "play animation" or "autoplay", anim does not play

elalish:
Fixes #1508 and a bunch of other small issues. I removed some UI that I find distracting (since there's hardly any reason to change it). Also removed USDZ upload since we do auto-generation now for Quick Look.

### DIFF
--- a/packages/space-opera/src/components/best_practices/styles.css.ts
+++ b/packages/space-opera/src/components/best_practices/styles.css.ts
@@ -75,7 +75,7 @@ export const arButtonCSS: CSSResult = css`
   left: 50%;
   transform: translateX(-50%);
   white-space: nowrap;
-  bottom: 132px;
+  bottom: 16px;
   padding: 0px 16px 0px 40px;
   font-family: Roboto Regular, Helvetica Neue, sans-serif;
   font-size: 14px;
@@ -112,7 +112,7 @@ export const arPromptCSS: CSSResult = css`
 model-viewer > #ar-prompt {
   position: absolute;
   left: 50%;
-  bottom: 175px;
+  bottom: 60px;
   animation: elongate 2s infinite ease-in-out alternate;
   display: none;
 }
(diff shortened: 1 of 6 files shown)


packages/space-opera/src/components/camera_settings/camera_settings.ts
```diff
@@ -246,14 +241,6 @@ export class CameraSettings extends ConnectedLitElement {
     return html`
     <me-expandable-tab tabName="Camera Setup" .open=${true}>
       <div slot="content">
-        <me-checkbox id="cam-controls-checkbox" label="Interactive camera"
```
chrismgeorge, line 249: So we're just removing the ability to change this entirely? (Making the assumption that 99% of the time, people want this enabled.)

packages/space-opera/src/components/best_practices/styles.css.ts
```diff
@@ -75,7 +75,7 @@ export const arButtonCSS: CSSResult = css`
   left: 50%;
   transform: translateX(-50%);
   white-space: nowrap;
-  bottom: 132px;
+  bottom: 16px;
```
chrismgeorge, line 78: What are the visual changes that this change is making? Here and below?

packages/space-opera/src/components/camera_settings/camera_settings.ts
elalish: Yeah, we already only expose a subset of our API as UX, and this didn't feel deserving of being included. Likewise, I have AR on by default instead of selectable.

packages/space-opera/src/components/best_practices/styles.css.ts
elalish: These were copied from the WebXR example, which has then really high off the bottom to make room for the carousel, but since that's not here, I decided to put them back down so they're less obtrusive.

PR #2656, opened 2021-08-05
adding powerPreference to Moelviewer static options

asemanh:


packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -109,6 +110,21 @@ export interface RendererInterface {
...
   protected static[$template]: HTMLTemplateElement|void;
+  private static _powerPreference: string = 'high-performance';
+
+  /** @export */
+  static set powerPreference(value: string) {
+    if (ModelViewerElementBase._powerPreference == value) {
+      return;
+    }
+    ModelViewerElementBase._powerPreference = value;
+    Renderer.resetSingleton();
```
elalish, line 121: There's some kind of bug here, but I think it's likely a long-standing issue with resetSingleton, since that's not a well-tested method. There are now errors, the but the render is just blank afterward. Let me see if I can debug (since I actually have two GPUs).

elalish (changes requested):

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -55,15 +56,27 @@ const DEFAULT_LAST_STEP = 3;
...
+      element.disconnectedCallback();
+    }
+
+    this._singleton = new Renderer({
+      powerPreference: ModelViewerElementBase.powerPreference,
+      debug: isDebugMode()
+    });
+
+    for (const element of elements) {
+      element.connectedCallback();
```
elalish, line 78: disconnecting any existing model-viewer element from the old singleton and reconnecting them to the new one was all that was required to get the state back in shape. If they already had a loaded model, I'd guess it'll still be broken until they reset their `src`, but I think I'm okay with that, as it'll serve as a reminder that they really shouldn't be resetting after they have models loaded.

packages/model-viewer/src/model-viewer-base.ts
asemanh: This is not applicable anymore after your commit I assume

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -308,6 +324,12 @@ export default class ModelViewerElementBase extends UpdatingElement {
       // at all times:
       this[$isElementInViewport] = true;
     }
+    const ModelViewerGlobalConfig: ModelViewerGlobalConfig =
+        (self as any).ModelViewerElement || {};
+    if (ModelViewerGlobalConfig.powerPreference != null) {
+      ModelViewerElementBase.powerPreference =
+          ModelViewerGlobalConfig.powerPreference;
+    }
```
elalish, line 332: So, after some testing, it turns out we can revert all the changes to this file. Your code is great and does what you want, but we're defeated by the browser. The powerPreference is only a preference, and it turns out if we create a 3D context on the high-performance GPU, then destroy it and create one with a "low-power" preference, the browser says "well, we're already using the high-performance GPU, so we don't need to bother switching just for a preference". So, it turns out we have to set powerPreference before the model-viewer library is loaded or it can't work at all, which we should make clear in the docs.

packages/model-viewer/src/test/model-viewer-base-spec.ts
```diff
@@ -99,6 +99,28 @@ suite('ModelViewerElementBase', () => {
...
+      });
+
+      test(
+          'validate threeRenderer is using the given powerPreference',
+          async () => {
+            const {threeRenderer} = Renderer.singleton;
+            expect(threeRenderer.getContextAttributes().powerPreference)
+                .to.be.equal('low-power');
+          });
+    });
```
elalish, line 122: Another side effect is that this will no longer be auto-testable. This also describes a common issue we have with testing in the browser, which is that a true end-to-end test that really checks what you care about can be pretty impossible in some situations.

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -56,15 +59,27 @@ const DEFAULT_LAST_STEP = 3;
  * the texture.
  */
 export class Renderer extends EventDispatcher {
-  private static _singleton = new Renderer({debug: isDebugMode()});
+  private static _singleton = new Renderer(
+      {powerPreference: DEFAULT_POWER_PREFERENCE, debug: isDebugMode()});
```
elalish, line 63: It turns out this is the only place we need to query for the setting. And setting the value ahead of time was the thing I needed help with. Turns out our docs have a typo (sort of). 

The docs say to set static properties before the library loads like so:
```
<script>
ModelViewerElement = self.ModelViewerElement || {};
ModelViewerElement.dracoDecoderLocation = 'http://example.com/location/of/draco/decoder/files/';
</script>
```

Which works, except when you use `<script type="module">`, which is now standard practice. It should say:

```
<script type="module">
self.ModelViewerElement = self.ModelViewerElement || {};
self.ModelViewerElement.dracoDecoderLocation = 'http://example.com/location/of/draco/decoder/files/';
</script>
```

Apparently in the olden days if you forgot to declare a variable (no `const`, `let`, or `var`) then it would just merrily attach it to the window, which is what `self` does properly. Modules won't let you get away with being that sloppy. 

So, since `self.ModelViewerElement` might have been declared by someone before we were loaded, we just need to query it to get the set value: `powerPreference: (self as any).ModelViewerElement?.powerPreference ||
        DEFAULT_POWER_PREFERENCE`

elalish (changes requested):

packages/model-viewer/src/model-viewer-base.ts
asemanh: Does that mean I should introduce a helper function like isDebugMode and retrieve the powerPreference that way?

packages/model-viewer/src/test/model-viewer-base-spec.ts
asemanh: Would that mean I should revert this change?

packages/model-viewer/src/model-viewer-base.ts
elalish: You can probably skip the helper, since unlike debug mode, we don't set this via URL query parameters.

packages/model-viewer/src/test/model-viewer-base-spec.ts
elalish: Yes

packages/model-viewer/src/three-components/Renderer.ts
asemanh: Thanks for the thorough explanation

elalish (approved):
Thanks for sticking with this!

PR #2657, opened 2021-08-05
fixed context restoration

### LINKED ISSUE #782: Models sometime turn completely black (after initially working)
### Description

Every once in a while when I leave a model-viewer tab open in Safari on iOS 13, I come back to it and the whole model is black. Sometimes it also does a weird tearing effect:

weird

#### Live Demo

I don't quite know what the repro steps are since it seems to happen randomly. I go here:
https://tryit.media

And then I open other tabs and come back to it some time later.

### Browser Affected

- [ ] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Helios
- [ ] IE
- [x] Safari

### OS

- [ ] Android
- [x] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### Versions
- model-viewer: v0.6.1
- three.js: r105

elalish:
Fixes #782 

This is tricky to test; you have to put a break point in the render loop, then set `window.renderer = this.threeRenderer` to keep a handle. Then you remove the break point and continue and in the console put `window.renderer.forceContextLoss()` and verify the render is gone. Then put `window.renderer.forceContextRestore()` and verify the render is working again (the bug was that the object would show up black instead of lit because our lighting was not getting properly restored). I tried to make an automated test out of this, but for some reason it is not behaving properly.

### DIFF
--- a/packages/model-viewer/src/three-components/Renderer.ts
+++ b/packages/model-viewer/src/three-components/Renderer.ts
@@ -471,6 +471,10 @@ export class Renderer extends EventDispatcher {
       this.textureUtils.dispose();
     }
 
+    if (this.roughnessMipmapper != null) {
+      this.roughnessMipmapper.dispose();
+    }
+
     if (this.threeRenderer != null) {
       this.threeRenderer.dispose();
     }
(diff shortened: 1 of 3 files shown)


packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -111,7 +112,6 @@ export class Renderer extends EventDispatcher {
 
     this.canvas3D = document.createElement('canvas');
     this.canvas3D.id = 'webgl-canvas';
-    this.canvas3D.addEventListener('webglcontextlost', this.onWebGLContextLost);
```
asemanh, line 114: Why are we removing this? Why it was added originally?

asemanh (approved):
I would need to spend some more time to understand the logic. Technically looks good.
And probably we need to add a task for testing such changes

PR #2677, opened 2021-08-10
Added all the remaining khronos models

asemanh:


packages/render-fidelity-tools/test/config.json
```diff
@@ -709,6 +869,22 @@
...
+      "name": "khronos-SimpleSkin",
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/SimpleSkin/glTF/SimpleSkin.gltf",
+      "orbit": {
+        "radius": 2.5
+      },
+      "target": {
+        "y": 1
+      },
+      "dimensions": {
+        "height": 1400
```
elalish, line 882: since this test isn't very interesting, let's reduce the width rather than increase the height.

packages/render-fidelity-tools/test/config.json
```diff
@@ -358,6 +358,166 @@
...
+      },
+      "exclude": [
+        "dspbr-pt"
+      ]
+    },
+    {
+      "name": "khronos-MetalRoughSpheresNoTextures",
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/MetalRoughSpheresNoTextures/glTF/MetalRoughSpheresNoTextures.gltf",
+      "orbit": {
+        "radius": 120.0
```
elalish, line 465: Any idea why this one wasn't working? I'm assuming it could use the same configuration from the other MetalRoughSpheres examples we have (and it could probably live next to them).

packages/render-fidelity-tools/test/config.json
```diff
@@ -358,6 +358,166 @@
...
+      },
+      "exclude": [
+        "dspbr-pt"
+      ]
+    },
+    {
+      "name": "khronos-IridescentDishWithOlives",
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/IridescentDishWithOlives/glTF-Binary/IridescentDishWithOlives.glb",
+      "orbit": {
+        "radius": 0.47
```
elalish, line 384: Can we angle the camera so that we can see some of the olives through the glass? That will help with testing refraction. I'd also like to put a background on this one; maybe spruit sunrise? Just pick one of the backgrounds we're already using in another test.

packages/render-fidelity-tools/test/config.json
```diff
@@ -358,6 +358,166 @@
         "dspbr-pt"
       ]
     },
+    {
+      "name": "khronos-ToyCar",
+      "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/ToyCar/glTF-Binary/ToyCar.glb",
+      "orbit": {
+        "radius": 0.037,
```
elalish, line 365: lol, looks like they're toy car is smaller than a matchbox car! I kind of doubt that was intentional... Probably worth filing a bug on their repo stating its dimensions (all glTF models are in meters).

elalish (changes requested):
Thanks for getting all these models in! Nice job with the framing and camera pointing - they look good.

packages/render-fidelity-tools/test/config.json
asemanh: I'm not really sure what is going on. Based on the picture, I thought maybe I'm too close, but even after playing around with Z and radius, I couldn't make it work. I did move it next to the other ones.

packages/render-fidelity-tools/test/config.json
asemanh: How do I file an issue related to this?
https://screenshot.googleplex.com/3Jqj646xCh3fm4E

What information should I include? Do you happen to have a sample issue?

elalish: No, you can just file an issue on their github: https://github.com/KhronosGroup/glTF-Sample-Models/issues

If you load the model in our editor, go to dev tools Elements section and select the model-viewer element, you can put `$0.getDimensions()` in the console to see the model's size. We really need to add this feature to our UX...

packages/render-fidelity-tools/test/config.json
elalish: Can you also set a skybox? You can use MetalRoughSpheresHDR as an example.

packages/render-fidelity-tools/test/config.json
elalish: I just loaded this in the editor and if you click "save current as initial" under the camera tab, you'll see the default radius is 0.015m in the snippet. This must be another accidentally tiny model.

packages/render-fidelity-tools/test/config.json
```diff
@@ -224,14 +224,16 @@
       "name": "khronos-MetalRoughSpheresNoTextures",
       "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/MetalRoughSpheresNoTextures/glTF/MetalRoughSpheresNoTextures.gltf",
       "orbit": {
-        "radius": 12.0
+        "radius": 0.0000000005
```
elalish, line 227: This feels like it's asking for a rounding error. Can you increase this to something reasonable by changing target z? If you subtract the same amount from z that you add to radius it should keep the camera in the same place (although I might have a sign wrong).

packages/render-fidelity-tools/test/config.json
```diff
@@ -394,6 +396,7 @@
     {
       "name": "khronos-IridescentDishWithOlives",
       "model": "../../../shared-assets/models/glTF-Sample-Models/2.0/IridescentDishWithOlives/glTF-Binary/IridescentDishWithOlives.glb",
+      "lighting": "../../../shared-assets/environments/spruit_sunrise_1k_HDR.hdr",
```
elalish, line 399: sorry, you also need `"renderSkybox": true` so the background shows up.

elalish (changes requested):

packages/render-fidelity-tools/test/config.json
asemanh: Sigh, I couldn't fully figure this out

elalish (approved):
Thanks!

PR #2679, opened 2021-08-10
Skips creating a material if it is uncorrelated (i.e. it does not map…

timmmeh:
… to a three.js material)

Fixes: #2660

### DIFF
--- a/packages/model-viewer/src/features/scene-graph/model.ts
+++ b/packages/model-viewer/src/features/scene-graph/model.ts
@@ -20,6 +20,8 @@ import {CorrelatedSceneGraph} from '../../three-components/gltf-instance/correla
 import {Model as ModelInterface} from './api.js';
 import {Material} from './material.js';
 
+
+
 const $materials = Symbol('materials');
 
 /**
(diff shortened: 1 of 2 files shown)


packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -36,11 +38,16 @@ export class Model implements ModelInterface {
...
     gltf.materials!.forEach(material => {
-      this[$materials].push(new Material(
-          onUpdate,
-          gltf,
-          material,
-          gltfElementMap.get(material) as Set<MeshStandardMaterial>));
+      const correlatedMaterial =
+          gltfElementMap.get(material) as Set<MeshStandardMaterial>;
+
+      if (correlatedMaterial) {
```
elalish, line 44: Our style is to use explicit null checks like `correlatedMaterial != null`.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -36,11 +38,16 @@ export class Model implements ModelInterface {
...
+        this[$materials].push(
+            new Material(onUpdate, gltf, material, correlatedMaterial));
+      } else {
+        console.warn(
+            `Unsupported material "${material.name}" removed from model.`);
```
elalish, line 49: Maybe unreferenced material?

elalish:
Putting @asemanh on as a reviewer, as she said she'd like to review more PRs to help ramp up.

packages/model-viewer/src/features/scene-graph/model.ts
asemanh: This style is also recommended in c++ to return early on error

asemanh (approved):

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: Done.
I did this to combine the check for null and undefined, but I can add explicit checks

elalish: ah, so this is why we allow `==` and `!=` for exactly one situation: comparing to `null`, because this one combines the check for `null` and `undefined`. However, it's still better than a truthiness check, because that will also catch things like empty string, zero, etc.

timmmeh: Done.

PR #2685, opened 2021-08-11
Upgraded filament version to Latest version: 1.11.0

asemanh:


elalish (approved):
Thanks! Looks like it fixed several of their bugs.

asemanh:
> Thanks! Looks like it fixed several of their bugs.

ah nice! What was an example of bugs which were fixed?

PR #2700, opened 2021-08-17
Variant fixes

timmmeh:
Fixes issue with all variant materials being correlated as the same material, meaning a change to one would affect all variants.
Fixes issue with materials being re-created every time a variant was re-selected.


packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -216,6 +217,7 @@ export class CorrelatedSceneGraph {
   private[$gltf]: GLTF;
   private[$gltfElementMap]: GLTFElementToThreeObjectMap;
   private[$threeObjectMap]: ThreeObjectToGLTFElementHandleMap;
+  private[$gltfVariantMap]: GLTFElementToThreeObjectMap|null = null;
```
timmmeh, line 220: Created a variant map to hold glTF elements belonging to a variant definition. This prevents them from being all being correlated to the default material variant.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -277,10 +283,14 @@ export class CorrelatedSceneGraph {
...
-          let threeObjects = this.gltfElementMap.get(gltfElement);
+          if (this[$gltfVariantMap] === null) {
+            this[$gltfVariantMap] = new Map<GLTFElement, ThreeObjectSet>();
+          }
+          let threeObjects = this[$gltfVariantMap]!.get(gltfElement);
           if (threeObjects == null) {
             threeObjects = new Set();
-            this.gltfElementMap.set(gltfElement, threeObjects);
+            // Maps the original gltfElement to a set of possible variants.
+            this[$gltfVariantMap]!.set(gltfElement, threeObjects);
```
timmmeh, line 293: correlation of element to objects for variants tracked separately, to prevent polluting the `gltfElementMap` when switching variants.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -23,14 +23,16 @@ import {Material} from './material.js';
...
 
 /**
  * A Model facades the top-level GLTF object returned by Three.js' GLTFLoader.
  * Currently, the model only bothers itself with the materials in the Three.js
  * scene graph.
  */
 export class Model implements ModelInterface {
-  private[$materials]: Array<Material> = [];
+  private[$materials]: Material[] = [];
+  private[$variants]: Map<number, Material> = new Map<number, Material>();
```
timmmeh, line 35: A map of variants materials is created instead of a list as variants are added out of order and the built-in gltf indexing would not work.

elalish: No need to type the Map; TS can infer from the constructor. So we're mapping a variant number to a material? Don't variants have multiple materials? Or is this some kind of per-mesh variant number?

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
elalish: Okay, so we're populating the `$gltfVariantMap` but not using it anywhere yet? What is its intended purpose, or is it just keeping things from getting garbage-collected?

elalish: Wait, I just found where it's used; scratch that.

packages/model-viewer/src/features/scene-graph/model.ts
elalish: Oh wait, I think I see now; this maps a glTFMaterial index to our Material wrapper object. I think the name  is confusing to me here; I understand this is helpful for variants, but I don't think anything about this data structure is variants-specific, correct? This seems pretty duplicate to our `$materials` array. Would it be possible to stick with the original array but populate it with `null` for any index that  hasn't been  created yet? Or else just remove $materials entirely (though that would be breaking).

asemanh (changes requested):

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -151,15 +151,23 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-              this[$getOnUpdateMethod](),
-              gltf,
-              material,
-              gltfElementMap.get(material) as Set<MeshStandardMaterial>);
+        const {gltf, gltfVariantMap} = threeGLTF.correlatedSceneGraph;
+
+        if (gltfVariantMap !== null) {
+          for (const index of updatedMaterials) {
+            let variant = this[$model]!.variants.get(index);
+            if (variant === null || variant === undefined) {
```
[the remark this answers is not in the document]
timmmeh: I don't think the `continue` works because in the case where `variant == null` there are still other duties to fulfill that must happen after this check which would get skipped if it were to continue, the null check is to just to create the variant on the fly/as-needed and to prevent it from being created more than once. Apologies if I'm misunderstanding you comment, sometimes I get tunnel vision after I've written something one way.

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: I was able to collapse into one list of materials by carving out a space for un-instantiated materials. It does require changing the type from `material: Array<Material>` to `material: Array<Material|null>` which created a compile in typescript where the list is accessed and assumed to be not-null, obviously the compiled javascript is the same.

packages/model-viewer/src/features/scene-graph.ts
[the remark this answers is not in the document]
timmmeh: This is moot with latest changes

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -119,7 +120,7 @@ export class CorrelatedSceneGraph {
     const originalGLTF = upstreamCorrelatedSceneGraph.gltf;
     const cloneGLTF: GLTF = JSON.parse(JSON.stringify(originalGLTF));
     const cloneThreeObjectMap: ThreeObjectToGLTFElementHandleMap = new Map();
-    const cloneGLTFELementMap: GLTFElementToThreeObjectMap = new Map();
+    const cloneGLTFElementMap: GLTFElementToThreeObjectMap = new Map();
```
elalish, line 123: I'm guessing this rename was unintentional?

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -277,10 +283,14 @@ export class CorrelatedSceneGraph {
          gltfMaterialIndex: number) => {
           updatedMaterials.add(gltfMaterialIndex);
           const gltfElement = this.gltf.materials![gltfMaterialIndex];
-          let threeObjects = this.gltfElementMap.get(gltfElement);
+          if (this[$gltfVariantMap] === null) {
+            this[$gltfVariantMap] = new Map<GLTFElement, ThreeObjectSet>();
```
elalish, line 287: Seems like `$gltfVariantMap` is duplicating `$gltfElementMap` in  much  the same way as was happing with the  array and map on `model`. Can we simplify this in the same way? Is there any reason we need both?

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
```diff
@@ -176,63 +174,63 @@ suite('scene-graph/material', () => {
...
+      expect(element.model!.materials[1]!.getDoubleSided()).to.be.equal(true);
 
-      element.model!.materials[1].setDoubleSided(false);
-      expect(element.model!.materials[1].getDoubleSided()).to.be.equal(false);
+      element.model!.materials[1]!.setDoubleSided(false);
+      expect(element.model!.materials[1]!.getDoubleSided()).to.be.equal(false);
 
-      element.model!.materials[1].setDoubleSided(true);
-      expect(element.model!.materials[1].getDoubleSided()).to.be.equal(true);
+      element.model!.materials[1]!.setDoubleSided(true);
```
elalish, line 203: Is there some way we could make some kind of a getter, where if you asked for a `null` material it would automatically load it so it would have something to give you? Maybe it's a helper method. Would be nice to avoid all these `!`s. But this can also be part of a later PR if you'd rather.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
timmmeh: This prevents bad correlations, I'm using the new map to prevent variant objects with the same key from being correlated together. Actually looking at this now I think I can see more clearly where the bug was coming from, I think I might be able to get rid of it.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
timmmeh: Yeah we can lazy load the missing material.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
timmmeh: Typo? Was it supposed to be ELementMap vs ElementMap?

elalish: oh, oops; I read it backwards. Thanks!

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -120,33 +142,49 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+  }
+
+  private[$ensureLoaded]() {
+    if (this[$lazyLoadGLTFInfo] !== undefined) {
+      this[$initialize](this[$lazyLoadGLTFInfo]!);
+    }
   }
 
   get name(): string {
+    this[$ensureLoaded]();
```
timmmeh, line 157: Calling `ensureLoaded` makes sure the material is properly configured. However, not strictly necessary, any operation done on the material would succeed, it just might not work as expected, this could be left out and we could assume that a valid material would always have been loaded correctly without having to 'ensure' things everywhere.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -13,42 +13,226 @@
...
+  constructor(name: string) {
+    this.name = name;
+  }
+}
+
+// Represents a primitive in a glTF mesh.
+class PrimitiveNode extends Node {
+  private[$mesh]: Mesh;
+  // Maps glTF material index number to a material that this primitive supports.
+  private[$materials] = new Map<number, Material>();
```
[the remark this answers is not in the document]
timmmeh: You would still need to keep a list of indices into the original list in order to track which materials belong to this primitive, so it would require a mapping of indices to materials to do the look-up, hence why I condensed the relationship into a map. Not apposed to creating a list of indices used to lookup a material in the 'master list' but I don't see how that is cleaner.

elalish (commented):
comment

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -13,42 +13,275 @@
...
+        }
+      }
+    }
+  }
+
+  get mesh() {
+    return this[$mesh];
+  }
+
+  async setActiveMaterial(material: number|Material):
```
elalish, line 147: Kind of an odd signature; maybe make this two separate functions?

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -13,42 +13,275 @@
...
+  private[$roots] = new Array<Node>();
+  private[$primitives] = new Array<PrimitiveNode>();
 
   constructor(
       correlatedSceneGraph: CorrelatedSceneGraph,
       onUpdate: () => void = () => {}) {
-    const {gltf, gltfElementMap} = correlatedSceneGraph;
+    const {gltf, threeGLTF, gltfElementMap} = correlatedSceneGraph;
+
+    for (let i = 0, size = gltf.materials!.length; i < size; ++i) {
```
elalish, line 207: I believe this is the syntax you're looking for: `for (const [i, value] of myArray.entries())`

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -13,42 +13,275 @@
...
+        const capturedMatIndex = i;
+        const materialLoadCallback = async () => {
+          const threeMaterial =
+              await threeGLTF.parser.getDependency(
+                  'material', capturedMatIndex) as MeshStandardMaterial;
+
+          // Adds correlation, maps the variant gltf-def to the
+          // three material set containing the variant material.
+          const threeMaterialSet = new Set<MeshStandardMaterial>();
+          gltfElementMap.set(gltfMaterialDef, threeMaterialSet);
```
elalish, line 230: Do we know that every invocation of this has a unique gltfMaterialDef? I just want to make sure we're safe always creating a new set rather than adding to an existing one.

packages/model-viewer/src/test/features/scene-graph/mesh-primitive-spec.ts
```diff
@@ -0,0 +1,134 @@
...
+import {assetPath, loadThreeGLTF} from '../../helpers.js';
+
+
+
+const expect = chai.expect;
+
+const BRAIN_STEM_GLB_PATH = assetPath(
+    'models/glTF-Sample-Models/2.0/BrainStem/glTF-Binary/BrainStem.glb');
+const SHEEN_CHAIR_GLB_PATH = assetPath(
+    'models/glTF-Sample-Models/2.0/SheenChair/glTF-Binary/SheenChair.glb');
```
elalish, line 29: We might want to pick smaller test models; If they take a long time to load it'll slow down  the whole test pass. Not a huge deal, just keep an eye on how long the CI takes to pass.

packages/model-viewer/src/test/features/scene-graph/mesh-primitive-spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+  return result.length > 0 ? result : null;
+};
+
+suite('scene-graph/model/mesh-primitives', () => {
+  suite('Static Primitive', () => {
+    test('Primitive count matches glTF file', async () => {
+      const threeGLTF = await loadThreeGLTF(SHEEN_CHAIR_GLB_PATH);
+      const model = new Model(CorrelatedSceneGraph.from(threeGLTF));
```
elalish, line 47: Looks like you could pull these into a `setup` method.

packages/model-viewer/src/test/features/scene-graph/mesh-primitive-spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    });
+
+    test('Should have variant info', async () => {
+      const threeGLTF = await loadThreeGLTF(SHEEN_CHAIR_GLB_PATH);
+      const model = new Model(CorrelatedSceneGraph.from(threeGLTF));
+      expect(findPrimitivesWithVariant(model, 'Mango Velvet')).to.not.be.null;
+      expect(findPrimitivesWithVariant(model, 'Peacock Velvet')).to.not.be.null;
+    });
+
+    test('Should not have variant info', async () => {
```
elalish, line 58: Well, apparently it should have two, from above. Can we be specific about what should not have info?

packages/model-viewer/src/test/features/scene-graph/mesh-primitive-spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      for (const primitive of primitives) {
+        materials.push(
+            primitive.enableVariant('Peacock Velvet') as MeshStandardMaterial);
+      }
+      expect(materials.find((material: MeshStandardMaterial) => {
+        return material.name === 'fabric Mystere Peacock Velvet';
+      })).to.not.be.null;
+    });
+
+    test('Switching back variant name', async () => {
```
elalish, line 108: I might be tempted to combine the previous two tests into this one.

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
```diff
@@ -74,5 +79,58 @@ suite('scene-graph/model', () => {
...
       expect(collectedMaterials.size).to.be.equal(materials.size);
     });
+
+    suite('Model Variants', () => {
+      test('Switch variant and lazy load', async () => {
+        const threeGLTF = await loadThreeGLTF(SHEEN_CHAIR_GLB_PATH);
+        const model = new Model(CorrelatedSceneGraph.from(threeGLTF));
+
+        expect(model[$materials][4][$correlatedObjects]).to.be.null;
+        expect(model[$materials][4][$lazyLoadGLTFInfo]).to.not.be.undefined;
```
elalish, line 89: I think there's a `.to.be.okay` or something like that. It would be a shame to get a null here and have the test pass.

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
```diff
@@ -74,5 +79,58 @@ suite('scene-graph/model', () => {
...
+
+        expect(model[$materials][5][$correlatedObjects]).to.be.null;
+        expect(model[$materials][4][$lazyLoadGLTFInfo]).to.not.be.undefined;
+
+        await model[$switchVariant]('Peacock Velvet');
+
+        expect(model[$materials][4][$correlatedObjects]).to.not.be.null;
+        expect(model[$materials][4][$lazyLoadGLTFInfo]).to.be.undefined;
```
elalish, line 97: This is because you've loaded all the variants, right?

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
```diff
@@ -74,5 +79,58 @@ suite('scene-graph/model', () => {
...
+            model[$materials][0][$correlatedObjects]!.values().next().value;
+        const sizeBeforeSwitch = model[$materials][0][$correlatedObjects]!.size;
+        await model[$switchVariant]('Does not exist');
+
+        expect(model[$materials][0][$correlatedObjects]!.values().next().value)
+            .equals(threeMaterial);
+        expect(model[$materials][0][$correlatedObjects]!.size)
+            .equals(sizeBeforeSwitch);
+      });
+    });
```
elalish, line 134: Might be nice to add a test that you can look up all the material names without causing them to load. Also that you can access an unloaded material without crashing (since that should cause it to load).

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -32,26 +33,49 @@ const $occlusionTexture = Symbol('occlusionTexture');
 const $emissiveTexture = Symbol('emissiveTexture');
 const $backingThreeMaterial = Symbol('backingThreeMaterial');
 const $applyAlphaCutoff = Symbol('applyAlphaCutoff');
+export const $lazyLoadGLTFInfo = Symbol('lazyLoadGLTFInfo');
+const $initialize = Symbol('initialize');
+export const $ensureLoaded = Symbol('ensureLoaded');
+
+export enum MaterialLoadMethod {
```
elalish, line 40: Where is this used?

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -57,7 +290,23 @@ export class Model implements ModelInterface {
    *
    * TODO(#1003): How do we handle non-active scenes?
    */
-  get materials(): Array<Material> {
+  get materials(): Material[] {
     return this[$materials];
   }
+
+  async[$switchVariant](variantName: string) {
```
elalish, line 297: So, one thing I don't love about the variant spec is that the initially loaded configuration is not necessarily the same materials as any one of the named variants. I'd say that's a poor practice, but we might want to think about how one might get back to the "default" state. That can wait for a follow-on if it's not simple.

elalish (changes requested):
Looking good! A couple more tests and some `!` removals I think we'll be there.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -13,42 +13,275 @@
...
+    // The gltf mesh array to sample from.
+    const meshElementArray = gltf['meshes'] || [];
+    // List of primitives under the mesh.
+    const gltfPrimitives =
+        (meshElementArray[meshIndex].primitives || []) as Primitive[];
+
+    for (const primitive of gltfPrimitives) {
+      // Maps the primitive default to a material.
+      this[$materials].set(
+          primitive.material!, mvMaterials[primitive.material!]);
```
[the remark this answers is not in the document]
timmmeh: It does, but you bring up a good point, added updates to account for this.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -13,42 +13,275 @@
...
+    // primitives.
+    let associationKey: Object3D = mesh;
+    // Work around 1, skinned meshes (glTF primitives) have no association but
+    // the parent (mesh) does, which maps to a 'node'.
+    if (mesh instanceof SkinnedMesh && mesh.parent != null) {
+      associationKey = mesh.parent;
+    }
+    let gltfMeshReference = correlatedSceneGraph.threeObjectMap.get(
+                                associationKey) as GLTFReference;
+    // Work around 2, the association value is a material not a node.
```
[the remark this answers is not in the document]
timmmeh: Yeah, I should add a comment, this was happening when a static model has multiple primitives per mesh.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -32,26 +33,49 @@ const $occlusionTexture = Symbol('occlusionTexture');
...
         .value;
   }
+
   constructor(
       onUpdate: () => void, gltf: GLTF, gltfMaterial: GLTFMaterial,
-      correlatedMaterials: Set<MeshStandardMaterial>) {
+      correlatedMaterials: Set<MeshStandardMaterial>,
+      lazyLoadInfo: LazyLoader|undefined = undefined) {
     super(onUpdate, gltfMaterial, correlatedMaterials);
+    if (lazyLoadInfo === undefined) {
```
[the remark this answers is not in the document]
timmmeh: done

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -32,26 +33,49 @@ const $occlusionTexture = Symbol('occlusionTexture');
...
         .values()
         .next()
         .value;
   }
+
   constructor(
       onUpdate: () => void, gltf: GLTF, gltfMaterial: GLTFMaterial,
-      correlatedMaterials: Set<MeshStandardMaterial>) {
+      correlatedMaterials: Set<MeshStandardMaterial>,
+      lazyLoadInfo: LazyLoader|undefined = undefined) {
```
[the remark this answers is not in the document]
timmmeh: oooh, thanks!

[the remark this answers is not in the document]
timmmeh: Well, that didn't work for the function signature but it does work on the field declaration, good to know.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -13,42 +13,275 @@
...
+    if (gltfMeshReference.type === 'materials') {
+      associationKey = mesh.parent!;
```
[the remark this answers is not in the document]
timmmeh: I worked shopped this a bit, if this code is running than the association type is wrong if this doesn't fix things up than the error/assertion at line 105 should catch it.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -13,42 +13,275 @@
...
+      if (primitive.extensions &&
+          primitive.extensions['KHR_materials_variants']) {
+        const variantsExtension =
+            primitive.extensions['KHR_materials_variants'] as
+            KHRMaterialsVariants;
+        const extensions = threeGLTF.parser.json.extensions;
+        const variantNames = extensions['KHR_materials_variants'].variants;
+        // Provides definition now that we know there are variants to support.
+        this[$variantInfo] = new Map<string, Material>();
+        for (const mapping of variantsExtension.mappings!) {
```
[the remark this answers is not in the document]
timmmeh: Well, the fact that we have a variantsExtension to start with is a pretty good indication that mappings exist unless the file is malformed.

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
timmmeh: Correct, when the variant loads the LazyLoader reference is removed.

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: Oh, it was from a different time, a simpler time.
Removed

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: Ahh, actually the previous implementation did that but we avoid it now because we are mapping variants to _model-viewer_ materials, after they list has been created, so the "default" material loaded will be always reference a material that was previously loaded.

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
timmmeh: Okay.

packages/model-viewer/src/features/scene-graph/model.ts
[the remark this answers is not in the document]
elalish: If that's true I  would expect variantsExtension.mappings to be a type  that doesn't include `null`/`undefined`. Maybe  just needs an update to the type? However, I think mappings can be absent for a mesh that has the same material in all variants?

packages/model-viewer/src/features/scene-graph/model.ts
elalish: Okay, but say I selected a variant and now I want to go back. Is there a `variantName` I can choose that will take me to the original configuration?

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: Yeah, actually I think passing a material is improper here as it could be any material from anywhere. It's good to be able to do that but it would bypass all the bookkeeping right now, that would need to be handled differently, going to kick the can down the road.

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: Yes, the only way it ends up here is if that was missing in the first place.

packages/model-viewer/src/test/features/scene-graph/mesh-primitive-spec.ts
timmmeh: Done.

packages/model-viewer/src/test/features/scene-graph/mesh-primitive-spec.ts
timmmeh: Done.

packages/model-viewer/src/test/features/scene-graph/mesh-primitive-spec.ts
timmmeh: Agreed

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: I think I see, the 'original' should be one of the variants yes? So you would have to know which one was the default then go back to that. If not, then I'd call that an authoring mistake, also the spec is letting you down. We could add a `revertToDefaultVariant` to help, but yeah the variant spec has some edge cases, it feels very bolted on.

packages/model-viewer/src/features/scene-graph/model.ts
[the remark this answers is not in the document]
timmmeh: The spec does not definitively say one way or the other but looking at the implementation it seems clear:
If mappings is not defined the variant is malformed, I'm going to remove the undefined option from our interface.
```
"primitives": [
      {
        ...,
        "extensions": {
          "KHR_materials_variants" : {
            "mappings": [
              {
                "material": 2,
                "variants": [0, 3],
              },
              {
                "material": 4,
                "variants": [1],
              },
              {
                "material": 5,
                "variants": [2],
              },
            ],
          }
        }
      },
      // ... more primitives ...
    ]
```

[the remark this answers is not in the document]
elalish: Hmm, I don't think that's true. From the spec:

```
Across the entire mappings array, each variant index must be used no more than one time.

When the active variant is referenced in a mapping, a compliant viewer will apply its material to the primitive.

When no mapping contains the active variant, or there is no active variant, a compliant viewer will fall back on vanilla glTF behaviour for the primitive.
```

I think when there's no mapping, you use the initially-loaded (non-extension) material. I think handling this fits into handing the return to initial case.

packages/model-viewer/src/features/scene-graph/model.ts
elalish: No, that's exactly the problem. I think it would be a good idea, but sadly the spec does not require any of the variants to equal the non-extension material setup. The extension mappings are completely independent of the non-extension materials, which are what you get when you first load it. I think a simple solution might just be to accept `null` and put it back then. But I'd guess you may have to do some work when creating your mappings to save this as a special state.

packages/model-viewer/src/features/scene-graph/api.ts
```diff
@@ -71,6 +71,16 @@ export declare interface Material {
...
+
+  /**
+   * Asynchronously loads the underlying material resource if it's currently
+   * unloaded, otherwise the method is a noop.
+   */
+  ensureLoaded(): void;
+  /**
+   * Returns true if the material is loaded and valid.
+   */
+  isValid: boolean;
```
elalish, line 83: Maybe `isLoaded`? What would cause a material to be invalid?

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -155,42 +151,60 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
     }
     return this[$correlatedObjects]!.values().next().value;
   }
 
+  [$ensureMaterialIsValid]() {
+    if (this[$lazyLoadGLTFInfo] == null) {
+      return;
+    }
+    throw new Error(`Material "${this.name}" has not been loaded, call 'await
+    myMaterial.ensureLoaded()' before using an unloaded material.`);
```
elalish, line 166: 👍

packages/model-viewer/src/features/scene-graph/nodes/node.ts
```diff
@@ -0,0 +1,12 @@
...
+export const $children = Symbol('children');
+
+// Defines the base level node methods and data.
+export class Node {
+  name: string = '';
+  [$children] = new Array<Node>();
+  constructor(name: string) {
+    this.name = name;
+  }
+}
```
elalish, line 12: This is too tiny for it's own file; Let's put it in with primitive-node for now; we can re-organize later if needed.

packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
```diff
@@ -0,0 +1,155 @@
+
```
elalish, line 1: Remember to put in our license header.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
```diff
@@ -235,4 +238,45 @@ suite('scene-graph/material', () => {
...
+
+    const loadModel = async (path: string) => {
+      element.src = path;
+      document.body.insertBefore(element, document.body.firstChild);
+      await waitForEvent(element, 'load');
+      model = element.model as Model;
+    };
+
+    test('Accessing the name getter does not cause throw error.', async () => {
+      await loadModel(CUBES_GLTF_PATH);
```
elalish, line 260: you can also just move this to `setup`

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
```diff
@@ -82,55 +81,52 @@ suite('scene-graph/model', () => {
 
     suite('Model Variants', () => {
       test('Switch variant and lazy load', async () => {
-        const threeGLTF = await loadThreeGLTF(SHEEN_CHAIR_GLB_PATH);
+        const threeGLTF = await loadThreeGLTF(CUBES_GLTF_PATH);
```
elalish, line 84: 👍

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
```diff
@@ -0,0 +1,136 @@
...
+      const threeGLTF = await loadThreeGLTF(CUBES_GLTF_PATH);
+      model = new Model(CorrelatedSceneGraph.from(threeGLTF));
+    });
+
+    test('Primitive count matches glTF file', async () => {
+      expect(model![$primitives].length).to.equal(2);
+    });
+
+    test(
+        'Primitives should have variant info that with expected mappings',
```
elalish, line 74: missing word?

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
```diff
@@ -235,4 +238,45 @@ suite('scene-graph/material', () => {
...
+      // tslint:disable-next-line:no-unused-expression
+      expect(() => {model.materials[2].name}).does.not.throw;
```
elalish, line 262: Might be easier to just check the name is what you expect. It'll still fail if it throws.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
```diff
@@ -235,4 +238,45 @@ suite('scene-graph/material', () => {
...
+      expect(model.materials[2][$lazyLoadGLTFInfo]).to.be.ok;
+    });
+
+    test(
+        'Accessing a getter of an unloaded material throws an error.',
+        async () => {
+          await loadModel(CUBES_GLTF_PATH);
+          expect(() => {model.materials[2].pbrMetallicRoughness}).to.throw;
```
elalish, line 270: Maybe also check that `isValid` is false?

elalish (changes requested):

packages/model-viewer/src/features/scene-graph/api.ts
timmmeh: If we do some validation step in the future - which always comes with materials, I've just done this dance many times with materials but maybe I'm getting too far ahead.

elalish: glTF validation is generally a pre-process anyway. If it managed to successfully load in three.js, I think it's plenty valid enough for us.

packages/model-viewer/src/features/scene-graph/nodes/node.ts
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
timmmeh: Done.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
timmmeh: Done.

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
timmmeh: Done.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
timmmeh: Done.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
timmmeh: Done

elalish (approved):
Looks great, thanks!

packages/model-viewer/src/features/scene-graph/model.ts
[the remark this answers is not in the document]
timmmeh: I think we are saying the same thing overall, I'm just trying to make the point that the spec does not make room for having a variant definition in a primitive that has an undefined `mappings` field.
like:
```
"extensions": {
          "KHR_materials_variants" : {
             // Can't have just emptiness here
            ],
          }
        }
```
I think there is a problem with the `return to initial case`
Say you have model that support variants `red` and `blue` for instance, where `red` is the default/initial material.
Then you call `model.switchVariant(blue)` it switches to blue, then you call `model.switchVariant(green)` then it switches to... `red`? This would be because technically there is no mapping for `green` so it defaults back to the initial material. But the spec says to fall back on the vanilla glTF _behavior_ not fall back to the initial material, which they could have easily written, the default _behavior_ does not allow switching materials at all so in other words: just ignore it.

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: Sounds like an implied 'extra' variant, I think that's a small change to just mark the 'initial' material, but I'd want to be very specific about switch back to it if it's the 'implied' material though, to avoid false positives, where we just end up switching back to the initial material because we got some noise that couldn't quite be parsed.

PR #2701, opened 2021-08-18
Editor camera state cleanup

elalish:
Follow-on to the big refactor that  removed material  state from the editor (relying instead on model-viewer's own state), this smaller  refactor removes the camera state from Redux as  its  information was already contained in the model-viewer config state in Redux. It also  fixes a  number of async bugs related to updating the UI, especially noticeable with the camera target.


packages/space-opera/src/components/camera_settings/camera_settings.ts
```diff
@@ -153,79 +131,57 @@ export class CameraSettings extends ConnectedLitElement {
...
+  async updateInitialCamera() {
+    const cameraState = await getCameraState();
+    this.cameraTargetInput.target = cameraState.target;
+    if (this.config.cameraOrbit == null) {
+      this.cameraOrbitEditor.style.display = 'none';
+    } else {
+      const currentOrbit = cameraState.orbit;
+      this.cameraOrbitEditor.yawInput.value = currentOrbit.thetaDeg;
+      this.cameraOrbitEditor.pitchInput.value = currentOrbit.phiDeg;
+      this.cameraOrbitEditor.style.display = '';
```
asemanh, line 164: Question: what is the difference of having display set to none vs ''?

packages/space-opera/src/components/camera_settings/components/limits_base.ts
```diff
@@ -53,9 +53,9 @@ export abstract class LimitsBase extends ConnectedLitElement {
     return 1;
   }
 
-  @query('#minimum') minimumInput?: SliderWithInputElement;
-  @query('#maximum') maximumInput?: SliderWithInputElement;
-  @query('#limit-enabled') enabledInput?: HTMLInputElement;
+  @query('#minimum') minimumInput!: SliderWithInputElement;
+  @query('#maximum') maximumInput!: SliderWithInputElement;
+  @query('#limit-enabled') enabledInput!: HTMLInputElement;
```
asemanh, line 58: What was the reasoning behind the original design?
It looks like previously it was ok for any of these variables to be undefined, and now we are saying it shouldn't?

packages/space-opera/src/components/poster_controls/poster_controls.ts
```diff
@@ -40,11 +38,12 @@ export class PosterControlsElement extends ConnectedLitElement {
...
   @internalProperty() toastBody: string = '';
-  @internalProperty() cameraSnippet: Camera = {};
+  @internalProperty() initialCamera: string = 'auto auto auto';
 
   stateChanged(state: State) {
-    this.poster = getConfig(state).poster;
-    this.cameraSnippet = getCamera(state);
+    const config = getConfig(state);
+    this.poster = config.poster;
+    this.initialCamera = config.cameraOrbit ?? 'auto auto auto';
```
asemanh, line 46: In js world they love to use all the possible signs lol ?? == === !=== =D

packages/space-opera/src/components/camera_settings/camera_settings.ts
elalish: setting an attribute to `''` mean unset, so it now just inherits its behavior from its parent HTML elements. I could set it to `'block'`, but this  is safer, since I don't have to worry about what it's originally computed style  was. In this context, `''` means show.

asemanh (dismissed):

packages/space-opera/src/components/camera_settings/components/limits_base.ts
elalish: Yeah, I've been slowly updating this style across the editor. Before when  a panel  was disabled, they replaced the html wholesale  with nothing, while now I'm preferring to set `display: 'none'`. In this way the elements always exist  and a lot of checking can be removed. In some cases, they had `?` even if  they already always existed, probably because  this is the TS default.

packages/space-opera/src/app.ts
```diff
@@ -21,7 +21,12 @@
 
 import '@material/mwc-button';
 import '@polymer/paper-dialog/paper-dialog';
-
+// The order of imports is important, as this is the order in which their
+// stateChanged() methods will be called. We put the preview first to apply
+// changes to the <model-viewer> element in case they need to be queried by
+// subsequent components. Keep in mind that those will still have to await
+// updateComplete since rendering is async in Lit Element.
+import './components/model_viewer_preview/model_viewer_preview.js';
```
timmmeh, line 29: Does the model_viewer_preview actually expose API to make changes to the model? Just trying to understand why it's called preview but its applying changes.

elalish: The preview is where the model-viewer instance lives, so yes, it effectively exposes the whole MV API to the editor features.

elalish: Other components only use getters on the MV  element. The preview  is the one place where the config Redux state is applied to the MV element, which happens when its  render method is called.

packages/space-opera/src/components/camera_settings/camera_settings.ts
```diff
@@ -66,23 +58,19 @@ class CameraOrbitEditor extends ConnectedLitElement {
...
             innerLabel="yaw"
-            value=${this.orbit.thetaDeg}
             min=-9999 max=9999
             style="min-width: 90px; max-width: 90px;"
             @change=${this.onChange}>
           </me-draggable-input>
           <me-draggable-input
             id="pitch"
             innerLabel="pitch"
-            value=${this.orbit.phiDeg}
```
chrismgeorge, line 85: Now that you're removing these values. Will they be wrong when the model is initially loaded, and only be correct once you start dragging?

elalish: Yeah, they aren't shown initially and I decided it quite a bit simpler to make them just sync one way. There are edge cases where you can update the snippet and it won't show in the UI, but I think this still overall has fewer bugs than where we were.

packages/space-opera/src/components/config/reducer.ts
```diff
@@ -95,49 +158,77 @@ export const getConfig = (state: State) =>
...
+    case SET_CONFIG:
       return action.payload;
-    }
     case SET_REVEAL:
       return {...state, reveal: action.payload};
     case SET_POSTER:
-      return {
-        ...state, poster: action.payload
-      }
+      return {...state, poster: action.payload};
```
timmmeh, line 166: I like the expansion syntax, not sure if this was a thing the last time got into javascript.

timmmeh (approved):

packages/space-opera/src/components/model_viewer_snippet/reducer.ts
```diff
@@ -52,18 +48,6 @@ export function applyRelativeFilePaths(
   }
 }
 
-/** Use when the user wants to load a new config (probably from a snippet). */
```
chrismgeorge, line 55: This reminds me, something we don't have a test for is ensuring the flow of when a user loads a model, then goes to edit the snippet manually, and then saves that. Just wanted to make sure that functionality was still preserved and caught somewhere.

PR #2706, opened 2021-08-23
Support EXT_meshopt_compression

donmccurdy:
Fixes https://github.com/google/model-viewer/issues/1164.

Screen Shot 2021-08-22 at 8 01 55 PM


donmccurdy:
@googlebot I signed it!

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -83,6 +87,19 @@ export class CachingGLTFLoader<T extends GLTFInstanceConstructor =
     return ktx2TranscoderLocation;
   }
 
+  static setMeshoptDecoderLocation(url: string) {
+    if (meshoptDecoderLocation !== url) {
+      meshoptDecoderLocation = url;
+      meshoptDecoder = fetchScript(url)
+        .then(() => MeshoptDecoder.ready)
+        .then(() => MeshoptDecoder);
```
elalish, line 95: We usually use async/await instead of .then; would that be reasonable here too?

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -155,6 +172,10 @@ export class CachingGLTFLoader<T extends GLTFInstanceConstructor =
     this.dispatchEvent(
         {type: 'preload', element: element, src: url} as PreloadEvent);
     if (!cache.has(url)) {
+      if (meshoptDecoder) {
```
elalish, line 175: We generally  use explicit combined null/undefined checks instead of truthiness: `if (meshoptDecoder != null)`.

packages/model-viewer/src/utilities/fetch-script.ts
```diff
@@ -0,0 +1,10 @@
+export function fetchScript(src: string): Promise<Event> {
+    return new Promise((resolve, reject) => {
+        const script = document.createElement('script');
+        document.body.appendChild(script);
+        script.onload = resolve;
+        script.onerror = reject;
+        script.async = true;
+        script.src = src;
+    });
+}
```
elalish, line 10: These  two new  files are pretty tiny (and would require the license header); can we just put these in CachingGLTFLoader instead?

packages/modelviewer.dev/examples/loading/index.html
```diff
@@ -330,7 +330,50 @@ <h4>
...
+          </example-snippet>
+          <p>
+            Enable support for Meshopt compression by providing
+            <code>meshoptDecoderLocation</code>:
+          </p>
+          <example-snippet inert-script highlight-as="html">
+            <template>
+<script>
+ModelViewerElement = self.ModelViewerElement || {};
+ModelViewerElement.meshoptDecoderLocation = 'https://unpkg.com/meshoptimizer@0.16.0/meshopt_decoder.js';
```
elalish, line 364: Can  we switch these  (and the previous example) from `ModelViewerElement` to `self.ModelViewerElement`? I've been meaning to do this anyway, since as written it doesn't compile as a module script.

packages/modelviewer.dev/examples/loading/index.html
```diff
@@ -411,7 +454,7 @@ <h4></h4>
       <div style="margin-top:24px;" class="copyright">©Copyright 2018-2020 Google Inc. Licensed under the Apache License 2.0.</div>
       <div id='footer-links'></div>
   </div>
-      
+
```
elalish, line 457: Thanks for coffeemat.glb; where is it from? Can we show some kind of attribution/license? It should probably have a listing here in this footer.

elalish (changes requested):
Thanks Don!

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
donmccurdy: I think we need to store a Promise, so additional invocations of this method don't make another request. Doing this with async/await would look something like:

```js
meshoptDecoder = new Promise(async (resolve, reject) => {
  try {
    await fetchScript(url);
    await MeshoptDecoder.ready;
    resolve(MeshoptDecoder);
  } catch (e) {
    reject(e);
  }
});
```

A bit more verbose, but could be wrapped in a helper function too, any preference here?

packages/modelviewer.dev/examples/loading/index.html
donmccurdy: It's originally from Sketchfab, under CC Attribution license – good point on attribution! Also included on this three.js example: https://threejs.org/examples/?q=compress#webgl_loader_gltf_compressed

Will add attribution for the artist similarly 👍

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
elalish: Thanks for  the explanation; I'm fine with it as is.

donmccurdy:
Thanks @elalish – I think I've addressed the feedback. 👍🏻

elalish (approved):
Looks great, thanks!

elalish:
Looks like I checked in a flaky test to space-opera. I'm going to merge this, since the failure is definitely unrelated to anything here.

PR #2711, opened 2021-08-24
Auto poster generation

elalish:
I removed the semi-functional poster UX in the editor and now the poster is auto-generated on any save and whenever it is deployed to mobile. In fact posters were not working on mobile at all before, but now they do, and they improve the experience since they demonstrate the use while the slow piping server downloads. I'm also generating only WebP images, since they are supported universally now (since IE is dead) and they have compact lossy compression with an alpha channel.


packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -237,13 +237,13 @@ export class Renderer extends EventDispatcher {
...
     const scaleStep = this.scaleStep;
-    if (this.avgFrameDuration > HIGH_FRAME_DURATION_MS &&
-        this.scaleStep < this.lastStep) {
+    if (this.avgFrameDuration > HIGH_FRAME_DURATION_MS) {
       ++this.scaleStep;
     } else if (
         this.avgFrameDuration < LOW_FRAME_DURATION_MS && this.scaleStep > 0) {
       --this.scaleStep;
     }
+    this.scaleStep = Math.min(this.scaleStep, this.lastStep);
```
elalish, line 246: This ensures minScale will be respected immediately on  the next render.

packages/model-viewer/src/features/loading.ts
```diff
@@ -469,6 +469,8 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
             this.dispatchEvent(new CustomEvent('poster-dismissed'));
           });
         }, {once: true});
+      } else {
+        this[$transitioned] = true;
```
elalish, line 473: Found a small bug where when  things  changed too quickly, the transitioned event would not fire and then the state would get messed up. This ensures the state is correct for the poster to dismiss.

packages/space-opera/src/components/mobile_view/open_mobile_view.ts
```diff
@@ -180,6 +182,10 @@ export class OpenMobileView extends ConnectedLitElement {
 
     await post(JSON.stringify(packet), getSessionUrl(this.pipeId, session.id));
 
+    await post(
+        posterBlob,
+        posterToSession(this.pipeId, session.id, updatedContent.posterId));
```
elalish, line 187: This post is what allows the poster  to function on  mobile (it was set  to an objectURL before, but that can't be accessed from a different device; it  must be  explicitly sent over the network).

packages/space-opera/src/components/model_viewer_snippet/reducer.ts
```diff
@@ -64,3 +58,30 @@ export function extraAttributesReducer(state: any = {}, action: Action): any {
...
   }
 }
+
+const SET_HEIGHT = 'SET_HEIGHT';
+export function dispatchHeight(height: number) {
+  return {type: SET_HEIGHT, payload: height};
+}
+
+const SET_MIMETYPE = 'SET_MIMETYPE';
+export function dispatchMimeType(type: ImageType) {
```
elalish, line 68: I put this in because I thought I'd allow choosing PNG, JPEG, WEBP, but then I realized WEBP is definitely the best, so I removed the UX. I left the logic in case it's useful down the road.

packages/space-opera/src/components/mobile_view/open_mobile_view.ts
```diff
@@ -207,12 +213,17 @@ export class OpenMobileView extends ConnectedLitElement {
...
 
   async sendSessionContentHolder(
-      session: MobileSession, updatedContent: EditorUpdates,
+      session: MobileSession, updatedContent: EditorUpdates, posterBlob: Blob,
       usdzBlob: Blob|undefined, gltfBlob: Blob|undefined,
       envBlob: Blob|undefined) {
     try {
       await this.sendSessionContent(
-          session, {...updatedContent}, usdzBlob, gltfBlob, envBlob);
+          session,
```
timmmeh, line 221: Looks like editor state is serialized and restored? Is that right?

timmmeh (approved):

packages/space-opera/src/components/mobile_view/open_mobile_view.ts
elalish: Yeah, in order to deploy to your mobile device, it  has to serialize everything and post it through the  piping-server for the mobile to fetch. It doesn't serialize all of the Redux state, since the mobile doesn't care; just enough to recreate the MV element.

packages/space-opera/src/components/utils/render_model_viewer.ts
```diff
@@ -34,6 +39,51 @@ export interface ModelViewerEventHandlers {
...
+  const poster = getPosterConfig(state);
+
+  const height = poster.height / window.devicePixelRatio;
+  modelViewer.style.width = `${height}px`;
+  modelViewer.style.height = `${height}px`;
+
+  // Set to beginning of animation
+  const oldTime = modelViewer.currentTime;
+  modelViewer.autoplay = false;
+  modelViewer.currentTime = 0;
```
asemanh, line 58: would it make sense to have reset or init function for this part?

elalish: Maybe, but since the reset is pretty specific to the poster generation I feel like it's simpler to contain the logic here. If it comes in hand anywhere else, I'll  definitely put it in a function.

Issue #2682, opened 2021-08-11
Attach camera-controls pointermove event listeners to window instead of element


lucadalli:
### Discussed in https://github.com/google/model-viewer/discussions/2678

<div type='discussions-op-text'>

<sup>Originally posted by **lucadalli** August 10, 2021</sup>
When rotating a model with `camera-controls` enabled, I often find myself getting frustrated at the model's unresponsiveness when my pointer falls out of the element's bounds.

When I'm interacting with the model, after the initial pointerdown event, I become far less conscious of where my pointer is and more focused on the actual orientation of the model. I'd be rotating the model, until suddenly, my inputs are no longer recognized, only for me to realize that my pointer is outside the element. This breaks my "immersion" as I have to reset my pointer position back onto the model and resume the interaction.

I strongly believe this is an inferior user experience to the one offered by THREE's OrbitControls, which attaches move listeners to window and hence allows an initiated camera interaction to continue anywhere on the page until the pointer is released.

https://threejs.org/examples/#misc_controls_orbit

</div>

elalish:
Is there a reason you posted this again rather than replying on the previous thread?

lucadalli:
While investigating this issue further I noticed that the behavior exhibited is dependent on browser and device.

On Win10 Chrome, the `mousemove` events are confined to the element's bounding box. 
Surprisingly, Win10 Firefox does not behave in the same way and `mousemove` events preceded by a `mousedown`  keep firing even outside the element's bounds.

On Android Chrome and in the emulated mobile environments of Win10 Chrome and Firefox, `touchmove` following a `touchstart` event which happened within element bounds keep firing even outside the element bounds. This means that all mobile environments I was able to test behave unanimously, in the way we prefer.

If iOS Chrome and Safari behave in the same way as Android Chrome, I think this can be leveraged in our implementation to keep the complexity low and avoid wastefully attaching and detaching listeners.

lucadalli:
> Is there a reason you posted this again rather than replying on the previous thread?
I intend to work on this so I thought it would be better to have an issue for it. We can close and resume discussion on the original thread if its preferred.

elalish:
Ah, sure thing. I would have just converted that discussion to an issue anyway, so this is fine. Thank you for taking this on!

lucadalli:
My pleasure :) Are you able to test the current behavior on iOS Chrome and Safari, by any chance?

elalish:
Yes, I have those devices locally; let me get back here with results.

[...]

PR #2722, opened 2021-08-25
Attach controls mousemove listener to window

lucadalli:
As per issue #2682, and as a follow-on to #2691, this PR aims to fix the model-viewer's unresponsiveness when the cursor leaves the element's bounds during a mouse interaction.

I would appreciate feedback on the following:
1. Despite the synthetic events used during testing having `{bubbles: true}`, some tests were failing because the `mousemove` events dispatched to the model-viewer element were not triggering the `mousemove` listener on `window`. I fixed this by dispatching the event directly to `window`.
2. [Originally I intended to change the way the cursor styles are applied](https://github.com/google/model-viewer/discussions/2678#discussioncomment-1153581), such that during an interaction, the `grabbing` cursor persists even outside the element's bounds. However, I decided against due to the increased code complexity for a marginal UX improvement. I noticed that other libraries which handle mouse events like Swiper also do not bother with this.
Once the cursor leaves the element's bounds, it is not being paid much attention to anyway as the user is more focused on the model-viewer element itself.


elalish (dismissed):
Tested and it looks great, on Firefox, Safari, Chrome, Android and iOS. Regarding your two questions, I'm totally satisfied with your solutions. 

However in testing I found a bug related to your previous change (not this PR). On iOS, if you put down one finger, then add a second, you get proper zoom behavior. If  instead you put down both  fingers at once, then it zooms the page  instead of the model. Hopefully a simple fix. Android doesn't have this problem.

elalish:
Looks like this has a conflict now; want to resolve so we can merge this?

lucadalli:
Conflicts with `master` resolved

elalish (approved):
Looks good to me, thanks!

PR #2730, opened 2021-08-27
Fix iOS zoom

lucadalli:
Based on the [logs (or lack thereof) produced on iOS](https://github.com/google/model-viewer/issues/2729#issuecomment-907334260) on branch `lucadalli:debug-ios-zoom`, it seems that if two touches happen at the same time, iOS only fires one `touchstart` event 😵🤯

The previous implementation was banking on the fact that each touch fires a `touchstart` event.
These changes should hopefully fix the issue.

### DIFF
--- a/packages/model-viewer/src/three-components/SmoothControls.ts
+++ b/packages/model-viewer/src/three-components/SmoothControls.ts
@@ -176,7 +176,7 @@ export class SmoothControls extends EventDispatcher {
       this.element.style.cursor = 'grab';
       this._interactionEnabled = true;
 
-      this.setTouchActionStyle();
+      this.updateTouchActionStyle();
     }
   }
 
@@ -200,7 +200,7 @@ export class SmoothControls extends EventDispatcher {
       this.touchMode = null;
       this._interactionEnabled = false;
 
-      this.setTouchActionStyle();
+      this.updateTouchActionStyle();
     }
   }
 
@@ -220,7 +220,7 @@ export class SmoothControls extends EventDispatcher {
         this.element.addEventListener('wheel', this.onWheel);
       }
 
-      this.setTouchActionStyle();
+      this.updateTouchActionStyle();
     }
   }
 
@@ -432,7 +432,7 @@ export class SmoothControls extends EventDispatcher {
     this.moveCamera();
   }
 
-  setTouchActionStyle() {
+  updateTouchActionStyle() {
     const {style} = this.element;
 
     if (this._interactionEnabled) {
--- a/packages/model-viewer/src/three-components/SmoothControls.ts
+++ b/packages/model-viewer/src/three-components/SmoothControls.ts
@@ -173,6 +173,8 @@ export class SmoothControls extends EventDispatcher {
           'touchstart', this.onTouchStart, {passive: true});
       element.addEventListener('touchmove', this.onTouchMove, {passive: false});
 
+      element.addEventListener('touchend', this.onTouchEnd);
+
       this.element.style.cursor = 'grab';
       this._interactionEnabled = true;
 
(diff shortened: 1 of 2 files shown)


packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -605,31 +608,37 @@ export class SmoothControls extends EventDispatcher {
 
   private onTouchStart = (event: TouchEvent) => {
     this.onPointerDown(() => {
-      if (event.touches.length === 1) {
-        self.addEventListener('touchend', this.onTouchEnd);
+      const {targetTouches, changedTouches, touches} = event;
+      if (targetTouches.length === changedTouches.length) {
```
lucadalli, line 612: This is how we determine the start of a new interaction, irrespective of the number of touches.
In the case of two touches happening at the same time, Android fires two separate `touchstart` events, each having `targetTouches.length`, `changedTouches.length` equal to `1`.
On the other hand, iOS fires one `touchstart` event. `targetTouches.length` and `changedTouches.length` should both be equal to `2`.

lucadalli, line 612: Determines the start of a new interaction, irrespective of the number of touches.
In the case of two touches happening at the same time, Android fires two separate `touchstart` events, each having `targetTouches.length`, `changedTouches.length` equal to `1`.
On the other hand, iOS fires one `touchstart` event. `targetTouches.length` and `changedTouches.length` should both be equal to `2`.

packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -173,10 +173,12 @@ export class SmoothControls extends EventDispatcher {
           'touchstart', this.onTouchStart, {passive: true});
       element.addEventListener('touchmove', this.onTouchMove, {passive: false});
 
+      element.addEventListener('touchend', this.onTouchEnd);
```
lucadalli, line 176: We can attach the `touchend` listener to `element` instead of `self`. We can afford to do this because as observed [here](https://github.com/google/model-viewer/issues/2682#issuecomment-896964938), `touchmove` and `touchend` events fire on the target at `touchstart` time, even if that touch leaves the target's bounds.
Since all our event listeners are now attached to `element`, we can now react to `targetTouches` instead of `touches`. `targetTouches` are more suitable because we are not interested in touches which do not start on `element`.
This also helps avoid littering the global object with event listeners.

packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -432,13 +434,13 @@ export class SmoothControls extends EventDispatcher {
...
 
-  setTouchActionStyle() {
+  updateTouchActionStyle() {
     const {style} = this.element;
 
     if (this._interactionEnabled) {
       const {touchAction} = this._options;
       if (this._disableZoom && touchAction !== 'none') {
-        style.touchAction = `${touchAction} pinch-zoom`;
+        style.touchAction = 'manipulation';
```
lucadalli, line 443: Produces more consistent and intuitive behavior on Android.
Hoping it's the same for iOS.

elalish (approved):
This looks great, thanks! Completely fixes the problem as far as I can tell; Android and iOS are both working great and consistently. As a follow-on, would you mind taking a look at our Pan example and see if you can apply what you've learned here to make it behave well too? I want to make sure there's a clear path for people to add their own interactions on top of ours.

lucadalli:
Fantastic!

Sure. As soon as I have some time to spare I'll go through the relevant examples and make sure there's no erratic behavior and that they reflect our updated implementation.

lucadalli:
Fantastic!

Sure. As soon as I have some time to spare I'll go through the relevant examples and make sure there's no erratic behavior and that they reflect the new changes.

PR #2760, opened 2021-09-07
Update controls-related examples

lucadalli:
Updates relevant examples to be in line with the changes to SmoothControls.
Favors simplicity over completeness.
For example: after swiping vertically on the [panning example](http://localhost:8080/packages/modelviewer.dev/examples/stagingandcameras/#panning) to scroll, it is possible to add a second touch and pan while scrolling. SmoothControls has guards against zoom while scrolling but an unsavory amount of redundancy is necessary to port them to the examples.
To avoid this, we should consider adding an API to allow for user-defined `touchModes`.

### DIFF
--- a/packages/modelviewer.dev/examples/stagingandcameras/index.html
+++ b/packages/modelviewer.dev/examples/stagingandcameras/index.html
@@ -269,15 +269,15 @@
   }, true);
 
   modelViewer.addEventListener('touchstart', (event) => {
-    startX = event.touches[0].clientX;
-    startY = event.touches[0].clientY;
-    panning = event.touches.length === 2;
+    startX = event.targetTouches[0].clientX;
+    startY = event.targetTouches[0].clientY;
+    panning = event.targetTouches.length === 2;
     if (!panning)
       return;
 
-    const {touches} = event;
-    lastX = 0.5 * (touches[0].clientX + touches[1].clientX);
-    lastY = 0.5 * (touches[0].clientY + touches[1].clientY);
+    const {targetTouches} = event;
+    lastX = 0.5 * (targetTouches[0].clientX + targetTouches[1].clientX);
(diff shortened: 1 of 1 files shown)


packages/modelviewer.dev/examples/stagingandcameras/index.html
```diff
@@ -290,22 +290,26 @@ <h4>Event listeners can cooperate with camera-controls</h4>
...
   }, true);
   
-  self.addEventListener('touchend', (event) => {
-    if (event.touches.length === 0) {
+  modelViewer.addEventListener('touchend', (event) => {
+    if (event.targetTouches.length === 0) {
       recenter(event.changedTouches[0]);
+
+      if (event.cancelable) {
+        event.preventDefault();
```
lucadalli, line 311: Prevents emulation of `mouseup` event which was invoking `recenter` a second time.

packages/modelviewer.dev/examples/stagingandcameras/index.html
```diff
@@ -247,8 +247,8 @@ <h4>Event listeners can cooperate with camera-controls</h4>
         Math.abs(pointer.clientY - startY) > tapDistance)
       return;
     const rect = modelViewer.getBoundingClientRect();
-    const x = event.clientX - rect.left;
-    const y = event.clientY - rect.top;
+    const x = pointer.clientX - rect.left;
```
lucadalli, line 250: Lack of type-checking bit us again. Surprisingly, the browser injects `event` as a variable in the event handler scope.
Try:
`window.addEventListener('click', () => console.log(event))`

It was working with mouse events because unlike `TouchEvent`, `MouseEvent` has properties `clientX` and `clientY`.

elalish (approved):
Spectacular! Those examples both feel great across all my devices now. Thank you!

PR #2764, opened 2021-09-07
Restore initial material

timmmeh:
Allows passing 'null' to enableVariant() API in order to switch the material back its initial setting.

### DIFF
--- a/packages/model-viewer/src/features/scene-graph.ts
+++ b/packages/model-viewer/src/features/scene-graph.ts
@@ -149,8 +149,7 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
           return;
         }
 
-        await this[$model]![$switchVariant](
-            variantName == 'null' ? null : variantName!);
+        await this[$model]![$switchVariant](variantName!);
         this[$needsRender]();
         this.dispatchEvent(new CustomEvent('variant-applied'));
       }
--- a/packages/model-viewer/src/features/scene-graph.ts
+++ b/packages/model-viewer/src/features/scene-graph.ts
@@ -47,7 +47,7 @@ interface SceneExportOptions {
 
 export interface SceneGraphInterface {
   readonly model?: Model;
-  variantName: string|undefined;
+  variantName: string|null|undefined;
   readonly availableVariants: Array<string>;
   orientation: string;
   scale: string;
@@ -149,8 +149,10 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
           return;
         }
 
-        await this[$model]![$switchVariant](variantName!);
+        await this[$model]![$switchVariant](
+            variantName! == 'null' ? null : variantName!);
         this[$needsRender]();
+        this.dispatchEvent(new CustomEvent('variant-applied'));
       }
 
       if (changedProperties.has('orientation') ||
--- a/packages/model-viewer/src/features/scene-graph.ts
+++ b/packages/model-viewer/src/features/scene-graph.ts
@@ -47,7 +47,7 @@ interface SceneExportOptions {
 
 export interface SceneGraphInterface {
   readonly model?: Model;
-  variantName: string|null|undefined;
+  variantName: string|null;
   readonly availableVariants: Array<string>;
   orientation: string;
   scale: string;
@@ -70,7 +70,7 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
     private[$originalGltfJson]: GLTF|null = null;
 
     @property({type: String, attribute: 'variant-name'})
-    variantName: string|undefined = undefined;
+    variantName: string|null = null;
 
     @property({type: String, attribute: 'orientation'})
     orientation: string = '0 0 0';
@@ -150,7 +150,7 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
         }
 
         await this[$model]![$switchVariant](
-            variantName! == 'null' ? null : variantName!);
+            variantName == 'null' ? null : variantName!);
         this[$needsRender]();
         this.dispatchEvent(new CustomEvent('variant-applied'));
       }
--- a/packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
+++ b/packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
@@ -154,9 +154,9 @@ export class PrimitiveNode extends Node {
     return this.mesh.material;
   }
 
-  async enableVariant(name: string):
-      Promise<ThreeMaterial|ThreeMaterial[]|null> {
-    if (name === 'null') {
+  async enableVariant(name: string|
+                      null): Promise<ThreeMaterial|ThreeMaterial[]|null> {
+    if (name == null) {
       return await this.setActiveMaterial(this[$defaultMaterialIdx]);
     }
     if (this[$variantInfo] != null) {
--- a/packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
+++ b/packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
@@ -12,7 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import {Material as ThreeMaterial, Mesh, SkinnedMesh} from 'three';
+import {Material as ThreeMaterial, Mesh, MeshStandardMaterial, SkinnedMesh} from 'three';
 import {GLTFReference} from 'three/examples/jsm/loaders/GLTFLoader';
 
 import {CorrelatedSceneGraph} from '../../../three-components/gltf-instance/correlated-scene-graph.js';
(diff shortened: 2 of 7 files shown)


packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
```diff
@@ -47,13 +48,23 @@ export class PrimitiveNode extends Node {
...
     this[$mesh] = mesh;
     const {gltf, threeGLTF} = correlatedSceneGraph;
+    // Captures the primitive's default material.
+    const materialRef = correlatedSceneGraph.threeObjectMap.get(
+        mesh.material as MeshStandardMaterial);
+    if (materialRef != null) {
+      this[$defaultMaterialIdx] = materialRef.index;
+    } else {
+      console.error(
+          `Primitive (${mesh.name}) missing default material reference.`);
```
elalish, line 66: Is this possible? If no material was  specified, we should have created a default material and then reference that, right? If it's just an assert, that's fine, I'm just curious.

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
```diff
@@ -119,12 +121,26 @@ suite('scene-graph/model/mesh-primitives', () => {
...
-            primitive.enableVariant('Purple Yellow') as MeshStandardMaterial);
+            await primitive.enableVariant('Purple Yellow') as
+            MeshStandardMaterial);
       }
       expect(materials.find((material: MeshStandardMaterial) => {
         return material.name === 'purple';
       })).to.not.be.null;
     });
+
+    test('Primitive switches to default material', async () => {
```
elalish, line 132: This is good, but it's not a public method. Let's also include a more end-to-end version that sets it via the public `variantName` API. And let's also update the documentation on that.

elalish (changes requested):

packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
timmmeh: Correct the default material works and I don't think it would be possible for this to be null, at least not with todays codebase.

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
timmmeh: Added a test using the `variantName` API, a bit tricky to verify without being able to access a gltf-primitive mapping on the model, have a look, updating the docs next...

packages/model-viewer/src/test/features/scene-graph-spec.ts
```diff
@@ -85,6 +89,41 @@ suite('ModelViewerElementBase with SceneGraphMixin', () => {
...
+                  primitive[$defaultMaterialIdx] == 0) {
+                // Override enable variant to watch for change.
+                const override: PrimitiveNode&{superSetActiveMaterial: any} =
+                    primitive as PrimitiveNode & {superSetActiveMaterial: any};
+                override.superSetActiveMaterial = primitive.setActiveMaterial;
+
+                override.setActiveMaterial = async(material: number):
+                    Promise<ThreeMaterial|ThreeMaterial[]|null> => {
+                      defaultIndex = material;
+                      return override.superSetActiveMaterial(material);
```
elalish, line 110: This seems pretty complicated. Might there be a way to just check the `element.model![primitive][0]` for its material, then check that it's different when you switch variants, then check it's back to the original value when you set `null`?

elalish (changes requested):

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -149,8 +149,10 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
           return;
         }
 
-        await this[$model]![$switchVariant](variantName!);
+        await this[$model]![$switchVariant](
+            variantName! == 'null' ? null : variantName!);
```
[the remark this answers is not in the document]
timmmeh: lol, yeah

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -47,7 +47,7 @@ interface SceneExportOptions {
 
 export interface SceneGraphInterface {
   readonly model?: Model;
-  variantName: string|undefined;
+  variantName: string|null|undefined;
```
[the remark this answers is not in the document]
timmmeh: Done.

packages/model-viewer/src/test/features/scene-graph-spec.ts
timmmeh: Yeah that's what I mean when I say it's a bit tricky, we can access the primitives as you show but the order is not guaranteed at the moment so I can't rely on that, tests get flaky. However, I simplified things quite a bit, in this next update.

packages/modelviewer.dev/data/docs.json
```diff
@@ -907,6 +907,14 @@
           "<a href=\"../examples/scenegraph/#variants\">Related examples</a>"
         ]
       },
+      {
+        "name": "variantName",
+        "htmlName": "variantName",
+        "description": "Corresponding property of the attribute <code>variant-name</code>.",
```
elalish, line 913: This is not necessary (you'll notice none  of our other attributes have property definitions, since all  HTML attributes always have corresponding properties). That said, I'm open to changing the overall  docs template to  make that more clear.

packages/modelviewer.dev/data/docs.json
```diff
@@ -867,8 +867,8 @@
     "Attributes": [
       {
         "name": "variant-name",
-        "htmlName": "variantName",
-        "description": "Selects a model variant by name.",
+        "htmlName": "variant-name",
```
elalish, line 870: Let's stick with our conventions here. Also, might be good to add that the default is `null` and options include variant name string or `null`, in the `default` json.

packages/model-viewer/src/test/features/scene-graph-spec.ts
```diff
@@ -93,35 +93,30 @@ suite('ModelViewerElementBase with SceneGraphMixin', () => {
           `Setting varianName to null results in primitive
            reverting to default/initial material`,
           async () => {
-            let defaultIndex = 0;
-            // Patches the first primitive with variant-info and a default
-            // material index of 0.
+            // let defaultIndex = 0;
```
elalish: remove?

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -149,8 +149,10 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
           return;
         }
 
-        await this[$model]![$switchVariant](variantName!);
+        await this[$model]![$switchVariant](
+            variantName == 'null' ? null : variantName!);
```
elalish, line 153: Did you see  my whole comment on this line?

packages/model-viewer/src/test/features/scene-graph-spec.ts
```diff
@@ -85,6 +89,35 @@ suite('ModelViewerElementBase with SceneGraphMixin', () => {
...
+            element.variantName = 'Yellow Red';
+            await waitForEvent(element, 'variant-applied');
+            expect((primitiveNode!.mesh.material as MeshStandardMaterial).name)
+                .equal('red');
+
+            // Switches to null variant.
+            element.variantName = null;
+            await waitForEvent(element, 'variant-applied');
+            expect((primitiveNode!.mesh.material as MeshStandardMaterial).name)
+                .equal('purple');
```
elalish, line 118: 👍

elalish (changes requested):
Looks good, just a couple minor things.

packages/model-viewer/src/features/scene-graph.ts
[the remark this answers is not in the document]
timmmeh: passing `variantName` as is, it's possible, this API responds to the actual value being null though, not a magical variant-name 'null' that's why this is here. Also if you assign the value `null` to the property `variantName` Lit apparently converts that into a string, don't know why, that's why I'm checking for a string at all.

packages/model-viewer/src/test/features/scene-graph-spec.ts
timmmeh: done.

packages/modelviewer.dev/data/docs.json
timmmeh: Wasn't sure if this was omitted on purpose or not, omitting it does assume that anyone can work out the fact that `variant-name` corresponds to a property called `variantName`.

packages/modelviewer.dev/data/docs.json
timmmeh: Is the "htmlName" supposed to be visible somewhere in on the documentation page? I don't see it.

timmmeh: Done.

packages/model-viewer/src/features/scene-graph.ts
timmmeh: Updated that comment.

packages/model-viewer/src/features/scene-graph.ts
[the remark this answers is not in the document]
timmmeh: Just works with strings now.

elalish (approved):
Thanks!

PR #2770, opened 2021-09-09
Fix editor mobile file open

### LINKED ISSUE #2672: [Editor] GLB selector doesn't open file selection dialogue on mobile
### Description

We occasionally have to view GLB files on-the-go and the modelviewer.dev/editor is a pretty good general viewer (and one of the few with AR capabilities), so I was trying to view a model on an iPad. 
The only (near-) blocker was a UX detail: the file selector doesn't seem to open on mobile.

My workaround was to use split screen and drag the file over to the page, which worked on iPad but didn't on an Android phone.

#### Live Demo
Open modelviewer.dev/editor on iPad
Try to upload a GLB file to view it
IMG_20210807_183253

### Browser Affected

- [x] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [x] Safari

### OS

- [x] Android
- [x] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### AR

- [ ] WebXR
- [ ] SceneViewer
- [ ] QuickLook

elalish:
Fixes #2672 

I'm not sure why it wasn't working on mobile, but simplifying it fixed it. I decided to take the opportunity to remove all the USDZ-related dead code as well, since I don't  want to maintain it. The editor can focus on the auto-generation  path.


timmmeh (commented):
Is the point of this PR to just remove USDZ support from the editor because it's support through THREE?

elalish:
The  point was to fix the bug (file modal on mobile). The  file modal  class turned out to be broken, so I  removed it. One thing using it was a bunch of dead  code for USDZ support I'd already  removed from the UI for the reason you noted. So I decided to strip out the rest of the related state while I  was there.

PR #2779, opened 2021-09-13
added joint count to editor validation

elalish:
This was an internal feature request for validating animated models.

### DIFF
--- a/packages/space-opera/src/components/model_viewer_snippet/components/validation.ts
+++ b/packages/space-opera/src/components/model_viewer_snippet/components/validation.ts
@@ -20,11 +20,12 @@ import {customElement, html, internalProperty, LitElement, property, query} from
 import {validationStyles} from '../../../styles.css.js';
 import {State} from '../../../types.js';
 import {ConnectedLitElement} from '../../connected_lit_element/connected_lit_element';
-import {getGltfUrl} from '../../model_viewer_preview/reducer.js';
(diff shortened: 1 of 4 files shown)


timmmeh (approved):

packages/space-opera/src/components/model_viewer_snippet/components/validation.ts
```diff
@@ -166,6 +183,18 @@ export class Validation extends ConnectedLitElement {
     }
   }
 
+  countJoints(gltf: GLTF) {
+    const jointSet = new Set();
+    if (gltf.skins != null) {
+      for (const skin of gltf.skins) {
```
asemanh, line 189: Do we care to have report.info when gltf.skins is null? If so, we could probably have an early return when gltf.skins is null

And, the three dots concept is cool. Didn't know about it. Though not sure what happens when this.report doesn't exist

asemanh (approved):

packages/space-opera/src/components/model_viewer_snippet/components/validation.ts
elalish: It looks nicer in the output when it says `0 joints` instead of `joints`, so I switched it to this way. 

I also love the ellipsis, and I'm also not sure what happens when the original in undefined, but I went ahead and  initialized `this.report = {}` to avoid that.

PR #2781, opened 2021-09-14
Fix glTF loading in the editor

elalish:
Now you can drag and drop a glTF folder or all the files that make up a glTF, or multi-select them in the dialogue and it all loads properly.

### DIFF
--- a/packages/render-fidelity-tools/package.json
+++ b/packages/render-fidelity-tools/package.json
@@ -23,8 +23,8 @@
   },
   "dependencies": {
     "@actions/core": "^1.2.4",
-    "@babylonjs/core": "^5.0.0-alpha.31",
-    "@babylonjs/loaders": "^5.0.0-alpha.31",
+    "@babylonjs/core": "*5.0.0-alpha.31",
+    "@babylonjs/loaders": "*5.0.0-alpha.31",
     "@google/model-viewer": "^1.8.0",
     "@khronosgroup/gltf-viewer": "^1.0.7",
     "@polymer/paper-button": "^3.0.1",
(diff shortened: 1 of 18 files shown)


packages/render-fidelity-tools/package.json
```diff
@@ -23,8 +23,8 @@
   },
   "dependencies": {
     "@actions/core": "^1.2.4",
-    "@babylonjs/core": "^5.0.0-alpha.31",
-    "@babylonjs/loaders": "^5.0.0-alpha.31",
+    "@babylonjs/core": "*5.0.0-alpha.31",
+    "@babylonjs/loaders": "*5.0.0-alpha.31",
```
elalish, line 27: This tried to update to a newer version when I added our simple-dropzone dependency, and broke something. I pinned the  version to fix it, but we'll have to deal with this when we update eventually.

packages/space-opera/src/components/model_viewer_snippet/components/open_button.ts
```diff
@@ -216,19 +217,57 @@ export class ImportCard extends LitElement {
...
-  async onUpload(files) {
-    if (!files) {
-      /// The user canceled the previous upload
-      return;
+  async onUpload(fileMap: Map<string, File>) {
+    const modelViewer = getModelViewer();
+    let rootPath: string;
+    for (const [path, file] of fileMap) {
+      const filename = file.name.toLowerCase();
+      if (filename.match(/\.(gltf|glb)$/)) {
```
timmmeh, line 230: Nit: It looks like maybe this loop is just finding the .gltf/.glb file and then doing some stuff with it? If so I think you can break at the bottom of _if_ statement yeah?

timmmeh: NM braces are difficult to follow in this view

timmmeh (approved):

PR #2799, opened 2021-09-16
Patch to resolve mapping issue on skinned meshes

timmmeh:
Patches association issue with skinned-mesh


elalish (approved):
Let's add a skinned mesh test in your next PR. Thanks for the quick turn around!

PR #2806, opened 2021-09-19
Update dspbr-pt renderer -> v0.3.2

bsdorra:
The pull request doesn't change the renderer config and it does not update golden renderings.


packages/render-fidelity-tools/src/components/renderers/dspbr-pt-viewer.ts
```diff
@@ -105,18 +118,17 @@ export class PathtracingViewer extends LitElement {
       numSamples = 2048;
     }
 
+    console.log(this[$camera].position);
+    console.log(this[$camera].matrixWorld);
```
elalish, line 122: do  we need all these console logs?

elalish (approved):
Thank you for the update! I'll update the config as a  follow-on. Are there  any glTF sample models that are not yet supported?

packages/render-fidelity-tools/src/components/renderers/dspbr-pt-viewer.ts
bsdorra: Not needed, removed them

bsdorra:
> Thank you for the update! I'll update the config as a follow-on. Are there any glTF sample models that are not yet supported?

Most of the new PBR tests should work just fine . I suggest to add:
- khronos-DragonAttenuation 
- khronos-IridescentDishWithOlives
- khronos-SheenChair
- khronos-SheenCloth
- khronos-TransmissionRoughnessTest
- khronos-TransmissionTest
- khronos-AttenuationTest
- khronos-SpecularTest

These two are broken for unknown reasons, need to investigate
- khronos-GlamVelvetSofa
- khronos-ToyCar

elalish:
Actually, @bsdorra there seems to be a problem here. I can't get it to update any screenshots and when  I go to `http://127.0.0.1:8080/packages/render-fidelity-tools/test/renderers/dspbr-pt/` and choose a scenario I get this in the console: 
```
100% loaded
renderer.js:6 Uncaught (in promise) ProgressEvent
bvh.js:319 Building BVH for 682 triangles.
bvh.js:327 BVH Depth = 10; NumNodes = 251
gl_utils.js:86 Create data texture: 502 x 1
gl_utils.js:86 Create data texture: 10230 x 1
gl_utils.js:86 Create data texture: 9 x 1
gl_utils.js:86 Create data texture: 22 x 1
renderer.js:768 Create texture array: 2048 x 2048 x 3
Failed to load resource: the server responded with a status of 404 (Not Found) :8080/packages/render-fidelity-tools/test/renderers/dspbr-pt/shader/pt.vert:1 
```
It looks like some kind of relative file path problem in your library, since it is somehow trying find its own internals at my web address.

bsdorra:
> It looks like some kind of relative file path problem in your library, since it is somehow trying find its own internals at my web address.

@elalish Did the npm `prepare` script run after the dspbr-pt update? If not, running it should solve the issue. 
[`prepare`]( https://github.com/google/model-viewer/blob/64080d53834243576f14a88a4748310cbdeeaa6e/packages/render-fidelity-tools/package.json#L12) copies the shader folder from _node_modules/dspbr-pt/dist/shader_ to the renderer directory.

elalish:
Okay, the prepare script did fix that, though it seems like a strange way to solve that problem. Anyway, I can now get it running, though it is incredibly slow. Have you tried generating any of these renders with e.g. `npm run update-screenshots dspbr-pt khronos-DragonAttenuation`? Is there any way we could get a progress indicator of some kind? Even the Avocado doesn't look like it's converging. Also, I'm getting GL context during scene setup on some models, which is caught in such a way that it doesn't notify the test, so it just hangs:
```
Create texture array: 2048 x 2048 x 1
localhost/:1 WebGL: CONTEXT_LOST_WEBGL: loseContext: context lost
gl_utils.js:47 Uncaught (in promise) Error: Could not create shader
    at createShader (gl_utils.js:47)
    at Object.createProgramFromSource (gl_utils.js:58)
    at PathtracingRenderer.<anonymous> (renderer.js:826)
    at Generator.next (<anonymous>)
    at fulfilled (renderer.js:5)
```

elalish:
Good to know. My laptop has a pretty powerful GPU, and it's what I've used before. I know it took awhile, but I feel like it's slowed down quite a bit. Did you up the number of bounces, or something else that may have hit performance significantly?

Do  you have access to any  machines that would be good for creating the goldens?

bsdorra:
May I ask which GPU you're using? 
Right, performance generally decreased with the addition of the new material features. Transmission, for example, does have quite an impact on the complexity of the BSDF sampling code. I didn't realize that it is that bad though. 

I've a RTX6000 in my work machine, that should do :) I'll create a another pull request when the goldens are ready.

elalish:
Ah, good call! I have  a dual GPU macbook and I just checked and it's using the weak intel one instead of the big AMD. I bet this is because Chrome switched their default to use 'low-power'. I bet if you set `powerPreference = 'high-performance'` on three's renderer it will be much improved. 

Anyway, a golden update PR would be much appreciated!

bsdorra:
Oh ok, this explains the performance issues... 
Actually, it's my bad. I remember we were having the high-performance flag problem/conversation before. The flag got lost when I ported the renderer from three to plain WebGL. I just realized that it is an actual parameter of the WebGL context creation, so I'm adding it back in...

PR #2816, opened 2021-09-22
Prepares Model-Viewer for Three.js glTF associations updates

timmmeh:
Updates apply the latest changes coming down from Three.js in regards to multiple-associations (https://github.com/mrdoob/three.js/pull/22544)


asemanh (approved):

packages/model-viewer/src/test/three-components/gltf-instance/correlated-scene-graph-spec.ts
```diff
@@ -109,11 +111,10 @@ suite('correlated-scene-graph', () => {
...
         let name;
         cloneCorrelatedSceneGraph.threeObjectMap.forEach(
-            (reference, threeObject) => {
+            (mappings, threeObject) => {
               if ((threeObject as MeshStandardMaterial).isMaterial) {
-                name =
-                    cloneCorrelatedSceneGraph.gltf.materials![reference.index]
-                        .name;
+                const index = mappings.materials!;
+                name = cloneCorrelatedSceneGraph.gltf.materials![index].name;
```
elalish, line 117: Good start; are there any end-to-end tests we should add? I recall you saying this was fixing  a bug (or removing a workaround) for loading skinned meshes?

elalish (commented):
We should probably talk about the three.js changes before we do much here.

packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
```diff
@@ -157,7 +131,8 @@ export class PrimitiveNode extends Node {
   async enableVariant(name: string|
                       null): Promise<ThreeMaterial|ThreeMaterial[]|null> {
     if (name == null) {
-      return await this.setActiveMaterial(this[$defaultMaterialIdx]);
+      // tslint:disable-next-line:no-return-await needed by FireFox
+      return await this.setActiveMaterial(this[$initialMaterialIdx]);
```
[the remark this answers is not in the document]
timmmeh: Thats what I thought but I tracked some flaky Firefox behavior here and this seemed to resolve it, I was thinking maybe there are circumstance where this gets treated differently on different browsers or is maybe a browser bug, or maybe I was just getting a false positive several times in a row 🤷

packages/model-viewer/src/test/three-components/gltf-instance/correlated-scene-graph-spec.ts
timmmeh: Will be adding these soon...

packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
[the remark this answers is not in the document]
elalish: When were you seeing this flakiness? I notice this change is only to add the `//tslint`, not the actual `await`. If there's a  race condition in Firefox, it could easily mean there's a race generally that just isn't  getting triggered because of perf differences between browsers. I'd like to make sure we understand the root cause.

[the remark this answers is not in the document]
timmmeh: I added the await first, then added the comment later to prevent removing the await accidentally during the course of changes. It's easy to repro just do `npm run watch:test` then open the tests in Firefox and run.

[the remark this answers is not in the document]
timmmeh: I just removed the await and tested again on Firefox and it had no problems, there were other synchronous issues that were addressed that probably resolved this, I think it's safe to remove.

packages/model-viewer/src/test/features/scene-graph-spec.ts
```diff
@@ -236,5 +240,20 @@ suite('ModelViewerElementBase with SceneGraphMixin', () => {
...
+      test('has a mapping for each primitive mesh', async () => {
+        element.src = RIGGEDFIGURE_GLB_PATH;
+
+        await waitForEvent(element, 'scene-graph-ready');
+
+        const gltf = (element as any)[$currentGLTF] as ModelViewerGLTFInstance;
+
+        for (const primitive of element.model![$primitives]) {
+          expect(gltf.correlatedSceneGraph.threeObjectMap.get(primitive.mesh))
+              .to.be.ok;
```
elalish, line 254: 👍

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
```diff
@@ -60,47 +60,62 @@ export class CorrelatedSceneGraph {
...
-      // glTF. In this case we append a default material to allow this to be
-      // operated upon.
-      if (gltfElementReference == null) {
+    const defaultReference = {type: 'materials', index: -1};
+
+    for (const threeMaterial of associations.keys()) {
+      // Note: GLTFLoader creates a "default" material that has no
+      // corresponding glTF element in the case that no materials are
+      // specified in the source glTF. In this case we append a default
+      // material to allow this to be operated upon.
```
elalish, line 74: Do we have a test for loading a glTF with no  material? Might be good to check that this gets created properly.

elalish (commented):
Looking  good.

packages/model-viewer/src/three-components/gltf-instance/correlated-scene-graph.ts
timmmeh: correlated-scene-graph does cover the default material, but I will add an e2e test covering a primitives configuration of the default material.

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
```diff
@@ -41,6 +45,28 @@ const findPrimitivesWithVariant = (model: Model, variantName: string) => {
...
+
+    teardown(() => {
+      document.body.removeChild(element);
+    });
+
+    test('has a default material', async () => {
+      const model = element.model!;
+      expect(model[$primitives].length).to.equal(1);
+      expect(model.materials.length).to.equal(1);
+      expect(model.materials[model[$primitives][0][$initialMaterialIdx]].name)
```
elalish, line 65: Seems like you  could just  put `model.materials[0]` and maybe separately check that  `$initialMaterialIdx  === 0`.

elalish (approved):
Love it when the update is smooth!

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
timmmeh: Done

PR #2847, opened 2021-10-07
Fix xr-environment

elalish:
I noticed recently that though the `xr-environment` attribute is supposed to be opt-in  for WebXR  estimated lighting, it was in fact enabled all the  time accidentally. This  rectifies that and  also takes care of properly disposing of  it at session end. 

A handy way to test this is to edit the snippet in our  editor and  add the  `xr-environment`  attribute. Even though its not exposed in  the UI, any MV attributes you add will be  respected by both  the  local view and the mobile deployments, which makes it  easy to test advanced features.


timmmeh:
Is there a way to control the frequency of lighting updates? I know in SceneViewer controlling this helped to reduce pressure on the GPU.

PR #2925, opened 2021-10-26
fixed animated picking

elalish:
Picking (raycasting) was occasionally not working for animated models. Turns out this is because the broad phase uses bounding boxes to cull meshes, but the bounding  boxes are wrong for animated models. If they're wrong by enough, then it  doesn't bother checking the triangles for intersection. This is fixed by setting both the bounding sphere  and  bounding box in such a  way as to circumvent the early-outs. 

I also changed the editor to default to `bounds="tight"`, which I'd simply forgotten to before since this is the more precise way to frame.


bokssssss:
WOW 😳

Issue #2915, opened 2021-10-24
v1.9.0 performance regression


lucadalli:
### Description

After upgrading to v1.9, I was able to observe a significant slowdown for the `setTexture` method. While I was creating a minimal reproduction demo for `setTexture`, I noticed that initial load performance also suffered a regression. I imagine this stems from the same issue.

#### Live Demo
https://codepen.io/citylights/pen/qBXRwvX

The Codepen above serves as a primitive way to measure load performance. Alternate between version 1.8.0 and 1.9.0 by changing the script tag at the top of the `template`. 
Top left of the view shows the visible model count followed by the duration taken to load those models.
On my machine I am consistently measuring ~1500ms for v1.8 and ~10,000ms for v1.9 for 10 models.

### Version

- model-viewer: v1.9.0

### Browser Affected

- [x] Chrome, Win 10 Version 94.0.4606.81 (Official Build) (64-bit)
- [ ] Edge
- [x] Firefox
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [x] Windows

### AR

- [ ] WebXR
- [ ] SceneViewer
- [ ] QuickLook

elalish:
Reopening, since I  had to revert the fix since it was causing a bigger problem: #2949. A proper fix is  coming in three.js, but  it'll take our next release cycle to resolve.

sprengerst:
Oh no :( 
When will the next release cycle happen? It makes whole pages unusable :-/

elalish:
@sprengerst you're welcome to pin to either v1.8 or v1.9.1 until our next release comes, e.g. `<script type="module" src="https://unpkg.com/@google/model-viewer@1.8.0/dist/model-viewer.min.js"></script>`. It's the  holidays, so it'll probably be after the new year.

sprengerst:
@elalish Thank you. Holidays? What country do you live in, I think I should join you. :D

PR #2935, opened 2021-10-29
removed texture cloning

elalish:
So, while looking into #2915, I noticed that our cloning of textures is in fact causing those copies to be independently uploaded to the GPU, which is causing much of the obscene slow-load  of  that example with  ten identical models. I  originally put in the cloning so that our `setURI` method on the scene-graph would change textures independently for two identical models. This change obviously regresses that, however I think it's the right call now that we have  a better texture creation API and since this increases performance and simplifies the code considerably. 

Note, this does not actually fix the real problem  in #2915, which appears to be caused by very slow shader  compilation, which has seen been resolved upstream in three.js. 

I have also removed `RoughnessMipmapper` here, as it has become too much of a headache to maintain. It also costs a little startup time and GPU memory for all models loaded in MV, while having a noticeable effect on very few. Instead, if someone has a  model that could use this (shiny material with high-frequency normal map), I will instead  recommend the use of KTX2 textures (which everyone should probably be using anyway); these require all mip levels to be present in the file in the first place, so you'll simply want to ensure you use a tool that takes appropriate note of the  normal map when generating roughness mips. FYI @pushmatrix @mikkoh.


timmmeh (dismissed):
Other than the shiny dog issue which also appears on master everything looks fine.

PR #2936, opened 2021-10-30
switch back to WebGL1Renderer

### LINKED ISSUE #2915: v1.9.0 performance regression

elalish:
Fixes #2915 

There's a bunch of discussion in three.js and apparently there's a chrome bug in webGL2 around a particular sRGB format. Hopefully a better fix will be coming soon, but until then, it  seems our best way around this performance issue is to go back to WebGL1 for the moment.

### DIFF
--- a/packages/model-viewer/src/three-components/Renderer.ts
+++ b/packages/model-viewer/src/three-components/Renderer.ts
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import {ACESFilmicToneMapping, Event, EventDispatcher, GammaEncoding, PCFSoftShadowMap, WebGLRenderer} from 'three';
+import {ACESFilmicToneMapping, Event, EventDispatcher, GammaEncoding, PCFSoftShadowMap, WebGL1Renderer} from 'three';
 import {RoughnessMipmapper} from 'three/examples/jsm/utils/RoughnessMipmapper';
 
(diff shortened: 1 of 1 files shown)


timmmeh:
👍

PR #2984, opened 2021-11-18
Variant editor

timmmeh:
Adds ability to create new variants to the editor.


packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
```diff
@@ -175,4 +174,63 @@ export class PrimitiveNode extends Node {
...
+    if (!this.validateMaterial(materialIndex) ||
+        !this.validateVariant(variantName)) {
+      return;
+    }
+    const material = this[$materials].get(materialIndex)!;
+    material.variants.add(variantName);
+    this.variantInfo.set(variantName, material);
+
+    this.updateVariantUserData(
+        variantName, this[$materials].get(materialIndex)!);
```
elalish, line 188: could you just use `material`?

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
```diff
@@ -166,6 +166,74 @@ suite('scene-graph/model', () => {
...
+          const onVariantApplied = async () => {
+            expect(primitive.getActiveMaterial().name)
+                .to.equal('test-material');
+          };
+
+          element.addEventListener('variant-applied', onVariantApplied);
+          element.variantName = 'test-variant';
+
+          await element.updateComplete;
+          element.removeEventListener('variant-applied', onVariantApplied);
```
elalish, line 215: I think this test will pass if `variant-applied` never fires?

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
```diff
@@ -166,6 +166,74 @@ suite('scene-graph/model', () => {
...
+        });
+
+        test('New variant is available to model-viewer', async () => {
+          await loadModel(CUBES_GLTF_PATH);
+          model.createVariantFromMaterial(0, 'test-material', 'test-variant');
+
+          expect(element.availableVariants.find(variant => {
+            return variant === 'test-variant';
+          })).to.not.be.undefined;
```
elalish, line 224: maybe just `to.be.ok`?

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
```diff
@@ -127,16 +133,7 @@ export class ModelViewerGLTFInstance extends GLTFInstance {
...
-      material: MeshStandardMaterial,
-      sourceUUIDToClonedMaterial: Map<string, Material>) {
-    // If we already cloned this material (determined by tracking the UUID of
-    // source materials that have been cloned), then return that previously
-    // cloned instance:
-    if (sourceUUIDToClonedMaterial.has(material.uuid)) {
-      return sourceUUIDToClonedMaterial.get(material.uuid)!;
-    }
-
+  static[$cloneAndPatchMaterial](material: MeshStandardMaterial): Material {
```
elalish, line 136: I think this is fine, but would you mind explaining briefly what these changes are about?

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -120,8 +132,33 @@ export class MaterialPanel extends ConnectedLitElement {
...
   }
 
-  getMaterial() {
-    return getModelViewer()!.model!.materials[this.selectedMaterialIndex];
+  getMaterial(options: GetMaterialOptions = GetMaterialOptions.Default) {
+    let material =
+        getModelViewer()!.model!.materials[this.selectedMaterialIndex];
+
+    // Creates a new instance if allowed.
+    if (options === GetMaterialOptions.AllowInstancing) {
```
elalish, line 140: Maybe this should just be its own  function, instead  of using options to change the behavior?

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -414,20 +471,22 @@ export class MaterialPanel extends ConnectedLitElement {
   }
 
   onBaseColorChange() {
-    this.getMaterial().pbrMetallicRoughness.setBaseColorFactor(
-        this.selectedBaseColor);
+    const material = this.getMaterial(GetMaterialOptions.AllowInstancing);
```
elalish, line 474: Maybe I'm just missing something, but where is the logic that changes `AllowInstancing`?

elalish (changes requested):
I haven't tried the editor yet with these changes, so this is just code review for now.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -245,4 +262,109 @@ export class Model implements ModelInterface {
...
+      }
+    }
+
+    if (!created) {
+      return null;
+    }
+
+    this.createVariant(variantName);
+
+    if (activateVariant) {
```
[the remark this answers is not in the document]
timmmeh: yes but it's async, I need to be able to create and activate the new variant and it can't be async.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -245,4 +262,109 @@ export class Model implements ModelInterface {
...
+    const gltfSourceMaterial =
+        JSON.parse(JSON.stringify(material[$sourceObject])) as GLTFMaterial;
+    gltfSourceMaterial.name = newMaterialName;
+
+    const clonedSet = new Set<MeshStandardMaterial>();
+    for (const [i, threeMaterial] of threeMaterialSet.entries()) {
+      const clone = ModelViewerGLTFInstance[$cloneAndPatchMaterial](
+                        threeMaterial) as MeshStandardMaterial;
+      clone.name =
+          newMaterialName + (threeMaterialSet.size > 1 ? '_inst' + i : '');
```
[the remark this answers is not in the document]
timmmeh: This is an edge case, if there are more than one three-material represented by the MV material than some extra decorating is applied to the name, they have unique names to start with but we are applying a new name to the material so this ensure each instance has a unique name.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -245,4 +262,109 @@ export class Model implements ModelInterface {
...
+      }
+
+
+      if (primitive.addVariant(variantMaterial, variantName)) {
+        created = true;
+      }
+    }
+
+    if (!created) {
+      return null;
```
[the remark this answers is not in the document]
timmmeh: If the variant already exists this prevents it from being created again.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -245,4 +262,109 @@ export class Model implements ModelInterface {
...
+
+  createVariant(variantName: string) {
+    // Updates the variants list in user data.
+    const threeGLTF = this[$correlatedSceneGraph].threeGLTF;
+    threeGLTF.userData.variants =
+        threeGLTF.userData.variants || new Array<string>();
+    const variants = threeGLTF.userData.variants as string[];
+
+    const found = variants.find(name => name === variantName);
+    if (found == null) {
```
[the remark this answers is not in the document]
timmmeh: done.

packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts
timmmeh: I refactored this to remove a random piece of dependent data from the method signature, which allows it to be called from anywhere. I did this to avoid code duplication when cloning material to create a variant.

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
timmmeh: done.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -245,4 +262,109 @@ export class Model implements ModelInterface {
...
+      variants.push(variantName);
+      // Updates variant names seen by the rest of ModelViewer.
+      this[$variants] = variants.slice();
+    }
+  }
+
+  addMaterialToVariant(materialIndex: number, targetVariantName: string) {
+    if (this[$variants].find(name => name === targetVariantName) == null) {
+      this.createVariant(targetVariantName);
```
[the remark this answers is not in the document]
timmmeh: done.

packages/space-opera/src/components/materials_panel/materials_panel.ts
timmmeh: Done.

packages/space-opera/src/components/materials_panel/materials_panel.ts
timmmeh: The enum is just used to identify places in code where I want to get a material that could be instanced (if the material does not exist for a variant) no logic for setting/changing it just gets passed in, but it's moot it's gone now.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -245,4 +262,109 @@ export class Model implements ModelInterface {
...
+  [$cloneMaterial](index: number, newMaterialName: string): Material {
+    const material = this.materials[index];
+
+    if (!material.isLoaded) {
+      console.error(`Cloning an unloaded material,
+           call 'material.ensureLoaded() before cloning the material.`);
+    }
+
+    const threeMaterialSet =
+        material[$correlatedObjects] as unknown as Set<MeshStandardMaterial>;
```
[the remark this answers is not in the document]
timmmeh: I think that is just left over from some code churn, removed.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -245,4 +262,109 @@ export class Model implements ModelInterface {
...
+      variantName: string, activateVariant: boolean = true): Material|null {
+    const variantMaterial =
+        this[$cloneMaterial](originalMaterialIndex, newMaterialName);
+
+    let created = false;
+    for (const primitive of this[$primitivesList]) {
+      if (primitive.variantInfo.has(variantName) ||
+          (primitive.variantInfo.has(variantName) &&
+           primitive.variantInfo.get(variantName)!.index !==
+               originalMaterialIndex)) {
```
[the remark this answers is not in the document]
timmmeh: done.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -245,4 +262,109 @@ export class Model implements ModelInterface {
     }
     await Promise.all(promises);
   }
+
+  [$cloneMaterial](index: number, newMaterialName: string): Material {
+    const material = this.materials[index];
+
+    if (!material.isLoaded) {
+      console.error(`Cloning an unloaded material,
+           call 'material.ensureLoaded() before cloning the material.`);
```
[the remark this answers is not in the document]
timmmeh: Correct just trying to avoid making it async as that generate a ton of changes in material_panel, which might be ok I just wouldn't want to do that in this PR.

timmmeh:
> I haven't tried the editor yet with these changes, so this is just code review for now.

So I have it working in the editor but is some glaring UI feedback problems (IMO) and some usage issues to address, but for now I just want it to be functional.

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
timmmeh: It looks `await element.updateComplete;` does what I want, which is to make this behave synchronously.

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -124,6 +131,34 @@ export class MaterialPanel extends ConnectedLitElement {
     return getModelViewer()!.model!.materials[this.selectedMaterialIndex];
   }
 
+  getMaterialVariant() {
+    let material =
+        getModelViewer()!.model!.materials[this.selectedMaterialIndex];
```
elalish, line 136: Perhaps just call `getMaterial()` here?

elalish (commented):
Thanks, that's looking better, though I think you missed a couple of my comments. Since merging is deployment for the editor, we'll need to polish up the  UX before we can merge this.

packages/space-opera/src/components/materials_panel/materials_panel.ts
timmmeh: done.

packages/model-viewer/src/features/scene-graph/nodes/primitive-node.ts
timmmeh: yes... like this?

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -191,7 +198,20 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
   }
 
   get name(): string {
-    return (this[$sourceObject] as Material).name || '';
+    return (this[$sourceObject] as Material).name;
+  }
+
+  set name(name: string) {
```
elalish, line 204: These should be added to api.ts.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -56,6 +61,19 @@ export class LazyLoader {
   }
 }
 
+/**
+ * Facades variant mapping data.
+ */
+export class VariantData {
+  name: string;
+  index: number;
+  materialVariants = new Array<number>();
```
elalish, line 70: maybe just `materials`? These are just the material indices this variant points to,  right?

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -56,6 +61,19 @@ export class LazyLoader {
...
+  constructor(name: string, index: number) {
```
elalish, line 71: I don't think you need this in TS.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -226,16 +267,18 @@ export class Model implements ModelInterface {
...
     await Promise.all(promises);
 
-    for (const material of this.materials) {
-      material[$setActive](false);
-    }
-    // Marks the materials that are now in use after the variant switch.
-    for (const primitive of this[$primitivesList]) {
-      this.materials[primitive.getActiveMaterial().index][$setActive](true);
-    }
+    const updateActiveState = async () => {
```
elalish, line 272: I  don't think this is doing what you think  it's doing. It's marked `async`, but it doesn't await anything, so it'll  return  immediately even if the  functions it calls are  also async.

packages/model-viewer/src/features/scene-graph/api.ts
```diff
@@ -49,6 +49,32 @@ export declare interface Model {
    * @returns the first material to whose name matches `name`
    */
   getMaterialByName(name: string): Material|null;
+
+  /**
+   * Creates a new material variant from an existing material.
```
elalish, line 54: Might be nice to explain when you'll get `null`.

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -121,11 +139,54 @@ export class MaterialPanel extends ConnectedLitElement {
...
+    return material;
   }
 
   getOriginalMaterial() {
-    return this.originalGltf!.materials[this.selectedMaterialIndex];
+    if (this.originalGltf!.materials.length > this.selectedMaterialIndex) {
+      return this.originalGltf!.materials[this.selectedMaterialIndex];
+    } else {
+      return this.originalGltf!.materials[this.originalMaterialFromCloneMap.get(
+          this.selectedMaterialIndex)!];
```
elalish, line 188: nit: this could probably just be a ternary op on the  index.

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -319,6 +513,23 @@ export class MaterialPanel extends ConnectedLitElement {
         this.onClick(event);
       }
     });
+
+    const self = this;
```
elalish, line 517: Can you  explain what's  going on here?

packages/space-opera/src/components/materials_panel/materials_panel.ts
```diff
@@ -256,29 +318,149 @@ export class MaterialPanel extends ConnectedLitElement {
...
+
+  renderEditMaterialNameDialog() {
+    return html`
+      <input-dialog id="edit-material-name" modal="true"
+        placeholder="Enter Material Name">
+      </input-dialog>
+    `;
+  }
+
+  onCreateVariant(newVariantName: string) {
```
elalish, line 443: I have no idea where  this  bug is, so I just picked a random line  for this comment. repro:
1) open the robot
2) create a variant
3) switch back to default variant
4) change a color
5) switch to new variant
note that both variants now have the new color instead of just the default.

elalish (approved):
Thanks, this looks pretty good. I'm approving in  case you'd rather address the remaining issues in your following PR.

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -48,7 +47,7 @@ interface SceneExportOptions {
 export interface SceneGraphInterface {
   readonly model?: Model;
   variantName: string|null;
-  readonly availableVariants: Array<string>;
+  readonly availableVariants: string[];
```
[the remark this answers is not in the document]
timmmeh: Oh this just satisfies Googles linter, apparently `field: Array<string>` is prohibited.

packages/model-viewer/src/features/scene-graph/api.ts
```diff
@@ -49,6 +49,32 @@ export declare interface Model {
...
+   * @param originalMaterialIndex index of the material to clone the variant
+   *     from.
+   * @param materialName the name of the new material
+   * @param variantName the name of the variant
+   * @param activateVariant activates this material variant, i.e. the variant
+   *     material is rendered, not the existing material.
+   */
+  createVariantFromMaterial(
+      originalMaterialIndex: number, newMaterialName: string,
+      variantName: string, activateVariant: boolean): Material|null;
```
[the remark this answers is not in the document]
timmmeh: this will actually ensure the variant is created and clone the material and assign the clone to the variant.
Using `createVariant` followed `setMaterialToVariant` is how to assign an existing material to a variant.

packages/model-viewer/src/features/scene-graph/material.ts
timmmeh: done

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: Yes, removing this field, it's not being used anywhere.

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: I wanted to ensure that neither of these inputs gets missed so TS is enforcing that they are assigned on construction.

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: Ahh your right, forgot to revert this.

packages/model-viewer/src/features/scene-graph/api.ts
timmmeh: Done

packages/space-opera/src/components/materials_panel/materials_panel.ts
timmmeh: Done

packages/space-opera/src/components/materials_panel/materials_panel.ts
timmmeh: That sounds like intended behavior, editing the material will create a new material for the variant, since the new variant has no change to the material it just utilizes the existing material, this allows the variant to only override a subset of material based on what is changed. You should see in this case that the actual selected material is the same material for both variants.

packages/space-opera/src/components/materials_panel/materials_panel.ts
timmmeh: Comments added to the code, update OnHandleValue to OnOK.

packages/space-opera/src/components/shared/dialog/input_dialog.css.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ */
+
+
+import {css, CSSResult} from 'lit-element';
+export const styles: CSSResult = css`
+ :host {
+   --material-color-google-blue-500: #4285f4;
+   --material-color-google-grey-300: #dadce0;
+ }
```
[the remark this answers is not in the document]
timmmeh: Done.

packages/model-viewer/src/features/scene-graph/api.ts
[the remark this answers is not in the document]
elalish: ah, so this is more about cloning the material than creating the variant? Like it'll just create the variant if it doesn't happen to exist yet? Maybe a  name like `cloneMaterialIntoVariant`?

packages/model-viewer/src/features/scene-graph/model.ts
elalish: I believe that's already enforced; you would need to use `name?: string;` to allow `name` to be  undefined.

packages/model-viewer/src/features/scene-graph/api.ts
elalish: Can you be a  little more specific than "when it doesn't succeed"? I would not generally expect this method to fail without reading its implementation. What restrictions are on the input to ensure it succeeds?

packages/model-viewer/src/features/scene-graph/model.ts
timmmeh: letting it be undefined is what I wanted to avoid and it if that's not allowed it needs to be defined on construction. I converted it to an interface, so now it can be defined with {} braces but I don't recommend this as those to values should always inserted at the same time.

packages/model-viewer/src/features/scene-graph/api.ts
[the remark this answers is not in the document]
timmmeh: That's right, done.

packages/model-viewer/src/features/scene-graph/api.ts
timmmeh: I reworked this to remove the fail case, so now it always succeeds.

packages/model-viewer/src/features/scene-graph/api.ts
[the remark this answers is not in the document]
timmmeh: changed to `createMaterialInstanceForVariant` and removed fail case.

packages/space-opera/src/components/materials_panel/materials_panel.ts
elalish: I'm seeing even stranger behavior now. Maybe just revert the  fix for now so we can  merge this and  then we can fix the bug. Unless you see a quick resolution.

elalish (changes requested):
This looks ready  to merge except for that bug. The previous  bug was less bad, so we can just revert and merge if you like.

timmmeh:
Resolved the bug, plus a bonus bug

elalish (approved):
Thanks!

PR #2985, opened 2021-11-18
Adding fix for alpha test issue #2931

### LINKED ISSUE #2931: Materials setAlphaMode appears to apply MASK instead of BLEND
### Description

Setting alpha mode to 'BLEND' on any material with alpha < 0.5 makes the material invisible (as though it was masked)

Correct with v1.8.0:
image

Wrong with v1.9.0:
image

Model Viewer version 1.8.0 works fine.

#### Live Demo

https://glitch.com/edit/#!/fluffy-skitter-move

### Version
- model-viewer: v1.9.0

### Browser Affected

- [x] Chrome, version: 95.0.4638.54
- [x] Firefox 93.0
- [x] Safari 15.1, previous versions affected too
- [x] Brave 1.30.89 

### OS

- [x] Linux
- [x] MacOS
- [x] Windows

### AR
Not applicable

timmmeh:
Adding fix for alpha test

Fixes #2931

### DIFF
--- a/packages/model-viewer/src/features/scene-graph/material.ts
+++ b/packages/model-viewer/src/features/scene-graph/material.ts
@@ -109,7 +109,7 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     }
 
     if (gltfMaterial.alphaCutoff == null) {
-      gltfMaterial.alphaCutoff = 0.5;
+      gltfMaterial.alphaCutoff = 0.0;
     }
 
     const {
@@ -238,7 +238,11 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     const gltfMaterial = this[$sourceObject] as DefaultedMaterial;
     for (const material of this[$correlatedObjects] as
          Set<MeshStandardMaterial>) {
-      material.alphaTest = gltfMaterial.alphaCutoff;
+      if ((this[$sourceObject] as DefaultedMaterial).alphaMode === 'MASK') {
+        material.alphaTest = gltfMaterial.alphaCutoff;
+      } else {
+        material.alphaTest = 0;
+      }
       material.needsUpdate = true;
     }
   }
--- a/packages/model-viewer/src/features/scene-graph/material.ts
+++ b/packages/model-viewer/src/features/scene-graph/material.ts
@@ -240,9 +240,8 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
          Set<MeshStandardMaterial>) {
       if ((this[$sourceObject] as DefaultedMaterial).alphaMode === 'MASK') {
         material.alphaTest = gltfMaterial.alphaCutoff;
-      } else {
-        material.alphaTest = 0;
       }
+
       material.needsUpdate = true;
     }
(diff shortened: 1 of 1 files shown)


packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -238,7 +238,11 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     const gltfMaterial = this[$sourceObject] as DefaultedMaterial;
     for (const material of this[$correlatedObjects] as
          Set<MeshStandardMaterial>) {
-      material.alphaTest = gltfMaterial.alphaCutoff;
+      if ((this[$sourceObject] as DefaultedMaterial).alphaMode === 'MASK') {
+        material.alphaTest = gltfMaterial.alphaCutoff;
+      } else {
+        material.alphaTest = 0;
```
elalish, line 244: This feels  like a hack to me; I don't think `alphaTest` should even  be used  by the shader  if  the mode is really set to `MASK`. I'd  guess something is wrong with setting the blend mode. Have you looked at what GLTFLoader does when it gets these different materials as input?

timmmeh: `material.alphaTest = 0;` was because I was paranoid, looks like it's not needed.

timmmeh:
@elalish the failing fidelity test has nothing to do with this change, it's occurring on the other PR too, it looks like we are producing the correct output, I think out golden is incorrect. It looks like the golden is actually matching the `incorrect.png` file from glTF-Sample-Models-master/2.0/TextureLinearInterpolationTest/screenshot

packages/model-viewer/src/features/scene-graph/material.ts
elalish: Hmm, looking at three's code I think you were actually on the right track, except they base their logic on its presence: https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLProgram.js#L637

Looks like  you need  to set it to undefined instead of zero.

elalish (changes requested):
Yeah, I'll look into the fidelity test  thing. You've  pulled master recently right?

elalish (approved):
Thanks!

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -238,7 +238,12 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     const gltfMaterial = this[$sourceObject] as DefaultedMaterial;
     for (const material of this[$correlatedObjects] as
          Set<MeshStandardMaterial>) {
-      material.alphaTest = gltfMaterial.alphaCutoff;
+      if ((this[$sourceObject] as DefaultedMaterial).alphaMode === 'MASK') {
+        material.alphaTest = gltfMaterial.alphaCutoff;
+      } else {
+        (material.alphaTest as number | undefined) = undefined;
```
elalish, line 244: Let's make a  note to update the  three.js types repo for  this.

PR #3004, opened 2021-11-30
update three.js to r135

### LINKED ISSUE #2915: v1.9.0 performance regression

### LINKED ISSUE #2958: Extremely slow performance with specific test file
### Description

The entire browser locks up when loading the comparatively small attached file.
This seems to be a threejs issue as it also happens in threejs editor and not only in model-viewer, tracked here for awareness.

See https://github.com/mrdoob/three.js/issues/22785 for full description and test file.

Might be related to the recent slowdowns mentioned in https://github.com/google/model-viewer/issues/2915 regarding shader compilation - the attached file will probably cause a number of different shaders to compile.

elalish:
Fixes #2915 
Fixes #2958 

Three.js has fixed the underlying problem that was giving us headaches with v1.9 and  its point releases, so  ideally pulling this  in has fixed our issues as well. However, since many of the  problems were only manifesting on particular OS/devices that  I don't  have handy for testing, it would be great to get some confirmation from the community that  this is working. Please test master using `https://modelviewer.dev/node_modules/@google/model-viewer/dist/model-viewer.min.js`  but don't use it for production!


timmmeh (approved):

hybridherbst:
- [x] Can confirm that opening https://modelviewer.dev doesn't freeze Chrome anymore for 1-2s, is smooth as expected again
- [x] Can confirm that https://github.com/mrdoob/three.js/issues/22785 is MUCH faster now both with modelviewer.dev/editor and threejs.org/editor. My understanding is that it still uses way more memory than it should (the same texture gets uploaded once per material that uses it) but that's a separate issue (https://github.com/mrdoob/three.js/pull/22846)

lucadalli:
Using the Pen from #2915 to measure performance on Win 10 Chrome:
v1.8 measures 4800ms
master measures 3000ms 🎉

Not only does master revert the performance regression in v1.9, but it appears to be consistently faster than v1.8 too.

elalish:
Thank you both, that's great to hear! @hybridherbst thanks for the pointer to that three.js change, though I think the more common issue is on our side (multiple texture uploads when  the same model  is shown multiple times), which I'm working on fixing in #2935, though it currently is causing some render problems I need to work through.

PR #3007, opened 2021-12-01
a11y changes

elalish:
Reverts #345 and a couple of other a11y labels that were deemed annoying. This is a draft so we can discuss the ramifications and if anyone has better ideas.


PeterSaathoff-Harshfield:
I've left comments on 461 and 462.

In general, I think aria announcements should be used sparingly, and when they are used, they should be more concise given the time constraint of consuming all information via a single line of text being read.

elalish:
Closed in favor or #3262.

PR #3018, opened 2021-12-02
Add Thor + Serpent cameraViews sample, asset license, and attribution

hybridherbst:
New sample as promised!
I added the relevant attribution note, and the full code for both view navigation and panning.

The latter isn't ideal: it would be better if the panning code was shared with the panning sample, but I'm not entirely sure where to move it then.

Alternatively, it could be hidden from the example-snippet to reduce the clutter there (with a reference to the panning sample). Simply moving it out of `<example-snippet>` wasn't enough though (that disabled the functionality of the script).

I also added a line about "Showcase by hybridherbst / prefrontal cortex", let me know if you'd prefer a different (or no) attribution here. For a future "showcase" section that would probably be needed anyways.

image


packages/modelviewer.dev/examples/annotations/index.html
```diff
@@ -312,11 +312,295 @@ <h4 id="intro"><span class="font-medium">Show Dimensions. </span></h4>
...
+      padding: 0.5em 1em;
+      position: absolute;
+      width: max-content;
+      height: max-content;
+      transform: translate3d(-50%, -50%, 0);
+    }
+</style>
+
+<!-- nearly identical to the "Staging & Cameras/Panning" example -->
+<script type="module">
```
elalish, line 450: I do like the idea of pulling this script out of the snippet and just referencing the other example in the text. I'm surprised pulling it out didn't work. Any idea what's going wrong?

elalish (approved):
This is awesome, thanks! If you can find a way to pull the panning code out, that would be great. If not, it's not a blocker. I'm considering enabling that through a `pan` attribute anyway.

hamid123ask (commented):
h55_90

Issue #424, opened 2019-03-12
Configurable animation iteration count


cdata:
As of #421 , we support playback of animations in models. The default behavior for this feature is to play a selected animation continuously. However, we may want to support user-configurable iteration count to limit the number of times an animation is looped. The default would remain what it is as of #421: infinity.

donmccurdy:
Perhaps worth considering this as a 'mode' rather than an integer property. For example, see [`THREE.AnimationAction.loopMode`](https://threejs.org/docs/#api/en/animation/AnimationAction.loop)'s three states:

- THREE.LoopOnce - playing the clip once,
- THREE.LoopRepeat - playing the clip with the choosen number of repetitions, each time jumping from the end of the clip directly to its beginning,
- THREE.LoopPingPong - playing the clip with the choosen number of repetitions, alternately playing forward and backward. 

Ping-pong could be helpful for something like a door opening, where the animation should be played in reverse before restarting.

elalish:
We have this now: https://modelviewer.dev/docs/#entrydocs-animation-methods-play

PR #3021, opened 2021-12-03
Add some options for play method about Animation

futahei:
### Reference Issue
This pull request is related to #424.

**Description**
- add options for `play` method & docs
  - control `repetitions` & `pingpong`
- add event for animation (`loop`, `finished`)


packages/model-viewer/src/test/features/animation-spec.ts
```diff
@@ -106,6 +106,39 @@ suite('ModelViewerElementBase with AnimationMixin', () => {
...
+          await animationsPause;
+        });
+
+        test('animations pause', () => {
+          expect(animationIsPlaying(element)).to.be.false;
+        });
+
+        test('changing currentTime triggers render', () => {
+          element.currentTime = 5;
+          expect(element[$scene].shouldRender()).to.be.true;
```
elalish, line 137: Thanks for adding tests! One question: if `autoplay` is enabled and then `playOnce()` is  called (after any number of  loops), what is the result?  I think  I would expect the animation to continue to the end, then  pause. Perhaps we should add a test for that?

elalish (commented):
Thank you, I think  this  will help a number of our users!

packages/model-viewer/src/test/features/animation-spec.ts
Futahei: I think so too about `autoplay` and `playOnce()`. I've added a test.
Please confirm that you have tested it correctly. (sorry for DeepL translation)

elalish (changes requested):
It looks like you accidentally deleted our package-lock file; can you please revert that?

Futahei:
unit test was failed. I'll fix.

packages/model-viewer/src/test/features/animation-spec.ts
```diff
@@ -137,6 +170,39 @@ suite('ModelViewerElementBase with AnimationMixin', () => {
         });
       });
 
+      suite('when playOnce is invoked', () => {
```
elalish, line 173: This appears to duplicate the suite you added above? Perhaps a copy/paste error?

packages/model-viewer/src/features/animation.ts
```diff
@@ -98,6 +100,19 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
       }
     }
 
+    playOnce() {
```
elalish, line 103: Considering the other features that were requested (number of loops, ping-pong) I think I'd prefer a more extensible API. Would you mind removing `playOnce` and instead adding a options object to the `play()` method? So we can  get `playOnce()` behavior  with `play({repetitions: 1})`, where `repetitions defaults to `Infinity`. Then we can easily  add other parameters to that  object without breaking anything.

elalish (changes requested):

packages/model-viewer/src/test/features/animation-spec.ts
futahei: This is a test to see that it works correctly if you call playOnce when configured to autoplay.
But it doesn't work correctly. I will fix it.

packages/model-viewer/src/features/animation.ts
futahei: I thought of playOnce because I wanted to emphasize simplicity.
But I think your opinion is very true.

packages/model-viewer/src/features/animation.ts
```diff
@@ -85,13 +90,13 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
 
-    play() {
+    play(options: PlayAnimationOptions = {repetitions: Infinity, pingpong: false}) {
       if (this[$paused] && this.availableAnimations.length > 0) {
         this[$paused] = false;
         this[$renderer].threeRenderer.shadowMap.autoUpdate = true;
 
         if (!this[$scene].hasActiveAnimation) {
-          this[$changeAnimation]();
+          this[$changeAnimation](options);
```
elalish, line 99: I think you're going to need to change this logic. When I `pause` and then `play({repetitions: 1})`, it loops because this line does not get hit. Likewise if it's already playing. You'll want to adjust your tests to catch this too.

elalish (changes requested):

packages/model-viewer/src/features/animation.ts
futahei: I removed the part that validates `hasActiveAnimation` in order to execute `this[$changeAnimation]()`.
To be honest, I don't know why we need to validate `hasActiveAnimation`.
I also removed the execution of `[$changeAnimation]` which seems unnecessary.

packages/model-viewer/src/test/features/animation-spec.ts
```diff
@@ -103,6 +103,22 @@ suite('ModelViewerElementBase with AnimationMixin', () => {
           element.currentTime = 5;
           expect(element[$scene].shouldRender()).to.be.true;
         });
+
+        suite('when play is invoked again with options', () => {
+          setup(async () => {
+            const animationsPlay = waitForEvent(element, 'play');
+            element.play({repitations: 10, pingpong: true});
```
elalish, line 110: Good start, but this still isn't really testing that what you've  asked for has occurred. I would recommend you serve our animations page locally, select the element in dev tools and try interacting with it by calling `play` and `pause` and try your different arguments. I still have the same problems: calling `play()` with arguments when it's already playing has no effect (this test passes because the animation continues  to play and you never check if it stops). If I pause and `play({repetitions: 1})`, it does stop correctly, but can never be restarted with `play`.

elalish (changes requested):
Thanks, but I think this one needs some more thorough testing and debugging before we can merge it. Also, our package-lock file was deleted again.

packages/model-viewer/src/test/features/animation-spec.ts
futahei: Thanks for the advice. I serve animations page locally and confirmed that it works.
However, I don't know how to write the test code. Is there any way to know when the animation has completed the specified number of times? Or should I add such an event? (for example `finished`)

futahei:
Two events (`loop` & `finished`) provided by three.js can now be listened.

packages/model-viewer/src/test/features/animation-spec.ts
elalish: I would modify this test to set repetitions to 1, then set some timeouts: after 80% of the animation duration, check that `animationIsPlaying`, after 180% duration, check that it's still playing and the current time is less than last time (playing backwards). After 220%, `animationIsPlaying` should be false and time back to zero. Maybe another version with pingpong false. Choose a nice short  animation (maybe "Punch" from the Robot) so  the test completes quickly.

futahei: Added a test to call `play({repetitions: 2, pingpong: true})` with `animationName` set to "Punch".
It verifies that the state at 80%, 180%, and 220% elapsed time is the intended one.

elalish (changes requested):
Are you running these  tests locally before you push? Once they pass  on your machine they will probably pass our  CI too. That  will make iteration a lot quicker. `npm run  test`. If you're  having any  trouble, please let me know.

futahei:
I'm sorry. I actually can't run the test. I tried it in two environments, windows and ubuntu, neither of which worked. I have written a detailed report [here](https://gist.github.com/futahei/adbf789522bdb6a7adf61ea8cba5f889).

elalish:
Thanks for the report; I replied with a tip; let me know if that helps or not.

futahei:
@elalish Thank you.
I was able to run the test successfully.

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -549,8 +554,16 @@ export class ModelScene extends Scene {
...
 
+      action.setLoop(loopMode, repetitionCount);
+
       action.enabled = true;
-      action.play();
+
+      if (repetitionCount === Infinity) {
+        action.play();
+      } else {
+        action.reset();
```
elalish, line 564: Is this necessary? I notice it makes for an  inconsistent behavior: normally, when  you `pause()`  and then `play()`, the animation picks up  where it left  off. However, if you instead do `play({repetitions: 1})` the animation is suddenly restarted from the beginning.

elalish (commented):
Thanks, this is looking much  better!

packages/model-viewer/src/three-components/ModelScene.ts
futahei: This is an unnecessary process. I think this is the result of trial and error when it did not work properly. Sorry.

elalish (approved):
Thanks for  sticking with this! It looks great; your contributions will always be welcome.

LazZiya:
Hello,

Is there any sample how to use `repetitions`?

I am setting its value to 1 but the animation keeps looping!

````js
modelViewer.play({ repetitions: 1 });
````

futahei:
Hello,
You mean, `play({repetitions: 1})` not working?
Please tell me your environment that this feature not working.
This is [sample page](
https://glitch.com/edit/#!/silicon-sun-swift) that play animation once.

LazZiya:
Hi @futahei ,

I checked the sample page, it is working fine as you provided. But if I specify animation name it keeps looping! 

Reported the issue here #3144

ArtCebola:
@LazZiya where you able to reach your goal here? I'm trying the same, but the play({repetitions: 1}) seems to not work properly when I have more than one animation on my model!

I have a button that should open a lid in my model, and another one that should close.
But once I run the opening animations, it keeps in loop even if I use play({repetitions: 1}).
And once I change it to the closing animation, it keeps in loop as well.

LazZiya:
@ArtCebola yes it worked fine, I did it as recommended [here](https://github.com/google/model-viewer/issues/3144#issuecomment-1019880926).

PR #3026, opened 2021-12-04
Twitter player draft

elalish:
This  adds a page to modelviewer.dev (not linked anywhere, so not particularly public yet) to serve  as a way  to easily create a twitter embed of any 3D model using their [player card](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/player-card) which  is now available universally. The generator.html page shows a preview of the embed and allows it to be customized.


timmmeh (approved):
Looks LGTM to me

PR #3043, opened 2021-12-09
Adding uvFromPoint (clientX, clientY)

BR14Nx:
Referring to [discussion](https://github.com/google/model-viewer/discussions/3006).
Added tests, but I couldn't sucessfully run  `npm run test` yet.
Added some description in the docs.json for explanation.
Own toString() Method for Vector2.

Works pretty much like the positionAndNormalFromPoint() method, but only returns a Vector2D.


packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -274,6 +282,20 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
 
       return this[$model]![$materialFromPoint](scene.raycaster);
     }
+
+    uvFromPoint(pixelX: number, pixelY: number): Vector2D|null {
+      const scene = this[$scene];
+      const ndcCoords = scene.getNDC(pixelX, pixelY);
+      scene.raycaster.setFromCamera(ndcCoords, scene.getCamera());
```
elalish, line 289: Would you mind shifting this over next to `positionAndNormalFromPoint` in annotations.ts instead? I think following that logic will also clean this up a shade (no need to access the raycaster out here).

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -679,6 +679,32 @@ export class ModelScene extends Scene {
     return {position: hit.point, normal: hit.face.normal};
   }
 
+  /**
+   * This method returns the x and y texture-coordinate of the point
+   * on the mesh corresponding to the input pixel coordinates given relative to
+   * the model-viewer element. If the mesh is not hit, the result is null.
+   */
+  uvFromPoint(ndcPosition: Vector2, object: Object3D = this): Vector2|null {
```
elalish, line 687: Since ModelScene is an internal class, we're free to refactor it. Let's combine this method with the one above and return a struct with everything to avoid some duplication.

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
```diff
@@ -186,6 +186,25 @@ suite('scene-graph/model', () => {
...
+              element[$scene].width / 2, element[$scene].height / 2)!;
+
+          expect(uv).to.be.ok;
+        });
+
+        test(
+            'uvFromPoint returns null when intersect fails', async () => {
+              await loadModel(ASTRONAUT_GLB_PATH);
+
+              await timePasses(1000);
```
elalish, line 202: Is this necessary?

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -83,6 +83,22 @@ export const toVector3D = (v: Vector3) => {
   };
 };
 
+export interface Vector2D {
+  x: number
+  y: number
```
elalish, line 88: Should this be u and v instead?

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -83,6 +83,22 @@ export const toVector3D = (v: Vector3) => {
...
+  toString(): string
+}
+
+export const toVector2D = (v: Vector2) => {
+  return {
+    x: v.x,
+    y: v.y,
+    toString() {
+      return `u:${this.x} v:${this.y}`;
```
elalish, line 97: what is `u:` for?

packages/modelviewer.dev/data/docs.json
```diff
@@ -952,6 +952,11 @@
         "links": [
           "<a href=\"../examples/scenegraph/#pickMaterialExample\">Related examples</a>"
         ]
+      },
+      {
+        "name": "uvFromPoint(clientX, clientY)",
+        "htmlName": "uvFromPoint",
+        "description": "Returns the x and y texture-coordinate of the point on the mesh corresponding to the input pixel coordinates given relative to the screen. The texture-coordinate is returned as Vector2D, which has a method toString(). The function returns null if no object is hit."
```
elalish, line 959: I was about to say it's not clear what a Vector2D is here, but then I realized we did the same thing with Vector3D. We should probably update those to describe the struct.

elalish (changes requested):
Thanks!

packages/model-viewer/src/features/scene-graph.ts
BR14Nx: Yes I was doing this in the beginning, but then I was not sure if it really fits under "annotations" since I couldn't see a use case relating the hotspots. Though I changed that now.

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
BR14Nx: Here I was following the test from "materialFromPoint()" for coherence. I removed it, and just check that the UV-coords are within 0-1 inside the positionAndNormalFromPoint() test.

packages/model-viewer/src/model-viewer-base.ts
BR14Nx: I changed that now, but wouldn't it be maybe better for other use cases in the future if it stays x and y? So it can be reused without creating confusion?

BR14Nx:
I made the changes you request. Let me know if it suits your proposals.

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -673,11 +674,43 @@ export class ModelScene extends Scene {
       return null;
     }
 
+    if (hit.uv == null) {
+      return null;
```
elalish, line 678: I think we don't want  this; a model can lack UV coordinates, in which case `uv = null` makes sense, and it shouldn't destroy the rest of the result.

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -673,11 +674,43 @@ export class ModelScene extends Scene {
...
+    }
+
     hit.face.normal.applyNormalMatrix(
         new Matrix3().getNormalMatrix(hit.object.matrixWorld));
 
-    return {position: hit.point, normal: hit.face.normal};
+    return {position: hit.point, normal: hit.face.normal, uv: hit.uv};
+  }
+
+  /**
```
elalish, line 687: Forgot to delete?

elalish (changes requested):
Thanks, I like that better. The method name ought to be more generic, but it's  not worth a breaking change. Just a little cleanup left.

packages/model-viewer/src/model-viewer-base.ts
elalish: Maybe, but I think we'll cross that bridge when we come to it. I don't see a lot of need to return `{x, y}`, so hopefully it's a non-issue.

packages/model-viewer/src/three-components/ModelScene.ts
BR14Nx: Indeed

elalish (approved):
Thanks!

PR #3060, opened 2021-12-14
Spherical harmonics

mikkoh:
One of the things we've been thinking about here at Shopify is adding support for Light Probes + Spherical Harmonics to light a Model rather than using an environment map.

I've created this Draft PR that shows how this could function. It should be noted:

- This PR does not include tests yet (if there's appetite tests will be added)
- The example and some core code has been disabled to fully leverage Light Probes + Spherical Harmonics
- Some work will need to be done to support specular reflections in Three.js
- There would need to be some discussion around how environment maps and spherical harmonics would work together

Some context. We've been interested in using Spherical Harmonics to light our Models because they are lower weight than downloading a large HDR image. 27 floats vs large image. They could also be leveraged before the environment map has been loaded.

Some questions we've had with Model Viewer and Spherical Harmonics:

1. Can lighting with Spherical Harmonics be expressive enough for selling Products
2. Given this is new functionality to Three.js does it function fully

Below is a video of the output from this PR:

https://user-images.githubusercontent.com/496903/146046350-9be604a8-726e-4e45-867a-76d04bdffc20.mp4

The above demo uses the same coefficients as the demo here:
https://threejs.org/examples/?q=lightpro#webgl_lightprobe

From this demo we've learned that:
1. Lighting is expressive enough for Merchant Products
2. The Three.js implementation isn't fully complete

There's a missing concept in Three.js which is the idea of Reflection probes. This is likely something we could help push forward but we wanted to see if there would be appetite for this kind of functionality in `<model-viewer>`. The current implementation of `LightProbes` in Three only calculate diffuse lighting. No specular lighting.

Anyway overall wanted to see if adding this kind of functionality would make sense as an optimization

### DIFF
--- a/packages/model-viewer/src/features/environment.ts
+++ b/packages/model-viewer/src/features/environment.ts
@@ -14,7 +14,13 @@
  */
 
 import {property} from 'lit-element';
-import {Event as ThreeEvent, Texture} from 'three';
+import {
+  Event as ThreeEvent,
+  LightProbe,
+  SphericalHarmonics3,
+  Texture,
+  Vector3,
+} from 'three';
 
 import ModelViewerElementBase, {$needsRender, $onModelLoad, $progressTracker, $renderer, $scene, $shouldAttemptPreload} from '../model-viewer-base.js';
 import {PreloadEvent} from '../three-components/CachingGLTFLoader.js';
@@ -24,10 +30,13 @@ import {Constructor, deserializeUrl} from '../utilities.js';
 export const BASE_OPACITY = 0.1;
 const DEFAULT_SHADOW_INTENSITY = 0.0;
 const DEFAULT_SHADOW_SOFTNESS = 1.0;
+const DEFAULT_LIGHT_PROBE_INTENSITY = 1.0;
 const DEFAULT_EXPOSURE = 1.0;
 
 const $currentEnvironmentMap = Symbol('currentEnvironmentMap');
+const $currentLightProbe = Symbol('currentLightProbe');
 const $applyEnvironmentMap = Symbol('applyEnvironmentMap');
+const $applyLightProbe = Symbol('applyLightProbe');
 export const $updateEnvironment = Symbol('updateEnvironment');
 const $cancelEnvironmentUpdate = Symbol('cancelEnvironmentUpdate');
 const $onPreload = Symbol('onPreload');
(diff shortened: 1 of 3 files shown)


packages/model-viewer/src/features/environment.ts
```diff
@@ -168,7 +233,7 @@ export const EnvironmentMixin = <T extends Constructor<ModelViewerElementBase>>(
         return;
       }
       this[$currentEnvironmentMap] = environmentMap;
-      this[$scene].environment = this[$currentEnvironmentMap];
+      // this[$scene].environment = this[$currentEnvironmentMap];
```
mikkoh, line 236: I've temporarily disabled this to show how Three.js handles a model with no environment map to define specular reflections.

packages/model-viewer/src/features/environment.ts
```diff
@@ -113,7 +144,41 @@ export const EnvironmentMixin = <T extends Constructor<ModelViewerElementBase>>(
       }
     }
 
-    async[$updateEnvironment]() {
+    [$applyLightProbe]() {
+      const coefficients = this.sphericalHarmonics?.split(' ').map(parseFloat);
```
mikkoh, line 148: Parsing `coefficients` should be more robust including the `for` loop below. This functionality likely should live in `packages/model-viewer/src/styles/parsers.ts`. I can move this functionality there if we'd like to move ahead with this functionality.

packages/modelviewer.dev/index.html
```diff
@@ -76,7 +76,7 @@ <h3 class="grouping-title grouping-title-new quick-start">Quick Start</h3>
...
+<model-viewer alt="Neil Armstrong's Spacesuit from the Smithsonian Digitization Programs Office and National Air and Space Museum" src="shared-assets/models/NeilArmstrong.glb" ar ar-modes="webxr scene-viewer quick-look" poster="shared-assets/models/NeilArmstrong.webp" seamless-poster shadow-intensity="1" camera-controls spherical-harmonics="0.46519692492969905 0.4241956607683382 0.4576711877824825 0.30826216320384436 0.27945656813465236 0.35219365313061146 0.03157084678012829 0.058602426333643216 0.07542049516981003 0.07828354031579374 -0.1251996711220649 -0.29448362136934575 0.1021822613078809 -0.12605373673262707 -0.32908371308308876 0.03329369607846006 0.06737488425270019 0.08468130688268709 -0.04482661611401552 -0.04074374804135103 -0.06285593291723642 0.06323356089983638 -0.029427772724582465 -0.07708623359768785 0.034812559526965166 0.03877050841818562 0.0025564137010261915" spherical-harmonics-intensity="2"></model-viewer>
```
mikkoh, line 79: This is all temporary to render using spherical harmonics.

FelipeR2U:
@mikkoh You made me very curious about Spherical Harmonics, do you have any materials to share and _give me some light_?

elalish:
I'm closing this after our conversation, since it appears there are better ways to address the environment map download time than spherical harmonics. If anyone else is interested in this topic, please start a  discussion thread.

mikkoh:
@FelipeR2U this is built into Three.js https://threejs.org/docs/#api/en/lights/LightProbe

We're ignoring this implementation for `<model-viewer>` given that Three.js does not implement direct lighting/reflections when using Light Probes.

The better solution here is to use a really low res HDR map.

PR #3077, opened 2021-12-23
Click triggering Progress when reveal strategy is auto fix

mehak08g:
### Reference Issue
When the reveal strategy is AUTO, clicking on the screen was triggering progress change as every click was added as a new ongoing activity.
@elalish


elalish (approved):
Thanks!

PR #3090, opened 2021-12-30
Fix for bug #3086

### LINKED ISSUE #3086: JS error when switching material variants if not all model parts have variant definitions (intentionally)
### Description

I noticed that some models with particular material variants cause an uncaught JS error when changing `variantName`.

It appears that it is caused when there are multiple nodes in the GLB and some nodes have variant definitions and some parts don't. 

This is an intentional scenario where there might be a model with multiple variants for some parts, but other parts don't change between variants. EG. For a model of a car, the bodywork might change paint colour, but the wheels wouldn't.

See demo:

https://modelviewer-materialvariant-test.glitch.me/

View the demo. Select the 'Primitives XOX' model from the first dropdown (default). Then select the 'Green' material from the second dropdown. A JS error is thrown. Cone and Cube should turn green, Sphere should remain black. But only Cube turns green. Now select another material. Now select 'Green' again, and this time no error occurs, both Cone and Cube are updated, as expected. So it appears to only affect the first time that variant is applied and subsequent applications work fine. 

See the JS Console for the following error:

Screenshot 2021-12-29 at 18 44 16

All the other models in the demo link work, as they all define variants for every node. 

#### Live Demo

https://modelviewer-materialvariant-test.glitch.me/

### Version

- model-viewer: v1.9.1

### Browser Affected

- [x] Chrome, version: 96.0.4664.110
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [x] Safari, 15.1

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [x] MacOS
- [ ] Windows

### AR

- [ ] WebXR
- [ ] SceneViewer
- [ ] QuickLook

ghost:
Resolve an async bug that occurs when two nodes reference the same material. The first node starts loading a material and marks it as loaded, then while waiting for Three to complete the actual material setup the second node begins loading the same material and which is now marked as loaded thus it tries to fetch the cached material which is not actually available because the first node was still waiting on Three.

Fixes #3086

### DIFF
--- a/packages/model-viewer/src/features/scene-graph/model.ts
+++ b/packages/model-viewer/src/features/scene-graph/model.ts
@@ -260,12 +260,9 @@ export class Model implements ModelInterface {
    * default/initial materials if 'null' is provided.
    */
   async[$switchVariant](variantName: string|null) {
-    const promises = new Array<Promise<ThreeMaterial|ThreeMaterial[]|null>>();
     for (const primitive of this[$primitivesList]) {
-      promises.push(primitive.enableVariant(variantName));
+      await primitive.enableVariant(variantName);
     }
-
-    await Promise.all(promises);
 
     for (const material of this.materials) {
       material[$setActive](false);


elalish (approved):
Thanks!

PR #3104, opened 2022-01-07
fix animation test

elalish:
I selected a shorter animation  and combined the tests to make them run quite a bit faster. Also fixed some incorrect formatting, since our auto-formatter wasn't set  up for the contributor who last touched this code.


elalish:
Well, that's odd. I had to update the  Sponza  golden because it was changed yesterday in the Github repo, but now our Sponza render still isn't matching (not even  particularly close) in our CI. This means SwiftShader  is doing something quite different  than the GPU, which is probably a bug outside our control. I'm merging this for now because it's definitely not  the fault of this PR. I may have to disable this test.

PR #3110, opened 2022-01-10
better math for progress-tracker

elalish:
We had a report that the progress-tracker was not passing through results when only a single activity was being tracked. Turns out the math in there was pretty weird, so I think this will be better. I also removed the  symbols and IE workarounds  since this file hadn't been touched in awhile.


elalish:
@mehak08g Can you verify this fixes your issue?

PR #3128, opened 2022-01-14
v1.10.1

### LINKED ISSUE #3126: Update to v1.10.0 causing an issue with 3D model textures
### Description

I created a personal project were i used Google modelviewer to display a 3D model in a webbrowser. This was done with v1.9.1 at the time. I noticed yesterday that my textures have problems. They look glossy and at some angles shows a strange patern.

I found out that there is an update to v1.10.0 and when i downgrade to v1.9.1 the issue is gone.
I'am not a developer but can somebody help me out how or what this problem is causing with version 1.10.0. 

I created the textures in substance painter and exported to gtlf 2.0 and then to an .glb file.
Next was to compress the textures to KTX2.0 and used KHR_draco_mesh_compression

It was working fine before the update.

I can fix it with downgrading to v1.9.1 but need some advice to what causing the issue in v1.10.0. The link below is showing the model with the issue. I wait for the reactions before i fix it online to an older version.

Model with v1.10.0
Google-model-viewer-v1 10 0

Model with v1.9.1
Google-model-viewer-v1 9 1

#### Live Demo

https://webxr.michelvdlinden.nl/

### Version

- model-viewer: v1.10.0

### Browser Affected

- [x] Chrome, version: 97.0.4692.71
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [x] Safari

### OS

- [ ] Android
- [x] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### AR

- [x] WebXR
- [ ] SceneViewer
- [ ] QuickLook

elalish:
Fixes #3126

Somehow I managed to build v1.10.0 wrong for the npm release (I had locally linked in an unstable version of three.js). I've  now updated our release process doc to avoid this in the future. So, v1.10.1 is actually identical to v1.10.0 in terms of what you'd get if you pull  these release tags from github and built them. But the npm versions will be different, and v1.10.1 will be  correct.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3320,9 +3320,9 @@
       "dev": true
     },
     "node_modules/fast-glob": {
-      "version": "3.2.10",
-      "resolved": "https://registry.npmjs.org/fast-glob/-/fast-glob-3.2.10.tgz",
-      "integrity": "sha512-s9nFhFnvR63wls6/kM88kQqDhMu0AfdjqouE2l5GVQPbqLgyFjjU5ry/r2yKsJxpb9Py1EYNqieFrmMaX4v++A==",
+      "version": "3.2.11",
+      "resolved": "https://registry.npmjs.org/fast-glob/-/fast-glob-3.2.11.tgz",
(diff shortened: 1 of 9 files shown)


elalish:
Also, I suspect something is going wrong with BrowserStack. We had no trouble with  this test until they day before yesterday, at which point it became a high-probability flake. Then yesterday it was back to passing 100%, and today it's back to a high  rate of failure. No actual changes  to the  build have happened during this time and I can't reproduce it locally, so I think the fault is in our CI.

PR #3129, opened 2022-01-17
Enable to change animation speed

### LINKED ISSUE #3125: Change animation speed
### Discussed in https://github.com/google/model-viewer/discussions/2795

<sup>Originally posted by **cw789** September 16, 2021</sup>
Hello everybody.

Is there a possibility to change the animation speed during the play?</div>

@elalish suggests using requestAnimationFrame, but there is `timeScale` in Three.js ( `AnimationAction` or `AnimationMixer`) that may allow you to change the animation speed.
Are there any plans to add `currentTimeScale` as property ?

futahei:
### Reference Issue
Fixes #3125 

### What I do
- add `timeScale` property
  - You can change animation speed. If you set to a negative value (ex. `-2`), the animation play reverse.
- fix last frame position when the animation played a specified number of time
  - Always stop the animation at the last frame when it ends.

### DIFF
--- a/packages/model-viewer/src/features/animation.ts
+++ b/packages/model-viewer/src/features/animation.ts
@@ -64,7 +64,6 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
         this.dispatchEvent(new CustomEvent('loop', {detail: {count}}));
       });
       this[$scene].subscribeMixerEvent('finished', () => {
-        this.currentTime = 0;
         this[$paused] = true;
         this[$renderer].threeRenderer.shadowMap.autoUpdate = false;
         this[$changeAnimation]({repetitions: Infinity, pingpong: false});
--- a/packages/model-viewer/src/features/animation.ts
+++ b/packages/model-viewer/src/features/animation.ts
@@ -41,6 +41,7 @@ export declare interface AnimationInterface {
   readonly paused: boolean;
   readonly duration: number;
   currentTime: number;
+  currentTimeScale: number;
   pause(): void;
   play(options?: PlayAnimationOptions): void;
 }
@@ -96,6 +97,16 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
 
     set currentTime(value: number) {
       this[$scene].animationTime = value;
+      this[$renderer].threeRenderer.shadowMap.needsUpdate = true;
+      this[$needsRender]();
+    }
+
+    get currentTimeScale(): number {
+      return this[$scene].animationTimeScale;
+    }
+
+    set currentTimeScale(value: number) {
+      this[$scene].animationTimeScale = value;
       this[$renderer].threeRenderer.shadowMap.needsUpdate = true;
       this[$needsRender]();
     }
--- a/packages/model-viewer/src/features/animation.ts
+++ b/packages/model-viewer/src/features/animation.ts
@@ -41,7 +41,7 @@ export declare interface AnimationInterface {
   readonly paused: boolean;
   readonly duration: number;
   currentTime: number;
-  currentTimeScale: number;
+  timeScale: number;
   pause(): void;
   play(options?: PlayAnimationOptions): void;
 }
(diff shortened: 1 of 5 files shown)


packages/model-viewer/src/features/animation.ts
```diff
@@ -101,6 +101,16 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
     }
 
+    get currentTimeScale(): number {
+      return this[$scene].animationTimeScale;
+    }
+
+    set currentTimeScale(value: number) {
+      this[$scene].animationTimeScale = value;
+      this[$renderer].threeRenderer.shadowMap.needsUpdate = true;
+      this[$needsRender]();
```
elalish, line 111: Are these renderer updates necessary? If the animation is playing, these are being called anyway, and if not, then changing the time scale shouldn't affect the  render, right?

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -556,6 +564,7 @@ export class ModelScene extends Scene {
       action.setLoop(loopMode, repetitionCount);
 
       action.enabled = true;
+      action.clampWhenFinished = true;
```
elalish, line 567: 👍

packages/model-viewer/src/features/animation.ts
```diff
@@ -41,6 +41,7 @@ export declare interface AnimationInterface {
   readonly paused: boolean;
   readonly duration: number;
   currentTime: number;
+  currentTimeScale: number;
```
elalish, line 44: maybe just `timeScale`?

packages/model-viewer/src/features/animation.ts
```diff
@@ -101,6 +101,16 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
       this[$needsRender]();
     }
 
+    get currentTimeScale(): number {
```
elalish, line 104: Could you add this to the docs and perhaps add/modify an animation example to demonstrate this?

elalish (changes requested):
Thanks! Just a few things...

packages/model-viewer/src/features/animation.ts
futahei: I was aware of the `currentTime`, but you're right.

packages/model-viewer/src/features/animation.ts
futahei: fix it

packages/model-viewer/src/features/animation.ts
futahei: added docs & sample

elalish (approved):
Thanks, and nice example!

futahei:
@elalish Thank you for merging it.
I use model-viewer in my products, and the fix I made is very useful. Do you have any plans to release the next version?

elalish:
@futahei Thanks for contributing! Ideally I try to  release every couple of months, though it varies a lot. If you need a feature sooner, I'd recommend you build locally and serve it yourself instead of using unpkg. Of course I'd also recommend you do some extra testing, but we generally try to keep master pretty stable.

Jayemby:
I wanted to raise this as an issue but the template terrified me.

I've just (23/02/22) done a local build to get the `timeScale` property to work (brilliant work, by the way). However, the `repetitions` option set to 1 in `play()` is broken for negative `timeScale`s

to recreate:
optionally, add a listener / bind a debug function to the `finished` event.
then:
```
modelViewer.timescale = -1;
modelViewer.play({repetitions:1});
```
observe that no animation plays (and that `finished` is triggered immediately triggered). I even tried -1 repetitions but that didn't work either.
to be clear: `modelViewer.play({repetitions:x});` still works like a charm for every positive `x` other than 1 for a negative `timeScale`. In fact, a workaround for this is to use `modelViewer.play({repetitions:0.9});` as it actually gives the desired affect of `modelViewer.play({repetitions:1});`.

elalish:
@Jayemby can you tell me what about the template terrified you? A new issue is definitely a better way to report a bug. And if you remix our Glitch to demonstrate your issue we'll be able to find and fix it much faster.

Jayemby:
Sorry, was mainly because the Glitch wouldn't load so I thought I was doing something wrong; I've not used Glitch before. The tone of the template was very imperative and said I should open a Q&A if I didn't have the info to fill out the template completely. I didn't want to get yelled at by a stranger on the internet 😅

I've managed to put something together and I'm going to open an issue

elalish:
Thanks for the feedback; I may need to tone it down a shade. We just had a *lot* of people deleting the template and asking a random question.

Jayemby:
maybe not; I might just be too insecure lol

PR #3171, opened 2022-02-02
The camera cannot be moved during AR

### LINKED ISSUE #3068: camera-target shouldn't work during webxr
### Description
I used the Annotation feature to add a process to change `camera-taget` and `camera-orbit` when the button is clicked, and it works fine in 3D-view, but when I click the button in AR (webxr), the model moves.
`camera-orbit` seems to be ignored, but changing `camera-target` causes the model to move.

I think that `camera-target` shouldn't work during AR mode (webxr).

#### Live Demo
**Preview**
https://held-chestnut-gauge.glitch.me
**Code**
https://glitch.com/edit/#!/held-chestnut-gauge

### Version
- model-viewer: v1.9.1

### Browser Affected
- [x] Chrome, version: 96.0.4664.104
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [ ] Safari

### OS
- [x] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### AR
- [x] WebXR
- [ ] SceneViewer
- [ ] QuickLook

futahei:
### Reference Issue
Fixes #3068 

### What I do
- Prevented the camera from moving during AR in WebXR mode

### DIFF
--- a/packages/model-viewer/src/features/controls.ts
+++ b/packages/model-viewer/src/features/controls.ts
@@ -593,7 +593,9 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
 
     [$syncCameraTarget](style: EvaluatedStyle<Vector3Intrinsics>) {
       const [x, y, z] = style;
-      this[$scene].setTarget(x, y, z);
+      if (!this[$renderer].arRenderer.isPresenting) {
+        this[$scene].setTarget(x, y, z);
+      }
       this[$renderer].arRenderer.updateTarget();
     }
 


packages/model-viewer/src/features/controls.ts
```diff
@@ -593,7 +593,9 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
 
     [$syncCameraTarget](style: EvaluatedStyle<Vector3Intrinsics>) {
       const [x, y, z] = style;
-      this[$scene].setTarget(x, y, z);
+      if (!this[$renderer].arRenderer.isPresenting) {
+        this[$scene].setTarget(x, y, z);
```
elalish, line 597: hmm, does this work? I think the `updateTarget()` call below does kind of the same thing. Also, I'm a little less sure about this bug now; what's wrong with the target moving the object around? In this case our state could be unsynced when we return from AR to 3D mode. Is it such a problem to disable your camera-target changes when in AR mode externally to MV? I could imagine situations where that motion could be useful, and I think it's consistent with how the turntable works.

futahei: It has been confirmed to work on devices running WebXR.
WebXR and 3DView are synchronized in terms of animation state and camera coordinates. However, I don't think that the camera in AR (your device) and the camera in 3DView are the same thing.
Zooming in and moving the camera should be done by moving your own device in AR.

elalish: Yeah, that's fair, but I might argue `camera-target` is just poorly named. It acts much more like the turntable (moving the model). Still, I see your point. I'm struggling to find a use case where shifting the model around programmatically during an AR session makes any sense (especially since we don't even expose the camera's location). 

Thanks!

elalish (approved):

packages/model-viewer/src/features/controls.ts
futahei: It's certainly not strange if you think of it as a turntable.
If you think about use cases, there are something that float in the air rather than AR that is placed on the ground.
For example, we can think of 3D maps that can be viewed in 3D using AR.

Thanks for the approval.

PR #3175, opened 2022-02-02
Set texture name

slightHYP:
### Reference Discussion #3132

Added the possibility to set a custom name to the model-viewer texture object.
The model-viewer image name has been set to the file name (plus extension) if the source field is valorized. 

The CreateTextures example has been updated in order to display these new values.

image

### DIFF
--- a/packages/model-viewer/src/features/scene-graph/image.ts
+++ b/packages/model-viewer/src/features/scene-graph/image.ts
@@ -47,7 +47,9 @@ export class Image extends ThreeDOMElement implements ImageInterface {
       onUpdate: () => void, texture: ThreeTexture|null,
       gltfImage: GLTFImage|null) {
     gltfImage = gltfImage ?? {
-      name: 'adhoc_image',
+      name: (texture && texture.image && texture.image.src) ?
+          texture.image.src.split('/').pop() :
+          'adhoc_image',
       uri: (texture && texture.image && texture.image.src) ?
           texture.image.src :
           'adhoc_image' + adhocNum++
@@ -77,6 +79,7 @@ export class Image extends ThreeDOMElement implements ImageInterface {
 
   async setURI(uri: string): Promise<void> {
     (this[$sourceObject] as GLTFImage).uri = uri;
+    (this[$sourceObject] as GLTFImage).name = uri.split('/').pop();
 
     const image = await new Promise((resolve, reject) => {
       loader.load(uri, resolve, undefined, reject);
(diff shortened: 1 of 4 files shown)


packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -71,8 +73,13 @@ export class Image extends ThreeDOMElement implements ImageInterface {
     return this.uri != null ? 'external' : 'embedded';
   }
 
+  set name(name: string) {
+    (this[$sourceObject] as GLTFImage).name = "image_" + name;
```
elalish, line 77: Do we want to prepend `image_` here? Seems odd to set one name and get something different back.

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -16,614 +16,578 @@
...
   <link type="text/css" href="../../styles/docs.css" rel="stylesheet" />
-  <link rel="shortcut icon" type="image/png" href="../../assets/favicon.png"/>
-
+  <link rel="shortcut icon" type="image/png" href="../../assets/favicon.png" />
   <!-- 💁 OPTIONAL: The :focus-visible polyfill removes the focus ring for some input types -->
   <script src="../../node_modules/focus-visible/dist/focus-visible.js" defer></script>
-
   <script>
-    window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
+    window.ga = window.ga || function () { (ga.q = ga.q || []).push(arguments) }; ga.l = +new Date;
```
elalish, line 31: I'm guessing you had auto-format set for this file. Can you revert it so our formatting stays? I know it's not great, but it's the only way we've found to keep the example text from having a bunch of tabs in front of every line. Also, I can't really see your changes when the diff is this bad.

elalish (changes requested):
Thanks! Just a couple small things...

packages/model-viewer/src/features/scene-graph/image.ts
slightHYP: Yep, this is a typo from development tests. Removed on next commit

packages/modelviewer.dev/examples/scenegraph/index.html
slightHYP: I noticed this also, reverted on next commit

elalish (approved):
Thanks!

PR #3182, opened 2022-02-04
Softer shadows

elalish:
Fixes #2907 

This is a better soft-shadowing approach; slightly more expensive, but not bad and significantly better results. API (intensity & softness) is unchanged, though obviously the look is a bit different. @mrdoob, since this is based on your work.


mrdoob (approved):
LGTM! Seems like a nice code clean up too.

mrdoob:
Here's a couple of befores and afters:

| Before | After |
| --- | --- |
| Screen Shot 2022-02-05 at 10 41 48 AM | Screen Shot 2022-02-05 at 10 41 55 AM |
| Screen Shot 2022-02-05 at 10 43 38 AM | Screen Shot 2022-02-05 at 10 43 46 AM |

I think it's looking good, but I would reduce the intensity by half:

| [New] shadow = 1 | [New] shadow = 0.5 | [Old] shadow = 1 |
| --- | --- | --- |
| Screen Shot 2022-02-05 at 10 43 46 AM | Screen Shot 2022-02-05 at 10 45 41 AM | Screen Shot 2022-02-05 at 10 43 38 AM |

Otherwise it gives a bit of a "dirty" look. Plus it matches better the intensity of the old shadow.

mrdoob:
This was the original model that made me start that shadow:

| Before | After (shadow = 0.5) |
| --- | --- |
| Screen Shot 2022-02-05 at 11 02 41 AM | Screen Shot 2022-02-05 at 11 02 46 AM |

elalish:
@mrdoob, thanks I really appreciate the comparisons! Yeah, I was messing around with the default intensities for a while on Friday trying to come up with what I liked best. I intentionally made it darker by default than the old one because I made the old one lighter by default than seemed reasonable in order to hide its artifacts. Since this one is so much smoother, this level of shadow seemed more reasonable and less likely to seem invisible. Still, I'm definitely open to changing it. Also, have you tried the softness slider? It's pretty fun in the editor.

elalish:
Actually, I already interpolate them between different values for that very reason. So they're independently adjustable. Anyway, keep looking at it in differently sized windows and such. I'll see if I can get our designers to take a look too. I find there's so much accidental anchoring that it's really helpful to come back to it after some time apart to see how I really feel.

mrdoob:
Another issue of the new intensity is that there's a "dark glow" around models that touch the ground.

| shadow = 1.0 | shadow = 0.5 |
| --- | --- |
| Screen Shot 2022-02-07 at 3 36 52 PM | Screen Shot 2022-02-07 at 3 37 02 PM |
| Screen Shot 2022-02-07 at 3 37 30 PM | Screen Shot 2022-02-07 at 3 37 39 PM |
| Screen Shot 2022-02-07 at 3 38 56 PM | Screen Shot 2022-02-07 at 3 38 49 PM |

Btw, the astronaut is kind of weird because you can see the hands casting a shadow in the ground, even if they are pretty high...

elalish:
Yeah, I think the "dark glow" is good; it looks more or less like the kind of ambient occlusion shadowing I'd expect in a 90-degree corner. And yeah, the astronaut hand shadows are what we get with your approximation here, but I still think it's decent. A more nuanced example of the same thing is to watch our animated examples of the robot expressive. Note the scrubber feature in the editor so you can easily stop on a frame now.

mrdoob:
Blender Cycles render for reference.

Screen Shot 2022-02-07 at 6 06 17 PM

mrdoob:
Another reference...

Screen Shot 2022-02-07 at 6 17 08 PM

I guess the right way of doing this would be by importing the IBL that modelviewer generates in Blender...
Has anyone made a HDR file out of it yet?

elalish:
Yeah, they're here: https://github.com/google/model-viewer/tree/master/packages/shared-assets/environments

`lightroom_14b.hdr` is default, `neutral.hdr` is "neutral".

elalish:
Personally I like the second set (darker shadows) than the first, aside from the noise.

mrdoob:
I guess this is how it's supposed to look like?

Screen Shot 2022-02-08 at 12 23 45 PM

arpu:
any Plans to include this in threejs?

mrdoob:
@arpu This has been an Three.js example for years:
https://threejs.org/examples/?q=soft#webgl_shadow_contact

arpu:
oh wau! @mrdoob Thanks to pointing me to this i always looked at the renderer.shadowMap

PR #3195, opened 2022-02-09
Fix out of range mesh when assigning variant userData

### LINKED ISSUE #3179: Model is not loading when a variant is defined on a mesh primitive other than the first one in the primitives array.
### Description
Basically I have a model with a mesh that has 3 primitives defined in the array. Each of these primitives have a different material assigned. What I want to do is to apply a different variant mapping to each primitive. Since, I switched to model-viewer 1.10.x,  I can only put it on the first primitive in the array, otherwise it will not load the model(blank page with no errors in the console). I have tried to put variant mappings in all 3 primitives, or only on any of the other 2 primitives which yielded the same result.
I have not used model-viewer 1.9.x in production, so I am not certain if this is related to latest version or the one previous to that, but I can confirm that my model worked in 1.8. 
Any idea on what could cause this changed behavior? I would be willing to help if you guys think this is a model-viewer specific issue.

**Here are some screenshots of the GLTF file to kinda explain what I mean**:
Renders: 
Screen Shot 2022-02-03 at 10 22 18 AM

Does not render:
Screen Shot 2022-02-03 at 10 24 56 AM

Screen Shot 2022-02-03 at 10 23 56 AM

### Version
- model-viewer: v.1.8.0 - Works
- model-viewer: v.1.9.x - Not Tested 
- model-viewer: v1.10.1 -> Fails

### Browser Affected
- [X ] Chrome, version:  97.0.4692.99
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [ ] Safari

### OS
- [ ] Android
- [ ] iOS
- [ ] Linux
- [ X] MacOS
- [ ] Windows

alexdaube:
### Description

There was a silent error occurring in the `VariantMaterialLoaderPlugin.afterRoot` method.  Basically, when traversing all elements of the loading scene and trying to assign variant materials to each mesh, we sometimes get out of range indexes for meshes with multiple primitives all having variants defined since 'primitives.length' and 'meshes.length' are not always corresponding in the code.

I thought it would be a good idea to add the catch callback in the loading promise to get some console feedback when the GLTF loading fails. 

### Reference Issue

Fixes #3179


alexdaube:
I will provide a testing file soon. Would it be a good idea to add a test in the fidelity project for this?

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -209,8 +209,10 @@ export class CachingGLTFLoader<T extends GLTFInstanceConstructor =
                                     .then((preparedGLTF) => {
                                       progressCallback(0.9);
                                       return new GLTFInstance(preparedGLTF);
-                                    });
-
+                                    }).catch((reason => {
+                                      console.error(reason);
```
elalish, line 213: 👍

packages/model-viewer/src/three-components/gltf-instance/VariantMaterialLoaderPlugin.ts
```diff
@@ -143,7 +143,8 @@ export default class GLTFMaterialsVariantsExtension implements
           if (!extensionsDef || !extensionsDef[this.name]) {
             continue;
           }
-          meshes[i].userData.variantMaterials =
+          const meshObjectIndex = i < meshes.length  ? i : 0;
```
elalish, line 146: This feels like a hack. I think you have discovered a real issue, but I'd like to fix it properly. Do you understand yet what circumstances lead to getting an out-of-bounds value?

packages/space-opera/package.json
```diff
@@ -15,7 +15,7 @@
     "watch:test": "karma start",
     "watch:tsc": "tsc -w",
     "watch:rollup": "rollup -c -w",
-    "serve": "./node_modules/.bin/http-server -a 127.0.0.1 -o /demo/ -c-1",
+    "serve": "./node_modules/.bin/http-server -a 127.0.0.1 -o /editor/ -c-1",
```
elalish, line 18: What does this change do?

.gitignore
```diff
@@ -3,4 +3,5 @@
 .DS_Store
 lib
 node_modules
-dist
\ No newline at end of file
+dist
+.idea
```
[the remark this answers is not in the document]
alexdaube: It's not necessary, but it's a bit annoying when you're working with Webstorm or another Jetbrain's product. It's a lot like the `..vscode` or `.DS_Store` folders...

packages/space-opera/package.json
elalish: Sure, that sounds good. I don't generally run `dev`, which is probably why I haven't noticed.

packages/model-viewer/src/three-components/gltf-instance/VariantMaterialLoaderPlugin.ts
alexdaube: Agreed. Sorry about that. 

Here is quick log for the execution inside the traverse callback for my test model.

Screen Shot 2022-02-10 at 10 44 48 PM

So when we traverse the scene recursively, we sometimes execute the code above for `Group` objects with `Mesh` objects(each representing a primitive of the original Mesh) as children and, then, we will execute for each of these `Mesh` object individually.  So that means that we actually assign `userData.variantMaterials` to each `Mesh` twice because of the nature of the traverse method. So using the index of the primitives array in the loop won't always work for accessing the corresponding mesh, since the `meshes` and `primitivesDef` arrays are only really corresponding when the object is a `Group` and we assign meshes to `object.children`. 

I think the loop could probably be completely removed here. We have all the necessary information in the `association` property of the `Mesh` objects to assign userData.variantMaterials without it. Also, in my case it seems that the last primitive's extension will be assigned to each mesh, which I didn't noticed before since the model and its variants is visually being rendered properly without any `userData.variantMaterials` whatsoever.

Does that make sense?

elalish: That sounds reasonable to me; can you go ahead and make those changes in this PR (and remove the comment if you're not sure it applies anymore)? I'd like to test it and see it in context.

elalish: We already have some variants regressions tests and if you add your new one as well, I'll feel fairly confident if everything passes.

alexdaube: You're talking about the regressions tests in `render-fidelity-tools` right?

elalish: No, I'm talking about these tests (and others) https://github.com/google/model-viewer/blob/master/packages/model-viewer/src/test/features/scene-graph-spec.ts#L88

elalish:
Sorry I've been ignoring this; I wasn't sure if you'd finished your changes. If you click re-request review, it'll ping me. Anyway, considering it's been almost a month, I'm guessing this is ready?

alexdaube:
> Sorry I've been ignoring this; I wasn't sure if you'd finished your changes. If you click re-request review, it'll ping me. Anyway, considering it's been almost a month, I'm guessing this is ready?

This completely slipped my mind. Thanks for reminding me. I just pushed some tests with the file I used for testing.
 Let me know if you require anything else. I will try to make time.

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
```diff
@@ -170,6 +171,79 @@ suite('scene-graph/model/mesh-primitives', () => {
...
+      const primitives = findPrimitivesWithVariant(model, 'Normal')!;
+      let materials = new Array<MeshStandardMaterial>();
+      for (const primitive of primitives) {
+        materials.push(
+            await primitive.enableVariant('Inverse') as
+                MeshStandardMaterial);
+      }
+      expect(materials.find((material: MeshStandardMaterial) => {
+        return material.name === 'STEEL BLACK';
+      })).to.not.be.ok;
```
elalish, line 220: This looks like the same test as above?

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
```diff
@@ -170,6 +171,79 @@ suite('scene-graph/model/mesh-primitives', () => {
...
+      const materials = new Array<MeshStandardMaterial>();
+      for (const primitive of primitives) {
+        materials.push(
+            await primitive.enableVariant('Inverse') as
+                MeshStandardMaterial);
+      }
+      expect(materials).to.not.be.empty;
+      expect(materials.find((material: MeshStandardMaterial) => {
+        return material.name === 'STEEL BLACK';
+      })).to.not.be.ok;
```
elalish, line 207: Can we check that it does equal something instead of doesn't, to make it more specific?

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
```diff
@@ -170,6 +171,79 @@ suite('scene-graph/model/mesh-primitives', () => {
...
+      const primitive = findPrimitivesWithVariant(model, 'Normal')![0];
+
+      // Gets current material.
+      const initialMaterial = await primitive.enableVariant('Normal');
+      // Switches to variant.
+      const variantMaterial = await primitive.enableVariant('Inverse');
+      expect(initialMaterial).to.not.equal(variantMaterial)
+      // Switches to initial material.
+      const resetMaterial = await primitive.enableVariant(null);
+      expect(resetMaterial).to.equal(initialMaterial);
```
elalish, line 243: 👍

packages/model-viewer/src/three-components/gltf-instance/VariantMaterialLoaderPlugin.ts
```diff
@@ -114,38 +114,28 @@ export default class GLTFMaterialsVariantsExtension implements
...
-          meshes[i].userData.variantMaterials =
-              mappingsArrayToTable(extensionsDef[this.name]);
+        const primitiveDef = primitivesDef[association.primitives];
+        const extensionsDef = primitiveDef.extensions;
+
+        if (!extensionsDef || !extensionsDef[this.name]) {
+          return;
         }
+
+        mesh.userData.variantMaterials = mappingsArrayToTable(extensionsDef[this.name]);
```
elalish, line 138: This is much simpler, which I like! I think this works now because of upstream changes in three.js; I believe `association.primitives` didn't exist at the time this was written. Thanks!

packages/model-viewer/src/test/features/scene-graph/nodes/primitive-node-spec.ts
alexdaube: You're right. I will only keep the switch and switch back. It covers both cases...

elalish (approved):
Thanks!

PR #3212, opened 2022-02-21
Remove unnecessary async/await in TextureUtils

sukolsak:
Async functions without an `await` expression are executed synchronously. So there is no need to make them `async`.

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function ("The body of an async function can be thought of as being split by zero or more await expressions. Top-level code, up to and including the first await expression (if there is one), is run synchronously. In this way, an async function without an await expression will run synchronously.")


elalish (approved):
Indeed, good catch. Thanks for the code cleanup!

PR #3222, opened 2022-02-23
Create texture from canvas

lucadalli:
A naïve and incomplete attempt at #3211

I've never explored this part of the codebase so I'd like to hear your thoughts, particularly about using the canvas `dataURL` as the `gltfImage uri`, before I proceed any further.

Todo:
- [x] Add a sister method alongside [`Image.setURI`](https://github.com/google/model-viewer/blob/95006dc49ba61be783848d07700308d5c13cba15/packages/model-viewer/src/features/scene-graph/image.ts#L80) which accepts `canvas` input
- [ ] Add example to docs

### DIFF
--- a/packages/model-viewer/src/features/scene-graph.ts
+++ b/packages/model-viewer/src/features/scene-graph.ts
@@ -23,7 +23,7 @@ import {NumberNode, parseExpressions} from '../styles/parsers.js';
 import {GLTF} from '../three-components/gltf-instance/gltf-defaulted.js';
 import {ModelViewerGLTFInstance} from '../three-components/gltf-instance/ModelViewerGLTFInstance.js';
 import GLTFExporterMaterialsVariantsExtension from '../three-components/gltf-instance/VariantMaterialExporterPlugin';
-import {Constructor} from '../utilities.js';
+import {Constructor, defineLazyMemoizedProperty} from '../utilities.js';
 
 import {Image, PBRMetallicRoughness, Sampler, TextureInfo} from './scene-graph/api.js';
 import {Material} from './scene-graph/material.js';
@@ -153,22 +153,10 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
 
       if (src instanceof HTMLCanvasElement) {
         const canvas = src;
+        const gltfImage = {name: 'canvas', uri: ''};
+        defineLazyMemoizedProperty(gltfImage, 'uri', () => canvas.toDataURL());
         return new ModelViewerTexture(
-            this[$getOnUpdateMethod](), texture, null, null, {
-              name: 'canvas',
-              // lazily set the canvas dataURL as the "uri"
-              get uri() {
-                // invoke own setter with dataURL
-                this.uri = canvas.toDataURL();
-                return this.uri;
-              },
-              set uri(uri) {
-                // delete getter and setter and set "uri" as a regular
-                // property
-                delete this.uri;
-                this.uri = uri;
-              }
-            });
+            this[$getOnUpdateMethod](), texture, null, null, gltfImage);
       }
       return new ModelViewerTexture(this[$getOnUpdateMethod](), texture);
     }
--- a/packages/model-viewer/src/features/scene-graph.ts
+++ b/packages/model-viewer/src/features/scene-graph.ts
@@ -52,8 +52,8 @@ export interface SceneGraphInterface {
   scale: string;
   readonly originalGltfJson: GLTF|null;
   exportScene(options?: SceneExportOptions): Promise<Blob>;
-  createTexture(canvas: HTMLCanvasElement): Promise<ModelViewerTexture|null>;
-  createTexture(uri: string, type?: string): Promise<ModelViewerTexture|null>;
+  createTexture(src: string|HTMLCanvasElement, type?: string):
+      Promise<ModelViewerTexture|null>;
   /**
    * Intersects a ray with the scene and returns a list of materials who's
    * objects were intersected.
@@ -120,12 +120,8 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
       };
     }
 
-    async createTexture(canvas: HTMLCanvasElement):
-        Promise<ModelViewerTexture|null>;
-    async createTexture(uri: string, type?: string):
-        Promise<ModelViewerTexture|null>;
     async createTexture(
-        src: HTMLCanvasElement|string,
+        src: string|HTMLCanvasElement,
         type: string = 'image/png'): Promise<ModelViewerTexture|null> {
       const currentGLTF = this[$currentGLTF];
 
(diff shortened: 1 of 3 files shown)


packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -36,6 +36,8 @@ export const $applyTexture = Symbol('applyTexture');
  * Image facade implementation for Three.js textures
  */
 export class Image extends ThreeDOMElement implements ImageInterface {
+  [$sourceObject]: GLTFImage;
```
elalish, line 39: I hadn't realized that TS had this way to narrow types; that's convenient!

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -119,11 +120,20 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
       };
     }
 
-    async createTexture(uri: string, type: string = 'image/png'):
-        Promise<ModelViewerTexture|null> {
+    async createTexture(canvas: HTMLCanvasElement):
+        Promise<ModelViewerTexture|null>;
+    async createTexture(uri: string, type?: string):
+        Promise<ModelViewerTexture|null>;
```
elalish, line 126: I think these are unnecessary; let's just stick with broadening the accepted input types.

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -141,6 +151,25 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+              // lazily set the canvas dataURL as the "uri"
+              get uri() {
+                // invoke own setter with dataURL
+                this.uri = canvas.toDataURL();
+                return this.uri;
+              },
+              set uri(uri) {
+                // delete getter and setter and set "uri" as a regular
+                // property
+                delete this.uri;
```
elalish, line 168: Huh, that's clever. I don't have any problem with this so long as it works. What's your plan for testing? Particularly, I'd like to ensure that `exportGLB` behaves sanely.

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -119,11 +120,20 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+        Promise<ModelViewerTexture|null>;
+    async createTexture(
+        src: HTMLCanvasElement|string,
+        type: string = 'image/png'): Promise<ModelViewerTexture|null> {
       const currentGLTF = this[$currentGLTF];
-      const texture: Texture = await new Promise<Texture>(
-          (resolve) => this[$textureLoader].load(uri, resolve));
+
+      const texture: Texture = src instanceof HTMLCanvasElement ?
+          new CanvasTexture(src) :
```
elalish, line 133: Will this behave as a video texture if the source canvas is updated every frame?

packages/model-viewer/src/features/scene-graph.ts
```diff
@@ -141,6 +151,25 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
...
 
+      if (src instanceof HTMLCanvasElement) {
+        const canvas = src;
+        return new ModelViewerTexture(
+            this[$getOnUpdateMethod](), texture, null, null, {
+              name: 'canvas',
+              // lazily set the canvas dataURL as the "uri"
+              get uri() {
+                // invoke own setter with dataURL
+                this.uri = canvas.toDataURL();
```
elalish, line 162: This is cool, though I'd like to understand better what its ramifications are.

packages/model-viewer/src/features/scene-graph/image.ts
lucadalli: I think there's a few classes than can be cleaned up by narrowing types in this manner

packages/model-viewer/src/features/scene-graph.ts
lucadalli: I've been testing with the "Export GLB" example in the docs and then loading the exported .glb into the editor to verify that the texture created from canvas was exported correctly and that the model can be validated without issues.
Everything works as expected but the getter/setter aren't invoked at all. In fact, the exported glb is the same even if the gltfImage given to the `ModelViewerTexture` constructor is `null`.
This makes sense to me considering that that the glb export is generated using THREE's `GLTFExporter` which I suspect does not interface with these gltfImages at all.
Since this is the case, what should my litmus test be to make sure that this lazy memoized getter pattern is not breaking anything?

```
  async function exportGLB(){
    let modelViewer = document.getElementById("static-model");

    const canvas = document.createElement('canvas')
    canvas.width = 1000
    canvas.height = 1000
    const ctx = canvas.getContext('2d')
    ctx.fillStyle = 'red'
    ctx.fillRect(0,0,canvas.width,canvas.height)
    const texture = await modelViewer.createTexture(canvas);
    const [material] = modelViewer.model.materials;
    material.pbrMetallicRoughness.baseColorTexture.setTexture(texture);
    
    const glTF = await modelViewer.exportScene();
    var file = new File([glTF], "export.glb");
    var link = document.createElement("a");
    link.download =file.name;
    link.href = URL.createObjectURL(file);
    link.click();
  }
```

packages/model-viewer/src/features/scene-graph.ts
lucadalli: No, unless `Image`'s `update` method is called on every frame.

packages/model-viewer/src/features/scene-graph/image.ts
lucadalli: I think there's a few classes that can be cleaned up by narrowing types in this manner

lucadalli: I think there's a few classes that can be cleaned up by narrowing types in this manner. A good opportunity for anyone that wants to get their feet wet with open source

lucadalli: I think there's a few classes that can be cleaned up by narrowing types in this manner. A good opportunity for anyone who wants to get their feet wet with open source

packages/model-viewer/src/features/scene-graph.ts
elalish: Ah, fair enough; in that case I'm not particularly worried about it. But yes, I like the test outlined above; maybe just draw an image to the canvas and then verify that the output glb has a reasonable file size. That will give me confidence the data has flowed through.

elalish:
Is this still a draft? Would you like to re-request review? I don't get pinged for commits, so I miss things sometimes...

lucadalli:
@elalish Still a draft, sorry. I'd like to add tests and an example to the docs and before requesting review.
Time I would've liked to spend on this PR went into creating a custom build based on v1.8 to work around issue #3193 :(

lucadalli:
@elalish Still a draft, sorry. I'd like to add tests and an example to the docs before requesting review.
Time I would've liked to spend on this PR went into creating a custom build based on v1.8 to work around issue #3193 :(

elalish:
Did you ever find the source of #3193? I'd like to get that fixed upstream. In any case, no rush and thanks for the contributions!

elalish:
I believe this has been addressed by #3886.

PR #3225, opened 2022-02-24
Workaround: parse animation index instead of falling back to 0 when animation name can't be found

### LINKED ISSUE #3224: Selecting animation by name fails with animations that have identical names
### Description

The glTF spec doesn't require animation names to be unique. Especially when combining multiple models into one file, they might share animation clip names (or not have an animation clip name at all). Currently, model-viewer only matches by name – so for these files animations can't be properly be chosen from.

In such cases, `availableAnimations` actually returns the correct array of animation names, they just happen to be all the same (or have duplicate entries, which is correct).

### Workarounds

Currently, this enforces files to have unique animation names. A good amount of other viewers have the same issue with that (but not all, and the spec doesn't require unique names).

A workaround I can imagine being useful here is that in addition to accepting a string as animation name. I started a PR for that.

#### Live Demo
This file has multiple animation clips that target different objects but have the same name (valid for glTF):
[MultipleInstancesSameClip.zip](https://github.com/google/model-viewer/files/8133643/MultipleInstancesSameClip.zip)

Calling `availableAnimations` on it will correctly return ["UpAndDown", "UpAndDown", "UpAndDown", "UpAndDown"], but the respective clips can't be played since it's only matched by name.

For comparison, here's the same file but with the workaround of having unique clip names:
[MultipleInstancesSameClip-uniqueNames.zip](https://github.com/google/model-viewer/files/8133659/MultipleInstancesSameClip-uniqueNames.zip)

### Notes

A number of other glTF viewers also struggle with this; I intend to report bugs for these as well 🕵️‍♀️.

hybridherbst:
Adds a workaround for a case we're seeing with industrial models and also larger, assembled scenes - the same animation clip name might be used multiple times for either the same animation or a different one. Currently model-viewer matches clips by name.

Adding a new property to set by index would be another option, but I thought this one was simple and non-obtrusive enough and covers these usecases. Matching by name still happens first, so even when clips would be named "5", "5", "348" this doesn't break existing functionality.

I'm currently having trouble setting up a working dev environment on WSL again, so unfortunately I wasn't able to fully test this :(

### Reference Issue
Fixes #3224


packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -542,6 +542,13 @@ export class ModelScene extends Scene {
 
     if (name != null) {
       animationClip = this.animationsByName.get(name);
+      
+      if (animationClip == null) {
+        const parsedAnimationIndex = parseInt(name);
+
+        if (!isNaN(parsedAnimationIndex) && parsedAnimationIndex >= 0 && parsedAnimationIndex < animations.length) {
+          animationClip = animations[parsedAnimationIndex];
+        }
```
elalish, line 551: I think you missed a `}`

hybridherbst: 🤦‍♂️ the wonders of incomplete dev environments and partial discards. Thanks!

Got a related question regarding dev env: it seems that `npm run dev` in packages/model-viewer now always runs karma; I think it used to serve modelviewer.dev for testing instead. What's the right steps nowadays to work on model-viewer but test in modelviewer.dev?

elalish: So, someone recently changed the `dev` script, but I didn't pay a lot of attention because I never use it personally. I just have one terminal open where I have `npm run serve` going non-stop, and then another terminal where I do `npm run build:dev` and `npm test` alternately in the model-viewer package directory. If you'd like the script to behave differently (or add others) feel free to toss up a PR. I'm not sure what other devs like in terms of automation.

elalish (approved):
Any chance you'd like to add a test for this?

hybridherbst:
I started adding some tests but am unsure about a couple of things:

- how do I figure out which animation _index_ is currently playing? `animationIsPlaying` would need to support passing in an int or string ("1") to properly test it. Would this require making more things public in ModelScene?
- other tests will incorrectly succeed - e.g. when you have ["Clip", "Clip"] and you tell it to play availableAnimations[1] it the test succeeds but actually plays the wrong animation

I still can't run tests locally unfortunately :( I pushed my attemps nonetheless, feel free to revert again if this doesn't make sense. 

Could also _just_ use the first test ("1" plays availableAnimations[1]) which is kind of cheating since it doesn't actually test for what the workaround is supposed to fix, just for the new code working, let me know what you prefer.

packages/model-viewer/src/test/features/animation-spec.ts
```diff
@@ -181,6 +182,32 @@ suite('ModelViewerElementBase with AnimationMixin', () => {
...
+      suite('a model with duplicate animation names', () => {
+        setup(async () => {
+          element.src = ANIMATED_GLB_DUPLICATE_ANIMATION_NAMES_PATH;
+          await waitForEvent(element, 'load');
+          element.animationName = "1";
+          await timePasses();
+        });
+
+        test('plays the specified animation', () => {
+          expect(animationIsPlaying(element, element.availableAnimations[1]))
```
hybridherbst, line 206: This will always succeed for a model with duplicate clip names. Correct would be to check that "1" plays animations[1].

elalish: In fact all the internals *are* available to our tests, since private things are just symbols; we often import a few symbols to our tests for this purpose. 

I'm impressed you're managing this without being able to run the tests locally; any idea what the trouble is? I'd like to ensure our project is runnable from WSL2. Granted our Windows 10 setup instructions are a bit light, but if something is missing we should definitely add it.

hybridherbst: I started a readme with some of the things that I tried, and some of them that worked - think I have a working test setup now and will attempt the above again.

Is there a good place to PR those notes to? Would be great to link to it from the main Readme WSL section.
(I could also just send you a link and you add it to the wiki?)

elalish: You're welcome to add them directly to the WSL section of the README if it's not crazy long. But yeah, I can add it to the wiki otherwise. I just feel like the wiki overall is not terrible up to date...

packages/model-viewer/src/test/features/animation-spec.ts
```diff
@@ -227,9 +242,21 @@ suite('ModelViewerElementBase with AnimationMixin', () => {
...
+        suite('when playing an animation by name', () => {
+          setup(async () => {
+            element.animationName = element.availableAnimations[1];
+            await timePasses();
+          });
+  
+          test('fails to play the specified animation', () => {
+            expect(animationWithIndexIsPlaying(element, 1))
+                .to.be.false;
+          });
```
hybridherbst, line 258: I'm not sure if this type of test - testing that a known failure case is still failing - is in line with testing for model-viewer, feel free to remove again.

hybridherbst:
OK, tests are in, was able to run them locally on WSL, and will make a separate PR for some Readme adjustments regarding that.

elalish (dismissed):

packages/model-viewer/src/test/features/animation-spec.ts
hybridherbst: Will you make that change or do you want me to? Sorry, wasn't quite clear

elalish: Oh, sorry that wasn't clear. If you could, that'd be great 👍

hybridherbst: OK, pushed and clarified the wording a bit

elalish (approved):
Thanks!

PR #3230, opened 2022-03-01
WSL troubleshooting

hybridherbst:
### Reference Issue

While working on #3224, I had again issues with setting up model-viewer development on WSL. This time, I took some notes, and added them to the Readme. There's a few lines that everyone using WSL should read (e.g. the magic trick to get started is "clone via https from inside WSL, not from inside Windows"), and a number of concrete error messages and issues I ran into that I added behind a foldout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,82 @@ Command                        | Description
 
 You should now be ready to work on any of the `<model-viewer>` projects!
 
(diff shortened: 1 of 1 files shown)


elalish (approved):
Thank you! I think this will definitely help with getting more contributors.

PR #3239, opened 2022-03-03
Add with-credentials attribute to configure three.js loader

ymoriaud:
### Reference Discussion
Can I use the withCredential option to load GLTF files ? #3236 

### Description
Allow the model-viewer tool to access non public .GLTF files. When files are served by a static server that can only be accessible with a specific HTTP cookie to authenticate client requests, model-viewer was failing to load the GLT file.

I integrated it as new attribute `with-credentials`, based on the current documentation and behavior of model-viewer.

This attribute name has been choosed, based on [XMLHttpRequest property](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) and [three.js documentation](https://threejs.org/docs/?q=setMeshoptDecoder#api/en/loaders/Loader.withCredentials)

### Context
This is my first PR, feel free to let me know if I did something in the wrong way or if my PR is not complete. 🙏 

### Steps

- [x] Add and integrate `with-credential` attribute
- [x] Write documentation

### DIFF
--- a/packages/model-viewer/src/features/loading.ts
+++ b/packages/model-viewer/src/features/loading.ts
@@ -83,7 +83,6 @@ export declare interface LoadingInterface {
   loading: LoadingAttributeValue;
   generateSchema: boolean;
   seamlessPoster: boolean;
-  withCredentials: boolean;
   readonly loaded: boolean;
   readonly modelIsVisible: boolean;
   dismissPoster(): void;
@@ -237,13 +236,6 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
      */
     @property({type: Boolean, attribute: 'seamless-poster'})
     seamlessPoster = false;
-
-    /**
-     * This attribute is a boolean value that indicates whether or not 3D model fetch
-     * should be made using credentials such as cookies, authorization headers or TLS client certificates.
-     */
-    @property({type: Boolean, attribute: 'with-credentials'})
-    withCredentials = false;
 
     /**
      * Dismisses the poster, causing the model to load and render if
@@ -423,10 +415,6 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
         } else {
           this[$posterContainerElement].classList.remove('quick');
         }
-      }
-
-      if (changedProperties.has('withCrendentials') || this.withCredentials === true) {
-        CachingGLTFLoader.setWithCredentials(this.withCredentials)
       }
     }
 
--- a/packages/model-viewer/src/features/loading.ts
+++ b/packages/model-viewer/src/features/loading.ts
@@ -83,6 +83,7 @@ export declare interface LoadingInterface {
   loading: LoadingAttributeValue;
   generateSchema: boolean;
   seamlessPoster: boolean;
+  withCredentials: boolean;
   readonly loaded: boolean;
   readonly modelIsVisible: boolean;
   dismissPoster(): void;
@@ -238,6 +239,13 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
     seamlessPoster = false;
 
     /**
+     * This attribute is a boolean value that indicates whether or not 3D model fetch
+     * should be made using credentials such as cookies, authorization headers or TLS client certificates.
+     */
+    @property({type: Boolean, attribute: 'with-credentials'})
+    withCredentials = false;
+
+    /**
      * Dismisses the poster, causing the model to load and render if
      * necessary. This is currently effectively the same as interacting with
      * the poster via user input.
@@ -360,6 +368,10 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
     connectedCallback() {
       super.connectedCallback();
 
+      if (this.withCredentials) {
+        CachingGLTFLoader.setWithCredentials(this.withCredentials)
+      }
+
       // Fired when a user first clicks the model element. Used to
       // change the visibility of a poster image, or start loading
       // a model.
@@ -414,6 +426,12 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
           this[$posterContainerElement].classList.add('quick');
         } else {
           this[$posterContainerElement].classList.remove('quick');
+        }
+      }
+
+      if (changedProperties.has('withCrendentials')) {
+        if (this.withCredentials) {
+          CachingGLTFLoader.setWithCredentials(this.withCredentials)
         }
       }
     }
--- a/packages/model-viewer/src/features/loading.ts
+++ b/packages/model-viewer/src/features/loading.ts
@@ -368,10 +368,6 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
     connectedCallback() {
       super.connectedCallback();
 
-      if (this.withCredentials) {
-        CachingGLTFLoader.setWithCredentials(this.withCredentials)
-      }
-
       // Fired when a user first clicks the model element. Used to
       // change the visibility of a poster image, or start loading
       // a model.
@@ -429,10 +425,8 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
         }
       }
 
-      if (changedProperties.has('withCrendentials')) {
-        if (this.withCredentials) {
-          CachingGLTFLoader.setWithCredentials(this.withCredentials)
-        }
+      if (changedProperties.has('withCrendentials') || this.withCredentials === true) {
+        CachingGLTFLoader.setWithCredentials(this.withCredentials)
       }
     }
 
--- a/packages/model-viewer/src/three-components/CachingGLTFLoader.ts
+++ b/packages/model-viewer/src/three-components/CachingGLTFLoader.ts
@@ -73,6 +73,8 @@ const ktx2Loader = new KTX2Loader();
 let meshoptDecoderLocation: string;
 let meshoptDecoder: Promise<typeof MeshoptDecoder>|undefined;
 
+let withCredentials: boolean;
+
 interface MeshoptDecoder {
   ready: Promise<void>;
   supported: boolean;
@@ -122,6 +124,14 @@ export class CachingGLTFLoader<T extends GLTFInstanceConstructor =
 
   static initializeKTX2Loader(renderer: WebGLRenderer) {
     ktx2Loader.detectSupport(renderer);
+  }
+
+  static setWithCredentials(state: boolean) {
+    withCredentials = state
+  }
+
+  static getWithCredentials() {
+    return withCredentials
   }
 
   static[$evictionPolicy]: CacheEvictionPolicy =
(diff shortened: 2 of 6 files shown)


packages/model-viewer/src/features/loading.ts
```diff
@@ -416,6 +428,12 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
           this[$posterContainerElement].classList.remove('quick');
         }
       }
+
+      if (changedProperties.has('withCrendentials')) {
+        if (this.withCredentials) {
```
elalish, line 433: I don't think you want this if statement; it would only allow you to set the value to `true`, but not back to `false`.

packages/modelviewer.dev/data/docs.json
```diff
@@ -67,6 +67,18 @@
           "default": "auto",
           "options": "auto, interaction, manual"
         }
+      },
+      {
+        "name": "with-credentials",
+        "htmlName": "withCredentials",
+        "description": "This attribute is a boolean value that indicates whether or not 3D model fetch should be made using credentials such as cookies, authorization headers or TLS client certificates.",
```
elalish, line 74: Can you link to the MDN article? Also, if you can give a little lay-person detail on what situation you'd need this for, that would be great. For instance, we could simply make it always true; presumably there's a reason that's not a good idea?

packages/modelviewer.dev/data/docs.json
```diff
@@ -67,6 +67,18 @@
...
+        "links": [
+          "<a href=\"../examples/loading/#withCredentials\">Related examples</a>"
+        ],
+        "default": {
+          "default": "false",
+          "options": "any legal boolean"
```
elalish, line 80: now about "true, false"?

elalish (changes requested):
Thanks for taking the initiative on this! A few comments...

packages/model-viewer/src/features/loading.ts
ymoriaud: That make sense ! I'm removing the if statement it. 👍

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -189,6 +199,9 @@ export class CachingGLTFLoader<T extends GLTFInstanceConstructor =
   async preload(
       url: string, element: ModelViewerElementBase,
       progressCallback: ProgressCallback = () => {}) {
+    if (withCredentials == true) {
```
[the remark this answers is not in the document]
ymoriaud: Thanks for the heads-up, I'm updating the code.

packages/modelviewer.dev/data/docs.json
ymoriaud: Sound much better ! I'm changing it.

packages/modelviewer.dev/examples/loading/index.html
```diff
@@ -433,6 +433,23 @@ <h4></h4>
...
+      <div id="withCredentials" class="demo"></div>
+      <div class="content">
+        <div class="wrapper">
+          <div class="heading">
+            <h2 class="demo-title">Enable with-credentials</h2>
+            <h4></h4>
+          </div>
+          <example-snippet stamp-to="withCredentials" highlight-as="html">
+            <template>
+<model-viewer src="../../shared-assets/models/shishkebab.glb" alt="A 3D model of a shishkebab" shadow-intensity="1" camera-controls auto-rotate with-credentials></model-viewer>
```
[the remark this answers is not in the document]
ymoriaud: No problem, I'm removing it. 😃

packages/modelviewer.dev/data/docs.json
ymoriaud: No problem, I'm updating the description.

> For instance, we could simply make it always true; presumably there's a reason that's not a good idea?

For backward compatibility and to follow default behaviour of `XMLHttpRequest.withCredentials` and three.js loader option, I kept them to `false` by default. Only a `true` value can change the value of three.js loader.

Regarding the always `true`, I imagine is to avoid sending authentication data if the request didn't ask for that and follow the principe : `privacy by design` ?

elalish: > Regarding the always `true`, I imagine is to avoid sending authentication data if the request didn't ask for that and follow the principe : `privacy by design` ?

Yeah, that sounds right to me. Let's make sure our users understand that.

ymoriaud: I changed the text to that :

> This attribute make the browser include credentials (cookies, authorization headers or TLS client certificates) in the request to fetch the 3D model. It's useful if the 3D model file is stored on another server that require authentication. By default the file will be fetch without credentials. Note that this has no effect if you are loading files locally or from the same domain.

packages/model-viewer/src/features/loading.ts
```diff
@@ -360,6 +368,10 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
     connectedCallback() {
       super.connectedCallback();
 
+      if (this.withCredentials) {
+        CachingGLTFLoader.setWithCredentials(this.withCredentials)
+      }
```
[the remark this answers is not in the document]
ymoriaud: I tried to make it work without this.

The `changedProperties.has('withCrendentials')` do not see any update. So the addition under `update` is not triggered.

But if I do something like that, it's working :

```js
if (changedProperties.has('withCrendentials') || this.withCredentials === true) {
   CachingGLTFLoader.setWithCredentials(this.withCredentials)
}
```

packages/model-viewer/src/features/loading.ts
```diff
@@ -416,6 +424,10 @@ export const LoadingMixin = <T extends Constructor<ModelViewerElementBase>>(
           this[$posterContainerElement].classList.remove('quick');
         }
       }
+
+      if (changedProperties.has('withCrendentials') || this.withCredentials === true) {
```
elalish, line 428: No need for the `||` statement; checking that it has changed is enough.

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
```diff
@@ -189,6 +199,9 @@ export class CachingGLTFLoader<T extends GLTFInstanceConstructor =
   async preload(
       url: string, element: ModelViewerElementBase,
       progressCallback: ProgressCallback = () => {}) {
+    if (withCredentials === true) {
+      this[$loader].setWithCredentials(withCredentials);
```
elalish, line 203: Instead of setting this every time, how about you just set it inside of `setWithCredentials()`, and then you can also remove `withCredentials`. I don't think the getter is used anywhere anyway, right?

packages/modelviewer.dev/data/docs.json
```diff
@@ -67,6 +67,18 @@
           "default": "auto",
           "options": "auto, interaction, manual"
         }
+      },
+      {
+        "name": "with-credentials",
+        "htmlName": "withCredentials",
+        "description": "This attribute make the browser include credentials (cookies, authorization headers or TLS client certificates) in the request to fetch the 3D model. It's useful if the 3D model file is stored on another server that require authentication. By default the file will be fetch without credentials. Note that this has no effect if you are loading files locally or from the same domain.",
```
elalish, line 74: nit: make -> makes.

elalish (changes requested):
Almost there; I think we can still simplify it a bit. I trust you've tested this works for you?

packages/modelviewer.dev/data/docs.json
ymoriaud: Fixed !

packages/model-viewer/src/features/loading.ts
ymoriaud: If I do not add the `||` statement, the object `changedProperties` do not have the `withCredentials` value. Maybe I'm missing something ?

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
ymoriaud: I already tried but I have a typescript error :

```
src/three-components/CachingGLTFLoader.ts:130:5 - error TS7053: Element implicitly has an 'any' type because expression of type 'unique symbol' can't be used to index type 'typeof CachingGLTFLoader'.
  Property '["./CachingGLTFLoader.js".$loader]' does not exist on type 'typeof CachingGLTFLoader'.

130     this[$loader].setWithCredentials(state)
        ~~~~~~~~~~~~~
```

I'm do not learned typescript yet, if you have a workaround or an advice to fixe that, do not hesitate !

ymoriaud: Or maybe I do not understand correctly your proposal. I integrated 

```js
    if (withCredentials === true) {
      this[$loader].setWithCredentials(withCredentials);
    }
```

In the 

```js
  static setWithCredentials(state: boolean) {
    this[$loader].setWithCredentials(state)
  }
```

packages/model-viewer/src/features/loading.ts
elalish: `this` always has `withCredentials`; the `changedProperties` is a subset. Follow the similar statements above. If it's not behaving as expected, let me know.

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
elalish: Oh, I think the problem is `[$loader]` is a class member, but you made `setWithCredentials()` a static member, which can't access class members. Just make `setWithCredentials` a class member and it should be fine. We only create one shared CachingGLTFLoader object anyway, so all methods are effectively static anyway. `setDRACODecoderLocation` and such are static because they are called by a static public API, which yours is not.  That's just tricky Lit-element stuff.

elalish (changes requested):

ymoriaud:
> Almost there; I think we can still simplify it a bit. I trust you've tested this works for you?

Yes I tried on my side, and it work fine. :)

packages/model-viewer/src/features/loading.ts
ymoriaud: image

`updated` is triggered, but the `changedProperties` subset do not have the `withCredentials` to true. It's undefined on the first trigger, and no preset in the other console.log.

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
ymoriaud: Understood, thanks !

So I moved the static setWithCredentials(state:boolean) inside the class. But now, in the `loading.ts` file, I can't access to the `setWithCredentials` :

```
src/features/loading.ts:430:27 - error TS2339: Property 'setWithCredentials' does not exist on type 'typeof CachingGLTFLoader'.

430         CachingGLTFLoader.setWithCredentials(this.withCredentials)
                              ~~~~~~~~~~~~~~~~~~
```

packages/model-viewer/src/features/loading.ts
ymoriaud: I think I found my issue !
I'm not adding that at the right place. I was editing the `loading.ts` file instead of `model-viewer-base.ts` one.

I'm pushing a new PR change.

packages/model-viewer/src/three-components/CachingGLTFLoader.ts
ymoriaud: Here is a new proposal, based on variable instead of functions.
Let me know what do you think about that. :)

elalish (approved):
Nice and simple, thank you!

ymoriaud:
🥳  Awesome !
Thanks @elalish for your help, feedback and to approve this PR.

Have a nice day :)

PR #3262, opened 2022-03-10
a11y improvements

### LINKED ISSUE #461: Separate aria-live region
### Description

As stated in #460, pairing `aria-label` with the `aria-live` attribute on the `canvas` element may not have the best support. Ideally there would be a separate, dedicated `aria-live` region that would be used for all update announcements.

Ensuring the element is available on page load is also required for best support as some assistive technology do not recognize a dynamically generated `aria-live` region after initial page load.

**Recommendation:** create a dedicated `aria-live` region, sending all announcement strings to this element. Set the element role as “[status](https://www.w3.org/TR/wai-aria-1.1/#status)” as this will feature the most accurate semantic meaning for this element.

Wrapping the “status” element with a parent element with a note attached via `aria-label` will help to provide the user with context on what they're interacting with when they inevitably find the announcement element when navigating via virtual cursor or swipe gestures.

```html
<div role="region" aria-label="Live announcements">
  <span role="status" class="visuallyhidden">
    View from stage back
  <span>
</div>
```

#### Live Demo
1. Go to [`<model-viewer>`](https://googlewebcomponents.github.io/model-viewer/index.html) with Edge on Windows
2. Load up [JAWS screen reader](https://www.freedomscientific.com/products/software/jaws/)
3. <kbd>Tab</kbd> to the `canvas` content

### Browser Affected

- [ ] Chrome
- [x] Edge
- [x] Firefox
- [ ] Helios
- [x] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [x] MacOS
- [x] Windows

### Versions

- model-viewer: v0.1.1

### LINKED ISSUE #462: Separate instruction element
### Description

Currently, the instructions on how to interact with the 3D model (_"Use mouse, touch or arrow keys to control the camera!"_) are announced after a short time via updating the `aria-label` of the canvas element.

With the changes suggested in #461, it will be required to set aside the instructions and include with the focus announcement in a different manner.

**Recommendation:** Set the instructions in a new, `hidden` element. Include the instructions on the `canvas` element via `aria-describedby` attribute, referencing the `id` of the instructions container. This will help to announce the 3D model with its `aria-label`, and then, the instructions will follow.

```html
<span id="instructions" hidden>
  Use mouse, touch or arrow keys to control the camera!
</span>
<!-- … -->
<canvas … aria-describedby="instructions">
  <!-- … -->
</canvas>
```

### Browser Affected

- [ ] Chrome
- [x] Edge
- [x] Firefox
- [ ] Helios
- [x] IE
- [ ] Safari

### OS

- [x] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [x] Windows

### Versions

- model-viewer: v0.1.1

elalish:
Instead of #3007 
Fixes #461 
Fixes #462 

I've removed the singleton loading status announcer in favor of similar status elements on each instance. They save their status internally, but only update the element when focused (they also clear it when blurred and display the current status when focus is received). This status takes care of both "Loading" and "View from" messages. The interaction prompt is now simply appended to the alt text, to avoid extra interruptions.


elalish:
@PeterSaathoff-Harshfield @svinkle I *think* I got this working more or less properly based on the feedback from both of you, though this was my first time using a screen reader for testing. I'm going to go ahead and merge this so that it's easy for you two to test if you have a spare moment at some point. This should be live in the next half an hour or so and the best page to test may be https://modelviewer.dev/examples/loading/index.html since it has multiple models with different loading strategies, so it tests most of the edge cases. Please file new issues if you notice anything still wrong and  I'll be happy to get it fixed up before our next release.

PR #3263, opened 2022-03-11
retain texture transform info

### LINKED ISSUE #3240: Texture transform parameters (offset, rotation and scale)
### Discussed in https://github.com/google/model-viewer/discussions/3226 (more repro detail in the discussion)

<sup>Originally posted by **fbenedetti06** February 25, 2022</sup>
Hi all,

I'm a very beginner with model-viewer and 3D model. So I apologize by advance if my question is stupid!

With Blender I made a simple 3D model.
I used an external image as a texture and I changed offset, rotation and scale parameters to repeated it all over the surface (the render look like a tiled wall).

With model-viewer I would like to replace programmatically initial texture image by another.
Unfortunatly it seems that all transform parameters are lost (now the render look like a single one big tile all over the wall).
I can't find a way to restore these parameters by program nor a function to set these values.

Here the part of Gltf 3D file I identified to be desired settings to resize and repeat the tiles (generated by Blender)
image

Here lines of JS code I use to change programmatically the image:
image

Any idea?

Fred

</div>

elalish:
Fixes #3240 

This maintains any texture transforms on the texture info slot and applies them to any textures that are set on that slot. We can also expose this as an API later, though I'll wait for user demand.

### DIFF
--- a/packages/model-viewer/src/features/scene-graph/texture-info.ts
+++ b/packages/model-viewer/src/features/scene-graph/texture-info.ts
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import {LinearEncoding, MeshStandardMaterial, sRGBEncoding, Texture as ThreeTexture, TextureEncoding} from 'three';
+import {LinearEncoding, MeshStandardMaterial, sRGBEncoding, Texture as ThreeTexture, TextureEncoding, Vector2} from 'three';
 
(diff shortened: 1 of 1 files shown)


fbenedetti06:
Yes for sure it will be very desirable to transform textures with an API
Scale and rotation at least are very useful

danilokrulj:
Agreed, would be amazing if we could change textures through code.

PR #3287, opened 2022-03-17
Prevent the code to fail when loading models without the "materials" property or having a mesh primitive without the material association

### LINKED ISSUE #3279: Support models lacking the "materials" property or primitive/material association
### Description
I have point cloud scans that yield models without any materials once converted to GLTF/GLB.
These models visually load, but then, break the code at a few places. This brought me to investigate further the situation.
As far as I can tell, I have encountered 2 situations where the code breaks with these unconventional files.

#### Case 1: File with no "materials" property in the GLTF

Screen Shot 2022-03-16 at 9 49 06 PM

In `model-viewer/src/features/scene-graph/model.ts`
Screen Shot 2022-03-16 at 9 49 44 PM

**Suggestion**: Don't use the `!` syntax in `gltf.materials!.entries()` and do a null check instead `gltf.materials?.entries() ?? []`

#### Case 2: File with the "materials" property in the GLTF but no material associated in the mesh primitives

Screen Shot 2022-03-16 at 9 55 48 PM

In `model-viewer/src/features/scene-graph/nodes/primitive-node.ts`
Screen Shot 2022-03-16 at 9 55 16 PM

**Suggestion**: Simple null check `materialMappings?.materials != null`

This is another case where I seem to be the only one having to deal with bizarre GLTF structures 😆 . I will open a PR for this if that is ok with you guys. 

### Version
- model-viewer: v1.10.1

### Browser Affected
- [x] Chrome, version: latest
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [ ] Safari

### OS
- [ ] Android
- [ ] iOS
- [ ] Linux
- [x] MacOS
- [ ] Windows

alexdaube:
Fixes #3279 

- Couple of null checks in the model and primitive-node constructors
- In space-opera, the material panel will shot "No materials" when the panel does not have any selectable materials found in the active model.

### DIFF
--- a/packages/space-opera/src/test/materials_panel/materials_panel_test.ts
+++ b/packages/space-opera/src/test/materials_panel/materials_panel_test.ts
@@ -28,6 +28,7 @@ import {waitForEvent} from '../utils/test_utils.js';
 const TEXTURE_CUBE_GLTF_PATH = '../base/shared-assets/models/textureCubes.gltf';
 const CUBES_GLTF_PATH = '../base/shared-assets/models/cubes.gltf';
 const TRIANGLE_GLTF_PATH = '../base/shared-assets/models/Triangle.gltf';
+const VASE_NO_TRIANGLE_GLTF_PATH = '../base/shared-assets/models/VaseNoMaterials.gltf';
 const TEXTURE_PATH = 'base/shared-assets/models/ORM.png';
 
 async function checkUpload(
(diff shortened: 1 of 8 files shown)


packages/space-opera/src/test/materials_panel/materials_panel_test.ts
```diff
@@ -75,20 +76,38 @@ describe('material panel test', () => {
...
-  it('selector reflects materials in GLTF, including defaults where undefined',
-     async () => {
-       panel.selectedMaterialIndex = 0;
-       await panel.updateComplete;
-       expect(panel.selectedBaseColor).toEqual([1, 0, 1, 1]);
-       expect(panel.selectedRoughnessFactor).toEqual(0.2);
-       expect(panel.selectedMetallicFactor).toEqual(1);
+  describe('Given a model without any materials', () => {
+    beforeEach(async () => {
+      reduxStore.dispatch(dispatchGltfUrl(VASE_NO_TRIANGLE_GLTF_PATH));
```
elalish, line 81: Do you mean no material instead of no triangle? Also, what's the difference between the new GLB and glTF files?

elalish (commented):
Thanks, this looks great. I especially appreciate all the added tests!

packages/space-opera/src/test/materials_panel/materials_panel_test.ts
alexdaube: I sure do mean material.
No difference between the 2 files. I can probably keep only a single one of the 2 formats.

elalish (approved):
Thanks! I wonder if khronos_SimpleTriangle would work; it's a lot smaller and will load faster in the tests.

alexdaube:
> Thanks! I wonder if khronos_SimpleTriangle would work; it's a lot smaller and will load faster in the tests.

On the surface it would seems so, but `khronos_SimpleTriangle` generate a `Default` material in the correlatedSceneGraph object, which is not the case of the vase glb.

elalish:
hmm, that's interesting. I thought we always generated a default material when none were specified. Can you take a look and let me know what the difference is? I'd like to make sure we're fixing this properly.

alexdaube:
It does seems a bit fishy. I will check it out tomorrow morning no problem. Today is a bit busy on my end.

alexdaube:
Looking at the associations in the gltf object generated from the ThreeGLTFParser, the triangle file has 2 associations (material and mesh) and the vase only has 1 (mesh). Looking at the code, the `gltf.materials` array and the default material is only added when there is a material in the associations.

Screen Shot 2022-03-22 at 2 36 36 PM

What would you suggest we do with this?

elalish:
Hmm, sounds like the bug might be in the GLTFLoader (since it's supposed to create a default material association, and apparently fails to). I think for now we should go ahead and create our default material also in the case there is no material association at all, for robustness. Separately, it'd be interesting to know why GLTFLoader has weird behavior here.

alexdaube:
I wanted to get to this last week so it could be included in the next release, but I have been indisposed by Covid unfortunately. 

Anyways, it seems like the associations entry to the DefaultMaterial is set in different places which explains the inconsistency. For the Triangle files it is done here, which is called when we load the meshes in the loader.
https://github.com/mrdoob/three.js/blob/8921bfaee413e809cb2d27f0aec917408d690a0f/examples/jsm/loaders/GLTFLoader.js#L3114

For the Vase(new file) it is done here, which is called when we load a material.
https://github.com/mrdoob/three.js/blob/8921bfaee413e809cb2d27f0aec917408d690a0f/examples/jsm/loaders/GLTFLoader.js#L3304

The problem that we have in model-viewer with this is that CorrelatedSceneGraph and PrimitiveNode objects are instantiated before `loadMaterial` is first called and this association is made in the Model constructor I believe.

I could probably address the issue by adding this block of code, which uses the parser's cached DefaultMaterial to setup the association.

Screen Shot 2022-03-30 at 6 47 21 PM

The cache attribute is not public in the interface, so I am guessing this snippet is not exactly how we should use the library. I am not sure what direction you want to take with this? Perhaps this should be only fix/addressed three.js?

elalish:
> I wanted to get to this last week so it could be included in the next release, but I have been indisposed by Covid unfortunately.

No problem! Get well soon.

> The cache attribute is not public in the interface, so I am guessing this snippet is not exactly how we should use the library. I am not sure what direction you want to take with this? Perhaps this should be only fix/addressed three.js?

Indeed, let's not mess with the internals of three.js. You make it sound like something is happening too soon? Might this just be a race condition of some kind on our side? Maybe we just haven't awaited the right thing.

elalish:
Ping; did you ever find a deeper cause here? Would love to merge a fix for the root problem.

alexdaube:
I haven't had much time to check this any further.  

Off memory, I think it was basically that in this particular case three.js is only creating the associations for the default material once loadMaterial is being called which in model viewer happens after the constructors cited above. So the race condition assessment sounds about right, though I find it a bit weird as of why this association is done in different places in the three.js GTLFLoader's code to be honest.

I will try to make some time on Friday to try to find a good solution for this.

elalish:
I'm still interested in what's going on here. Have you gotten stuck?

alexdaube:
Sorry for the slow response (on vacation for the past few weeks).

I didn't put any time into this since March as my plate got really full, but I am interested in finding a good solution here. Let me put this back into my personal task board and come back to you with this.

elalish:
It looks like this is actually a three.js bug that needs to be fixed upstream. Should we merge the rest of this PR?

PR #3320, opened 2022-03-29
Custom prompt

elalish:
The idea here is to make an API so you can create your own custom interaction prompt, potentially including pan and zoom motions in addition to rotation. It's working here, including an example and docs.


elalish:
Ugh, okay this concept of using synthetic touch events works on Chrome, but not on other browsers - if touch is not supported on a device, they don't let you construct touch events. Looks like I need to switch to PointerEvents first, which I wanted to anyway, but had trouble with last time. I was hoping a synthetic pointer event might trigger a mouse/touch event pair, but alas no.

elalish:
@diegoteran @samaneh-kazemi Probably one of you will end up implementing Tomi's custom prompts using this API, so I figured you might be interested in how it works. I'll write up a design doc for it too.

packages/model-viewer/src/features/controls.ts
```diff
@@ -34,22 +34,28 @@ import {timeline} from '../utilities/animation.js';
...
-]);
-
-const fade = timeline(0, [
-  {frames: 1, value: 1},
-  {frames: 5, value: 1},
-  {frames: 1, value: 0},
-  {frames: 6, value: 0}
-]);
+// and a "value" is a target value for the Frame.
+const wiggle = timeline({
```
samaneh-kazemi, line 38: Out of curiosity, based on what we choose these numbers and number of frames?

elalish: These didn't change; the default prompt was designed with input from the Shopify UX folks a few years ago.

packages/model-viewer/src/features/controls.ts
```diff
@@ -353,11 +366,12 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
-    protected[$promptElement] =
-        this.shadowRoot!.querySelector('.interaction-prompt') as HTMLElement;
     protected[$promptAnimatedContainer] =
-        this.shadowRoot!.querySelector(
-            '.interaction-prompt > .animated-container') as HTMLElement;
+        this.shadowRoot!.querySelector('#prompt') as HTMLElement;
+    protected[$fingerAnimatedContainers]: HTMLElement[] = [
+      this.shadowRoot!.querySelector('#finger0')!,
+      this.shadowRoot!.querySelector('#finger1')!
+    ];
```
samaneh-kazemi, line 374: what is each finger0 and finger1 for?

packages/modelviewer.dev/data/docs.json
```diff
@@ -701,6 +706,14 @@
         "name": "zoom(keyPresses)",
         "htmlName": "zoom",
         "description": "Applies the default zoom behavior (a combination of radius and field of view change), mimicking keyboard or mouse wheel inputs. If keyPresses = 1.0, it will zoom in as much as a single keystroke; negative values zoom out."
+      },
+      {
+        "name": "interact(duration, finger0, finger1)",
+        "htmlName": "interact",
```
samaneh-kazemi, line 712: nit: Creates

samaneh-kazemi (approved):

PR #3354, opened 2022-04-07
ensure pan target appears and disappears correctly

elalish:
Fixes https://github.com/google/model-viewer/issues/3334#issuecomment-1086174404

I managed to repro with `pan-y`, where dragging up/down doesn't start a 3D interaction - that also caused the pan target to erroneously show and not hide. Now fixed; thanks for the head's up @hybridherbst. It was tricky for me to repro at first.

### DIFF
--- a/packages/model-viewer/src/three-components/SmoothControls.ts
+++ b/packages/model-viewer/src/three-components/SmoothControls.ts
@@ -647,6 +647,9 @@ export class SmoothControls extends EventDispatcher {
   }
 
   private recenter(pointer: Pointer) {
+    const {scene} = this;
+    (scene.element as any)[$panElement].style.opacity = 0;
+
     if (!this.enablePan ||
         Math.abs(pointer.clientX - this.startPointerPosition.clientX) >
             TAP_DISTANCE ||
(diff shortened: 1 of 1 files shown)


hybridherbst:
Nice - were you also able to reproduce the "pointer leaving case" that I mentioned, e.g. starting any interaction and then taking a screenshot?

elalish:
It doesn't repro with my mac's screen-shotting system after this fix. Let me know if you still have trouble.

hybridherbst:
Yup, seems to still happen on modelviewer.dev/editor (since it's not zooming in on tap anymore I assume that's always using latest, right?)

20220407-224332-JPQU-chrome_6Zbi

elalish:
Oh, that's a separate and old (I believe) issue. Not sure there's much I can do about that - seems more OS-level.

hybridherbst:
Interesting.

Quick test on other pages:
- ✔️ doesn't happen here https://www.w3schools.com/html/html5_draganddrop.asp
- ❌ happens here as well https://interactjs.io/
- ✔️ doesn't happen here https://web.dev/drag-and-drop/

mrdoob:
Not 100% sure, but I think listening to `pointercancel` solves this.

elalish:
Thanks, I think this is fixed by #3404.

PR #3399, opened 2022-04-28
fixed ImageBitmap memory leak

elalish:
As helpfully noted by @deenns: https://github.com/mrdoob/three.js/issues/23953#issuecomment-1110742946, we had a memory leak due to GLTFLoader switching to `ImageBitmap` while we had not updated our `dispose()` method. That's done here, along with a package-lock update to pull in the latest three.js types.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1929,9 +1929,9 @@
       }
     },
     "node_modules/async": {
-      "version": "1.5.2",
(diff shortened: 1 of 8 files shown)


elalish:
Ugh, well I don't love merging things while our tests are timing out, but it's pretty clearly unrelated to our code changes, so I'm going to move on for now. Hopefully we get this BrowserStack stuff sorted soon.

PR #3404, opened 2022-04-30
Switch to pointer events

elalish:
I finally decided to switch us over from mouse/touch events to pointer events, as this was blocking us from making synthetic events in a cross-device way for #3320. This supersedes #1645, where I had been stymied by two strange WebKit issues: first, touch-action: pan-y switches back to scrolling even if the motion starts in X, unlike on Android. Second: this is hard to work around because their touch events are sometimes not cancelable (randomly). I finally found the [solution](https://github.com/SortableJS/Sortable/issues/973#issuecomment-383813976), which works great. Also of note, even though we've switched to pointer events, they don't have a `preventDefault()` method to cancel scroll behavior, so a `touchmove` listener is still required, hence the above solution.


elalish:
FYI @lucadalli I'm pretty sure I have this all working consistently with your last refactor, but more testing is always appreciated.

elalish:
This is looking good to me; Firefox is still having trouble on BrowserStack, but that appears to be a test infra problem on their side.

PR #3427, opened 2022-05-05
make target change fire camera-change event

elalish:
Fixes https://github.com/google/model-viewer/discussions/3386#discussioncomment-2618327

This is important now that we have pan causing tap-to-recenter. This change will allow that type of user-interaction to be listened for.


elalish:
FYI @diegoteran and @samaneh-kazemi Not sure if you've noticed this yet, but I wouldn't be surprised if this bug affects us too. You can repro it [here](https://modelviewer.dev/examples/stagingandcameras/#customPrompt) by tapping on the interaction prompt. It will stop, but then start up again in a few seconds, because my `camera-change` listener didn't get triggered. This fixes it.

PR #3428, opened 2022-05-05
Override MWC tab to expose min-width

e111077:
Overrides `mwc-tab` and creates `me-tab` and exposes the min-width property to fix space opera's seemingly missing 5th, inspect, tab

Workaround for:
material-components/material-web#2609


elalish (approved):
I have no idea how long it would have taken me to discover that little solution. Learning a little more about web components every day... Thanks!

PR #3434, opened 2022-05-09
Support extended AR QuickLook parameters when auto generating USDZ blob

milesgreen:
### Reference Issue
#3425

### Approach
This change allows extended AR QuickLook parameters (EG `custom`, `checkoutTitle`, `checkoutSubtitle`, etc) to be appended after the `src` path as `#` hash fragment when using auto USDZ generation and there is no `ios-src` property. 

Any / all defined `#` params get appended to USDZ blob without discrimination (AR QuickLook will ignore any invalid params).

### Usage
Define `src` as something like:
`https://example.com/model.gltf#custom=path-to-banner.html`
Omit `ios-src`
Include `"quick-look"` in `ar-modes` to trigger auto USDZ generation.
When opening AR on iOS, USDZ will be auto generated and displayed with custom banner in AR.

### Tests
Two tests created:
- replicate src hash to usdz blob url
- replicate src hash to usdz blob and set hash for fixed scale

### Demo
https://modelviewer-auto-usdz-params.glitch.me/

### Compatibility
Note that passing # params after blob url only works in Safari iOS 15.5 beta 4 onwards. Prior to that, passing # params after blob url were either ignored (https://bugs.webkit.org/show_bug.cgi?id=226265) or caused model to never load (https://bugs.webkit.org/show_bug.cgi?id=236069).

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -373,6 +373,15 @@ configuration or device capabilities');
       const objectURL = generateUsdz ? await this.prepareUSDZ() : this.iosSrc!;
       const modelUrl = new URL(objectURL, self.location.toString());
 
+      if(generateUsdz){
+        const location = self.location.toString();
+        const locationUrl = new URL(location);
+        const srcUrl =  new URL(this.src!, locationUrl);
+        if(srcUrl.hash){
+          modelUrl.hash = srcUrl.hash;
+        }
+      }
+
       if (this.arScale === 'fixed') {
         if (modelUrl.hash) {
           modelUrl.hash += '&';
--- a/packages/model-viewer/src/test/features/ar-spec.ts
+++ b/packages/model-viewer/src/test/features/ar-spec.ts
@@ -143,44 +143,38 @@ suite('ModelViewerElementBase with ARMixin', () => {
         test('replicate src hash to usdz blob url', async () => {
           element.src = assetPath('models/cube.gltf')+'#custom=path-to-banner.html';
           element.arModes = 'webxr scene-viewer quick-look';
-          
-          console.log('element.src:'+element.src);
-          console.log('element.arModes:'+element.arModes);
-          console.log('element.iosSrc:'+element.iosSrc);
-          console.log('intentUrls:'+intentUrls);
 
           await (element as any)[$openIOSARQuickLook]();
 
-          console.log('intentUrls:'+intentUrls);
-          console.log('intentUrls.length:'+intentUrls.length);
-          console.log('intentUrls[0]:'+intentUrls[0]);
-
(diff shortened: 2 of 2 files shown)


milesgreen:
@elalish - please let me know if you'd like me to tackle in a different way. 

I couldn't manage to get the unit test working though :(

packages/model-viewer/src/test/features/ar-spec.ts
```diff
@@ -139,6 +139,21 @@ suite('ModelViewerElementBase with ARMixin', () => {
           expect(url.hash).to.equal(
               '#custom=path-to-banner.html&allowsContentScaling=0');
         });
+
+        test('replicate src hash to usdz blob url', () => {
+          element.src = 'https://example.com/model.gltf#custom=path-to-banner.html';
+          element.arModes = 'webxr scene-viewer quick-look';
+
+          (element as any)[$openIOSARQuickLook]();
```
elalish, line 147: The reason your test fails is that this is an async function, so you need to `await` it. It happens to work in the other tests because they aren't triggering its async nature (that was added for `prepareUSDZ`, and apparently these were not updated).

packages/model-viewer/src/features/ar.ts
```diff
@@ -373,6 +373,15 @@ configuration or device capabilities');
       const objectURL = generateUsdz ? await this.prepareUSDZ() : this.iosSrc!;
       const modelUrl = new URL(objectURL, self.location.toString());
 
+      if(generateUsdz){
+        const location = self.location.toString();
+        const locationUrl = new URL(location);
+        const srcUrl =  new URL(this.src!, locationUrl);
+        if(srcUrl.hash){
+          modelUrl.hash = srcUrl.hash;
```
elalish, line 381: What is the behavior if someone specifies `allowsContentScaling=1` and `ar-scale="fixed"`?

elalish (commented):
Thanks! See, that wasn't so bad, was it? 😁

packages/model-viewer/src/features/ar.ts
milesgreen: In testing, it appears the last instruction wins. QuickLook doesn't complain if there are duplicate params appended to the url, but respects the last one added. In this case, the model-viewer `ar-scale` property is appended after copying the `src` hash. So in the event someone specifies `allowsContentScaling=1` and `ar-scale="fixed"` the model will be presented at fixed scale.

elalish: That sounds like good behavior to me, thanks for checking.

elalish (approved):
Looks great, thanks! I marked this as ready for review, because I think it's all done now, right?

milesgreen:
Right. I think that's it...! If you're happy, I'm happy.

Issue #3455, opened 2022-05-12
Cannot export Hotspots from the editor


Gamigre:
### Description

The hotspot feature doesn't works anymore for me. I have this issue since last week or so. I can place hotspots but the snippet will not update to show their position nor create a new < button> or < div> for them [see picture]. When using the export function, the scene created doesnt features any of the hotspots placed with editor. The index.html file is akin to the snippet and devoid of any hotspot mention. Hotspot placed manually/pasted from another scene works. 

### Version

- last online verion of the editor

### Browser Affected

- [X] Chrome, version: 101.0.4951.54
- [X] Edge
- [X] Firefox last version
- [X] Safari for iPad

### OS

- [X] Windows 10
- [x] Tested on safari for ipad with a default model for the same results
![image](https://user-images.githubusercontent.com/43046028/168174999-1db25762-e7c1-4be2-96b2-59de8bdce320.png)

e111077:
addressed in https://github.com/google/model-viewer/pull/3462

PR #3462, opened 2022-05-14
Do not greedily strip lit comments in snippet-viewer

e111077:
Lit comments were being stripped greedily in `snippet-viewer`. This means when doing manual DOM insertion like with hotspots, the `<!--.*-->` regex was matching the entire line of:

```html
<!----><!----><button slot="ar-button" id="ar-button">View in your space</button><!----><!---->
```

rather than just individual comments. See:

| Before | After |
| -- | -- |
| image | image |

### Reference Issue
fixes: #3455


elalish (approved):
Thanks for adding tests!

PR #3496, opened 2022-05-25
Add three-gpu-pathtracer to Fidelity Tests

gkjohnson:
### Reference Issue
--

Hello! I'm the developer of the [three-gpu-pathtracer](https://github.com/gkjohnson/three-gpu-pathtracer) project (a generalized path tracer library built on top of three.js) and have added support for it to the "render-fidelity-tools" package to help find problems with the implementation. The project is still fairly new so there are issues with a number of models and some but it's great to see where it's failing and which glTF parameters are not properly supported, yet.

Also I see now that you'd like an issue for each PR - I'm happy to make one now if you guys would like. And if this looks acceptable then I can make whatever other changes are needed / generate images! Thanks for the awesome project!

Here are some of the screenshots including cases where it's clearly not behaving correctly:

cc @bhouston


elalish:
Thanks, this is fantastic! Sorry I didn't get back to you sooner; this came in just as I caught COVID and I was out for awhile. Even if some of the renders aren't perfect, can you please add the golden images? This will allow us to show your current results publicly. Of course you can also set `exclude` in the config on any scenarios that are known to be not supported, or cause a crash/hang.

gkjohnson:
Thanks @elalish! I've just updated the goldens with a few tests excluded (`khronos-sponza`, LDR Metal Rough Spheres, and `khronos-boxInterleaved`). I'll plan to look into the issues with sponza box interleaved at some point but I don't think I'll plan to support LDR jpg environment maps. I've only set the renders to 100 samples at the moment to save on render time but maybe I'll bump that up in a future update.

As mentioned above there are still quite a few issues that will make the three-gpu-pathtracer images stick out quite a bit but I think it's a good first step to getting these problems addressed!

gkjohnson:
Great - this is good feedback. When I have a little time I'll go through and mark the obviously incorrect ones and add them over time as the renderer improves. The project is still fairly new so part of this effort is finding where the result is meaningfully different from other renders so we can improve the fidelity of three pathtracer (including things like being too dark in some cases). Definitely understand that the intent of the project is to measure the correctness model-viewer, though.

gkjohnson:
Hey @elalish! I've made quite a few fixes and releases over the last several months that have made significant improvements to the fidelity of the examples. You can see the current state of the screenshots here:

https://gkjohnson.github.io/three-gpu-pathtracer/example/bundle/screenshotList.html#model-viewer

There are still some improvements to make specifically to rough metallic materials but there will always be room for improvement. Do you have any further thoughts now on including three-gpu-pathtracer into the fidelity tests? If you're open to it I can update the PR to the latest version.

elalish:
Incredible progress! Yes, I would love to see these included in our fidelity tests. As long as you can commit to updating them here semi-regularly, please go ahead and add them. Hopefully this benefits your process too.

bhouston:
Amazing work @gkjohnson!  BTW something is weird when I first load the comparison page.  It says it is comparing to dspbr-pt, but it is actually comparing to filament.  That confused the heck out of me initially.

hybridherbst:
This is so cool to see :) great progress!

gkjohnson:
Thanks guys! I've just updated the PR with the latest version of the path tracer, new screenshots, and updated exemptions for the config.

I noticed some issues with handling of clearcoat maps and env maps which will be fixed in the next release - so I'll be able to remove some of the excluded screenshots then. Some of the other excluded screenshots include the dark triangles which are a result of the lack of multibounce-accommodation noticeable in rough metal materials, some specular transparent materials with noticeable fireflies, and demos with features that aren't supported (unlit, LDR env map, etc).

@elalish let me know if there's anything else I can do to get this merged!

gkjohnson:
Oh also @bhouston

> BTW something is weird when I first load the comparison page. It says it is comparing to dspbr-pt, but it is actually comparing to filament. That confused the heck out of me initially.

Good point - I just pushed a fix for this. It defaults to comparing to `model-viewer` but it takes a second for the JS to update the dropdown. I've adjusted the page to hide the UI until the JS has been able to run.

bhouston:
> Good point - I just pushed a fix for this. It defaults to comparing to model-viewer but it takes a second for the JS to update the dropdown. I've adjusted the page to hide the UI until the JS has been able to run.

It appears to be a Safari only bug and it remains.  I can not reproduce in Chrome.

gkjohnson:
> It appears to be a Safari only bug and it remains. I can not reproduce in Chrome.

Looks like I used a non-standard method for changing the dropdown. It should (hopefully) be fixed now.

elalish (approved):
Thanks, amazing work on your path tracer! I look forward to seeing updates and watching the improvement. Also, I believe more models have been added to the sample repo. If you'd like to help us keep up by adding them to our config, I'd be very grateful!

gkjohnson:
Great thanks @elalish! Maybe I'll take a look at some of the other sample models when I get a chance.

PR #3505, opened 2022-06-01
Remove visibilitychange event listener if interaction is cancelled

valdrinkoshi:
Remove `visibilitychange` listener when we cancel the `interact` animations because of user interactions.


valdrinkoshi:
Will close this PR in favor of https://github.com/google/model-viewer/pull/3507

PR #3513, opened 2022-06-06
Add RhodoniteTS

emadurandal:
This PR integrates my RhodoniteTS library into the render fidelity tools.

This library is still in the early stages of development and has more limited support for glTF extensions than other libraries, so some samples may be significantly less expressive than others.

I have not included images of the rendering results in the PR because they are too large. If necessary, I can render and include them.

This PR may have problems before merging. I will try to resolve it and would appreciate feedback.


elalish:
Looks great, thanks! Yes, please go ahead and run `npm run update-screenshots rhodonite` and commit the golden images, which is what we will use to show on our public fidelity results page. If any scenarios crash or are not supported, you're welcome to put `exclude` in the config to skip those.

packages/render-fidelity-tools/package.json
```diff
@@ -43,6 +43,7 @@
     "mkdirp": "^1.0.4",
     "pngjs": "^6.0.0",
     "puppeteer": "^10.1.0",
+    "rhodonite": "github:actnwit/RhodoniteTS#d42809b7b20976d35c9ef682d749b9c2b2887fb8",
```
elalish, line 46: Is there a chance you'd be willing to publish an npm release (even an alpha or something) rather than using a specific commit? It's nice ideally to be able to look at release notes and see what changes are expected when upgrading.

emadurandal: Sure. I'll do it.

emadurandal: @elalish I'm currently having an unknown problem with rhodonite that is unable to publish the new version to npm. I am sorry, but can I keep this SHA designation for the time being until the problem is resolved?

elalish: Sure, that's fine for now. I'll just wait on the images.

emadurandal: @elalish I pushed the rendering images.

elalish: Have you compared the images to the other renderers? It looks like the camera / lighting / tone mapping is not consistent yet. I understand shader differences, but at least the scene setup should match before we merge this. I would also exclude any scenarios where your renderer is deeply disfunctional and add them back in as you release improvements.

emadurandal: @elalish Sure. I will check the other renderer's results and make some more adjustments.

emadurandal: @elalish Adjusted lighting environment and fixed bugs. I think I have minimized the differences from output images of other libraries. Also excluded scenarios that did not work enough.
I succeeded in publishing 0.8.5 to npm, so I replaced rhodonite with it.

elalish (approved):
Thank you! Please come back and update as you fix various scenarios.

elalish:
@emadurandal Having looked through the [fidelity comparisons](https://modelviewer.dev/fidelity/#khronos-Suzanne) in more detail, Rhodonite is clearly standing out as less accurate. Do you have the same ACES tone mapping applied that the rest of the renderers do? Also, it looks like your roughness is not consistent with the spec. 

Also, I'm getting a lot of strange spew from rhodonite when I build it, like:
```
@google/model-viewer-render-fidelity-tools: 61962: 936918E3,2847714899,3736837829,1202900863,817233897,3183342108,3401237130,1404277552,615818150,3134207493,3453421203,1423857449,601450431,3009837614,3294710456,1567103746,711928724,3020668471,3272380065,1510334235,755167117],C=y?new Uint32Array(ca):ca;if(aa.Uint8Array!==r)try{eval("String.fromCharCode.apply(null, new Uint8Array([0]));")}catch(ea){String.fromCharCode.apply=function(a){return function(b,c){return a.call(String.fromCharCode,b,Array.prototype.slice.call(c))}}(String.fromCharCode.apply)};function D(a){var b=a.length,c=0,d=Number.POSITIVE_INFINITY,f,h,k,e,g,m,p,s,q,x;for(s=0;s<b;++s)a[s]>c&&(c=a[s]),a[s]<d&&(d=a[s]);f=1<<c;h=new (y?Uint32Array:Array)(f);k=1;e=0;for(g=2;k<=c;){for(s=0;s<b;++s)if(a[s]===k){m=0;p=e;for(q=0;q<k;++q)m=m<<1|p&1,p>>=1;x=k<<16|s;for(q=m;q<f;q+=g)h[q]=x;++e}++k;e<<=1;g<<=1}return[h,c,d]};var F=[],G;for(G=0;288>G;G++)switch(!0){case 143>=G:F.push([G+48,8]);break;case 255>=G:F.push([G-144+400,9]);break;case 279>=G:F.push([G-256+0,7]);break;case 287>=G:F.push([G-280+192,8]);break;default:l("invalid literal: "+G)}
```

Are you planning to fix these issues in the near term?

emadurandal:
@elalish Sorry. I noticed your mentions too late.
I agree with your decision to withdraw the public disclosure at this time, as I think it will be a steady and time-consuming process to resolve the issue of the accuracy of the rendering.

I may send another PR at another opportunity, but that may be a while yet.

PR #3607, opened 2022-07-28
fix karma.conf

elalish:
Hopefully solving the strange concurrency issues we've been having on BrowserStack.


elalish:
So, hostname: 'bs-local.com' makes BS work better, but causes WebXR tests to fail on Android (maybe not HTTPS?). Still getting a lot of random type-2 errors, `Chrome 100.0.4896.127 (Android 11): Executed 0 of 0 SUCCESS (0.002 secs / 0 secs)` see https://github.com/karma-runner/karma/issues/3806. Mostly on Safari (Mac and iOS), but now also sometimes on Android and even Windows. Very strange, but seems to be longstanding.

PR #3519, opened 2022-06-09
Extension to keyboard controls

clochardM33:
### Reference Issue

Fixes #3517

This relates to discussion Keyboard controls https://github.com/google/model-viewer/discussions/3484
In our specific world we are involved with technical drawings, and now with model-viewer, 3D models. The techy nature of it lends itself to wanting to view models at specific views (the orthogonal ones are the obvious examples) and wanting to be able to put models in specific orientations (although most of the users I've observed don't realise that is what they are trying to do with the mouse painfully sliding back and forth over the same spot).
It doesn't seem like an obvious need when looking at a model of a show or a tangerine as they don't have the inherent precision that a CAD dwg brings to mind.
The current keyboard controls are exactly the right sort of thing, but they are tantalisingly short of functionality.

In working this though there are some code changes in SmoothControl.ts (and a quick one in the package list that is unrealted).
These extend the keyboard controls from orbit changes to orbit, pan, step size & preset views and also remove the null hit behavior.
The removal of null hit behavior should really be conditional on an exposed state, but I didn't know how to do that.
They are contained entirely within SmoothControls.ts.

Keyboard controls are:
Arrows - orbit
Shift + Arrows - pan
Ctrl + Arrows - orbit at small step (ratio from unmodified set as a const, but could ideally be passed in)
Ctrl + Shift + Arrows - pan at small steps

Pressing Shift displays the same spot cursor as used on mouse-pan. Shift now behaves as a defacto 'about to pan' indicator. It also accidentally solved a user comment that they didn't really know where it was orbiting around.

Keys 1, 2, 3, 4 (and Numpad 1, 2, 3, 4) set theta to 0, 90, 180, 270 degrees at current phi and radius
Keys 7, 8, 9 (and Numpad 7, 8, 9) set phi to +45, 0 and -45 degrees at current theta and radius
This gives the ability to 'square up' a view which works in conjunction with the orbit keys.

The ration of small steps to big step should probably be configurable, but as above, that is something that I can't see how to get from the top level interface down to this script. 

Only experimental testing has been done. No automated test.

Live Demo
Video clips in Keyboard controls https://github.com/google/model-viewer/discussions/3484

PR #3703, opened 2022-08-10
Pan keys 1

### LINKED ISSUE #3517: Code changes re. Keyboard controls #3484
### Description

This relates to discussion Keyboard controls #3484
In working this though there are some code changes in SmoothControl.ts that you can have if you want them.
These extend the keyboard controls from orbit changes to orbit, pan, step size & preset views and also remove the null hit behavior.
The removal of null hit behavior should really be conditional on an exposed state, but I didn't know how to do that.
They are contained entirely within SmoothControls.ts. 

Keyboard controls are:
Arrows - orbit
Shift + Arrows - pan
Ctrl + Arrows - orbit at small step (ratio from unmodified set as a const, but could ideally be passed in)
Ctrl + Shift + Arrows - pan at small steps

Pressing Shift displays the same spot cursor as used on mouse-pan. Shift now behaves as a defacto 'about to pan' indicator. It also accidentally solved a user comment that they didn't really know where it was orbiting around.

Keys 1, 2, 3, 4 (and Numpad 1, 2, 3, 4) set theta to 0, 90, 180, 270 degrees at current phi and radius
Keys 7, 8, 9 (and Numpad 7, 8, 9) set phi to +45, 0 and -45 degrees at current theta and radius
This gives the ability to 'square up' a view which works in conjunction with the orbit keys.

Only experimental testing has been done. No automated test.

#### Live Demo
Video clips in  Keyboard controls #3484

### Version

- model-viewer: v1.12.0

### Browser Affected

- [ 1 ] Chrome, version: xx.x.xxxx.xx
- [ 1] Edge
- [ 1] Firefox
- [ 0] IE
- [ 0] Safari

### OS

- [ 0] Android
- [ 0] iOS
- [ 0] Linux
- [ 0] MacOS
- [ 1] Windows

### AR

- [ ] WebXR
- [ ] SceneViewer
- [ ] QuickLook

clochardM33:
Fixes #3517
Relates primarily to discussion #3484 and subsiquently #3517 and #3519 

This is an PR to add panning to the keyboard controls.
If enablePan is true and the Shift key is down the arrow keys implement panning rather than orbiting.
The step size is set by a constant in SmoothControls.ts


packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -823,23 +827,45 @@ export class SmoothControls extends EventDispatcher {
...
+
+  /**
+   * Handles the orbit and Zoom key presses
+   * Uses constants for the increment.
+   * @param event The keyboard event for the .key value
+   * @param isUserChange Is this a user initiated event
+   * @returns boolean to indicate if the key event has been handled
+   */
+   private orbitZoomKeyCodeHandler(event: KeyboardEvent, isUserChange: boolean) {
+    // releventKey as the return value is to get the condition back to the calling function as don't have pass by reference arguments in js
```
elalish, line 850: Actually, all arguments in js are passed by reference (except literals like number). How about you move the `event.preventDefault();` into these functions and avoid the return?

packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -823,23 +827,45 @@ export class SmoothControls extends EventDispatcher {
...
+    let relevantKey = true;
+    switch (event.key) {
+      case 'PageUp':
         this.userAdjustOrbit(0, 0, ZOOM_SENSITIVITY);
         break;
-      case KeyCode.PAGE_DOWN:
+      case 'PageDown':
```
elalish, line 856: I'd been meaning to update these for ages; thank you!

packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -823,23 +827,45 @@ export class SmoothControls extends EventDispatcher {
     const {isUserChange} = this;
     this.isUserChange = true;
 
-    switch (event.keyCode) {
-      case KeyCode.PAGE_UP:
+    // Handle modifier key conditions of:
+    // Shift & enablePan == PAN
+    // else == ORBIT/ZOOM
```
elalish, line 832: Honestly, this logic is equally clear from reading the code, so I think we can skip this comment. Self-documenting code is ideal!

elalish (changes requested):
Looking good! Thanks for splitting things up. Also, did you run clang-format on this? We have a .clang-format file that should help keep the style consistent.

clochardM33:
Ooh no I forgot clang. It's on a different pc and it's a ground up wsl and vs code install.
Will try ASAP. Any other that I can catch at the same time?

elalish:
That and the other two requested changes should be enough to merge this.

elalish:
Also, the tests are failing; did you run `npm test` locally? Should be an easy fix.

clochardM33:
Nope, will do.

clochardM33:
> Also, the tests are failing; did you run `npm test` locally? Should be an easy fix.

Easier to say that to fix!
Help required. The fails all occur with lines such as this...
`dispatchSyntheticEvent(element, 'keydown', { keyCode: KeyCode.UP });`
Presumably the fail relates to the keyhandler no longer looking for `KeyCode.XYZ` and is going after `event.key`
If so what would be the change of syntax in the custom parameters?

elalish:
Agreed; I think you just need to update the tests accordingly to set `key` instead of `KeyCode`. You might also want to add some similar tests for pan.

elalish:
To be more specific: `dispatchSyntheticEvent(element, 'keydown', { keyCode: KeyCode.UP });` becomes `dispatchSyntheticEvent(element, 'keydown', { key: 'ArrowUp' });`, but to fix TS, you also need to change [this line](https://github.com/google/model-viewer/blob/master/packages/model-viewer/src/test/helpers.ts#L55) to `key?: string`.

clochardM33:
Great. I'll get on it. Still suffering from holiday / WiFi / time this
week.
Thanks.

clochardM33:
Today's question.
Key events and helper.ts updated, but how do I get past the two cases of:
`import {ChangeSource, KeyCode, SmoothControls} from '../../three-components/SmoothControls.js';`
without putting a placeholder reference to a KeyCode somewhere?
If the line is commented out the error panel lights up with missing Smooth Controls references.
If it is left in with some null tests to make a reference to a KeyCode then it ends up with 6 skipped tests rather than 6 failed tests (which is some kind of progress).
P.S. Working from a carpark is not conducive to creative thinking so I'm all up for being spoon fed.
Thanks.

elalish:
You should be able to update *all* usages of `KeyCode` to `key` and then you won't need to import it at all. If you're still having trouble, just push your latest and I'll take a look.

clochardM33:
Progress. The snag was that I didn't spot that the arguments in `import {ChangeSource, ...` was just a list of arguments.
Next question. How do you pass a the shift key in to the keydown event so that the pan function can be tested?
I have dropped a place holder test in, but it is currently just a copy of the orbit test.
Next commit coming up asap.
Thanks.

clochardM33 (commented):
Revision of test to swap KeyCode to key and all the related tidyups.
But how do you pass a modifier key (ie Shift) into the keydown event?

elalish:
From [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/KeyboardEvent) it looks like `{ key: 'ArrowUp', shiftKey: true }` should do the trick.

elalish:
And remember to pull master and fix the conflicts :)

clochardM33:
> And remember to pull master and fix the conflicts :)

This is just another thing on the don't-know-how-to-do-that list. 
Typed git pull origin master after updating my master branch on fork and all hell broke loose.



> From [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/KeyboardEvent) it looks like `{ key: 'ArrowUp', shiftKey: true }` should do the trick.

Perfect. That is in, but I can't make any sense of the expect condtion. Pls see commit.

clochardM33 (commented):
Added Pan key test, but can't make sense of the expect logic.
Using logging the numeric values are the same as the logic conditions, but the test fails saying the numbers are not different, but they are! Please sprinkle magic understanding powder on it.

elalish (approved):
Okay, there were some tricky merge conflicts, so I went ahead and took care those (they were my fault anyway). You test was failing because it had in fact unearthed an existing bug in `getTarget()`, so that's now fixed too. No problems in your code. Thanks for sticking with this!

clochardM33:
Thank you so much for finishing this off. The last hurdle was way out of my league.
My initial plan was to have a test for each individual key. If needed I can add them, if not then job done.

elalish:
Thanks! Always welcome, but honestly beyond our current level of testing 😄

PR #3706, opened 2022-08-15
Disable non-conforming path tracers

elalish:
FYI @emadurandal and @bsdorra, I'm disabling Rhodonite and dppbr-pt in our render fidelity tests for now. The code is still there, so it'll be easy to reenable them if anyone wants to. I'm removing Rhodonite primarily because of my [comments](https://github.com/google/model-viewer/pull/3513#issuecomment-1206673414) which have yet to be acknowledged or fixed. I'm removing dspbr-pt because we now have the Stellar renders with is a more official Dassault product, and running dspbr-pt was prohibitively slow. 

In both cases many of the tests were disabled, indicating these renderers are not yet generally glTF-compliant. I'd prefer to keep our official results to those renderers that either support the whole spec, or are actively working toward that goal. Please open a PR if you'd like to address these issues and get back on our public page. 

Thank you for your contributions! It's fantastic to have more web renderers, but I rely on owners to maintain these in our repo, as I'm limited in time. Thanks @bsdorra especially for demonstrating a three.js-based path tracer; it seems that several projects are now following in your footsteps, which is great to see for our ecosystem!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -156,9 +156,9 @@
       "dev": true
     },
     "node_modules/@humanwhocodes/config-array": {
-      "version": "0.9.5",
-      "resolved": "https://registry.npmjs.org/@humanwhocodes/config-array/-/config-array-0.9.5.tgz",
-      "integrity": "sha512-ObyMyWxZiCu/yTisA7uzx81s40xR2fD5Cg/2Kq7G02ajkNubJf6BopgDTmDyc3U7sXpNKM8cYOw7s7Tyr+DnCw==",
+      "version": "0.10.4",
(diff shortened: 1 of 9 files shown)


emadurandal:
@elalish My library is still in its infancy, and I felt it was premature to include it in this project.
I think it is unavoidable to withdraw my library from the project.

Since I am working alone as an amateur activity, my know-how and development resources are limited, which is a problem.

Nevertheless, I will continue my efforts, so please let me send you another PR when the time comes.

elalish:
@emadurandal Absolutely! I completely understand - best of luck in your efforts.

PR #3718, opened 2022-08-19
add model-viewer.d.ts file

### LINKED ISSUE #3707: switch module to lib
Switch our package.json module to point to lib instead of dist. For JS users, main will still point to our bundle in dist, but this change will send TS users to our lib so they can build and tree-shake it themselves, which is best practice. Also, our lib already contains all our *.d.ts files.

See #3460

elalish:
Fixes #3707 

It looks like a bunch of changes in rollup.config.js, but it's just formatting. I just added a new build step to create the model-viewer.d.ts file. It does contain all of our internal API rather than just what's public, but the symbols should make clear to people when they are leaving the stable API. I've also updated the `main` and `module` paths in our package.json, which is always slightly terrifying since it's unclear exactly how they are used by all the tools. Hopefully this works better.


elalish:
@e111077 would you mind taking a look at this, since I'm trying to implement what we talked about? Particularly the package.json stuff.

elalish:
In particular, I'm basically reverting this: https://github.com/google/model-viewer/pull/1470, which was done ideally to help with React, but I also didn't really know what I was doing.

e111077:
Will tackle this Monday!

e111077:
switching back to `lib/...` for `module:` looks like it should handle anyone doing `import {...} from 'model-viewer'`.

The dts package looks like it _should_ add TS compat for people accessing `model-viewer/dist/*`.

This looks okay to me, but would it be possible to do an NPM canary release. This stuff is hard to 100% check without that.

PR #3774, opened 2022-09-03
Added typings in package.json

subhankar-trisetra:
This PR fixes #3773 .

PR #3791, opened 2022-09-07
add types to package.json

elalish:
@e111077 Does this look right to you? We can do an `rc2` to test it (now that I know how to not screw it up). Note that #3774 also added `typings`, which seems to be a bit like `types`?

### DIFF
--- a/packages/model-viewer/package.json
+++ b/packages/model-viewer/package.json
@@ -35,6 +35,8 @@
     "dist/model-viewer.d.ts"
   ],
   "typings": "dist/model-viewer.d.ts",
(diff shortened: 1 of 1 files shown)


e111077:
Can we do an RC2 to be safe?

elalish:
The `"type": "module"` change is breaking our karma testing, so we'll have to sort that out first, but yes, RC2 sounds good.

e111077:
we can move type: module out of this PR and TODO that for later

e111077 (approved):

e111077 (approved):

PR #3835, opened 2022-09-27
fix adjacent visibility

elalish:
The IntersectionObserver by default logs elements that are adjacent to the viewport (zero intersection) as intersecting: https://github.com/w3c/IntersectionObserver/issues/134#issuecomment-224365108

This is a problem for some carousels which have the element in view having the full width: it would cause the invisible neighboring carousel elements to render and steal frame budget. Just a one-line fix, but added tests to catch this subtlety.


samaneh-kazemi (approved):
Thank you for the quick fix

diegoteran (approved):

PR #3844, opened 2022-09-28
Hotspot data query API

jukibom:
### Reference Issue

Discussion at https://github.com/google/model-viewer/discussions/3833

This is a first pass to add the ability to query a read-only snapshot of in-scene information about a hotspot by name, most crucially the screen space (canvas) position. 

Entirely open to suggestions about the API (in particular it does feel odd that the returned position Vector2D is `u` and `v`, though it makes sense where it's currently only used for UV lookups) and how on earth to test a screen space position given that it relies on so many model-viewer defaults...

The PR includes a working example to aid discussion, which may or may not be complete overkill for this 😅

### DIFF
--- a/packages/model-viewer/src/features/annotation.ts
+++ b/packages/model-viewer/src/features/annotation.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import {Matrix3, Matrix4, Vector2, Vector3} from 'three';
+import {Matrix3, Matrix4, Vector3} from 'three';
 
 import ModelViewerElementBase, {$needsRender, $scene, $tick, toVector2D, toVector3D, Vector2D, Vector3D} from '../model-viewer-base.js';
 import {Hotspot, HotspotConfiguration} from '../three-components/Hotspot.js';
@@ -32,7 +32,7 @@ const worldToModelNormal = new Matrix3();
 export declare type HotspotData = {
   readonly position: Vector3D;
   readonly normal: Vector3D;
-  readonly screenPosition: Vector2D;
+  readonly screenPosition: Vector3D;
   readonly facingCamera: boolean;
 }
 
@@ -159,13 +159,14 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
       vector.x = (vector.x * widthHalf) + widthHalf;
       vector.y = -(vector.y * heightHalf) + heightHalf;
 
-      const screenPosition = toVector2D(
-        new Vector2(
+      const screenPosition = toVector3D(
+        new Vector3(
           vector.x,
-          vector.y
+          vector.y,
+          vector.z
         ));
 
-      if (!Number.isFinite(screenPosition.u) || !Number.isFinite(screenPosition.v)) {
+      if (!Number.isFinite(screenPosition.x) || !Number.isFinite(screenPosition.y)) {
         return null;
       }
 
--- a/packages/model-viewer/src/features/annotation.ts
+++ b/packages/model-viewer/src/features/annotation.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import {Matrix3, Matrix4} from 'three';
+import {Matrix3, Matrix4, Vector2, Vector3} from 'three';
 
 import ModelViewerElementBase, {$needsRender, $scene, $tick, toVector2D, toVector3D, Vector2D, Vector3D} from '../model-viewer-base.js';
 import {Hotspot, HotspotConfiguration} from '../three-components/Hotspot.js';
@@ -29,8 +29,16 @@ const $removeHotspot = Symbol('removeHotspot');
 const worldToModel = new Matrix4();
 const worldToModelNormal = new Matrix3();
 
+export declare type HotspotData = {
+  readonly position: Vector3D;
+  readonly normal: Vector3D;
+  readonly screenPosition: Vector2D;
+  readonly facingCamera: boolean;
+}
+
 export declare interface AnnotationInterface {
   updateHotspot(config: HotspotConfiguration): void;
+  queryHotspot(name: string): HotspotData | null;
   positionAndNormalFromPoint(pixelX: number, pixelY: number):
       {position: Vector3D, normal: Vector3D, uv: Vector2D|null}|null
 }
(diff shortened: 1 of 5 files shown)


packages/model-viewer/src/features/annotation.ts
```diff
@@ -124,6 +132,46 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+      const vector = new Vector3();
+
+      vector.setFromMatrixPosition(hotspot.matrixWorld);
+      vector.project(camera);
+
+      const widthHalf = scene.width / 2;
+      const heightHalf = scene.height / 2;
+
+      vector.x = (vector.x * widthHalf) + widthHalf;
+      vector.y = -(vector.y * heightHalf) + heightHalf;
```
elalish, line 160: I wonder if we could pull this info out of the `CSS2DRenderer` instead of duplicating the math here? I can't remember if that all gets hidden by a closure or not.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -124,6 +132,46 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+
+      const screenPosition = toVector2D(
```
elalish, line 162: Regarding U,V - would it be better to return a `Vector3D` instead? It would be an OpenGL-style screen-space depth which most users would probably ignore, but it might still be useful for something.

packages/model-viewer/src/features/annotation.ts
```diff
@@ -124,6 +132,46 @@ export const AnnotationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+        new Vector2(
+          vector.x,
+          vector.y
+        ));
+
+      if (!Number.isFinite(screenPosition.u) || !Number.isFinite(screenPosition.v)) {
+        return null;
+      }
+
+      return {position, normal, screenPosition, facingCamera};
```
elalish, line 172: I like the API :+1:

packages/model-viewer/src/test/features/annotation-spec.ts
```diff
@@ -96,6 +96,22 @@ suite('Annotation', () => {
...
+      expect(hotspot?.normal.toString()).to.equal(toVector3D(new Vector3(0, 0, -1)).toString());
+      expect(hotspot?.facingCamera).to.be.true;
+
+      // Testing the screen space coordinate is contingent on lots of default values
+      // (canvas size, camera position and camera fov).
+      // Just test to see that it's valid.
+      expect(hotspot?.screenPosition?.u).to.satisfy(Number.isFinite);
+      expect(hotspot?.screenPosition?.v).to.satisfy(Number.isFinite);
+      expect(hotspot?.screenPosition?.u).to.satisfy(Number.isFinite);
+      expect(hotspot?.screenPosition?.v).to.satisfy(Number.isFinite);
```
elalish, line 112: duplicate?

packages/model-viewer/src/test/features/annotation-spec.ts
```diff
@@ -96,6 +96,22 @@ suite('Annotation', () => {
...
+    test('querying it returns valid data', () => {
+      const hotspot = element.queryHotspot('hotspot-1');
+
+      expect(hotspot?.position.toString()).to.equal(toVector3D(new Vector3(1, 1, 1)).toString());
+      expect(hotspot?.normal.toString()).to.equal(toVector3D(new Vector3(0, 0, -1)).toString());
+      expect(hotspot?.facingCamera).to.be.true;
+
+      // Testing the screen space coordinate is contingent on lots of default values
+      // (canvas size, camera position and camera fov).
+      // Just test to see that it's valid.
```
elalish, line 108: How about a simple symmetric case? If you use a cube centered at the origin and set `phi = 90deg`, then a hotspot at the origin should show up in the center of the viewport, right?

packages/modelviewer.dev/examples/annotations/index.html
```diff
@@ -320,6 +320,218 @@ <h4 id="intro"><span class="font-medium">Show Dimensions. </span></h4>
...
+      <div id="overlays" class="demo"></div>
+      <div class="content">
+        <div class="wrapper">
+          <h4 id="intro"><span class="font-medium">Custom Overlays. </span></h4>
+          <p>
+            Hotspots can be queried to build custom overlays in screen space.
+          </p>
+          <p>
+            For example, by using an SVG canvas and updating the line nodes we can draw rudimentary dimensions lines.
+            Note that this kind of extended functionality cannot support occlusion in the 3d scene as it is drawn on top.
```
elalish, line 333: Love the example, thanks!

elalish (changes requested):
Thank you!

packages/model-viewer/src/features/annotation.ts
jukibom: Yep, that seems reasonable! The depth is a bit of a weird number presented raw because I'm pretty sure it's some sort of logarithmic scale (so it's almost always ~0.98) but it could still be used to sort hotspots to affect overlay draw order or something

packages/model-viewer/src/test/features/annotation-spec.ts
jukibom: Huh, yep! 🙃

packages/model-viewer/src/test/features/annotation-spec.ts
jukibom: Ah, yeah that works! Done

packages/model-viewer/src/features/annotation.ts
jukibom: Yeah I wondered this too but it seems the class [does the work and applies the transform css directly, otherwise discarding the result](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/renderers/CSS2DRenderer.js#L112-L154) (Unless I'm missing something)

elalish (approved):
Looks great, thanks!

packages/model-viewer/src/features/annotation.ts
elalish: Yeah, extracting it from the CSS sounds worse that computing it. No problem.

PR #3886, opened 2022-10-17
Video texture

elalish:
Adds support for video, canvas, and lottie textures via `createVideoTexture`, `createCanvasTexture`, and `createLottieTexture`. The elements are available under `texture.source.element` along with an `update()` method for the canvas textures. The lottie library lazy loads (it's large), but requires the user to set an importmap, which unfortunately can't be done automatically:
```
  <script type="importmap">
    {
      "imports": {
        "three": "path/to/model-viewer.min.js"
      }
    }
  </script>
```


eduardconstantin:
hello, when will this be available?

elalish:
@eduardconstantin This'll be in the next release, sometime in January.

PR #3895, opened 2022-10-21
try updating karma

elalish:
hopefully this version wasn't actually the browserstack problem...


elalish:
Turns out Karma v6 breaks BrowserStack pretty badly - making it wildly unreliable.

Issue #213, opened 2018-11-15
Add post-processing capabilities to renderer


cdata:
We are planning to base it on https://github.com/vanruesc/postprocessing

As part of this change, include anti-aliasing

Issue #210, opened 2018-11-15
Add bloom post-processing effect


cdata:

jcguarinpenaranda:
Hello, I've seen this has been opened for a while. May I help with something?

elalish:
@jcguarinpenaranda Absolutely! Contributions are very welcome, and this should be reasonably doable given that three.js already supports it: https://threejs.org/examples/webgl_postprocessing_unreal_bloom.html

We'll want it to be off by default since it has render cost. What kind of parameterization are you considering?

ti-hardikshah:
Is there any example of how can we use bloom with the model viewer?

elalish:
No, but you can fork us and add Bloom in our three.js code using their examples. If you'd like to contribute it back when you get it working, that would be much appreciated.

AhrenFullStop:
nobody up for the task? I might have a stab

elalish:
@AhrenFullStop By all means! Happy to review. I should really expose a rendering performance metric so we can track how expensive these these features are. I don't have a good sense for it at the moment. As long as it  defaults to off it should be fine though.

braedenf:
This would be an amazing feature to integrate. I was thinking of trying something like this example - https://p3d.in/jCj24

Beilinson:
hi, will this be implemented any time soon?

elalish:
Not on my roadmap, but PRs welcome!

Beilinson:
> Not on my roadmap, but PRs welcome!

It should be possible by simply replacing the [`threeRender.render(scene, scene.camera)` line in the `Renderer`](https://github.com/google/model-viewer/blob/61435b8b2aaf160b92317b8d27e00cf8e86eefb3/packages/model-viewer/src/three-components/Renderer.ts#L494) with `composer.render()` from this [example](https://threejs.org/docs/#manual/en/introduction/How-to-use-post-processing), right? Or is there something else I'm missing in the rendering stack?

Issue #3796, opened 2022-09-10
allow three access


makc:
for the case when the page using model-viewer needs three to do some other job, since it is already bundled with model-viewer - it would be a shame to have to download three.js twice.

I see somebody have [thought about it before](https://github.com/google/model-viewer/commit/4f6bae88be94fffb722234713614036962ef819a#diff-fcd9619b063d76c34a55c5211623b5ad7490d00acf5a86cd40ad5c6836cd4111R28) but having official support for this would be nice.

elalish:
Agreed! I just need to find the proper way to expose it and ensure the bundler doesn't tree-shake anything out. For this reason, I think we'll make a separate bundle for this, like `model-viewer-three.min.js`.

elalish:
@makc Have you looked at #3948? The proposal is to distribute an additional mv package without three.js at all - then you would pull that in with an import map along with whatever other code you need it for. Does that look like a workable solution for you?

makc:
@elalish that does sound reasonable, but then would you not have false bug reports in mv if the user imports incompatible three revision?

elalish:
Yeah, but at least we aren't claiming to validate versions like with npm. I prefer no handrail to a handrail you can't trust. We'll just have to put it in the docs and remind people to post their versions in the issue template.

elalish:
This happened as part of [v3.2](https://github.com/google/model-viewer/releases/tag/v3.2.0), try model-viewer-module.min.js, and follow the import map instructions at https://modelviewer.dev/examples/postprocessing/#setup

PR #3948, opened 2022-11-18
Post processing

Beilinson:
Deals with #213, #210

I started working with the inbuilt three.js post-processing pipeline, but switched to the [pmndrs/postprocessing](https://github.com/pmndrs/postprocessing) library instead after seeing that it has a [significant performance increase over the three.js](https://pmndrs.github.io/postprocessing/public/demo/#performance), and seems to have a better developed API overall.

I however have come upon some issues for the implementation for the model-viewer (note, I use bloom-effect here in reference to any possible post-processing effect, as thats what I tested with)
1. Due to the ThreeRenderer being a singleton shared across all modelviewer elements, enabling/disabling an effect will affect it globally for all modelviewer elements. This behavior isn't desirable in my opinion, but it's up for discussion
2. Due to 1, the current implementation where the effect is able to be activated or deactivated through adding the attribute to the DOM element ,i.e, `<model-viewer ... bloom-effect></model-viewer`. then we have the following issue: 
    If multiple modelviewer elements exist in the same window, the last one in the DOM decides whether or not bloom is activated:
    - if it doesn't have the bloom-effect attribute, then it disables the effect globally, regardless of if previous elements declared the bloom-effect attribute
    - same issue but vice versa, it may have it declared despite a previous element not wanting the bloom effect
    As such, my current implementation only allows activating the effects, while disabling them has to be done manually through code. This isn't perfect though, would love some input on a better solution
  4. Specifically the BloomEffect has a well-known issue where adding bloom on a scene with a transparent background when the background has a high lightness value (i.e white) [causes an unsightly effect](https://github.com/mrdoob/three.js/issues/14104). ThreeJS specifically solves the issue by just making the background black with their bloom effect, while the postprocessing library I used doesn't affect the background which is technically better, but still no real solution.


Beilinson:
An idea I considered for solving 1 and 2 is using the `preRender()` method in the `Renderer` to enable/disable effects on each render pass, as its called for every for every scene during the `render()` method, and hopefully there won't be a performance hit (from my understanding there shouldn't be)

packages/model-viewer/tsconfig.json
```diff
@@ -10,8 +10,8 @@
     "sourceMap": true,
     "inlineSources": true,
     "strict": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
```
elalish, line 14: Let's set these back.

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -270,26 +270,30 @@ <h2 class="demo-title">Change Material Metalness and Roughness Factors</h2>
...
+  //   material.pbrMetallicRoughness.setMetallicFactor(event.target.value);
+  //   metalnessDisplay.textContent = event.target.value;
+  // });
+
+  // document.querySelector('#roughness').addEventListener('input', (event) => {
+  //   //material.pbrMetallicRoughness.setRoughnessFactor(event.target.value);
+  //   // material.setEmissiveFactor([event.target.value, 0, event.target.value]);
+  //   //material.pbrMetallicRoughness.setRoughnessFactor(event.target.value);
+  //   roughnessDisplay.textContent = event.target.value * 5;
+  // });
```
elalish, line 296: Perhaps copy this into a new example instead? You'll need to add an entry in examples.json.

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -141,21 +145,28 @@ export class Renderer extends EventDispatcher {
       this.threeRenderer = new WebGLRenderer({
         canvas: this.canvas3D,
         alpha: true,
-        antialias: true,
+        antialias: false,
```
elalish, line 148: Is this necessary?

packages/model-viewer/src/features/post-processing.ts
```diff
@@ -0,0 +1,46 @@
...
+  vignetteEffect: boolean;
+  ssaoEffect: boolean;
+}
+
+export const PostProcessingMixin = <T extends Constructor<ModelViewerElementBase>>(
+    ModelViewerElement: T): Constructor<PostProcessingInterface>&T => {
+  class PostProcessingModelViewerElement extends ModelViewerElement {
+    @property({type: Boolean, attribute: 'bloom-effect'})
+    bloomEffect: boolean = false;
+    @property({type: Boolean, attribute: 'vignette-effect'})
```
elalish, line 33: Remind me what this does? Can it not be achieved with a CSS overlay outside of MV?

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -409,6 +409,12 @@ export default class ModelViewerElementBase extends ReactiveElement {
     }
   }
 
+
+  /** @export */
+  getRenderer() { 
+    return this[$renderer];
```
elalish, line 415: unnecessary?

packages/model-viewer/src/three-components/EnvironmentScene.ts
```diff
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import {BackSide, BoxGeometry, Mesh, MeshBasicMaterial, MeshStandardMaterial, PointLight, Scene} from 'three';
+import {BackSide, BoxGeometry, Mesh, MeshBasicMaterial, MeshStandardMaterial, PointLight, AmbientLight, DirectionalLight, Scene} from 'three';
```
elalish, line 16: Used?

packages/model-viewer/src/three-components/PPRenderPipeline.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const SSAO_EFFECT = new SSAOEffect();
+
+export const BLOOM_EFFECT = new BloomEffect({
+    blendFunction: BlendFunction.ADD,
+    mipmapBlur: true,
+    luminanceThreshold: 0.85,
+    luminanceSmoothing: 0.025,
+    intensity: 3,
+    kernelSize: KernelSize.LARGE
```
elalish, line 16: Do any of these parameters need to be part of the public API? I haven't used bloom much; I love things being simple, but curious how much adjustment tends to be needed.

packages/model-viewer/package.json
```diff
@@ -73,8 +73,9 @@
     "3d"
   ],
   "dependencies": {
-    "three": "^0.146.0",
-    "lit": "^2.2.3"
+    "lit": "^2.2.3",
+    "postprocessing": "^6.29.1",
```
elalish, line 77: How big is this dependency? What's the before/after on our gzipped bundle?

elalish (changes requested):
Thanks a lot for this! Very cool possibilities here, but this is significant change (from 2 dependencies to 3) so this'll require a lot of checking. Also a significant API change that we'll want to get right so we don't have to break it down the road. Have you tried the SSAO? And how many devices have you tested this PR on?

packages/model-viewer/tsconfig.json
Beilinson: Of course, only for local testing

packages/model-viewer/src/three-components/Renderer.ts
Beilinson: I saw many examples using it so I set it for the beginning, but I would like to do some tests to see if it's really needed. 
If it is then the solution is to use the default anti aliasing when no post processing is enabled, otherwise use the anti-alias pass with the post processing pipeline

packages/modelviewer.dev/examples/scenegraph/index.html
Beilinson: If this becomes part of the API then as a separate mixin it would obviously require separate examples, I just set this for minimal testing during dev

packages/model-viewer/src/features/post-processing.ts
Beilinson: Definitely, I just wanted to test some basic effects that would be easily noticeable

packages/model-viewer/src/model-viewer-base.ts
Beilinson: For personal testing

packages/model-viewer/src/three-components/PPRenderPipeline.ts
Beilinson: It's definitely something to be considered, I too want the api to be kept to a minimum simplicity, when I'll make some tests I'd love for your input on how much of the parameters to expose in the api

packages/model-viewer/package.json
Beilinson: I'll check and get back to you on that

Beilinson:
I have checked the SSAO it's very subtle and not a dramatic visual impact (which is good imo). So far I've tested only on Mac (firefox/chrome). I'll test on windows and linux, also edge, but there's nothing that shouldn't work because all of the post processing effects just use the inbuilt three renderer to do some image manipulation, in fact the implementation of most effects is quite similar to the three inbuilt postprocessing

packages/model-viewer/src/features/post-processing.ts
```diff
@@ -0,0 +1,46 @@
...
+export declare interface PostProcessingInterface {
+  bloomEffect: boolean;
+  vignetteEffect: boolean;
+  ssaoEffect: boolean;
+}
+
+export const PostProcessingMixin = <T extends Constructor<ModelViewerElementBase>>(
+    ModelViewerElement: T): Constructor<PostProcessingInterface>&T => {
+  class PostProcessingModelViewerElement extends ModelViewerElement {
+    @property({type: Boolean, attribute: 'bloom-effect'})
```
elalish, line 31: Starting a thread here for the bloom on a transparent background issue. That's a long thread you linked to: can you summarize and maybe provide a few images? How bad and what are the mitigations?

elalish:
Please show off some before/after visuals of these various effects; that will help convince people to use them! Also, how is it on mobile? How significant is the performance reduction?

packages/model-viewer/src/three-components/Renderer.ts
elalish: You can't enable/disable alpha on the fly, nor per MV element, so I hope we can keep this `true`.

Beilinson:
> Please show off some before/after visuals of these various effects; that will help convince people to use them! Also, how is it on mobile? How significant is the performance reduction?

I haven't checked on mobile, nor have I checked AR as I'm not sure how I can do that with a local-dev server. Testing on my old Macbook Air 2013 (1.3 Dual-Core i5, 4gb DDR3, Intel HD 5000 1.5gb vram) I see little to no performance impact from Bloom and SSAO combined.

elalish:
When I do local testing on my mac, I just do `npm run serve` and then go to the 192.168... address it serves to on my iphone, which works as long as both devices are on the same wifi. For Android, you can also just plug in via USB, and open chrome://inspect to set up port forwarding.

Beilinson:
Ok sounds great, will test this weekend and upload some screenshots

elalish:
Where are we with this? I think I'm waiting for some screenshots and some details about transparent backgrounds. No rush, just making sure no one is waiting to me?

Beilinson:
Uni is a bit busy right now, I'll start working on this again probably next week. Haven't sent screenshots coz I had some trouble getting the project to run on my vmware ubuntu and I don't have access to the macbook anymore

elalish:
No problem. If you're on windows, we have directions for setup on WSL2: https://github.com/google/model-viewer#windows-1011-setup if that helps.

Beilinson:
Planning on returning to this in a couple days, any updates that I should be aware of? Haven't been able to get it running on the Ubuntu VM so I'll try the WSL route. 
Had some other ideas on keeping the effect passes performant with many viewers, not that I noticed too many issues before but just in case. 
If you've had any thoughts about the API in general I'd really appreciate any input!

packages/model-viewer/package.json
elalish: This is the main thing I need an answer on - if this has significant size, then it should be optionally, lazily loaded like `lottie-web`. Also note that `three` has been switched to a peer dependency.

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,149 @@
...
+const ssao = modelViewer.querySelector('#ssao');
+
+function toggleEffect(checkbox, property) {
+    modelViewerPostprocessing[property] = checkbox.checked;
+}
+
+bloom.addEventListener('change', () => toggleEffect(bloom, 'bloomEffect'));
+pixelate.addEventListener('change', () => toggleEffect(pixelate, 'bloomEffect'));
+outline.addEventListener('change', () => toggleEffect(outline, 'bloomEffect'));
+ssao.addEventListener('change', () => toggleEffect(ssao, 'bloomEffect'));
```
elalish, line 111: Update property names?

packages/model-viewer/package.json
elalish: According to [bundlephobia](https://bundlephobia.com/package/postprocessing@6.30.1) it's twice the size of three.js (presumably three.js is part of this bundled value). This means we definitely need a way for it to be optionally loaded. Figuring that out will be the main API-level design challenge.

blumelume: I am actually seeing very different numbers there. Bundlephobia says, it's pretty much exactly half of [three's](https://bundlephobia.com/package/three@0.150.1) size. Am I looking at it wrong?

donmccurdy: postprocessing doesn't bundle three.js AFAIK, but it's a big library of effects and shaders, so tree-shaking is going to be critical. If the build system here is configured for it, tree-shaking should work well.

Beilinson: image
in vs-code i see that the bundle size according to [Import Cost ](https://marketplace.visualstudio.com/items?itemName=wix.vscode-import-cost) is ~100kb, compared to ~400kb for three.js. 
Regardless, as @donmccurdy said if tree-shaking is supported I don't believe we need to import all of the possible effects and shaders, so that could definitely be brought down. Lazy loading probably will require additional checks because the postprocessing sits very much in the Renderer component with three.js, I'm not sure thats an optimal solution.

elalish: Yes, we use rollup for our builds, which tree shakes. So: can I have a before/after of our gzipped bundle size? If it's too much, then it may be better to make a `<model-viewer++>` component that depends on both. I'm open to options, but I'd like to see some numbers in order to make a decision.

packages/model-viewer/src/three-components/Renderer.ts
Beilinson: According to the pmndrs wiki:

*The default WebGL antialiasing solution has no effect in post processing pipelines because it only applies to the canvas. It's recommended to set the <code>antialias</code> flag to <code>false</code> when creating a <code>WebGLRenderer</code> for a post processing setup.*

So the answer is we can keep it enabled, but it would be a possible performance benefit if it could be disabled when postprocessing is running. Since there is no way to dynamically enable/disable it, the solution is to keep it on permanently.

packages/model-viewer/package.json
elalish: @blumelume Ah yes, I read it wrong, so only 2/3 the size (gzipped). Still, that's far too big, so let's see where tree-shaking gets us. Our bundle is 210kb all up (gzipped), though I just realized that bundlephobia now shows us as 55kb due to three.js now being a peer dependency. 🙄 I would say bloating that by 10% (20kb) would definitely be too much unless we had a lot of users of this feature.

Beilinson: image

<html>
<body>


build/postprocessing.esm.js | 102.33KB | 11.5%
-- | -- | --



</body>
</html>

note that I am still on the older version (@2.1.1) as I haven't merged the latest changes from master which make three into a peer-dependency, so the size without the postprocessing is ~800kb).

Beilinson: postprocessing is tree-shakeable according to bundlephobia

elalish: Thank you @Beilinson! Okay, so between this increase and the antialiasing difference above, this feels like it should be somehow optional, since this won't be used by most of our users (and certainly there is performance impact, particularly on mobile). I'm not sure the best path forward - a simple option from my perspective is you can simply fork our repo into `<model-viewer-postprocessing>` (or something) and make this advanced version. I'll be happy to point advanced users your way. If we can find a decent way to make this optional in the repo here, that would likely be preferable, but I'm not quite sure how. @donmccurdy any thoughts here?

elalish: I do like the idea of child components, and I'm considering a similar pattern for a multi-model API. @donmccurdy Can you elaborate on why bundle-splitting and CDN distribution is unappealing (I'm a bit ignorant in this domain)? 

@Beilinson Does this sound like a viable path forward to you?

donmccurdy: By bundle-splitting, I mean having your build process create multiple JavaScript files for a single npm package, and dynamically importing parts as needed. Great idea in theory but in practice there are a lot of obstacles with build tools, CDN caching, etc. Works well for *application* authors, but I haven't seen library authors doing this successfully before, and wouldn't be excited to learn why the hard way.

Unless the tooling situation improves and there are good examples to follow, I think it's better and safer to just have discrete packages, or discrete exports, and let the user explicitly import what they need. I think the child component approach would fit well with that design.

Beilinson: I also like the idea of composition, especially as it solves the issues of new effects and maintenance as well as allows for simple definition of how to expose the effect api, each effect is just a web component and as much as the entire settings interface could be exposed.

I am wondering if its already doable using the <code>externalRenderer</code> on the ModelScene, as that is already called using the internal render loop in the <code><Renderer></code>.  From my testing running both render functions of the effectComposer and the threeRenderer still works, though its a shame from a performance standpoint to run them both. 

I can create a quick mockup using a similar idea to the externalRenderer, i.e <code><postprocessingRenderer></code> or something along those lines, so that the user created externalRenderer will not be overridden by the postprocessing, as well as a basic web component effect or two. 

@donmccurdy @elalish Do you think its best to keep the <code><model-viewer-effects></code> as a separate package then, or just be additional exports by <code><model-viewer></code>? With the second option I believe tree-shaking would work only if the package was imported using JS, with regular html script tag it would probably import everything and that brings us back to square one.

donmccurdy: I would not use additional ESM exports, for the reason you mention. Using additional [package.json#exports entrypoints](https://nodejs.org/api/packages.html#package-entry-points) would be the way to go if @elalish would like for this to be part of the same npm package, so there are multiple scripts on the CDN, but just one npm package. Otherwise effects could be a separate package, either within this repository or elsewhere.

elalish: I would avoid `externalRenderer` - it voids much of the useful code we have for managing multiple MV elements performantly and such. That's more for something like cloud streaming. 

We may simply need to expose our internal `Renderer` class (or part of it) to a post-processing web component. And we may want to rely on some static configuration to take care of e.g. setting antialiasing (since it needs to happen before the `Renderer` is created). 

Another advantage to making a more generic plug-in model is we may also be able to incorporate effects like these: https://github.com/0beqz/realism-effects

thibaud-be: this might be of interest https://webgi.xyz/docs/intro
a threejs based solution with a quite exhaustive post-processing pipeline the editor is worth cheking out

Beilinson: image

I just tested the basic composition idea, it works relatively well out of the gate, will need some work but its definitely the way to go.

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,140 @@
...
+  </script>
+
+  <!-- Documentation-specific dependencies: -->
+  <script type="module"
+      src="../built/dependencies.js">
+  </script>
+
+  <!-- Loads <model-viewer> on modern browsers: -->
+  <script type="module"
+      src="../../node_modules/@google/model-viewer/dist/model-viewer.js">
```
elalish, line 137: This example code is looking great from an API perspective. The next thing will be to import the library defining `<mv-effects-composer>` as its own `<script>` that is loaded in addition to `model-viewer.js`. That'll require another output from our rollup config. That'll be the key in making sure this is all feasible for us to ship. Do you feel unblocked there?

donmccurdy:
Naming notes — The three.js class is THREE.EffectComposer, so we may want to use a singular `*-effect-composer` rather than plural `*-effects-composer` here. I'm not sure what is considered best practice on prefixing web components. Does the `mv` prefix make sense or would it be better to match `<model-viewer>` more closely, with something like `<model-effect-composer />`?

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: I tried getting the separate library to link with packages/modelviewer.dev with rollup, it doesn't seem to be working yet so I'd like some help with that. Will make a push with the new library this week

Beilinson:
I ag

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -66,6 +66,7 @@ export class ModelScene extends Scene {
   public element: ModelViewerElement;
   public canvas: HTMLCanvasElement;
   public annotationRenderer = new CSS2DRenderer();
+  public effectsRenderer: EffectsRendererInterface|null = null;
```
Beilinson, line 69: Starting a thread here for performance and testing:

OS's tested: Windows, WSL2/vmware, macOS, IOS | Effects work identically on all systems
Devices: 
- Mac Air 2013 (1.3 ghz dual core i5, intel HD Graphics 5000, 4gb ram)
- Gigabyte G5 (i5-10500H, intel UHD 630, 16gb ram)
- iPhone 11

Browsers: Chrome, Firefox, Safari | Effects work identically on all browsers

I haven't noticed a significant performance reduction with any of the above device/browser configurations.

packages/model-viewer/src/three-components/PPRenderPipeline.ts
Beilinson: You can now take a look at all the effects I've implemented (mostly those I think would be useful, i.e outline for product showcases, bloom/ssao/color-grade/smaa for visual fidelity, glitch and pixelate for more niche users). I haven't exposed every parameter in most of the effects, only those I think would be useful to edit.

packages/model-viewer/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@google/model-viewer",
-  "version": "2.1.1",
+  "name": "@beilinson/model-viewer",
```
Beilinson, line 2: temporary, so I can test with a published version. Change back before merged.

donmccurdy:
I'll defer to @elalish on the prefix, that is more of a project branding question I guess. Either no prefix, or `<model->`, both seem clean and clear enough to me.

The `postprocessing` package uses both Effects and Passes, which are slightly different. Effects can be composed into Passes. I would probably keep a suffix, as it may be necessary to distinguish between those.

packages/model-viewer/src/model-viewer-base.ts
```diff
@@ -541,6 +570,12 @@ export default class ModelViewerElementBase extends ReactiveElement {
   }
 
   [$onModelLoad]() {
+    this.dispatchEvent(new CustomEvent('beforeRender'));
```
Beilinson, line 573: not sure this is needed, might be possible with the 'progess' event.

Beilinson:
@donmccurdy For now only effects are implemented, as I'm not sure how much I want this library to be a web-component wrapper for `postprocessing` like `react-postprocessing`, rather a useful simple library, similar to how <model-viewer> is for `three`. 

I don't see a scenario where a simple user would want to run ShaderPasses or the like, and if they would I expose an `addPass` and `addEffect` functions for users wanting to work with the postprocessing library directly.

donmccurdy:
Ok, yes — totally defer to you on scope. I do think keeping the term "effect" in the name is probably a good thing, though.

packages/modelviewer.dev/examples/postprocessing/index.html
elalish: Sorry, been a bit busy/sick. Will try to take a look next week.

Beilinson:
> WARNING: Multiple instances of Three.js being imported.

Like `<model-viewer>`, the `postprocessing` library requires a decent chunk out of three.js, and up until now I've been bundling it together with my lib as well, which has led to some interesting bugs. 

#3796 You mentioned here making your own version of the three bundle with only the exports model-viewer needs, however that may not necessarily align with what postprocessing requires. The current solution of exporting only a few of the required things directly from model-viewer and then import-mapping three as model-viewer is an unsatisfactory solution seeing as there might always be something missing, requiring either constant updating or forcing users to bring another instance of three.js anyways.

I think the proper solution would be to not bundle three together with model-viewer, and bring it through an import-map for all other scripts to share. 

Bringing the full `three.module.js` is a [whopping 1.2mb](https://www.npmjs.com/package/three?activeTab=code) however, which is a shame for anyone who genuinely just wants to use model-viewer on its own.

I propose two solutions:
1. As you mentioned in #3796, bundle a model-viewer only tree-shaken version of three, which I think should be possible automatically with rollup. All users will be required to import-map the three dependency, and theres also a good polyfill for older browsers: https://www.bram.us/2021/09/13/support-import-maps-in-any-browser-with-es-module-shims/
2. Create two versions of model-viewer dists, one with bundled three.js and one without, the second for those who want to use three additionally. 

I have tested the first solution by excluding three from the rollup options and bringing it through an import-map, it works for me.

Beilinson:
Obligatory awesome effects screenshot:
The Cosmic Web made with (VIVID)[https://github.com/GalaxyHunters/VIVID3D]

Before:
image

After:
image

elalish:
> I propose two solutions:
> 
> 1. As you mentioned in [allow three access #3796](https://github.com/google/model-viewer/issues/3796), bundle a model-viewer only tree-shaken version of three, which I think should be possible automatically with rollup. All users will be required to import-map the three dependency, and theres also a good polyfill for older browsers: https://www.bram.us/2021/09/13/support-import-maps-in-any-browser-with-es-module-shims/
> 2. Create two versions of model-viewer dists, one with bundled three.js and one without, the second for those who want to use three additionally.

Am I correct that number 2 is basically to add number 1 to our current distribution without changing the original? That sounds workable - I don't want to saddle the vast majority of our users with thinking about import maps when they aren't using these new features.

packages/model-viewer/src/features/controls.ts
```diff
@@ -786,7 +786,7 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
           Math.max(this[$scene].boundingSphere.radius, radius);
 
       const near = 0;
-      const far = 2 * maximumRadius;
+      const far = Math.abs(2 * maximumRadius);
```
elalish, line 789: How did you get a negative maximumRadius?

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -321,7 +321,6 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
     const enableTransparency =
         (material: MeshStandardMaterial, enabled: boolean): void => {
           material.transparent = enabled;
-          material.depthWrite = !enabled;
```
elalish, line 324: Is this change intentional?

packages/model-viewer-effects/test/fidelity/goldens/AlphaBlendLitmusTest/filament-golden.png
elalish, line 1: Mind removing the goldens from this PR?

packages/model-viewer/src/three-components/ModelScene.ts
elalish: Do you have a cheap old Android phone to test? That's where performance tends to make a big difference.

elalish (commented):
Thanks for digging into this - think you might be able to set up a simple Glitch so we can see how these libraries would be packaged and used in practice? I guess it would require a model-viewer-no-three.min.js, three.min.js, and post-processing.min.js and an import map?

Beilinson:
> Thanks for digging into this - think you might be able to set up a simple Glitch so we can see how these libraries would be packaged and used in practice? I guess it would require a model-viewer-no-three.min.js, three.min.js, and post-processing.min.js and an import map? 

The postprocessing examples are already using the import-map, do you have any input on how to call the no-three package bundle?

Where do you think the documentation and examples for this library should sit btw? If it's on the base modelviewer website that'd be easiest so I can use all the same templates you're using.

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,222 @@
...
+
+  <!-- Loads <model-viewer> on modern browsers: -->
+  <script type="module"
+      src="../../node_modules/@google/model-viewer/dist/model-viewer-no-three.min.js">
+  </script>
+
+  
+  <!-- Loads <model-viewer-effects> on modern browsers: -->
+  <script type="module"
+      src="../../node_modules/@beilinson/model-viewer-effects/dist/model-viewer-effects.js">
```
elalish, line 219: >The postprocessing examples are already using the import-map, do you have any input on how to call the no-three package bundle?
>Where do you think the documentation and examples for this library should sit btw? If it's on the base modelviewer website that'd be easiest so I can use all the same templates you're using.

Honestly, this example you've built is actually doing exactly what I wanted to see, and I think this is a great place for it. We'll just need to link it in as another examples page. I like this because it's clear there are three scripts coming in instead of one: `three.module.js` (in the import map), `model-viewer-no-three.min.js`, and `model-viewer-effects.js`. What are the sizes of those (gzipped & minified)?

packages/model-viewer-effects/src/effect-composer.ts
```diff
@@ -0,0 +1,321 @@
...
+export const $selection = Symbol('selection');
+export const $setSelection = Symbol('setSelection');
+export const $resetEffectPasses = Symbol('resetEffectPasses');
+export const $userEffectCount = Symbol('userEffectCount');
+
+/**
+ * Light wrapper around {@link EffectComposer} for storing the `scene` and `camera
+ * at a top level, and setting them for every {@link Pass} added.
+ */
+export class EffectRenderer extends EffectComposer {
```
elalish, line 48: I'm super impressed with the work you've put in here - this is a powerful new set of capabilities. If someone wanted to extend this with new effects, but wanted to do it outside of our repo, what directions would you give them? Two cases: also depending on the `postprocessing` lib, or using something else.

donmccurdy: I think it would be practical to require the `postprocessing` library's EffectComposer, rather than the original three.js EffectComposer. Supporting additional effects that don't ship with `postprocessing` is worth considering, though.

packages/model-viewer/src/features/scene-graph/material.ts
elalish: Yes, there was a long discussion years ago regarding the best way to render transparency and this was the result. @donmccurdy do you remember that? I wonder though how much it has been superseded by the actual glTF transparency extension which came along afterwards.

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,306 @@
...
+    {
+      "imports": {
+        "three": "https://cdn.jsdelivr.net/npm/three@0.150.1/build/three.module.js"
+      }
+    }
+  </script>
+
+  
+  <!-- Import the no-three component -->
+  <script type="module" src="../../node_modules/@google/model-viewer/dist/model-viewer-no-three.js"></script>
```
elalish, line 40: duplicate?

packages/model-viewer-effects/src/effect-composer.ts
```diff
@@ -0,0 +1,342 @@
...
+  /**
+   * A selection of all {@link Mesh}'s in the ModelScene.
+   */
+  get selection() {
+    return this[$selection];
+  }
+
+  /**
+   * Creates a new MVEffectComposer element.
+   * The EffectComposer instance is created only on connection with the dom so that
```
elalish, line 154: ...?

packages/model-viewer/src/three-components/Renderer.ts
```diff
@@ -324,7 +325,7 @@ export class Renderer extends EventDispatcher {
         scene.scaleStep = 0;
         this.rescaleCanvas(scene);
       } else {
-        return false;
+        return scene.dirtyRender; // this is usually false, unless specified otherwise
```
elalish, line 328: What about skipping this change and giving a `beforeRender()` callback for the effect that's called in `$tick`? That's how we call `queueRender` in e.g. our animationMixer. Might be helpful in case a custom effect takes input of some kind.

elalish (commented):
This is a big PR, so It's going to take a few passes before I manage to review it all... Here's a start.

packages/model-viewer/src/features/scene-graph/material.ts
donmccurdy: It would be a pretty major regression to enable depth write for all alpha blended materials I think.   There's probably a way to solve this in post-processing, like a depth pre-pass... perhaps we could leave that for a separate PR?

Beilinson: I don't think something like that exists in postprocessing, if not then this is definitely , but I guess its possible to do a depth pre-pass to writes all materials as depthWrite = true, saving that to the texture which is used by the depth-requiring effects.

packages/model-viewer/src/three-components/Renderer.ts
Beilinson: Interesting, I didn't see that. Sounds like a better solution to me

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: Here is where I'm actually bringing the module in, in the example it's just to show how it works. Bringing it in the example seems to cause some bugs sometimes where the postprocessing just doesn't work, not sure why

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,309 @@
...
+  </effect-composer>
+</model-viewer>
+<script type="module">
+import * as PostProcessing from 'https://cdn.jsdelivr.net/npm/postprocessing@6.30.1/build/postprocessing.esm.js';
+const customComposer = document.querySelector("effect-composer#customComposer");
+
+const grid = new PostProcessing.GridEffect();
+const sepia = new PostProcessing.SepiaEffect();
+const noisePass = new PostProcessing.EffectPass(undefined, grid, sepia);
+customComposer.addPass(noisePass);
```
elalish, line 264: 💯

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,309 @@
...
+
+  <div class="examples-container">
+    <div class="sample">
+      <div id="setup" class="demo"></div>
+      <div class="content">
+        <div class="wrapper">
+          <div class="heading">
+            <h2 class="demo-title">Setup PostProcessing</h2>
+            <p>The &lt;model-viewer-effects&gt; library addon for &lt;model-viewer&gt; provides
+            you with a easy-to-use postprocessing workflow for spicing up your models.<br/><br/>
```
elalish, line 91: Maybe add a list of all the included effects here?

packages/model-viewer-effects/screenshot.png
elalish, line 1: Maybe with some effects applied?

elalish (changes requested):
This is looking great, just some small nits.

packages/model-viewer-effects/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+### PostProcessing
+`<model-viewer-effects>` uses the [postprocessing](https://github.com/pmndrs/postprocessing) library under the hood, for its superior [performance](https://github.com/pmndrs/postprocessing#performance) and support.
+
+While not all effects are wrapped by this library, you can add any custom effects/passes that follow the postprocessing spec.
+
+There is no documentation yet, but you may refer to the types for all available properties.
+
+### *XR Support*
+The effects are not supported in the `<model-viewer>` XR modes, which will render as usual.
```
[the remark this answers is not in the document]
Beilinson: There's a couple threads about it on the postprocessing git:

https://github.com/pmndrs/postprocessing/issues/452

It links to another issue with a jsfiddle example of it maybe working, but I haven't tested that and I'm not sure how well it actually works. 

In general the developers answer was they would consider supporting it after their next version release which is a WIP for some time now

packages/model-viewer-effects/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+### HTML
+
+`<model-viewer-effects>` and `<model-viewer>` share a [Three.js](https://threejs.org/) dependency. In order to avoid version conflicts, you should bring Three through an `import-map`:
+
+```html
+<script type="importmap">
+  {
+    "imports": {
+      "three": "https://cdn.jsdelivr.net/npm/three@0.150.1/build/three.module.js"
```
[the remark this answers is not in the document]
Beilinson: How best to do that?

packages/modelviewer.dev/data/docs.json
```diff
@@ -1089,5 +1104,35 @@
         ]
       }
     ]
+  },
+  {
+    "Title": "PostProcessing Effects",
+    "htmlName": "postprocessing",
+    "Attributes": [
```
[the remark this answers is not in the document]
Beilinson: I would want the same docs style for the effects, plus the examples aren't listing all the parameters and how they affect the look of the final postprocessing. I was thinking it could be just another docs page that is referenced from the main docs page and from the examples, with the same structure as the main docs page.

packages/model-viewer-effects/package.json
```diff
@@ -0,0 +1,113 @@
+{
+  "name": "@beilinson/model-viewer-effects",
```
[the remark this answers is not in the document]
Beilinson: When you'll merge it of course!

elalish (changes requested):
This looks great - and you keep adding more! How close are you to feeling ready to merge? I think we just need it updated to @google first, and it looks like there's a typo making the tests fail.

packages/model-viewer-effects/package.json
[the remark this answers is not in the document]
elalish: Well, it needs to be part of the PR that we merge, so I think now is the time.

packages/model-viewer-effects/src/effects/selective-bloom.ts
```diff
@@ -0,0 +1,93 @@
...
+   * The strength of the bloom effect.
+   */
+  @property({ type: Number, attribute: 'strength', reflect: true })
+  strength = 3;
+
+  /**
+   * Value in the range of (0, 1). Pixels with a brightness above this will bloom.
+   */
+  @property({ type: Number, attribute: 'threshold', reflect: true })
+  threshold = 0.85;
```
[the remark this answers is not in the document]
Beilinson: The postprocessing library documentation is a bit weak on what exactly some of the params mean. 

From my testing it treats the pixel RGB brightness, which is clamped from 0 to 1 (a completely white pixel), and thats also what the docs recommend (to use sRGBEncoding).

Beilinson:
Not sure why the test isn't finding the renamed function, I see that I pushed the rename.

Beilinson:
On a different thread, I tested the new three@0.151.0 release since it has some features I personally requested, and they added a warning for something that the postprocessing lib does with ShaderMaterials.

The effects still work despite the warning, and I'll report it to the postprocessing issues page, just an FYI.

packages/model-viewer-effects/package.json
```diff
@@ -73,11 +73,11 @@
     "postprocessing": "^6.30.0"
   },
   "peerDependencies": {
-    "@beilinson/model-viewer": "^3.1.4",
+    "@google/model-viewer": "^3.1.0",
```
elalish, line 76: I think our latest is 3.0.2

packages/model-viewer-effects/src/effects/mixins/selective.ts
```diff
@@ -40,31 +40,32 @@ export const SelectiveMixin = <T extends Constructor<IEffectBaseMixin & Reactive
...
+      if (!this.effectComposer) return;
       this.effects.forEach((effect: ISelectionEffect) => effect.selection?.clear());
       if (this.selection?.length > 0) {
         const scene = this.effectComposer[$scene];
         scene?.traverse(
           (obj) => this.selection.includes(obj.name) && this.effects.forEach((effect: ISelectionEffect) => effect.selection?.add(obj))
         );
       } else {
-        this.effects.forEach((effect: ISelectionEffect) => effect.selection?.set(this.effectComposer[$selection].values()));
+        this.effects.forEach((effect: ISelectionEffect) => effect.selection?.set(this.effectComposer![$selection].values()));
```
elalish, line 68: Minor, but you can avoid `!` by using `const {effectComposer} = this;` at the top. If you null-check a local, TS will understand that, but if you null-check a member, it figures it might have changed since you checked it.

elalish (dismissed):
More docs would be excellent! One option would be to a make another page copying the framework of `docs/index.html` or `faq.html` specifically for this effects-composer element and everything under it. We can link it from the front page and you only need to populate the JSON data.

packages/model-viewer-effects/package.json
Beilinson: yes all tests work and are passing locally, the CI is downloading the latest release but i've already bumped the model-viewer version in this PR, do you want me to change it back for now?

Beilinson:
> More docs would be excellent! One option would be to a make another page copying the framework of `docs/index.html` or `faq.html` specifically for this effects-composer element and everything under it. We can link it from the front page and you only need to populate the JSON data.

Sounds good to me, will set it up this week :D

Beilinson:
Btw, I really don't like the naming scheme of \<model-viewer-no-three\>, it seems arbitrarily long. 

Was thinking \<model-viewer-standalone\>, though thats less descriptive and not really shorter.

Also, three released a couple patches since yesterday, the new 0.151.2 release works without errors/warnings for me.

packages/model-viewer-effects/package.json
elalish: Yes, I bump versions when I do a release - I think that may be confusing the CI.

elalish:
> Btw, I really don't like the naming scheme of <model-viewer-no-three>, it seems arbitrarily long.
> 
> Was thinking <model-viewer-module> instead

So, it'll be `<model-viewer>` either way for the tag. It's just the library file name that changes. `model-viewer.js` is already an ES6 module, so I think `model-viewer-module.js` may be confusing.

Beilinson:
> So, it'll be `<model-viewer>` either way for the tag. It's just the library file name that changes. `model-viewer.js` is already an ES6 module, so I think `model-viewer-module.js` may be confusing. 

Bump for same tag, I meant the name of the library itself. I think module is fine as it means "the model viewer module on its own", while the regular model viewer bundle is a "standalone", but it's your call at the end of the day :)

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,364 @@
...
+            <p>All Effects are reactive components, so all properties can be changed during runtime. 
+            This example demonstrates the BlendMode API, which allows you to blend effects like you would
+            layers in Photoshop. We use a &lt;color-grade-effect&gt;, which allows you to alter the look of your model.
+            </p>
+          </div>
+          <example-snippet stamp-to="effects" highlight-as="html">
+            <template>
+<model-viewer id="blendViewer" camera-controls touch-action="pan-y" ar src="../../shared-assets/models/Astronaut.glb" alt="A 3D model of an astronaut">
+  <effect-composer>
+    <color-grade-effect id="color-grade" contrast="0.5" saturation="0.4"></color-grade-effect>
```
elalish, line 157: Why is there heavy aliasing and strange transparency effects here? 
image

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,364 @@
...
+      <div class="content">
+        <div class="wrapper">
+          <div class="heading">
+            <h2 class="demo-title">Selective Effects</h2>
+            <p>Some of the implemented effects are selective and may be applied to only certain meshes in your model. 
+            Currently, this is exclusive to &lt;selective-bloom-effect&gt; and &lt;outline-effect&gt;. The default behavior for selective
+            effects is to use the default <code>selection</code> provided by the &lt;effect-composer&gt;, which contains a set of
+            all meshes in the model.<br/>
+            By providing an array of mesh names to the <code>selection</code> attribute, you may enable the effects only for those meshes.<br/>
+            This example demonstrates the application of the Bloom effect exclusively to Meshes with an emissive material:</p>
```
elalish, line 200: Are all the materials emissive? It doesn't look very selective...
image

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,364 @@
...
+    <ssao-effect blend-mode="skip"></ssao-effect>
+    <glitch-effect blend-mode="skip"></glitch-effect>
+    <smaa-effect blend-mode="skip"></smaa-effect>
+    <bloom-effect blend-mode="skip"></bloom-effect>
+    <pixelate-effect granularity="5" blend-mode="skip"></pixelate-effect>
+  </effect-composer>
+  <div class="controls">
+    <label for="outline">Outline Effect</label>
+    <input type="checkbox" id="outline">
+    <label for="ssao">SSAO Effect</label>
```
elalish, line 257: This one is giving me an exception: 
```
postprocessing.esm.js:6338 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'replace')
    at EffectPass.updateMaterial (postprocessing.esm.js:6338:48)
    at EffectPass.initialize (postprocessing.esm.js:6437:10)
    at EffectComposer.addPass (postprocessing.esm.js:7756:10)
    at EffectComposer.addPass (effect-composer.js:58:15)
    at HTMLElement.updateEffects (effect-composer.js:217:39)
    at HTMLElement.updated (blend-mode.js:58:88)
```

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,364 @@
...
+    <input type="checkbox" id="ssao">
+    <label for="glitch">Glitch Effect</label>
+    <input type="checkbox" id="glitch">
+    <label for="smaa">SMAA Effect</label>
+    <input type="checkbox" id="smaa">
+    <label for="bloom">Bloom Effect</label>
```
elalish, line 263: Bloom is excellent, but do you know how to remove the hard edge at large radius?
image

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,364 @@
...
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+-->
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <title>&lt;model-viewer&gt; PostProcessing Examples</title>
```
elalish, line 20: Do these examples still render transparently? As in DOM content behind MV shows through? If not, we may want to mention what limitations users should be aware of. I imagine this might be particularly noticeable for bloom.

elalish (changes requested):
Okay, I pulled your branch, merged our update to r151 and resolved the dependencies. Looks to be working well - just a failing test in effects. I also tried out your examples locally. Awesome stuff, but I mentioned a few that appear broken - I suppose this might be due to a newer three.js version?

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: image

Transparency still works for all effects

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: image

I'm not seeing this artifact. I have noticed that such a hard edge occurs when combining color-grade and bloom (logically, increasing contrast or changing brightness of the canvas itself will cause such artifacts), and also banding is an issue on dark backgrounds (for which there is the option on the effect composer: render-mode="quality" which is active in this example)

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: image

Not getting this issue either in firefox, managed to reproduce using chrome. I'll look into it, the selection also works weirdly for outline so its probably a bug on my end.

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: should be fixed now

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: Interesting bug, turns out that despite the documentation that lit-element array properties update only on changing the reference https://lit.dev/docs/components/properties/#mutating-properties, the above example where I simply pushed values into the array triggered the update loop on firefox on each push. On chrome this was not the case.

elalish:
Hmm, I'm guessing there's some typo in the last commit? Pulling you latest none of the examples run, giving this exception:
```
Uncaught TypeError: Cannot read properties of null (reading 'getDrawingBufferSize')
    at EffectComposer.addPass (postprocessing.esm.js:7751:40)
    at EffectComposer.addPass (effect-composer.js:60:15)
    at HTMLElement.connectedCallback (effect-composer.js:167:31)
```
Presumably the same problem with the tests.

packages/modelviewer.dev/examples/postprocessing/index.html
elalish: Interesting, it works correctly for me on Firefox, but has this same problem on both Chrome and Safari (I'm on an M1 Macbook Pro). I can see the artifact on the Robot example too, around the eyes. It looks like a very imprecise depth buffer to me, which is often caused by near and far planes that are much too far apart. Not sure where the browser-specific part comes in though...

packages/modelviewer.dev/examples/postprocessing/index.html
elalish: This one is consistent for me on Firefox, Chrome, and Safari (again on my M1 Macbook Pro).

packages/model-viewer-effects/src/effects/fxaa.ts
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FXAAEffect } from 'postprocessing';
+import { MVEffectBase } from './mixins/effect-base.js';
+
+export class MVFXAAEffect extends MVEffectBase {
```
elalish, line 19: Do you want to add this one to your examples? And maybe comment on the pros and cons of MSAA, SMAA, and FXAA?

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: Seems to be a WebGL implementation difference then, I've been mainly testing on windows

elalish (commented):
Hey, tests passing, congrats! I'm going to go ahead and push a release before we merge this, since I'd like plenty of time for our users to find bugs through our docs pages before we release it. However, I also think we need to get pretty consistent rendering across browsers before we merge, or we'll get flooded with issues. Do you have access to any other test machines?

Beilinson:
I do have a Mac air 2013 that I'll have access to next weekend

packages/model-viewer-effects/src/effects/fxaa.ts
Beilinson: I'm actually considering whether I should include it at all, mainly because the quality is so much worse than the default antialiasing of the WebGLRenderer or of the SMAA effect

packages/modelviewer.dev/examples/postprocessing/index.html
```diff
@@ -0,0 +1,394 @@
...
+            in separates Passes. Effects can be disabled by setting the <code>blend-mode</code> to <code>skip</code>.
+            <br>
+            It is recommended to follow the official <a href="https://github.com/pmndrs/postprocessing/wiki/Effect-Merging#effect-execution-order" target='_blank'>Postprocessing Wiki</a>
+            for best practices in combining effects. The <code>&lt;model-viewer-effects&gt;</code> handles effect compatibility issues by 
+            placing incompatible effects in a separate Pass.
+            </p>
+          </div>
+          <example-snippet stamp-to="combining-effects" highlight-as="html">
+            <template>
+<model-viewer id="effectsViewer" camera-controls touch-action="pan-y" src="../../shared-assets/models/EmissiveStrengthTest.glb" ar alt="A 3D ">
```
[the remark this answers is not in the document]
Beilinson: Even when an empty effect composer is used, the rendering is not actually using the Three render function, rather the base RenderPass of the effect composer. As I mentioned in some earlier comment, the default antialiasing in your renderer isn't really active then.

packages/model-viewer-effects/src/effect-composer.ts
```diff
@@ -0,0 +1,373 @@
...
+  renderMode: RenderMode = 'performance';
+
+  /**
+   * Anti-Aliasing using the MSAA algorithm. Doesn't work well with depth-based effects.
+   *
+   * Recommended to use with a factor of 2.
+   * @default 0
+   */
+  @property({ type: Number, attribute: 'msaa' })
+  msaa: number = 0;
```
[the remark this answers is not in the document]
Beilinson: The above comment answers this as well

packages/modelviewer.dev/examples/postprocessing/index.html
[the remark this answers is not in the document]
elalish: I think the effect on the depth buffer is the critical thing to understand. Somehow firefox and windows are unaffected, but chrome and safari (on mac) are. I'd also like to see examples of the ugly artifacting with certain effects. I'm curious if it's related, and also if it varies by platform.

packages/model-viewer-effects/src/effect-composer.ts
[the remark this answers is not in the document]
elalish: Hmm, but you pass our `WebGLRenderer` into the postprocessing function, and it's already been set to `antialias: true`, so it seems like it must have an effect? Because once the WebGL context has been created, that can't be changed.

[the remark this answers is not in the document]
Beilinson: @vanruesc the developer of the postprocessing library probably will be able to answer this much better than me.

packages/modelviewer.dev/examples/postprocessing/index.html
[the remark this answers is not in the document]
Beilinson: image

Artifacts like this for example in selective-bloom, which uses the depth buffer to choose which objects are emissive. Seems that the msaa solution provided by the EffectComposer causes artifacts in the depthBuffer, which confuses the effect and makes the pixels glow. 

This happens also if the alternative smaa-effect is applied before the depth based effect, for example here is the same example with msaa disabled and with smaa placed before the selective bloom:
image

And here again with msaa disabled, and smaa placed *after* the selective-bloom:
image

Finally, both msaa and smaa disabled for reference:
image

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: I've set the <code>render-mode='quality'</code> which uses a higher precision HalfFloatType buffers, this maybe might fix those issues for you?

packages/modelviewer.dev/examples/postprocessing/index.html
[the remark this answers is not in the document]
vanruesc: These artifacts occur because the masking technique of the `SelectiveBloomEffect` relies on depth which contains aliasing artifacts. When the color image is antialiased, the depth mask doesn't match the pixels anymore which causes these artifacts to linger. The solution is to render the selected objects to a separate render target during the main geometry render step via MRT instead of masking after the fact.

MSAA causes more fundamental issues for depth-based image effects like SSAO or DoF and, as far as I know, the solution is to implement a custom MSAA resolve step, but that's not possible in WebGL. This [wiki entry](https://github.com/pmndrs/postprocessing/wiki/Antialiasing) shows the artifacts that occur when rendering SSAO with MSAA enabled. This [article](https://mtnphil.wordpress.com/2013/06/26/know-your-ssao-artifacts/) touches on the subject as well (further down, heading "MSAA and SSAO together").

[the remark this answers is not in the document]
elalish: @vanruesc Thanks for the info, and it seems we get pretty nice results with the selections @Beilinson has posted for our examples. However, do you have any insight regarding the fairly extreme depth buffer imprecision artifacts we're getting above (the astronaut)? It seems very odd that it's browser-dependent.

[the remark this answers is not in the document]
Beilinson: @vanruesc Could you maybe explain why there seems to be less issues with SMAA artifacting when the outputEncoding is set to linearEncoding (both examples are without msaa, although that causes more or less the same issues obviously). setting the frameBufferType to UnsignedByteType or HalfFloatType doesn't make a difference.

Case in point (sRGBEncoding):
image

vs linearEncoding:
image

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: Writing this mainly for myself:

Safari and Chrome are embedded webkit while firefox is built on top of gecko, my assumption is that the WebGL2 implementation with Metal is different in gecko vs webkit and thats what is causing the issues. 

Test https://webglreport.com/?v=2 on the different platforms, see if any implementation is missing
Test different frameBufferTypes
Test outputEncoding sRGB vs Linear

Most likely there is not much that we can do, especially since these artifacts seem to be platform specific (IOS/windows do not have these issues). If this is too big of an issue, we can provide a flag to fail on the Metal backend using the renderer context:

```javascript
const gl = WebGLRenderer.getContext();
const debugInfo = gl.getExtension("WEBGL_debug_renderer_info");
const vendor = gl.getParameter(debugInfo.UNMASKED_VENDOR_WEBGL);     // I.e, "Google. Inc"
const renderer = gl.getParameter(debugInfo.UNMASKED_RENDERER_WEBGL); // I.e, "ANGLE (Intel, Intel(R) HD Graphics Direct3D11 vs_5_0 ps_5_0)" 
```

Additionally, I am considering adding the same WebGLRenderer flag `failIfMajorPerformanceCaveat` to the `<effect-composer>`, since model-viewer always sets it to false

elalish: No, Chrome is only WebKit on iOS - on Mac it's Blink, just like Chrome on Windows & Linux. I feel your pain - I've debugged many driver-specific GPU rendering issues. But this is the main value that model-viewer provides: a consistent cross-platform rendering experience, so our users don't have to worry about things like this. There must be some users of `postprocessing` who have run into this - hopefully there are already some workarounds described. If not, we need to file a bug on them. 

I'm also curious if this is a general Mac problem, or specific to the M1 chip.

packages/modelviewer.dev/examples/postprocessing/index.html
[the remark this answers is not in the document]
vanruesc: >  So the question is, what is different about our rendering when we have an empty effect composer, compared to no effect-composer?

I'm not sure, but you might be rendering directly to the canvas with an "empty" effect composer (see [autoRenderToScreen](https://pmndrs.github.io/postprocessing/public/docs/class/src/core/EffectComposer.js~EffectComposer.html#instance-member-autoRenderToScreen)) in which case `antialias` might be `false`.

> However, do you have any insight regarding the fairly extreme depth buffer imprecision artifacts we're getting above (the astronaut)? It seems very odd that it's browser-dependent.

This could be a driver issues as you said. Or it could be related to depth precision. Postprocessing uses `UnsignedIntType` for depth textures by default which can technically result in 16 bit precision although it really shouldn't. Try enabling the stencil buffer: `new EffectComposer(renderer, { stencilBuffer: true })` to force 24 bit precision via `UnsignedInt248Type`.

> Could you maybe explain why there seems to be less issues with SMAA artifacting when the outputEncoding is set to linearEncoding

The aliasing artifacts are probably very similar in shape, but less noticable with darker colors. Are you using a transparent canvas? Postprocessing effects struggle with alpha because WebGL doesn't provide a way to properly blend colors with the background.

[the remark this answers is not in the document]
Beilinson: > I'm not sure, but you might be rendering directly to the canvas with an "empty" effect composer (see [autoRenderToScreen](https://pmndrs.github.io/postprocessing/public/docs/class/src/core/EffectComposer.js~EffectComposer.html#instance-member-autoRenderToScreen)) in which case `antialias` might be `false`.

Actually that fixed it, the issue was that an "empty" effect composer actually had more than just a renderPass, now with only a renderPass it is rendering identically to no effectComposer at all. Thanks!

> This could be a driver issues as you said. Or it could be related to depth precision. Postprocessing uses `UnsignedIntType` for depth textures by default which can technically result in 16 bit precision although it really shouldn't. Try enabling the stencil buffer: `new EffectComposer(renderer, { stencilBuffer: true })` to force 24 bit precision via `UnsignedInt248Type`.

I have pushed a version with `stencilBuffer: true`, on my machine despite not getting such extreme artifacts like @elalish posted I am getting minor white pixel artifacts like in the outputEncoding example above. Not sure if it'll fix the more extreme mac artifacting

> The aliasing artifacts are probably very similar in shape, but less noticable with darker colors. Are you using a transparent canvas? Postprocessing effects struggle with alpha because WebGL doesn't provide a way to properly blend colors with the background.
 
Yes it is using a transparent canvas, I also noticed that setting `alpha: false` fixes the artifacts but thats not an option for release because there is one WebGLRenderer instance for all the scenes. 

I've also attempted https://discourse.threejs.org/t/using-post-processing-through-effectcomposer-causes-jagged-artifacts-on-model-browser-safari/22918/5 as this and one or two other discussions suggested setting the `logarithmicDepthBuffer: true`, again I'm not sure if it'll fix the mac issues, but it doesn't fix the white pixel outline artifacts from my examples.

I'm mostly okay with the white artifacts in my example, they are really only noticeable on a light background with the outline-effect. The Mac artifacting is a much bigger problem however, hopefully the `stencilBuffer` fixes it

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: Not getting this on my macbook air 2013 (tested on chrome and firefox)

packages/modelviewer.dev/examples/postprocessing/index.html
Beilinson: Getting this as well on my macbook air (both chrome and firefox). I'll try to debug and see if theres a difference threejs/postprocessing wise with regards to buffers, its most likely an alpha blend issue.
Interestingly I'm only getting this with the bloom-effect, the selective-bloom-effect works fine on mac on both firefox and chrome without that hard edge.

elalish: @mrdoob This might be an interesting one to bring up with your GPU Chromies. I know color profiles are still fairly recent, so I imagine there are some wrinkles to iron out still.

Beilinson: image

I have reproduced it on windows as well by setting the Color Profile to anything but the default **sRGB IEC61966-2.1**, so this isn't a platform specific issue at all.

elalish: Good to know... have you filed a bug on `postprocessing` yet?

Beilinson: https://github.com/pmndrs/postprocessing/issues/475

elalish (approved):
I think this one is ready, what do you think? With a PR of this magnitude I'm sure we could keep reviewing it forever, but I think I'd rather get it merged so users can test it out on our docs pages. And of course, further PRs are certainly welcome, and it'll make reviewing easier if they're separated.

packages/modelviewer.dev/examples/postprocessing/index.html
elalish: Looks like this completely fixed it for me, thanks!

packages/modelviewer.dev/examples/postprocessing/index.html
elalish: Okay, interesting discussion on that bug. Anyway, your rocket looks perfect for me now. This example we've been talking about still has the issue, unless I select either of the Reinhard2 tonemappers, in which case it looks great. Especially considering this is restricted to a relatively small number of devices and there are various workarounds and fixes probably coming down the pipe, I think this is looking just fine. Thanks for the effort!

Beilinson:
I think it's ready as well! 🎆

zelghrabi-edu:
Hello guys,
Great job 👏🏻
Is there any custom version released for testing purpose? I can do QA for this new feature
Thanks

elalish:
@zelghrabi-edu Since I just released v3.1.1, I'm merging this now to make it easier to get feedback on. It'll be visible on modelviewer.dev, and you can pull the pre-release libraries by adjusting the URLs from https://modelviewer.dev/docs/faq.html#entrydocs-general-questions-api.

elalish:
@Beilinson FYI: https://twitter.com/modelviewer/status/1651007361862225921 Happy to mention you if you have a twitter handle.

Beilinson:
> @Beilinson FYI: https://twitter.com/modelviewer/status/1651007361862225921 Happy to mention you if you have a twitter handle.

That would be nice! I'm part of a programmer collective @galaxyhuntersil . Thanks!

zelghrabi-edu:
Hey @Beilinson 
did you have time to check the issue I raised above?
Thanks in advance

Beilinson:
Hey @zelghrabi-edu, there are two main issues with implementing postprocessing effects on AR:
1. For the native WebXRRenderer for three.js, it requires some tinkering with the renderers that isn't fully supported yet by pmndrs/postprocesssing, which shouldn't enter model-viewer as it is meant to be a fully stable API.
2. On most mobile devices `<model-viewer>` actually defaults to the native AR viewer, meaning `<model-viewer-effects>` actually has no control over the rendering. On IOS for example, this is because WebXR is mostly unsupported.

samaneh-kazemi:
Hi @Beilinson, I'm trying to introduce a new lit component to model-viewer and was looking at your changes as an example. I noticed that the main `model-viewer` class uses a template, while your changes don't. 

https://github.com/aardvarkk/model-viewer/blob/master/packages/model-viewer/src/template.ts

Am I reading your design correctly? If yes, why did you decide to do it differently?

Beilinson:
Hi @samaneh-kazemi, I recommend that you read a bit about Lit (the web component library powering model-viewer and model-viewer-effects). The template is required only if you are adding or affecting dom elements in the document. Model-viewer-effects only adds behavior to model-viewer through JS, and does not require any template.

PR #3954, opened 2022-11-21
Oobe UI

mchiang-goo:


packages/modelviewer.dev/examples/OOBE/JSInterface.js
```diff
@@ -0,0 +1,149 @@
...
+        cardTitle.innerHTML = titleStringID;
+        cardBody.innerHTML = bodyStringID;
+    }
+}
+
+function animateCard() {
+    // Remove the animation class
+    card.classList.remove("run-slide-in-animation");
+    // Trigger reflow
+    void card.offsetWidth;
```
elalish, line 58: I've never seen this JS syntax; what does this do?

elalish (approved):
Looks great, thanks!

packages/modelviewer.dev/examples/OOBE/JSInterface.js
mchiang-goo: HTMLElement.offsetWidth is a read-only property that returns the layout width of an element as an integer. Calling it forces a reflow. Prob a better way to do this performance-wise.

PR #3984, opened 2022-12-02
Fix crash on browers that use projection layers

cabanier:
Three.js uses projection layers on the Quest browser. The code currently always expects a XRWebGLLayer

### DIFF
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -418,10 +418,17 @@ export class ARRenderer extends EventDispatcher {
       const scale = view.recommendedViewportScale;
       view.requestViewportScale(Math.max(scale, MIN_VIEWPORT_SCALE));
     }
-    const layer = this.currentSession!.renderState.baseLayer;
-    const viewport = layer!.getViewport(view)!;
-    this.threeRenderer.setViewport(
-        viewport.x, viewport.y, viewport.width, viewport.height);
+    const layer = xr.getBaseLayer();
+    if (layer instanceof XRProjectionLayer) {
+      const subimage = xr.getBinding().getViewSubImage(layer, view);
+      this.threeRenderer.setViewport(
+        subimage.viewport.x, subimage.viewport.y, subimage.viewport.width, subimage.viewport.height);
+
+    } else {
+      const viewport = layer!.getViewport(view)!;
+      this.threeRenderer.setViewport(
+          viewport.x, viewport.y, viewport.width, viewport.height);
+    }
   }
 
   private placeInitially() {


cabanier:
@elalish how do I add a reviewer?

elalish:
> @elalish how do I add a reviewer?

Don't worry about it, I get notified anyway.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -418,10 +418,17 @@ export class ARRenderer extends EventDispatcher {
       const scale = view.recommendedViewportScale;
       view.requestViewportScale(Math.max(scale, MIN_VIEWPORT_SCALE));
     }
-    const layer = this.currentSession!.renderState.baseLayer;
-    const viewport = layer!.getViewport(view)!;
-    this.threeRenderer.setViewport(
-        viewport.x, viewport.y, viewport.width, viewport.height);
+    const layer = xr.getBaseLayer();
+    if (layer instanceof XRProjectionLayer) {
+      const subimage = xr.getBinding().getViewSubImage(layer, view);
```
elalish, line 423: maybe `const viewport = layer instanceof XRProjectionLayer ? xr.getBinding().getViewSubImage(layer, view).viewport : layer!.getViewport(view)!` and avoid the whole branch?

elalish (changes requested):
@klausw Would you mind taking a look at this? I'm guessing this is fine, but you're more familiar with the latest WebXR APIs than I am. 

@cabanier I assume we still need to fix the hit-test thing to get this working on Quest? Do you think there are any other blockers?

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -418,10 +418,17 @@ export class ARRenderer extends EventDispatcher {
       const scale = view.recommendedViewportScale;
       view.requestViewportScale(Math.max(scale, MIN_VIEWPORT_SCALE));
     }
-    const layer = this.currentSession!.renderState.baseLayer;
-    const viewport = layer!.getViewport(view)!;
-    this.threeRenderer.setViewport(
-        viewport.x, viewport.y, viewport.width, viewport.height);
+    const layer = xr.getBaseLayer();
+    if (layer instanceof XRProjectionLayer) {
```
elalish, line 422: Any idea why our CI is throwing "`XRProjectionLayer` is undefined"?

cabanier: No, but maybe I can change it to `XRWebGLLayer`

cabanier:
> @cabanier I assume we still need to fix the hit-test thing to get this working on Quest? 

No, I'm working on getting it enabled in the Quest browser so you don't have to work around it.
I expect it to ship early 2023.

packages/model-viewer/src/three-components/ARRenderer.ts
cabanier: done

cabanier:
@elalish  can you point me to some other sites that are using model-viewer? Also what do we need to do to have the results show up in a google search.
It's exciting to see this come alive on our headset!

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -418,8 +418,10 @@ export class ARRenderer extends EventDispatcher {
       const scale = view.recommendedViewportScale;
       view.requestViewportScale(Math.max(scale, MIN_VIEWPORT_SCALE));
     }
-    const layer = this.currentSession!.renderState.baseLayer;
-    const viewport = layer!.getViewport(view)!;
+    const layer = xr.getBaseLayer();
```
elalish, line 421: Looks like we need to null-check `layer`, based on the CI failure. But the weirder part is that it's failing on Chrome Headless, which I can't imagine supports WebAR. I thought that would fail back in `supportsSession`. Any idea what's going on?

elalish (changes requested):

packages/model-viewer/src/three-components/ARRenderer.ts
cabanier: No, I don't. Is this running a static analysis?

elalish: No, just regular JS tests in a headless browser. If you go to `packages/model-viewer` and run `npm test` locally I wonder if it'll repro? Maybe only on Chrome Linux though...

cabanier: I tried running it locally but it didn't work. I suspect those ports are blocked on my machine.

elalish (approved):
Thanks! It passes all the automation, so I'm merging it in time for our manual test pass tomorrow. I'll let you know if we see any regressions.

PR #4005, opened 2022-12-12
set default anisotropy to 4

elalish:
Fixes #2218 

@hybridherbst Thanks for pushing three.js on this! I tested with your example model and the difference is stark indeed.


hybridherbst:
You may want to hold merging this until 
- https://github.com/mrdoob/three.js/issues/25067

is resolved, which for us only surfaced after switching the default to 4.

Actually kind of curious as to why the goldens didn't catch this! - but maybe there's just not a lot of nearest filtering going on in the context of model viewer or you're already catching this elsewhere?

elalish:
Merging since your fix has been pulled in; thanks @hybridherbst!

PR #4030, opened 2023-01-03
Quicklook ar placement

### LINKED ISSUE #3989: ar-placement ignored on quick look
### Description

It seems that the `ar-placement` option gets ignored on AR Quick Look. Is this a bug or intended behaviour? If it's intended, the documentation should mention that imo. If it's not implemented yet, it should at least be possible to make it available when converting the .glb to .usdz file, because the placement info is embedded into the .usdz file for Quick Look.

#### Live Demo
on iOS doesn't show the vertical wall, but the horizontal floor at the onboarding animation:
https://modelviewer.dev/examples/augmentedreality/index.html#wall

### Version
- model-viewer: v2.1.1

### Browser Affected

- [ ] Chrome, version: xx.x.xxxx.xx
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [x] Safari

### OS

- [ ] Android
- [x] iOS 16.0 iPhone XR
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### AR

- [ ] WebXR
- [ ] SceneViewer
- [x] QuickLook

pixlhero:
Fixes #3989

Also had to fix the boom_2_.glb model because the model was offset quite a lot, which messed with the rotation required for quicklook AR.

I tried a few different approaches to reset the rotation of the model (like keeping a copy of the matrix). Only rotating it back worked for me though.

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -434,13 +434,34 @@ configuration or device capabilities');
 
       updateSourceProgress(0.2);
 
-      const exporter = new USDZExporter();
-      const arraybuffer = await exporter.parse(model);
+      const usdzOptions = {
+        ar: {
+          anchoring: { type: 'plane' },
+          planeAnchoring: {
+            alignment: this.arPlacement === 'wall' ? 'vertical' : 'horizontal',
+          },
+        },
+      };
+
+      // necessary because quicklook internally rotates model when placing vertical. 
+      // See https://github.com/google/model-viewer/issues/3989
+      if (usdzOptions.ar.planeAnchoring.alignment === 'vertical') {
+        model.rotateX((-90 * Math.PI) / 180);
+        model.updateMatrixWorld();
+      }
+
+      const exporter = new USDZExporter() as any;
+      const arraybuffer = await exporter.parse(model, usdzOptions);
       const blob = new Blob([arraybuffer], {
         type: 'model/vnd.usdz+zip',
       });
 
       const url = URL.createObjectURL(blob);
+      
+      if (usdzOptions.ar.planeAnchoring.alignment === 'vertical') {
+        model.rotateX((90 * Math.PI) / 180);
+        model.updateMatrixWorld();
+      }
 
       updateSourceProgress(1);
 
--- a/packages/modelviewer.dev/examples/augmentedreality/index.html
+++ b/packages/modelviewer.dev/examples/augmentedreality/index.html
@@ -360,6 +360,8 @@
           <div class="heading">
             <h2 class="demo-title">Placing on a Wall</h2>
             <h4>This demonstrates the <code>ar-placement</code> attribute, which defaults to "floor", but using "wall" gives a different AR placement experience.</h4>
+            <p>QuickLook will ignore this attribute if a .usdz or .reality model is supplied through <code>ios-src</code> and instead look at the <code>planeAnchoring:alignment</code> option
+              inside the USDZ file.</p>
           </div>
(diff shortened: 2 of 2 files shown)


packages/model-viewer/src/features/ar.ts
```diff
@@ -432,14 +432,35 @@ configuration or device capabilities');
...
+          planeAnchoring: {
+            alignment: this.arPlacement === 'wall' ? 'vertical' : 'horizontal',
+          },
+        },
+      };
+
+      // necessary because quicklook internally rotates model when placing vertical. 
+      // See https://github.com/google/model-viewer/issues/3989
+      if (usdzOptions.ar.planeAnchoring.alignment === 'vertical') {
+        model.rotateX((-90 * Math.PI) / 180);
```
elalish, line 447: Let's not update the GLB; it's a good test precisely because many models are not aligned to the origin. I believe you'll need to also apply the `scene.target` translation to make it work.

elalish: It sounds like USDZ may have particular assumptions about where the origin should be relative to the placement point, so you may also need to make use of `$scene.boundingBox` to find the translation you need.

elalish (changes requested):
Thank you! Probably a good idea to test with several different off-origin models to ensure the transformations are consistent between WebXR and QuickLook.

elalish:
@pixlhero Ping - are you planning to come back to this?

pixlhero:
> @pixlhero Ping - are you planning to come back to this?

Not in the near future unfortunately. We worked around this issue by just not placing it on the wall in our project. So until I have some spare time between projects I won't be able to work on this.

h36ahmed:
@pixlhero - has this been fixed or still being worked on?

ekokotov:
any news how to enable vertical placement on IOS Quicklook?

elalish:
If anyone would like to take over this PR, it would be great to merge it. I may get to it eventually, but not sure when.

pixlhero:
Sorry about turning silent on this PR. I think I might be able to come back to this during next month. Probably not in the next two weeks though, but I put it in my calendar.

elalish:
Thanks!

pixlhero:
Working on this now 🚀

elalish:
Thanks! Let me know if you have any questions.

packages/model-viewer/src/features/ar.ts
elalish: Would you like to take over this PR or make your own version?

vincentfretin:
I tested the changes here with model-viewer 3.4.0, it works fine for my models.
But it seems we have an issue when auto-rotate attribute is used on model-viewer tag, the usdz generated has a different y rotation each time we generate the usdz, and after the 90° rotation here it gives really weird results. The bounding box seems wrong and will result of bad snapping on the wall, the shadow is ok but the model is not correctly on the wall.

packages/model-viewer/src/features/ar.ts
vincentfretin: I didn't understand what you were suggesting in the comments above with the translation and bounding box, and also every time I tried to do threejs operation like that translation followed by rotation I always do things wrong and lose my hairs, so I probably can't do any change related to that.
But not sure if we should go this way if removing the lines https://github.com/mrdoob/three.js/blob/aaafb4ffe8830932e31aa44003376457265d126a/examples/jsm/exporters/USDZExporter.js#L209-L210 also fixes the issue.

vincentfretin: related comment https://github.com/google/model-viewer/issues/3989#issuecomment-2030217299

PR #4041, opened 2023-01-10
fixed NaN cameraOrbit

elalish:
There are two fixes here, one for the actual problem, and a second to act as a backstop against similar problems. Both were verified to independently fix the issue via the included test.

I also verified that no matter what nonsense you put into `camera-orbit` or `cameraOrbit`, it will never produce NaN, it instead falls back to the default initial values.

### DIFF
--- a/packages/model-viewer/src/three-components/SmoothControls.ts
+++ b/packages/model-viewer/src/three-components/SmoothControls.ts
@@ -337,6 +337,10 @@ export class SmoothControls extends EventDispatcher {
       return false;
     }
 
+    if (!isFinite(nextTheta) || !isFinite(nextPhi) || !isFinite(nextRadius)) {
+      return false;
+    }
+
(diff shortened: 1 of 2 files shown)


packages/model-viewer/src/three-components/SmoothControls.ts
```diff
@@ -508,7 +512,7 @@ export class SmoothControls extends EventDispatcher {
   }
 
   private pixelLengthToSphericalAngle(pixelLength: number): number {
-    return 2 * Math.PI * pixelLength / this.element.clientHeight;
+    return 2 * Math.PI * pixelLength / this.scene.height;
```
elalish, line 515: I have no idea why `clientHeight` was used here; `scene.height` is correct as this is the one that is assured to never be < 1.

samaneh-kazemi (approved):
Thanks for the fix.

PR #4066, opened 2023-01-26
three.js r149

elalish:
Fixes transparent rendering regression and specular/ior export.


elalish:
@mrdoob Looking good!

PR #4070, opened 2023-01-30
Add new glTF Sample Model tests

gkjohnson:
Adds the following models from the glTF Sample Models repo:
- ABeautifulGame
- IridescenceLamp
- LightsPunctualLamp
- MosquitoInAmber
- EmissiveStrengthTest
- IridescenceDielectricSpheres
- IridescenceMetallicSpheres

I've also generated screenshots for the new models but it seems that the "stellar" screenshots did not render.

Also it seems that there are a few cases that are clearly incorrect here - ie Babylon's chessboard, emissive test, and amber renders and Filaments lack of iridescence support. Should I exclude those renderers in those cases?


elalish (changes requested):
Thank you so much for adding these! Stellar is run offline and added manually, so just exclude it for now and I'll ping @bsdorra to upload the Stellar versions for us.

I bet the other renderers are painfully out of date; they may already have support for some of this. Hopefully its as easy as updating `package.json` and running `npm run update-package-lock`. If you're up for trying, that would be great. If there's a bunch of breaking changes, we can look at it on a later PR. But I think it's fine to keep these images all included.

gkjohnson:
> I bet the other renderers are painfully out of date; they may already have support for some of this. Hopefully its as easy as updating package.json and running npm run update-package-lock. If you're up for trying, that would be great. 

Looks like there's at least one breaking change in Filament. I've updated Babylon but it seems to have just affected the amber mosquito model. I think we should save filament and rerunning the rest of the Babylon screenshots for a future PR.

elalish (approved):
Perfect, thank you!

PR #4078, opened 2023-01-31
Add stellar scenarios and renderings

bsdorra:
Add stellar renderings for recently added scenarios


elalish (approved):
Thank you!

elalish:
@bsdorra This one looks like a Stellar bug, are you aware of it? https://modelviewer.dev/fidelity/#khronos-TransmissionRoughnessTest

bsdorra:
@elalish if you're referrering to the darker color, then it is expected behavior.
The pathtracing integrator evaluates the albedo two times for these volumes, on enter and on exit. This effectively gives you albedo^2.

elalish:
@bsdorra @gkjohnson The purpose of these tests is to find rendering inconsistencies and it's found one even between these two path-tracers. I just figured it might be worth bringing up at PBR to decide who has the spec-compliant behavior.

gkjohnson:
My path tracer is only evaluating color when entering the front face of the surface but looking at the spec again it's probably actually correct to multiply in the color on the way in and out of the volume. I'll put it on the list to make that change:

> Base color changes the color of light at the volume boundary.

Thanks!

elalish:
It looks like three.js is making an update that will bring it more in line with Stellar here too: https://github.com/mrdoob/three.js/pull/25502

I guess that means we'll have to talk to Filament, Babylon, and the gltf-sample-viewer.

PR #4085, opened 2023-02-01
set withCredentials for other loaders

### LINKED ISSUE #4083: `with-credentials` is not respected for skybox-image
### Description

I have `with-credentials` attribute set and it works well both for model and poster image but seems it's not respected for skybox-image as it gives me a 403 error and `Cookie` header is not set.

### Version

- model-viewer: v2.1.1

### Browser Affected

- [X] Chrome, version: xx.x.xxxx.xx
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [X] Linux
- [ ] MacOS
- [ ] Windows

### AR

- [ ] WebXR
- [ ] SceneViewer
- [ ] QuickLook

elalish:
Fixes #4083 

This is a pretty straight-forward fix, but I don't have a great way to test it. @robertsLando I'd appreciate it if you can verify this is working for you once it's in.

### DIFF
--- a/packages/model-viewer/src/features/loading.ts
+++ b/packages/model-viewer/src/features/loading.ts
@@ -34,7 +34,7 @@ const DEFAULT_KTX2_TRANSCODER_LOCATION =
     'https://www.gstatic.com/basis-universal/versioned/2021-04-15-ba1c3e4/';
 
 const DEFAULT_LOTTIE_LOADER_LOCATION =
-    'https://cdn.jsdelivr.net/npm/three@0.148.0/examples/jsm/loaders/LottieLoader.js';
+    'https://cdn.jsdelivr.net/npm/three@0.149.0/examples/jsm/loaders/LottieLoader.js';
 
 const RevealStrategy: {[index: string]: RevealAttributeValue} = {
   AUTO: 'auto',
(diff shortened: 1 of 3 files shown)


robertsLando:
@elalish Thanks for this but I just tested v3.0.0 and it's still not working 😢  Keep getting 403 error, this only happens with skybox image BTW, poster and model src are loaded with credentials correctly

elalish:
Hmm, can you post a repro link?

robertsLando:
@elalish I wish but I would need a repro with a backend to also authenticate the client, it's not that easy to create... I checked the code and your changes looks good I have no clue what it does differently, let me try to debug it on my end. Any tips?

robertsLando:
This is the call stack when skybox image is loaded:

Schermata da 2023-02-03 11-20-43

robertsLando:
By debugging this seems that `this.crossOrigin` value is `"anonymus"` when it reaches `three.model.js` `load` method instead of `true`

robertsLando:
I think you should use `loader.setWithCredentials(value)` function instead of directly set the property, seems that setting it in this way the value isn't persisted to the parent instance, when it comes to the parent `withCredentials` is false but is `true` in `hdrLoader` and `ldrLoader` instances

elalish:
Ah, good call! That's just a typo on my part, but I thought the `@types/three` would catch that kind of thing for me. Okay, should be an easy fix. Really appreciate the debug.

robertsLando:
Patch release?

elalish:
Yeah, probably - got a few other things to fix first.

Issue #2747, opened 2021-09-01
Including a Build Without Bundling Dependencies


AmitMY:
### Description

The package `@google/model-viewer` has an entry point at `dist/model-viewer.min.js`.
This file includes all the code required to run `model-viewer`, including (an estimate) of `1MB` for the `three` package, and a few `kb`s for the actual package code.

I would like you to include an additional build in the distribution, that does not bundle all of the dependecies, but instead, statically imports from `three` without a resolution.

This will decrease my bundle size drastically, as right now I am using `three` for one thing, and `model-viewer` for another, instead of both just referring to the same `three` reference.

(This is how most packages are packaged in NPM, and probably just a small modification to your build system)

----

Ideally, as the web moves to using javascript `module`s, if you would also include a build that imports `three` dynamically, that would be great, as this entire huge dependency can be lazy-loaded if needed. (however, I believe this would require major changes)

### Version
- model-viewer: v1.8.0

### Browser Affected
- [x] Chrome, version: all
- [x] Edge
- [x] Firefox
- [x] IE
- [x] Safari

### OS

- [x] Android
- [x] iOS
- [x] Linux
- [x] MacOS
- [x] Windows

elalish:
Indeed, we've been talking about this a bit in a related issue: https://github.com/google/model-viewer/issues/1502

We do in fact also publish our unbundled library on npm (in the `lib` folder next to `dist`). However, I think there are some problems with our `package.json` for this use case. We're trying to find the best practice to support the various toolchains out there. If you can give any pointers as to what fields in `package.json` your tooling makes use of, that would be helpful. They all seem to do things a little differently. FYI @timmmeh.

As for lazy-loading three, yes that would be nice to do  at  some point.

AmitMY:
I tried importing `@google/model-viewer/lib/model-viewer` but typescript errors ensued.

My build system is the default Angular build system which uses Webpack with the following main files priorities:
```js
mainFields: ['browser', 'es2015', 'module', 'jsnext:main', 'main'],
```

elalish:
Pretty sure that is the right way; perhaps you can open a new issue with the specific typescript errors and we can help.

AmitMY:
Hi @elalish 
I think this issue is still relevant.
There are no typescript issues, just that if I import both `model-viewer` and `three` in my app, `three` is bundled twice, and the console says 
> WARNING: Multiple instances of Three.js being imported.

[...]

Issue #4090, opened 2023-02-02
model-viewer@v3 uses `lottie-web` as a dev dependency 


AmitMY:
### Description

- I installed v3.0.0
- I use model-viewer globally (`import(/* webpackChunkName: "@google/model-viewer" */ '@google/model-viewer');`)

The error I get is:
```
Error: node_modules/@google/model-viewer/lib/features/scene-graph/api.d.ts:1:31 - error TS2307: Cannot find module 'lottie-web' or its corresponding type declarations.

1 import { AnimationItem } from 'lottie-web';
```

### Version
- model-viewer: v3.0.0

### OS

- [x] MacOS

elalish:
Yes, we have lottie-web listed as a dev dependency specifically so you get the required types. How are you installing us? Is there a reason the dev dependencies are not being pulled down?

anders8:
Same error from the same issue. I use model-viewer in an Angular project, so it was previously (2.x.x) installed with just 

npm install @google/model-viewer

So are you saying that lottie-web needs to be added manually in dev dependencies? (I see it listed now in dev at https://www.npmjs.com/package/@google/model-viewer?activeTab=dependencies but I didn't look last night.)

elalish:
Interesting; I don't use Angular so I'll need your help to figure out the right solution here. My understanding of npm dev-dependencies was exactly that when you are developing with the project and you `npm install` it will pull them down for exactly this reason, similar to our @types/three dev dependency. If that's not happening, I'd love to understand why and what the best practice is here to solve it at the library level.

muuvmuuv:
Here is a repl: https://github.com/muuvmuuv/google-model-viewer-4090

1. pnpm install
2. pnpm start

elalish:
@muuvmuuv Thank you - I notice that your pnpm-lock.yaml file has no entry for lottie-web (hence the error) and also that its model-viewer entry has `dev: false`, while others have `dev: true`, which I'm guessing tells it whether or not to include its dev-dependencies. I've never configured pnpm, so might you know of a way to indicate that the dev-dependencies are important?

robertsLando:
I had this issue too but not using pnpm but npm and webpack shows that warning

AmitMY:
v3.0.1:

> Error: node_modules/@google/model-viewer/lib/features/scene-graph/api.d.ts:1:36 - error TS2307: Cannot find module 'lottie-web' or its corresponding type declarations.
> 
[...]

PR #4112, opened 2023-02-10
fix(model-viewer): update `three` version to correctly allow everything but a major version change #2747

AmitMY:
### Reference Issue
Adresses #2747

### DIFF
--- a/packages/model-viewer/package.json
+++ b/packages/model-viewer/package.json
@@ -73,7 +73,7 @@
     "3d"
   ],
   "dependencies": {
-    "three": "^0.149.0",
+    "three": ">=0.149.0 <= 1",
     "lit": "^2.2.3"
   },
   "devDependencies": {
@@ -84,7 +84,7 @@
     "@types/chai": "^4.3.1",
     "@types/mocha": "^9.1.1",
     "@types/pngjs": "^6.0.1",
-    "@types/three": "^0.148.1",
+    "@types/three": ">=0.148.1 <= 1",
     "@ungap/event-target": "^0.2.3",
     "chai": "^4.3.6",
     "focus-visible": "^5.2.0",


donmccurdy:
I think it'd be common to put `three` in `peerDependencies` and `devDependencies` but omit it from production dependencies. Something like these:

https://github.com/donmccurdy/three-pathfinding/blob/ece8962996f962ea893d9622f77f8e6149eaa676/package.json#L46-L59

https://github.com/pmndrs/react-three-fiber/blob/7b256898c1b0de8ecdae3374b2ebbd1c4b158220/packages/fiber/package.json#L54-L62

elalish:
@donmccurdy Thanks, I've actually been looking for advice on this for some time. That sounds like a reasonable change, but I don't want to take it too lightly. Can you give me a mental model of how to think about the three different kinds of dependencies? Sometimes a three.js version brings in a change that breaks MV and requires an update on our side. What would happen in that case after your proposed change?

donmccurdy:
What I'd aim for would be:

- In `peerDependencies`, give the range of three.js versions that MV intends to be compatible with. `>=0.149.0 <= 1` seems reasonable, I tend to be pretty liberal here. Users will see a warning about the peer dependency if their three.js version is missing or outside the range. 
- In `devDependencies`, give the _exact_ three.js version that you want to use when developing, and testing, within the MV repository. This has no effect on users installing from npm, but would affect builds to a CDN.
- In `dependencies`, omit three.js entirely. Because `dependencies` will override `peerDependencies`, any version specified here would create restrictions (and possible duplicate packages!) that you probably don't want. If the user had their own copy of three.js and the version didn't match `dependencies`, they'll likely end up with two copies in the bundle.

That would probably need to be considered a breaking change, though. Users must install three.js themselves when using npm, whereas three.js comes in automatically with MV on npm now. The benefit of the change is (1) flexibility in the version of three.js supported, and (2) eliminated risk of duplicating three.js in the bundle.

elalish:
Okay, so I guess the main risk is that someone updates their peer dependency to a newer version of three that has a serious breaking change for us and they don't understand why MV seems to be broken (subtly or otherwise). That happens less often now than it used to, but still. If the peer dependency thing is just a warning, I'm tempted to make that version exact too - would that cause other problems?

Also, should I take this same approach to our other dependency, `lit`? Or is there something that makes that a special case? @e111077, do you have opinions here?

e111077:
Lit is interoperable with other versions of Lit and in model-viewer's case, essentially an implementation detail. Lit team's guidance is to simply include it as a normal dep and not a peer dep

donmccurdy:
> If the peer dependency thing is just a warning, I'm tempted to make that version exact too - would that cause other problems?

Does MV have APIs that accept or return three.js objects? Or are the three.js objects just used internally, not in public interfaces? 

For a library-shaped thing (not sure if that describes MV?), an exact version is problematic — it only takes two three.js-related libraries that each require a specific version number before you have a pretty complicated conflict. If I'm using three.js non-trivially elsewhere in the app, I wouldn't want a dependency dictating the version that way.

To be conservative, you could specify r145–r150 or something, and just update the range when necessary. Incrementing the range to r145–r151 would just be a patch release, easy enough.

To be even more conservative, and specify an exact version, I'd recommend using `dependencies` instead, and just accept the risk of duplicating three.js in the bundle.

elalish:
> To be even more conservative, and specify an exact version, I'd recommend using `dependencies` instead, and just accept the risk of duplicating three.js in the bundle.

Yeah, I think this is basically how we arrived where we are. No, we don't have three.js objects as part of our public API, but we touch so many parts of three.js that we often get hit with breaking changes anyway. Most recent was the loss of SpecGloss support, not to mention the whole `TwoPassDoubleSided` thing. Since I only have testing for one version of three.js, I'm a little worried about saying others are fine. 

@AmitMY Regarding your original issue, v3.0 does use three.js r149, so you should be fine if you simply upgrade MV version - we always state what three.js version we're on in the release notes. Is there a reason you prefer to upgrade them independently?

AmitMY:
Currently, I was not using v3.0.0 because of #4090 (which is now fixed).
In my project, I am using three.js independently of `model-viewer` (to do a lot of mathy vector things), and we have a policy that every dependency must be updated to latest unless it is breaking the build/tests. Being "in sync" with MV would require someone to always check your versions, rather than let npm's installation process resolve the packages.

I think that @donmccurdy's solution is the best one:
1. `peerDependencies`: `three=>=0.149.0 <= 1`
2. `devDependencies`: `three=0.149.0`

If this was even a 100kb dependency, I would not bother, but since this is `three`, and is `>1MB`, it is harmful to use it.

Another option is for me to use tree shaking on my three dependency, so MV is not affected, but that would require MV to somehow publically export three, so I can use the installed MV three to interact with the models.

elalish:
> (Another, more difficult option is for me to use tree shaking on my three dependency, so MV is not affected, but that would require MV to somehow publically export three, so I can use the installed MV three to interact with the models.)

Actually, I've wanted to do this for awhile, and I don't think it's too hard. Basically I need another version of our bundle: model-viewer-three.js that would export all of THREE. It will be a little larger since we can't tree shake anything out, but probably not bad. 

Do you at least understand my reasoning above why I'm concerned about mixing and matching with different versions of three?

AmitMY:
I understand your reasoning, and it is true for almost all cases that use peer dependencies, however, this is so costly, that it should be the developer's choice IMO. it is nearly 1MB of bundle increase, which is (at least for me), quite a lot

elalish:
@AmitMY true, but as you said, all you need to do is break your latest-version policy to keep three.js and mv in lockstep. We're generally no more than two versions behind, and usually less. I think for the stability of your project, this is a good idea.

elalish:
@AmitMY Unless I've misunderstood - does putting three in `dependencies` mean you can't share it in your project at all? Or just if your other version doesn't match? If the former, then switching to `peerDependencies` with a strict version number would be fine with me.

elalish:
I'm working on putting out a patch release ideally tomorrow - if we can reach consensus by then, I can include the change.

AmitMY:
To be honest I'm a bit lost.

elalish:
Okay, we've done some testing and it looks like switching to a peer dependency is the right way to go (thanks @e111077 
and @donmccurdy). I'm going to pin the minor three.js version, since that will keep people on what we've tested and can stand behind. But you can still force your way around and use a different non-matching version with `--legacy-peer-deps`, which seems like the right level of warning for using a version of three with potentially breaking changes.

AmitMY:
Thank you, that seems like a good solution to me.
Can close this.

elalish:
Closing in favor of #4135.

PR #4118, opened 2023-02-13
Update three-gpu-pathtracer to v0.0.13

gkjohnson:
### Reference Issue
--

Updates three-gpu-pathtracer to v0.0.13 add support for lights to the renderer hook. This makes a few changes to the three-gpu-pathtracer renderer related to https://github.com/mrdoob/three.js/pull/17370#issuecomment-1426968852, as well.


elalish (approved):
Nice work, thank you! Your renderer is very impressive.

PR #4120, opened 2023-02-13
Babylon.js renderer: Fixing some models being rendered black

Popov72:
This is a small PR to fix some models that are rendered black with the Babylon.js renderer.

### DIFF
--- a/packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
+++ b/packages/render-fidelity-tools/src/components/renderers/babylon-viewer.ts
@@ -121,18 +121,17 @@ export class BabylonViewer extends LitElement {
     const modelFileName =
         scenario.model.substring(lastSlashIndex + 1, scenario.model.length);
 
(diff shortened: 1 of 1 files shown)


elalish (changes requested):
Looks good! Can you upload any babylon renders that have changed? Use `npm run update-screenshots babylon`. Also, any trouble signing the CLA?

elalish:
Okay, I rescanned it and it passed, so it looks like whatever you did worked. We have instructions for setting up the project under [WLS2](https://github.com/google/model-viewer#windows-1011-setup) - if that's not working it would be great to know what needs to be updated in our instructions.

Failing that, we can also merge this and I can update the screenshots here in a follow-up.

Popov72:
The problem is with the chrome instance used by pupeeter: it does not support WebGL/WebGL2 and I wasn't able to enable the support.

elalish:
Oh, right, I've actually been running into this problem too - the Chrome 109 update introduced this problem (even on my mac). Okay, I'll merge this and see if I can work around it from here. Thanks!

elalish (approved):

PR #4122, opened 2023-02-14
updated babylon goldens

elalish:
Followup to #4120


elalish:
Let's not reupload *all* the goldens....

PR #4141, opened 2023-02-24
fix video texture on iOS

elalish:
The videoTexture would not automatically play on iOS - turns out to be a simple fix.

### DIFF
--- a/packages/model-viewer/src/features/scene-graph.ts
+++ b/packages/model-viewer/src/features/scene-graph.ts
@@ -150,8 +150,9 @@ export const SceneGraphMixin = <T extends Constructor<ModelViewerElementBase>>(
       const video = document.createElement('video');
       video.src = uri;
       video.muted = true;
-      video.play();
+      video.playsInline = true;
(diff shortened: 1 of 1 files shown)


planetMatrix:
Hi there, does that mean Video Textures are working in Quick Look as well?

elalish:
No, I don't think USDZ supports video textures at all, and that's the only format Quick Look supports currently. FYI @grorg.

planetMatrix:
Yeah I thought so. I'm wondering why it's mentioned in the release page 3.1.0?

elalish:
iOS != QuickLook. This fixed it so the textures play in 3D mode on iOS. AR is a relatively narrow use case compared to what we offer for general 3D on the web.

donmccurdy:
Aside – USD (and USDZ) do support video textures. The limitation is Apple's Quick Look not supporting these textures. Ideally Safari would implement the WebXR API, so that libraries like modelviewer, three.js, and babylon.js could provide video texture support in AR/VR/XR independently.

PR #4151, opened 2023-03-03
fix annotations example bug

y-71:
This PR aims to fix a small bug that was present in the annotations example.


elalish (approved):
Thanks!

PR #4186, opened 2023-03-31
updated to three.js r151

elalish:
Lots of rendering improvements this time, most notably that multiple UV channels and texture transforms are fully and properly supported now, which is a huge win especially for models using a tiled texture swatch as well as a non-tiled AO map, e.g. Sheen Chair:
image
Further proof:
image
Also significant improvement of back-side reflections of transmissive materials:
image
I need to dig a little deeper into what's changed with punctual lights:
image
And also with the Glam sofa:
image


elalish:
@mrdoob This is looking great overall! I can't quite figure out why the last two renders changed. Not saying it's a regression, but I'd like to understand for sake of glTF compliance and other renderers. Any thoughts?

elalish:
Yeah, @mrdoob there's definitely something going in here; these used to look a lot more similar: https://modelviewer.dev/fidelity/#khronos-LightsPunctualLamp

This one's also looking not great - it was too dark before, and now it's much too light: https://modelviewer.dev/fidelity/#khronos-GlamVelvetSofa

elalish:
@mrdoob Okay, my best guess is whatever is causing that change to the LightsPunctualLamp also blew out our default lighting, so that's become a blocker.

elalish:
Okay, nevermind, I realized I just got hit by https://github.com/mrdoob/three.js/pull/25479. The one-liner fixed it and it only took an hour to figure it out, so not bad overall. And apparently GlamVelvetSofa has a punctual light (!?!).

ocrnalic:
Hi, this works great on model viewer and android, but when autogenerated usdz file is opened on Iphone second UV map doesn't work. Is there some naming rule for UV maps, or it just isnt supported for USDZ autogeneration?

mrdoob:
The current state of USDZ is a bit of a mess:
https://twitter.com/mrdoob/status/1654494500230778887

mrdoob:
In theory this should work but I don't think it does.
https://github.com/mrdoob/three.js/pull/25996

ocrnalic:
> The current state of USDZ is a bit of a mess:
> https://twitter.com/mrdoob/status/1654494500230778887

In gerneral autogeneration works perfectly except for models with animations and multiple uv maps

ocrnalic:
@elalish do you maybe have some info about multiple UV sets and USDZ autogeneration. Does this work or still isnt supported? I have tried with only AO on the second set and I cant make it work… thanks!

elalish:
I'm guessing you're looking for [this](https://github.com/mrdoob/three.js/pull/26023), which we can pull down at the end of the month when it's released.

mrdoob:
These two PRs implemented multiple uv sets:

* https://github.com/mrdoob/three.js/pull/25996
* https://github.com/mrdoob/three.js/pull/26009

ocrnalic:
@elalish @mrdoob thank you for help, I'm not a developer, so maybe I don't understand everything the best...
But I have made some tests.. 
I have created a very simple scene wher AO is baked on second UV map
image
It works correct in model viewer but after autogeneration second uv-map is not displayed correctly
image

But when I convert glb file using this online convertor (https://arize.io/tools/usdz-converter/) to usdz 
On Iphone it works correctly! It also works right on https://www.usdz-viewer.net/
image

Here are the files:
[cube.zip](https://github.com/google/model-viewer/files/11452826/cube.zip)

So somehow this online converter is converting in the right way comparing to autogeneration.

So if I understand correctly, this will be fixed by the end of the month as elalish wrote (https://github.com/mrdoob/three.js/pull/26023)?
or i got it wrong? :) 

Thanks!

ocrnalic:
Great!! off course I will test in detail when it will be released

PR #4191, opened 2023-03-31
Revert "updated to three.js r151 (#4186)"

elalish:
This reverts commit 640b55471ecf95d435bffb247b34a6df99da825a.

Okay, apparently I need to tighten up my testing because this broke nearly every usage, which is to stay anything using our default or "legacy" lighting, though it's fine with a supplied HDR (like our front page astronaut). You can see the problem here:
image
The problem appears to be that our generated lighting is wildly blown out, causing the renders to be highly overexposed. My best guess is that because our lighting scene contains a bunch of bright emissive objects to act as lights, and these never illuminated the room itself before, but they seem to be now. 

FYI @mrdoob I don't see anything in the release notes that points me to where this changed. I think I'll be asking for a point release next week, but I'm almost done for the day and I don't have any kind of solution yet. Let's pick it up Monday.


elalish:
Nevermind, figured it out: #4192.

Issue #4195, opened 2023-04-04
meshopt_decoder.js cannot be resolved due to unpkg.com being down


cpiggott:
### Description

At the time of this issue, Unpkg.com has now been down for almost 5-hours. Model-Viewer still references unpkg.com to pull the MeshOptimizer. Overall, jsDelivr.com seems to be a more stable option for pulling packages. This error can be seen by opening [Editor](https://modelviewer.dev/editor/)
#### Live Demo

https://modelviewer.dev/editor/

### Version

- model-viewer: v3.0.2 or Latest

### Browser Affected

- [X] Chrome, version: xx.x.xxxx.xx
- [X] Edge
- [X] Firefox
- [X] IE
- [X] Safari

### OS

- [X] Android
- [X] iOS
- [X] Linux
- [X] MacOS
- [X] Windows

### AR

- [ ] WebXR
- [ ] SceneViewer
- [ ] QuickLook

elalish:
Fixed by #4196

PR #4196, opened 2023-04-04
Remove references to unpkg.com and update to jsDelivr.com for stability

cpiggott:
### Reference Issue
Fixes: Issue #4195 

The MeshOptimizer cannot be resolved due to unpkg.com being down. This changes where meshopt_decoder.js is pulled from to jsDelivr.com which is a more stable option for a CDN.


elalish (approved):
Thanks! I switched us off unpkg earlier, but I must have missed these.

PR #4237, opened 2023-04-21
fix: Remove doubled period from start of alt text.

### LINKED ISSUE #4236: Alt text can add double periods
### Description

When you provide alt text using the `alt` attribute ". Use mouse, touch or arrow keys to move." is added to the `aria-label` added to a div wrapping the canvas. However, if the alt text provided ends in a "." the period ends up doubled. Resulting in something like `A 3D model of an astronaut.. Use mouse, touch or arrow keys to move.` 

#### Live Demo

https://glitch.com/edit/#!/soapy-quickest-leather

### Version

- model-viewer: v3.0.2 (version in my app)
- model-viewer: v3.1.0 (version in the repro)

### Browser Affected

- [x] Chrome, version: 112.0.5615.121 
- [x] Edge
- [x] Firefox
- [x] IE
- [x] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [x] MacOS
- [x] Windows

ashleahhill:
Removes ending period from custom alt text if it exists.

### Reference Issue
Fixes #4236

### DIFF
--- a/packages/model-viewer/src/features/controls.ts
+++ b/packages/model-viewer/src/features/controls.ts
@@ -808,7 +808,7 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
     }
 
     get[$ariaLabel]() {
-      return super[$ariaLabel] +
+      return super[$ariaLabel].replace(/\.$/, '') +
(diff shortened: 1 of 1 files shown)


elalish (approved):
Thanks!

PR #4259, opened 2023-04-27
three.js r152

elalish:
I'm not doing the [colorspace update](https://discourse.threejs.org/t/updates-to-color-management-in-three-js-r152/50791) yet, since the old way is only deprecated, but still works. I prefer to wait until the TS types are updated so we don't need to fill the code with `as any`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1538,18 +1538,18 @@
       }
     },
     "node_modules/@nrwl/cli": {
-      "version": "15.9.2",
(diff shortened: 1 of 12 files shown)


Beilinson:
Like you stated in the postprocessing pr the three versions will quickly get out of sync, I made a postinstall script that updates them to be matched to the package.json. #4260

elalish:
@Beilinson Thanks, works like a charm!

PR #4260, opened 2023-04-28
Added postinstall script that updates the three importmap version

Beilinson:
Whenever npm install is run in `packages/model-viewer`, `packages/model-viewer-effects/README.md` and the examples in `packages/modelviewer.dev/examples/postprocessing/index.html` are automatically updated to be in sync with the three version in `package.json`.

### DIFF
--- a/packages/model-viewer/package.json
+++ b/packages/model-viewer/package.json
@@ -45,6 +45,7 @@
   "typings": "lib/model-viewer.d.ts",
   "types": "lib/model-viewer.d.ts",
   "scripts": {
+    "postinstall": "node ../../scripts/postinstall.js",
     "clean": "rm -rf ./lib ./dist",
     "prepare": "if [ ! -L './shared-assets' ]; then ln -s ../shared-assets ./shared-assets; fi && ../shared-assets/scripts/fetch-khronos-gltf-samples.sh",
     "build": "npm run build:tsc && npm run build:rollup",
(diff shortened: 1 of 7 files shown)


packages/model-viewer/package.json
```diff
@@ -45,6 +45,7 @@
   "typings": "lib/model-viewer.d.ts",
   "types": "lib/model-viewer.d.ts",
   "scripts": {
+    "postinstall": "node ../../scripts/postinstall.js",
```
elalish, line 48: That's pretty magical, thanks! Should we add this into our `update:package-lock` script? That's what I run anytime I update our versions.

Beilinson: Checked it right now, it already runs since the update:package-lock runs npm install in every sub project.

elalish: Ah, great; I didn't realize `postinstall` was a standardized npm command. Thanks!

elalish:
We're starting to get requests already for post-processing in AR - would you like to take a look at hooking it up in WebXR? I'm hoping it shouldn't take too much.

Beilinson:
I'll open a separate issue for it with the information I wrote in the original pull request, I'm not sure I'll be able to test it however as I only have access to an Iphone which only experimentally supports WebXR, it would be preferable to test on android.

PR #4268, opened 2023-05-02
Playwright CI

elalish:
Since our Karma test runner is deprecated, we're switching to modern-web, which also recommends playwright as an alternative to BrowserStack, which is free and simpler, so let's give it a shot!

Turns out this was also a good opportunity to swap `jasmine` for `mocha` in the `space-opera` tests; it was driving me crazy that we were using two different test frameworks.


elalish:
Wonderful, Firefox doesn't work because of a 6-year-old bug: https://github.com/microsoft/playwright/issues/21783

packages/space-opera/src/test/animation_controls/animation_controls_test.ts
```diff
@@ -26,13 +27,13 @@ import {Dropdown} from '../../components/shared/dropdown/dropdown.js';
...
-const ANIMATED_GLB_PATH = '../base/shared-assets/models/RobotExpressive.glb';
+const ANIMATED_GLB_PATH = 'packages/shared-assets/models/RobotExpressive.glb';
 
-describe('animation controls test', () => {
+suite('animation controls test', () => {
   let preview: ModelViewerPreview;
   let animationControls: AnimationControls;
 
-  beforeEach(async () => {
+  setup(async () => {
```
elalish, line 36: We inherited the space-opera code base from another team that used `jasmine` as their test framework instead of `mocha`. I did a find-replace and then some more re-jiggering to get it all updated to be consistent with our other packages.

packages/model-viewer/src/features/controls.ts
```diff
@@ -594,8 +595,8 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
             pointerId: i - 5678,  // help ensure uniqueness
             pointerType: 'touch',
             target: inputElement,
-            clientX: width * position.x,
-            clientY: height * position.y,
+            clientX: width * position.x + rect.x,
+            clientY: height * position.y + rect.y,
```
elalish, line 599: This was an actual bug that our test didn't catch on our old CI, but did catch in this new framework, apparently because the default HTML has some padding now.

diegoteran (approved):

Issue #4273, opened 2023-05-09
Add PBR Next properties to our material API


elalish:
While model-viewer supports reading and rendering all the PBR Next extensions (thanks to three.js), we lack the ability to modify most of them through our [material API](https://modelviewer.dev/examples/scenegraph/#pickMaterialExample). Probably the easiest way to find everything, particularly the gltf - three.js equivalents is by perusing the extensions in [GLTFLoader](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/loaders/GLTFLoader.js#L683). 

Let's test these APIs by also adding them to our editor's materials panel, following the existing [conventions](https://github.com/google/model-viewer/blob/master/packages/space-opera/src/components/materials_panel/materials_panel.ts#L956). Once enabled, it'll be easy to test end-to-end by dropping the various glTF sample models (which can be found in `shared-assets`) into the editor and checking that the new textures show up and can be modified.

PR #4319, opened 2023-06-13
EmissiveStrength and Clearcoat extensions

diegoteran:
Change from MeshStandardMaterial to MeshPhysicalMaterial

Feature: #4273

### DIFF
--- a/packages/model-viewer/src/features/scene-graph/image.ts
+++ b/packages/model-viewer/src/features/scene-graph/image.ts
@@ -51,6 +51,12 @@ export class Image extends ThreeDOMElement implements ImageInterface {
     if (!this[$threeTexture].image.src) {
       this[$threeTexture].image.src = 'adhoc_image' + adhocNum++;
     }
+    if (!this[$threeTexture].image.name) {
+      this[$threeTexture].image.name =
+          (texture && texture.image && texture.image.src) ?
+          texture.image.src.split('/').pop() :
+          'adhoc_image';
+    }
   }
 
   get name(): string {
--- a/packages/model-viewer/src/features/scene-graph/image.ts
+++ b/packages/model-viewer/src/features/scene-graph/image.ts
@@ -16,18 +16,17 @@
 import {Mesh, MeshBasicMaterial, OrthographicCamera, PlaneGeometry, Scene, Texture as ThreeTexture, WebGLRenderTarget} from 'three';
 
 import {blobCanvas} from '../../model-viewer-base.js';
-import {Image as GLTFImage} from '../../three-components/gltf-instance/gltf-2.0.js';
 import {Renderer} from '../../three-components/Renderer.js';
 
 import {Image as ImageInterface} from './api.js';
-import {$correlatedObjects, $onUpdate, $sourceObject, ThreeDOMElement} from './three-dom-element.js';
+import {$correlatedObjects, $onUpdate, ThreeDOMElement} from './three-dom-element.js';
 
 
 const quadMaterial = new MeshBasicMaterial();
 const quad = new PlaneGeometry(2, 2);
-let adhocNum = 0;
 
 export const $threeTexture = Symbol('threeTexture');
+export const $threeTextures = Symbol('threeTextures');
 export const $applyTexture = Symbol('applyTexture');
 
 /**
@@ -41,30 +40,25 @@ export class Image extends ThreeDOMElement implements ImageInterface {
     return this[$correlatedObjects]?.values().next().value as ThreeTexture;
   }
 
+  get[$threeTextures](): Set<ThreeTexture> {
+    return this[$correlatedObjects] as Set<ThreeTexture>;
+  }
+
   constructor(
-      onUpdate: () => void, texture: ThreeTexture|null,
-      gltfImage: GLTFImage|null) {
-    gltfImage = gltfImage ?? {
-      name: (texture && texture.image && texture.image.src) ?
-          texture.image.src.split('/').pop() :
-          'adhoc_image',
-      uri: (texture && texture.image && texture.image.src) ?
-          texture.image.src :
-          'adhoc_image' + adhocNum++
-    };
-    super(onUpdate, gltfImage, new Set<ThreeTexture>(texture ? [texture] : []));
+      onUpdate: () => void, texture: ThreeTexture|null) {
+    super(onUpdate, new Set<ThreeTexture>(texture ? [texture] : []));
   }
 
   get name(): string {
-    return (this[$sourceObject] as GLTFImage).name || '';
+    return this[$threeTexture].image.name || '';
   }
 
   get uri(): string|undefined {
-    return (this[$sourceObject] as GLTFImage).uri;
+    return this[$threeTexture].image.src;
   }
 
   get bufferView(): number|undefined {
-    return (this[$sourceObject] as GLTFImage).bufferView;
+    return this[$threeTexture].image.bufferView;
   }
 
   get element(): HTMLVideoElement|HTMLCanvasElement|undefined {
@@ -88,7 +82,9 @@ export class Image extends ThreeDOMElement implements ImageInterface {
   }
 
   set name(name: string) {
-    (this[$sourceObject] as GLTFImage).name = name;
+    for (const texture of this[$threeTextures]) {
+      texture.image.name = name;
+    }
   }
 
   update() {
--- a/packages/model-viewer/src/features/scene-graph/image.ts
+++ b/packages/model-viewer/src/features/scene-graph/image.ts
@@ -35,9 +35,6 @@ export const $applyTexture = Symbol('applyTexture');
  */
 export class Image extends ThreeDOMElement implements ImageInterface {
   get[$threeTexture]() {
-    console.assert(
-        this[$correlatedObjects] != null && this[$correlatedObjects]!.size > 0,
-        'Image correlated object is undefined');
     return this[$correlatedObjects]?.values().next().value as ThreeTexture;
   }
 
@@ -45,7 +42,7 @@ export class Image extends ThreeDOMElement implements ImageInterface {
     return this[$correlatedObjects] as Set<ThreeTexture>;
   }
 
-  constructor(onUpdate: () => void, texture: ThreeTexture|null) {
+  constructor(onUpdate: () => void, texture: ThreeTexture) {
     super(onUpdate, new Set<ThreeTexture>(texture ? [texture] : []));
 
     if (!this[$threeTexture].image.src) {
--- a/packages/model-viewer/src/features/scene-graph/image.ts
+++ b/packages/model-viewer/src/features/scene-graph/image.ts
@@ -24,6 +24,7 @@ import {$correlatedObjects, $onUpdate, ThreeDOMElement} from './three-dom-elemen
 
 const quadMaterial = new MeshBasicMaterial();
 const quad = new PlaneGeometry(2, 2);
+let adhocNum = 0;
 
 export const $threeTexture = Symbol('threeTexture');
 export const $threeTextures = Symbol('threeTextures');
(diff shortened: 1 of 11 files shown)


diegoteran:
Looking for feedback on the correct way to set up the API before completing other PBR next properties + unit tests.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -340,4 +381,78 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+  get clearcoatNormalScale(): number {
+    this[$ensureMaterialIsLoaded]();
+    return this[$backingThreeMaterial].clearcoatNormalScale.x;
+  }
+
+  setEmissiveStrength(emissiveStrength: number) {
+    this[$ensureMaterialIsLoaded]();
+    for (const material of this[$correlatedObjects] as
+         Set<MeshPhysicalMaterial>) {
+      material.emissiveIntensity = emissiveStrength;
```
elalish, line 427: LGTM, but I believe these are simplified compared to the existing setters right (recall our duplicated state)? We should change the others to match this style and check to ensure that doesn't break anything before we adopt it for these. Let's add that in before we get too deep into testing these new ones.

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -41,30 +41,28 @@ export class Image extends ThreeDOMElement implements ImageInterface {
...
-      name: (texture && texture.image && texture.image.src) ?
-          texture.image.src.split('/').pop() :
-          'adhoc_image',
-      uri: (texture && texture.image && texture.image.src) ?
-          texture.image.src :
-          'adhoc_image' + adhocNum++
-    };
-    super(onUpdate, gltfImage, new Set<ThreeTexture>(texture ? [texture] : []));
+  get[$threeTextures](): Set<ThreeTexture> {
+    return this[$correlatedObjects] as Set<ThreeTexture>;
```
elalish, line 45: Were you still planning to rename `$correlatedObjects`?

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -41,30 +41,28 @@ export class Image extends ThreeDOMElement implements ImageInterface {
...
+  }
+
+  constructor(onUpdate: () => void, texture: ThreeTexture|null) {
+    super(onUpdate, new Set<ThreeTexture>(texture ? [texture] : []));
+
+    if (!this[$threeTexture].image.src) {
+      this[$threeTexture].image.src = 'adhoc_image' + adhocNum++;
```
elalish, line 52: Should this be the `name` as well, as before?

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -79,65 +81,34 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
     const correlatedMaterials =
         this[$correlatedObjects] as Set<MeshPhysicalMaterial>;
 
-    if (gltfMaterial.extensions &&
-        gltfMaterial.extensions['KHR_materials_pbrSpecularGlossiness']) {
-      console.warn(`Material ${gltfMaterial.name} uses a deprecated extension
-          "KHR_materials_pbrSpecularGlossiness", please use
-          "pbrMetallicRoughness" instead. Specular Glossiness materials are
-          no longer supported; to convert to metal-rough, see 
-          https://www.donmccurdy.com/2022/11/28/converting-gltf-pbr-materials-from-specgloss-to-metalrough/.`);
```
elalish, line 97: I'm a little sad to loose this warning, but I think it's okay because three.js throws its own anyway.

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -277,7 +232,9 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
 
   get emissiveFactor(): RGB {
     this[$ensureMaterialIsLoaded]();
-    return (this[$sourceObject] as DefaultedMaterial).emissiveFactor;
+    return (
+        this[$backingThreeMaterial]
+            .emissive.toArray() as [number, number, number]);
```
elalish, line 237: Does it not work with `as RGB`?

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -305,18 +262,29 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
          Set<MeshPhysicalMaterial>) {
       material.emissive.set(color);
     }
-    (this[$sourceObject] as DefaultedMaterial).emissiveFactor =
-        color.toArray() as [number, number, number];
     this[$onUpdate]();
   }
 
+  [$getAlphaMode](): string {
+    // Follows implementation of GLTFExporter from three.js
```
elalish, line 269: 👍

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -348,13 +321,13 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
     }
-    (this[$sourceObject] as DefaultedMaterial).doubleSided = doubleSided;
     this[$onUpdate]();
   }
 
   getDoubleSided(): boolean {
     this[$ensureMaterialIsLoaded]();
-    return (this[$sourceObject] as DefaultedMaterial).doubleSided;
+    // 0 for FrontSide, 2 for DoubleSide.
+    return this[$backingThreeMaterial].side == 2;
```
elalish, line 330: Use the `DoubleSided` enum from three, or whatever it's called.

packages/model-viewer/src/features/scene-graph/model.ts
```diff
@@ -281,14 +271,6 @@ export class Model implements ModelInterface {
     const threeMaterialSet =
         material[$correlatedObjects] as Set<MeshPhysicalMaterial>;
 
-    // clones the gltf material data and updates the material name.
-    const gltfSourceMaterial =
-        JSON.parse(JSON.stringify(material[$sourceObject])) as GLTFMaterial;
-    gltfSourceMaterial.name = newMaterialName;
-    // Adds the source material clone to the gltf def.
-    const gltf = this[$correlatedSceneGraph].gltf;
-    gltf.materials!.push(gltfSourceMaterial);
```
elalish, line 290: 💯  Always nice to remove code!

packages/model-viewer/src/features/scene-graph/pbr-metallic-roughness.ts
```diff
@@ -40,61 +37,43 @@ export class PBRMetallicRoughness extends ThreeDOMElement implements
...
-    if (pbrMetallicRoughness.roughnessFactor == null) {
-      pbrMetallicRoughness.roughnessFactor = 1;
-    }
-    if (pbrMetallicRoughness.metallicFactor == null) {
-      pbrMetallicRoughness.metallicFactor = 1;
-    }
+  private get[$threeMaterial]() {
+    console.assert(
+        this[$correlatedObjects] != null && this[$correlatedObjects]!.size > 0,
+        'Sampler correlated object is undefined');
```
elalish, line 43: Is this a Sampler?

packages/model-viewer/src/features/scene-graph/sampler.ts
```diff
@@ -49,29 +49,29 @@ const isWrapMode = (() => {
              wrapModes.indexOf(value as WrapMode) > -1;
 })();
 
-const isValidSamplerValue = <P extends 'minFilter'|'magFilter'|'wrapS'|'wrapT'|'rotation'|'repeat'|'offset'>(
-    property: P, value: unknown): value is DefaultedSampler[P] => {
```
elalish, line 53: I missed this `value is` thing before - I've never seen that syntax. I know it's not yours, but if there's a simpler way, that would be great. Seems like it should just be `P, value: DefaultedSampler<P>): boolean =>`

elalish (changes requested):
Looking good, just a few things.

packages/model-viewer/src/features/scene-graph/image.ts
diegoteran: I think the "correlated" name comes from the default correlation through the gltfMap from ThreeJS Loader: https://github.com/google/model-viewer/blob/8546f41450af57ee37fe7c00e3ef210138f6e3a0/packages/model-viewer/src/features/scene-graph/model.ts#L96

I'm not sure if the name should change.

packages/model-viewer/src/features/scene-graph/material.ts
diegoteran: RIP

packages/model-viewer/src/features/scene-graph/image.ts
diegoteran: Done

packages/model-viewer/src/features/scene-graph/pbr-metallic-roughness.ts
diegoteran: ahh good copy-paste catch

packages/model-viewer/src/features/scene-graph/sampler.ts
diegoteran: It says DefaultedSampler is not generic. Probably no support for <P>?

packages/model-viewer/src/features/scene-graph/image.ts
```diff
@@ -41,30 +41,34 @@ export class Image extends ThreeDOMElement implements ImageInterface {
...
-  constructor(
-      onUpdate: () => void, texture: ThreeTexture|null,
-      gltfImage: GLTFImage|null) {
-    gltfImage = gltfImage ?? {
-      name: (texture && texture.image && texture.image.src) ?
+  get[$threeTextures](): Set<ThreeTexture> {
+    return this[$correlatedObjects] as Set<ThreeTexture>;
+  }
+
+  constructor(onUpdate: () => void, texture: ThreeTexture|null) {
```
elalish, line 48: I know this predates you, but do we know when `|null` happens? There's an assert above that seems to imply `null` would be an error. Doesn't have to happen in this PR, but it would be great to clean this up if we can.

packages/model-viewer/src/test/features/scene-graph/material-spec.ts
```diff
@@ -249,11 +248,6 @@ suite('scene-graph/material', () => {
       document.body.removeChild(element);
     });
 
-    test('Accessing the name getter does not cause throw error.', async () => {
-      expect(model.materials[2].name).to.equal('red');
-      expect(model.materials[2][$lazyLoadGLTFInfo]).to.be.ok;
-    });
```
elalish, line 255: Can we leave this now that `name` is fixed?

packages/model-viewer/src/test/features/scene-graph/model-spec.ts
```diff
@@ -150,6 +150,7 @@ suite('scene-graph/model', () => {
 
       test('getMaterialByName returns material when name exists', async () => {
         await loadModel(CUBES_GLTF_PATH);
+        await model.materials[2].ensureLoaded();
```
elalish, line 153: Likewise, is this necessary anymore?

elalish (approved):

packages/model-viewer/src/features/scene-graph/image.ts
diegoteran: Yes, or at least it should work the same way as https://github.com/google/model-viewer/blob/8546f41450af57ee37fe7c00e3ef210138f6e3a0/packages/model-viewer/src/features/scene-graph/sampler.ts#L80C2-L100C56

Maybe another refactor where we remove `|null` from all ThreeDOMElements (Sampler, Image, Texture, etc...) is something worth looking into.

elalish: Yes, I think that would be ideal.

elalish (approved):
This looks great, thanks for helping to clean up our code base!

PR #4348, opened 2023-07-07
Add pbr iridescence API

diegoteran:
Follows threejs implementation.


elalish (dismissed):
Thanks!

packages/model-viewer/src/features/scene-graph/material.ts
```diff
@@ -592,4 +592,71 @@ export class Material extends ThreeDOMElement implements MaterialInterface {
...
+    this[$ensureMaterialIsLoaded]();
+    return this[$backingThreeMaterial].iridescenceThicknessRange[0];
+  }
+
+  get iridescenceThicknessMaximum(): number {
+    this[$ensureMaterialIsLoaded]();
+    return this[$backingThreeMaterial].iridescenceThicknessRange[1];
+  }
+
+  get iridescenceThicknessTexture(): TextureInfo {
```
elalish, line 620: I just noticed this function signature doesn't match the API: one returns `TextureInfo`, while the other returns `TextureInfo|null`. I noticed because of the `!` below. This is probably true of the rest as well - let's make sure we know which is correct and then make them match.

elalish: Have these been added to the pbr map with `createTextureInfo`? Maybe we need something more type-safe than a map...

diegoteran: yes, these are all added at $initialize

elalish (approved):
Thanks! LGTM.

PR #4363, opened 2023-07-12
check postinstall file exists https://github.com/google/model-viewer/…

btopro:
…issues/4362

### Reference Issue

Should fix #4362 introduced in version 3.2.0

This is a highly tooling specific script request which in `yarn` will cascade when projects implement 1 of the packages but not the entire mono-repo. I am using the model viewer tag in my own monorepo and yarn will call all the `postinstall` hooks it finds, looking then for `whatever/node_modules/scripts/postinstall.js` which is not a thing for downstream consumers.

I can't install w/ a fresh lock file currently but this is pattern off the other verifications of a file existing to avoid errors.

PR #4368, opened 2023-07-14
Remove lerna

### LINKED ISSUE #4380: npm update for 3.2.0 fails on Windows 10: can't find postinstall.js
### Description
I've been using model-viewer for a year with no trouble in an Angular app, this shows up only with the new 3.2.0 release. I see that there's a new script... but it's not being found. Take a look at _where_ it's trying to find it though: is it about two directories too high?

Perhaps there's a difference between OSs where the previous command is leaving the working directory?

Perhaps there's a slight OS difference in the working directory for the command vs the execution?
`"postinstall": "if [ -f '../../scripts/postinstall.js' ]; then node ../../scripts/postinstall.js; fi",`

I assume that the intent here is to be finding 
D:\_mystuff_\node_modules@google\model-viewer\scripts\postinstall.js 

But if I do a 
`npm pack --dry-run @google/model-viewer`
there isn't a scripts directory, or the postinstall.js script coming down anywhere. And to that, one could reasonably say "well, it's only supposed to try to run it **if** it finds it" and yeah, I don't have an answer for you there.

Additional notes: 
* same result with and without the "three" package installed
* same result in cmd and powershell, regular user and admin
* same result for npm update and npm install if I scrap node_modules and package_lock

image

### Version
- model-viewer: v3.2.0

### Browser Affected
None, npm install issue

### OS
- [x] Windows 10

### Env
npm: 9.6.7
node: 18.12.1

elalish:
Fixes #4380

The main purpose of this was to fix the postinstall script that's been giving us trouble - I think the main problem is it really needs to be at the root level rather than in the model-viewer package. However, during this update I found myself in a nasty catch-22 where the bootstrap process was somehow getting the old package.json and still looking for a non-existant postinstall script. I remembered that lerna bootstrap has been deprecated for some time, so figured this might be the right time to update to npm workspaces. 

Since this is related to #4260 and #4363, @Beilinson and @btopro can you review this for me?


Beilinson:
I looked into the issues and this seems like the best solution to me as well. Didn't even think about the yarn issue but it totally makes sense.

tobimori:
still running into this issue, would be great to get some progress here

hilyas786:
> still running into this issue, would be great to get some progress here 

Same using NPM, can't install v3

PR #4405, opened 2023-08-01
await openIOSARQuickLook when activating AR

### LINKED ISSUE #4404: openIOSARQuickLook is not awaited in activateAR
The `openIOSARQuickLook` function that's called internally inside `activateAR()` is asynchronous, but it is not awaited.
When an iOS src is provided it looks like it will resolve immediately, but when generating a local USDZ version of a GLTF file it will resolve after that process is complete.

https://github.com/google/model-viewer/blob/6af7342d5bc8cde108636b2d95ecbc71a596de93/packages/model-viewer/src/features/ar.ts#L369-L374

https://github.com/google/model-viewer/blob/6af7342d5bc8cde108636b2d95ecbc71a596de93/packages/model-viewer/src/features/ar.ts#L197-L202

heatblayze:
### Reference Issue
Fixes #4404

The function `openIOSARQuickLook` is asynchronous and should be awaited.

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -197,7 +197,7 @@ export const ARMixin = <T extends Constructor<ModelViewerElementBase>>(
     async activateAR() {
       switch (this[$arMode]) {
         case ARMode.QUICK_LOOK:
-          this[$openIOSARQuickLook]();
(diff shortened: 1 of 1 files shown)


elalish (approved):
Good catch! I can't believe I missed that.

PR #4412, opened 2023-08-02
Add zoom sensitivity and pan sensitivity attributes

nicolas-daures:
Fixes https://github.com/google/model-viewer/discussions/4401

When the distance between the minimum radius and the maximum radius is big, the mousewheel zooms too much. Exposing the zoom-sensitivity and pan-sensitivity attributes would allow for better control.


elalish (approved):
Looks great, thank you!

PR #4436, opened 2023-08-23
Update scenegraph examples

samaneh-kazemi:
### Reference Issue

in the browser
2. Go to Create Textures link
3. Change the normals and play around with the model in 3D and AR
4. Please try damaged helmet , lantern pole and water bottle options.

Expected Result:
-Asset should launch and render properly with Laptops & Mobile devices
-The text values of all options should be observed inside the screen.

Actual Result:
Belonging to the selected model; the picture name part does not fit on the screen and remains at the bottom of the screen.


elalish (approved):
Thanks!

PR #4473, opened 2023-09-21
Convert to ThreeJS.Vector2 internally in public api.

diegoteran:
Fixes #4403

### DIFF
--- a/packages/model-viewer/src/features/scene-graph/api.ts
+++ b/packages/model-viewer/src/features/scene-graph/api.ts
@@ -32,9 +32,9 @@ export declare interface ThreeDOMElementMap {
 }
 
 /** A 2D Cartesian coordinate */
-export interface Vector2D {
-  x: number;
-  y: number;
+export interface Vector2DInterface {
+  u: number;
+  v: number;
 }
 
 /**
@@ -334,12 +334,12 @@ export declare interface Sampler {
   /**
    * The texture scale.
    */
-  readonly scale: Vector2D|null;
+  readonly scale: Vector2DInterface|null;
 
   /**
    * The texture offset.
    */
-  readonly offset: Vector2D|null;
+  readonly offset: Vector2DInterface|null;
 
   /**
    * Configure the minFilter value of the Sampler.
@@ -371,12 +371,12 @@ export declare interface Sampler {
    * Sets the texture scale, or resets it to (1, 1) if argument is null.
    * As the scale value increases, the repetition of the texture will increase.
    */
-  setScale(scale: Vector2D|null): void;
+  setScale(scale: Vector2DInterface|null): void;
 
   /**
    * Sets the texture offset, or resets it to (0, 0) if argument is null.
    */
-  setOffset(offset: Vector2D|null): void;
+  setOffset(offset: Vector2DInterface|null): void;
 }
 
 
--- a/packages/model-viewer/src/features/scene-graph/api.ts
+++ b/packages/model-viewer/src/features/scene-graph/api.ts
@@ -32,7 +32,7 @@ export declare interface ThreeDOMElementMap {
 }
 
 /** A 2D Cartesian coordinate */
-export interface Vector2 {
+export interface Vector2D {
   x: number;
   y: number;
 }
@@ -334,12 +334,12 @@ export declare interface Sampler {
   /**
    * The texture scale.
    */
-  readonly scale: Vector2|null;
+  readonly scale: Vector2D|null;
 
   /**
    * The texture offset.
    */
-  readonly offset: Vector2|null;
+  readonly offset: Vector2D|null;
 
   /**
    * Configure the minFilter value of the Sampler.
@@ -371,12 +371,12 @@ export declare interface Sampler {
    * Sets the texture scale, or resets it to (1, 1) if argument is null.
    * As the scale value increases, the repetition of the texture will increase.
    */
-  setScale(scale: Vector2|null): void;
+  setScale(scale: Vector2D|null): void;
 
   /**
    * Sets the texture offset, or resets it to (0, 0) if argument is null.
    */
-  setOffset(offset: Vector2|null): void;
+  setOffset(offset: Vector2D|null): void;
 }
 
 
--- a/packages/model-viewer/src/features/scene-graph/sampler.ts
+++ b/packages/model-viewer/src/features/scene-graph/sampler.ts
@@ -15,10 +15,11 @@
 
 import {Texture as ThreeTexture, Vector2} from 'three';
 
+import {toVector2D} from '../../model-viewer-base.js';
 import {Filter, MagFilter, MinFilter, Wrap, WrapMode} from '../../three-components/gltf-instance/gltf-2.0.js';
 import {Sampler as DefaultedSampler} from '../../three-components/gltf-instance/gltf-defaulted.js';
 
-import {Sampler as SamplerInterface, Vector2D} from './api.js';
+import {Sampler as SamplerInterface, Vector2DInterface} from './api.js';
 import {$correlatedObjects, $onUpdate, ThreeDOMElement} from './three-dom-element.js';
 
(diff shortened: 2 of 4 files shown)


packages/model-viewer/src/features/scene-graph/sampler.ts
```diff
@@ -18,7 +18,7 @@ import {Texture as ThreeTexture, Vector2} from 'three';
 import {Filter, MagFilter, MinFilter, Wrap, WrapMode} from '../../three-components/gltf-instance/gltf-2.0.js';
 import {Sampler as DefaultedSampler} from '../../three-components/gltf-instance/gltf-defaulted.js';
 
-import {Sampler as SamplerInterface} from './api.js';
+import {Sampler as SamplerInterface, Vector2 as Vector2Interface} from './api.js';
```
elalish, line 21: Maybe we should just rename `Vector2` to avoid this confusion with three.js? Also, we have a `Vector2D` and `Vector3D` in model-viewer-base.js - can we be consistent with those (accept them as valid input)?

packages/model-viewer/src/features/scene-graph/sampler.ts
```diff
@@ -145,20 +145,20 @@ export class Sampler extends ThreeDOMElement implements SamplerInterface {
...
   }
 
-  setScale(scale: Vector2|null): void {
+  setScale(scale: Vector2Interface|null): void {
     if (scale == null) {
       // Reset scale.
       scale = new Vector2(1, 1);
     }
-    this[$setProperty]('repeat', scale);
+    this[$setProperty]('repeat', new Vector2(scale.x, scale.y));
```
elalish, line 153: Do we have an example that tests this? Did it work before?

elalish (changes requested):

packages/model-viewer/src/features/scene-graph/sampler.ts
diegoteran: Yes we have a transform Textures example. It did work before on android and Mac, still works. I need to test with iOS.

packages/model-viewer/src/features/scene-graph/sampler.ts
diegoteran: I initially tried to use Vector2D from model-viewer-base in the api.js file and it wouldn't work. I can give it another go at trying out deleting the api.js Vector2 interface.

`rename Vector2 to avoid this confusion with three.js? `
You mean rename the Vector2 in api.js?

elalish: I think the interface is necessary so that users don't have to include a `toString()`. And yes, I mean renaming in api.js. Also check the members for compatibility with `Vector2D`.

packages/model-viewer/src/features/scene-graph/sampler.ts
elalish: Wow, I guess three.js must automatically turn a scalar into a vector?

packages/model-viewer/src/features/scene-graph/sampler.ts
diegoteran: Yeah Vector2D from base.js requires additional imports. Just uploaded a new commit

packages/model-viewer/src/features/scene-graph/api.ts
```diff
@@ -32,7 +32,7 @@ export declare interface ThreeDOMElementMap {
 }
 
 /** A 2D Cartesian coordinate */
-export interface Vector2 {
+export interface Vector2D {
```
elalish, line 35: Okay, so now instead of shadowing three's construct, you're shadowing our existing public class from model-viewer-base. Any reason not to use `Vector2DInterface` like you did before?

elalish, line 35: Also, I think you may have missed this one:

> Also check the members for compatibility with Vector2D.

I believe they're `u` and `v`, not `x` and `y`?

elalish (changes requested):

packages/model-viewer/src/features/scene-graph/api.ts
diegoteran: Not really, I think this is trying to be an interface for threejs's vector2, so maybe Vector2Interface is better?

diegoteran: Do we want that? I though the purpose of this Vector Interface was to not expose threejs' vector2, and ThreeJs Vector2 uses `x` and `y`. It seems to me that this interface and `Vector2D` in base have different purposes. wdyt?

elalish: Sure, SG

elalish: Exactly, we have zero interest in an API consistent with three.js (we're hiding their API). We want an API that is self-consistent and consistent with glTF. We already have an API called Vector2D that uses u and v, so using x and y here would be confusing. Besides, this feature is about textures, which are mapped in a u-v space.

diegoteran: got it, sgtm

packages/model-viewer/src/features/scene-graph/sampler.ts
```diff
@@ -113,11 +113,11 @@ export class Sampler extends ThreeDOMElement implements SamplerInterface {
     return this[$threeTexture].rotation;
   }
 
-  get scale(): Vector2 {
+  get scale(): Vector2D {
     return this[$threeTexture].repeat;
   }
 
-  get offset(): Vector2|null {
+  get offset(): Vector2D|null {
```
elalish, line 120: I just realized this also needs to change - you're returning a pointer to an internal three.js construct, which means they could then use it to change out internals (in JS every object is pass by reference). You need to use our `toVector2D()` to make a proper copy.

packages/model-viewer/src/test/features/scene-graph/texture-info-spec.ts
```diff
@@ -99,10 +98,8 @@ suite('scene-graph/texture-info', () => {
                                        .pbrMetallicRoughness['baseColorTexture']
                                        .texture?.sampler!;
           expect(exported_sampler.rotation).to.be.eq(0.1, 'rotation');
-          expect(exported_sampler.offset)
-              .to.be.eql(new Vector2(0.2, 0.3), 'offset');
-          expect(exported_sampler.scale)
-              .to.be.eql(new Vector2(0.4, 0.5), 'scale');
+          expect(exported_sampler.offset).to.be.eql({u: 0.2, v: 0.3}, 'offset');
```
elalish, line 101: I think the trouble is you're actually returning a `Vector2D` which has a toString method. I would just separate these lines to check the u and v separately.

packages/model-viewer/src/features/scene-graph/sampler.ts
```diff
@@ -113,12 +114,12 @@ export class Sampler extends ThreeDOMElement implements SamplerInterface {
...
 
-  get scale(): Vector2 {
-    return this[$threeTexture].repeat;
+  get scale(): Vector2DInterface {
+    return toVector2D(this[$threeTexture].repeat);
   }
 
-  get offset(): Vector2|null {
-    return this[$threeTexture].offset;
+  get offset(): Vector2DInterface|null {
```
elalish, line 121: nit: unlike set, here you actually get a real `Vector2D`.

diegoteran: do we want the set and get to have different parameter/return value type?

elalish: Technically that's the most accurate - we give a specific object back, but accept a broader interface class as input.

elalish (approved):
Thanks! Mind poking around other code you've worked on just to see if you notice any similar issues?

PR #4489, opened 2023-10-02
r157

elalish:
Updated three.js. The new types are pickier about typing events, so this is updated to fix that. Also removed an unnecessary testing-only event.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5020,9 +5020,9 @@
       "dev": true
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001542",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001542.tgz",
-      "integrity": "sha512-UrtAXVcj1mvPBFQ4sKd38daP8dEcXXr5sQe6QNNinaPd0iA/cxg9/l3VrSdL73jgw5sKyuQ6jNgiKO12W3SsVA==",
+      "version": "1.0.30001543",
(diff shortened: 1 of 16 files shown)


elalish:
@Beilinson I'm finding an odd test failure upon updating three.js from 156 to 157. It seems that when the ColorGrade is applied and a render is triggered it somehow makes a render target without a colorspace and then segfaults. However, all the examples seem to work fine, so it seems to be specific to the tests. Any thoughts?

Beilinson:
Does that still happen if you upgrade postprocessing to the latest version? 

Looking at the r157 commit it seems some preliminary support for wide gamut p3 color space was added, maybe the issue is related to that?

donmccurdy:
Hm, I haven't heard of any regressions yet, and the changes in r157 _shouldn't_ have any effect on normal usage... what do you mean by "undefined" color space? Is that `THREE.NoColorSpace`, or something else?

donmccurdy:
`.colorSpace = undefined` (note capital "S") is not a valid state. If something within three.js is causing that assignment, I would consider it a bug in three.js. If something in `postprocessing` or model-viewer causes that assignment, I believe it would be a bug in those projects.

elalish:
Okay, I figured it out - turns out it was because it was grabbing r148 as a peer dependency in the tests instead of r157. I have no idea why - a different workspace package was set to use r148, but this one specified r157, yet it somehow got 148 anyway. Feels like an npm workspaces bug, but I worked around it by updating so that all our packages use the same three.js version. If it breaks three-gpu-pathtracer, such is life...

diegoteran (approved):
awesome

PR #4499, opened 2023-10-05
Add fix for Wrapping exposed

diegoteran:
Fixed #4494

### DIFF
--- a/packages/model-viewer/src/three-components/gltf-instance/gltf-2.0.ts
+++ b/packages/model-viewer/src/three-components/gltf-instance/gltf-2.0.ts
@@ -1,4 +1,4 @@
-import {TextureFilter, Wrapping} from 'three';
+import {TextureFilter} from 'three';
 
 export type RGB = [number, number, number];
 
@@ -111,7 +111,7 @@ export type ClampToEdgeWrap = Wrap.ClampToEdge;
 export type MirroredRepeatWrap = Wrap.MirroredRepeat;
 export type RepeatWrap = Wrap.Repeat;
 
-export type WrapMode = RepeatWrap|ClampToEdgeWrap|MirroredRepeatWrap|Wrapping;
+export type WrapMode = RepeatWrap|ClampToEdgeWrap|MirroredRepeatWrap;
 
 export interface Sampler {
   name?: string;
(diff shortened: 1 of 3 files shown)


packages/model-viewer/src/three-components/gltf-instance/gltf-2.0.ts
```diff
@@ -1,4 +1,4 @@
-import {TextureFilter, Wrapping} from 'three';
+import {TextureFilter} from 'three';
```
elalish, line 1: We need to remove this too.

elalish (changes requested):
Can you update your Transform Textures example to demonstrate this is working now? Just add a little drop-down for the wrap mode and apply same to both axes.

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -499,6 +499,12 @@ <h2 class="demo-title">Transform textures</h2>
       <input type="range" min="0.5" max="1.5" value="1" step="0.01" id="scaleSlider">
       <p>Offset</p>
       <input type="range" min="0" max="1" value="0" step="0.01" id="offsetSlider">
+      <p>WrapMode</p>
+      <select id="wrapMode">
+        <option value="10497">Repeat</option>
```
elalish, line 504: Can you switch this to `Wrap.Repeat` etc? I want to ensure our enum is properly exported.

elalish: If it doesn't work, we probably have to re-export it from model-viewer.ts.

elalish (approved):
Thanks

packages/model-viewer/src/three-components/gltf-instance/gltf-2.0.ts
diegoteran: Leaving this for a follow up PR.

packages/modelviewer.dev/examples/scenegraph/index.html
diegoteran: We realized at the office that making this work might take a bit more time of work, leaving it for another index.html cleanup :/

PR #4500, opened 2023-10-05
updated encoding to colorspace

elalish:
I just noticed the new three.js deprecated an API we were using - this should keep warnings out of our console.


diegoteran (dismissed):
nice

samaneh-kazemi (dismissed):

PR #4501, opened 2023-10-06
Material example update

elalish:
Switching cube models, as this one makes it clearer what happens in the case of clamp to edge - thanks @samaneh-kazemi for bringing it to my attention. I also fixed a few small typos in our mode API docs. 
image


packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -829,9 +823,9 @@ <h2 class="demo-title">Materials API</h2>
   update(): void;
 }
 
-interface Vector2 {
-  readonly x: number;
-  readonly y: number;
+interface Vector2D {
```
diegoteran, line 826: nice catch

diegoteran (approved):

PR #4528, opened 2023-10-24
Added example to toggle between mesh variants

samaneh-kazemi:
### Reference Issue

link to [glitch](https://glitch.com/edit/#!/toggle-mesh-variants):


packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -560,6 +560,111 @@ <h2 class="demo-title">Transform textures</h2>
...
+const mvTextures = document.querySelector("model-viewer#manifold");
+const textures = document.querySelector('#textures');
+
+mvTextures.addEventListener("load", () => {
+  const materials = mvTextures.model.materials
+  for(const material of materials) {
+    const option = document.createElement('option')
+    option.value = material.name;
+    option.textContent = material.name;
+    textures.appendChild(option);
```
elalish, line 588: If it's possible to do something declaratively (HTML) rather than procedurally (JS), always prefer HTML. Let's put these options in the `<div>` above.

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -560,6 +560,111 @@ <h2 class="demo-title">Transform textures</h2>
...
+  }
+
+  let startTimestamp;
+  const animationDuration = 1000; // 1 second
+  let alpha0 = 1;
+  let alpha1 = 0;
```
elalish, line 594: Do these need to be declared outside the function? We should do our best to avoid `let`.

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -560,6 +560,111 @@ <h2 class="demo-title">Transform textures</h2>
...
+
+  function updateAlphaValues(texture0, texture1, elapsed) {
+      alpha0 = 1 - (elapsed / animationDuration);
+      alpha1 = elapsed / animationDuration;
+      texture0.pbrMetallicRoughness.setBaseColorFactor([
+        texture0.pbrMetallicRoughness.baseColorFactor[0],
+        texture0.pbrMetallicRoughness.baseColorFactor[1],
+        texture0.pbrMetallicRoughness.baseColorFactor[2],
```
elalish, line 602: This is pretty verbose - have you considered something like this?
```
const pbr = texture0.pbrMetallicRoughness;
const baseColor = pbr.baseColorFactor;
baseColor[3] = alpha0;
pbr.setBaseColorFactor(baseColor);
```

elalish: Since you do it twice, you might also change this function to `updateAlphaValue`, then call it once for each texture.

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -560,6 +560,111 @@ <h2 class="demo-title">Transform textures</h2>
...
+        texture1.pbrMetallicRoughness.baseColorFactor[0],
+        texture1.pbrMetallicRoughness.baseColorFactor[1],
+        texture1.pbrMetallicRoughness.baseColorFactor[2],
+        alpha1,
+      ]);
+    }
+
+    mvTextures.addEventListener("load", () => {
+      const materials = mvTextures.model.materials;
+      for (const material of materials) {
```
elalish, line 615: This function appears to be a duplicate?

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -560,6 +560,111 @@ <h2 class="demo-title">Transform textures</h2>
...
+        const option = document.createElement('option');
+        option.value = material.name;
+        option.textContent = material.name;
+        textures.appendChild(option);
+      }
+    });
+
+    textures.addEventListener('input', (event) => {
+      const selectedValue = event.target.value;
+      const ball = mvTextures.model.materials[0];
```
elalish, line 625: Let's use `model.getMaterialByName` instead - if you put the names in your HTML options you can even iterate over those.

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -560,6 +560,111 @@ <h2 class="demo-title">Transform textures</h2>
...
+      } else {
+        texture0 = ball;
+        texture1 = stick;
+      }
+
+      startTimestamp = null; // Reset the start timestamp
+
+      function animate(currentTimestamp) {
+        if (!startTimestamp) {
+          startTimestamp = currentTimestamp;
```
elalish, line 642: Can you set the `startTimestamp` outside the function instead of in here?

packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -560,6 +560,111 @@ <h2 class="demo-title">Transform textures</h2>
...
+        option.value = material.name;
+        option.textContent = material.name;
+        textures.appendChild(option);
+      }
+    });
+
+    textures.addEventListener('input', (event) => {
+      const selectedValue = event.target.value;
+      const ball = mvTextures.model.materials[0];
+      ball.setAlphaMode('BLEND')
```
elalish, line 626: `BLEND` mode is actually more costly in the renderer - let's revert both materials to `OPAQUE` once the animation has finished. A comment to that effect would also be nice, to instruct our users.

elalish (changes requested):
Nice work, the functionality is great! And thanks for including the Glitch link - very helpful for review. 

Since this is a public example, code clarity and simplicity are of the utmost importance - I've put in several pointers, but please look through for any ways you can make it simpler and easier to read.

packages/modelviewer.dev/examples/scenegraph/index.html
samaneh-kazemi: Thanks for the feedback. Updated the select to have the options declaratively.

packages/modelviewer.dev/examples/scenegraph/index.html
samaneh-kazemi: I moved them around. Lmk what you think.

packages/modelviewer.dev/examples/scenegraph/index.html
samaneh-kazemi: Great suggestion, thank you! Sent a new commit to reflect these updates.

packages/modelviewer.dev/examples/scenegraph/index.html
samaneh-kazemi: Great catch. Thanks.

packages/modelviewer.dev/examples/scenegraph/index.html
samaneh-kazemi: I used `model.getMaterialByName` lmk if you prefer the other approach.

packages/modelviewer.dev/examples/scenegraph/index.html
samaneh-kazemi: I wasn't sure what you're suggesting here. Cause if I change it to OPAQUE it won't change.

elalish: Oh, sorry, I'm wrong - OPAQUE ignores the alpha value entirely. This actually relates to a new glTF extension we're discussing now, so maybe that will help eventually.

elalish: So yeah, what you have now is fine.

packages/shared-assets/models/manifold.glb
elalish, line 1: I just made exactly this mistake - you need to also add this file to `ci-before-deploy.sh` or else the public example page won't be able to find it.

elalish (changes requested):
Looks great, just one small update needed.

packages/shared-assets/models/manifold.glb
samaneh-kazemi: hmm wouldn't this cover it?
  shared-assets/models/*.* \

elalish: Oh shoot, you're right! Thanks.

elalish (approved):

PR #4534, opened 2023-10-24
Switch from glTF-Sample-Models to the newer glTF-Sample-Assets

bhouston:
Per a previous discussion [here](https://github.com/google/model-viewer/pull/4493#discussion_r1346071701), I have switched from the obsolete [glTF-Sample-Models git repository](https://github.com/KhronosGroup/glTF-Sample-Models) for the various models used in the fidelity test to the newer [glTF-Sample-Assets repository](https://github.com/KhronosGroup/glTF-Sample-Assets).  I have also switched from a custom git checkout script to just using git submodules.

I had to remove a few tests from config.json unfortunately as they were relying on models that are not available in glTF-Sample-Assets because of licensing issues.


bhouston:
Render Fidelity tests were initially failing because of submodule support issues.  But I got those fixed.

Now they are failing because of this difference in the Khronos-TextureTransformMulti-Test - specifically the normal rows look different:

Screenshot 2023-10-25 at 9 37 15 AM

I am investigating it now.

bhouston:
It looks like this change to the underlying asset makes the different renders appear to render more similar results.  Thus I suggest we update the goldens for model-viewer & filament and go with this.

This is based on this analysis - that Babylon and gltf-model-viewer do not change their results, but the end results for everyone is much closer.  I suspect that the way the normals were done before were exposing a bug in model-viewer and filament that was not handled but now that the asset is a bit more standard, that bug is no longer exposed.

BabylonJS's results on this test do not differ:

Screenshot 2023-10-25 at 9 41 57 AM

And filament results differ like model-viewer/three.js:

Screenshot 2023-10-25 at 9 43 41 AM

But gltf-model-viewer does not differ:

Screenshot 2023-10-25 at 9 45 31 AM

elalish:
I agree with your analysis and suggestion - this is a pretty common type of update we need to do as sample bugs get fixed.

elalish (approved):
Upon looking closer at the replaced images, I'm pretty sure there's still some kind of bug, either in the asset or in some of the renderers. Still, that's independent of this change, but worth filing a bug on probably both the sample asset repo and three.js to get to the bottom of it. We should also remove the no-longer-used goldens, but that can happen at a later cleanup stage.

PR #4542, opened 2023-10-26
enable complex arguments for "npm run render-goldens" command

bhouston:
This implements this idea: https://github.com/google/model-viewer/issues/4536

```
Options:
      --version       Show version number                              [boolean]
  -c, --config        Path to configuration json             [string] [required]
  -r, --renderer      Limit to specific renderers                        [array]
  -s, --scenario      Limit to specific scenarios                        [array]
  -p, --port          Port for web server               [number] [default: 9040]
  -m, --missing-only  Only render if an output image is missing
                                                      [boolean] [default: false]
  -d, --dry-run       Lists which images would be rendered but doesn't render
                                                      [boolean] [default: false]
  -q, --quiet         Hide the puppeteer controlled browser
                                                      [boolean] [default: false]
  -h, --help          Show help                                        [boolean]
```

To run a subset of renders or scenarios do something like this:

```
% npm run render-goldens -- --renderer=filament --renderer=model-viewer --scenario=clearcoat -q 
```

Explanation of the new features:
 --renderer, this now allows you to specify multiple renderers in the whitelist, rather than only one.
 -- scenario, this now also allows you to specify multiple scenarios in the whitelist, rather than only one.  It now also selects scenarios based on being a subset of the config.json scenario name, thus you can run all tests that contain say "transform" or "clearcoat".
 --missing-only is very useful when I add new tests and I don't want to render all other tests just to get goldens for the new ones.
 --dry-run, when I want to figure out which tests will run given that command line.
 --quiet, I can run update-screenshots in the background while working without Chrome constantly popping up to disrupt my flow.  Also gives about a 10% speed boost on my machine.
 --port, allows me to run two of these at the same time, useful when I want to run two different renderers at the same time, or work on two separate branches simultaneously.


elalish (changes requested):
This looks great, but can you update our [README](https://github.com/google/model-viewer/tree/master/packages/render-fidelity-tools#render-fidelity-tools) with the descriptions from your comment? I'm going to give you the benefit of the doubt that you've tested this yourself, since you're using it more than I am at the moment anyway.

bhouston:
@elalish I've updated the README for "npm run update-screenshots" here: https://github.com/google/model-viewer/blob/aa653126a2faa6a4d0073c65f5d7b97da2c1fee2/packages/render-fidelity-tools/README.mdhttps://github.com/google/model-viewer/blob/aa653126a2faa6a4d0073c65f5d7b97da2c1fee2/packages/render-fidelity-tools/README.md

bhouston:
I've merged in the "npm run test" CLI update into this PR to make it simpler.  I've also further updated the README to contain both sets of options in it.

bhouston:
I've renamed "npm run update-screenshots" to "npm run render-golden", it is much more accurate/descriptive, e.g. the script renders the images that go into the goldens folder.

bhouston:
In the latest push I've made it so that the renderer you are testing for fidelity is configurable. It defaults to "model-viewer" (for full backwards compatibility) but you can specify any of the web-based renderers (filament, babylon, gltf-sample-viewer, three-gpu-renderer) and it works.

packages/render-fidelity-tools/README.md
```diff
@@ -25,9 +25,49 @@ Command                         | Description
...
+
+For `npm run render-goldens`, we support these command line options:
+
+Command            | Default             | Description
+--------------------|----------- | -----------
+  -c, --config      |  | Path to configuration json.
+  -r, --renderer    |  | Limit to specific renderers. This now allows you to specify multiple renderers in the whitelist, rather than only one.
+  -s, --scenario    |  | Limit to specific scenarios. This now also allows you to specify multiple scenarios in the whitelist.  You can use a full name or a partial name of scenarios and it will match against all that contain that scenario substring.
+  -p, --port        | 9040  | Port for web server.
+  -m, --missing-only | false | Only render if an output image is missing.  Very useful when adding new tests.
```
elalish, line 63: 👍

elalish (approved):
Thank you, this looks like a huge improvement in usability!

Issue #4536, opened 2023-10-25
Update render-fidelity "update-screenshots" CLI to have more options.


bhouston:
### Description

We need better CLI options for the update-screenshot CLI tool.

Specifically I would like to see:

- allow for picking one or more render engines, not just one.
- allow for picking any subset of scenarios via string matches.
- enable headless or not.  Because having Chrome pop up repeated is disruptive to local development.
- enable parallel renderer runs (excluding Stellar, and V-Ray), because it will be faster and there is amble CPU available for this.

Something like:
```bash
% npm run update-screenshots --renderer=filament --renderer=model-viewer --scenario=anisotropy --headless=true --parallel=true
```

Or something like this:
```
% npx update-screenshots -r filament -r model-viewer -s anisotropy -h -p
```

I was thinking of using commander for the CLI parsing as it is sort of industry standard.

elalish:
Agreed, that would be a fantastic update!

elalish:
Fixed by #4542, thanks!

PR #4544, opened 2023-10-26
enable complex arguments for fidelity test command

bhouston:
This is very similar to this other PR where I add CLI options to the "update-screenshots" command: #4542 

This implements this idea #4536, but for the "test" command:

```
Options:
      --version   Show version number                                  [boolean]
  -c, --config    Path to configuration json                 [string] [required]
  -s, --scenario  Limit to specific scenarios                            [array]
  -p, --port      Port for web server                   [number] [default: 9030]
  -d, --dry-run   Checks that all comparison images exist
                                                      [boolean] [default: false]
  -q, --quiet     Hide the puppeteer controlled browser
                                                      [boolean] [default: false]
  -h, --help      Show help                                            [boolean]
```

To run the new test ci command, use this command line:

```bash
% npm run test -- --scenario=texture --quiet
```

Explanation of the new features:
--scenario, this now also allows you to specify multiple scenarios in the whitelist, rather than only one. It now also selects scenarios based on being a subset of the config.json scenario name, thus you can run all tests that contain say "transform" or "clearcoat".
--missing-only is very useful when I add new tests and I don't want to render all other tests just to get goldens for the new ones.
--dry-run, when I want to figure out which goldens are missing.
--quiet, I can run the test in the background.  Also gives about a 10% speed boost on my machine.
--port, allows me to run two of these at the same time, useful when I want to run two different renderers at the same time, or work on two separate branches simultaneously.

### DIFF
--- a/.github/workflows/fidelity-tests.yml
+++ b/.github/workflows/fidelity-tests.yml
@@ -29,4 +29,4 @@ jobs:
     - name: Fidelity tests
       uses: GabrielBB/xvfb-action@v1.0
       with:
-        run: npm run test --workspace=@google/model-viewer-render-fidelity-tools
+        run: npm run test --workspace=@google/model-viewer-render-fidelity-tools -- --quiet
--- a/package-lock.json
+++ b/package-lock.json
@@ -3247,6 +3247,21 @@
         "@types/node": "*"
       }
     },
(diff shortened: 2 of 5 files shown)


bhouston:
I've merged this PR into here: https://github.com/google/model-viewer/pull/4542

PR #4546, opened 2023-10-27
Update example assets

elalish:
@bhouston Our tests were broken because the Sample-Models hadn't been updated to Sample-Assets yet. You probably didn't noticed because like me, you hadn't deleted the Sample-Models directory and the unit tests weren't running on the CI for your PRs. I *think* I've fixed that here too - we'll see when you pull it into your open PRs.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,6 +1,6 @@
 name: Unit tests
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   full_test_run:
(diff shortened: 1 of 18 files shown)


diegoteran (approved):

elalish:
This is super strange - the CI is 404ing on BrainStem.glb and Triangle.gltf, though they work fine locally, and all the other glTF-Sample-Assets work fine in the other tests on the CI. @bhouston do you have any thoughts?

.github/workflows/unit-tests.yml
```diff
@@ -10,6 +10,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+      with:
+        submodules: true
```
elalish, line 14: Ah, I think this is what I was missing.

diegoteran (approved):
nice

PR #4578, opened 2023-11-27
UltraHDR lighting support

elalish:
This is an exciting one! Custom lighting environments with 10-30x less file size, using [UltraHDR](https://developer.android.com/guide/topics/media/platform/hdr-image-format) (a backward-compatible, HDR version of JPEG) rather than .hdr. This is pulling in a dependency on https://github.com/MONOGRID/gainmap-js (thanks @daniele-pelagatti!), which is a very fast and lightweight decoder that leverages the browser's in-built JPEG decoding that ends up faster than the JS .hdr decoder we use now. 

Monogrid has also helpfully made available a simple [webpage](https://gainmap-creator.mono-grid.com/) that will do local conversion of .hdr to UltraHDR .jpg files. This brought Spruit Sunrise down from 1.5MB to 122KB - and the compression is even more impressive with grayscale studio lighting. We're only going to support the single-file .jpg approach for now - separate WebP files gives a bit of additional compression, but it seems too inconvenient to our users. 

We put an upstream fix into three.js to make this work better, so I'll wait to merge this until we've updated to the next release.


packages/model-viewer/src/three-components/TextureUtils.ts
```diff
@@ -81,8 +82,11 @@ export default class TextureUtils {
 
   async loadImage(url: string): Promise<Texture> {
     const texture: Texture = await new Promise<Texture>(
-        (resolve, reject) =>
-            this.ldrLoader.load(url, resolve, () => {}, reject));
+        (resolve, reject) => this.imageLoader.load(url, (result) => {
+          const {texture} = result.renderTarget;
+          result.dispose(false);
+          resolve(texture)
```
elalish, line 88: Thanks for the fixes @daniele-pelagatti! All seems to be working now. One thing I was curious about: if I change this to `result.dispose(true)`, my textures go black. However as it is, I believe I lose my handle to the `renderTarget` and can therefore never dispose of it. Not sure that's a huge deal, but I'd like to know what your recommended pattern is here?

PR #4604, opened 2023-12-21
added Grounded Skybox option

elalish:
The three.js ground-projected-skybox example I found did not work well in practice, so I made my own simpler version. This goes along nicely with our new UltraHDR support because now we can transmit high-fidelity HDR lighting environments to use as skyboxes. 

As with ground-projected skyboxes generally, it'll only look decent if the image is taken on fairly flat ground without nearby stuff that sticks up. The `skybox-height` attribute is used to get the proper scale of the image on the ground. If you know the height at which the image was taken, simply put that in - otherwise experiment until it looks reasonable. Ideally you'll want a height that is on the same order as the size of the model for it to look good.


diegoteran (dismissed):
Awesome, I'll keep looking into it

PR #4635, opened 2024-01-12
Add CreateVideoTextureOptions to support crossOrigin videos

### LINKED ISSUE #4634: Calling createVideoTexture with cross origin video url cause CORS error
#### How To Reproduce
1. Run this [example](https://modelviewer.dev/examples/scenegraph/#animatedTexturesExample) on localhost or glitch.com.
1. Switch to video mode.
1. The below error is logged in the console.
```
THREE.WebGLState: DOMException: Failed to execute 'texImage2D' on 'WebGL2RenderingContext': The video element contains cross-origin data, and may not be loaded.
```

#### Live Demo

https://glitch.com/edit/#!/stellar-abstracted-verdict

#### Solution
As I searched about this issue, I come across this [issue](https://github.com/ml5js/ml5-library/issues/217) that addresses the same problem which solved with [this](https://github.com/ml5js/ml5-library/issues/217#issuecomment-424085565). And I submitted a PR that fixes this issue by letting us call the `createVideoTexture` like this:
```javascript
createVideoTexture(url, {crossOrigin: "anonymous"})
```
The detailed Implementation will be explained in the PR. I would be glad to know if there are anything else needed for this feature. or it can be solved via another solution.

### Version

- model-viewer: v3.3.0

### Browser Affected

I only checked the issue on the chrome. I guess the issue exists in all browsers.
- [X] Chrome, version: 120.0.6099.109
- [ ] Edge
- [ ] Firefox
- [ ] IE
- [ ] Safari

### OS

- [ ] Android
- [ ] iOS
- [ ] Linux
- [ ] MacOS
- [ ] Windows

### AR

- [ ] WebXR
- [ ] SceneViewer
- [ ] QuickLook

navidadelpour:
### Reference Issue
Fixes #4634

### Features
- Adds `option: CreateVideoTextureOptions` argument to `createVideoTexture`. Allowing to call it like `createVideoTexture(url, {crossOrigin: "anonymous"})`

Please let me know if there is any change needed if the solution is accepted (like updating documents, etc).

### DIFF
--- a/packages/modelviewer.dev/data/docs.json
+++ b/packages/modelviewer.dev/data/docs.json
@@ -1122,9 +1122,9 @@
         ]
       },
       {
-        "name": "createVideoTexture(uri, options?)",
+        "name": "createVideoTexture(uri)",
         "htmlName": "createVideoTexture",
-        "description": "Returns a Texture object for use with the <code>setTexture</code> method on <code>TextureInfo</code> objects in the Materials API. The video element created is available through <code>texture.source.element</code> for controlling playback. for cross origin resources, you can set `options.crossOrigin` to `anonymous`.",
+        "description": "Returns a Texture object for use with the <code>setTexture</code> method on <code>TextureInfo</code> objects in the Materials API. The video element created is available through <code>texture.source.element</code> for controlling playback.",
         "links": [
           "<a href=\"../examples/scenegraph/#animatedTexturesExample\">Create animated textures example</a>"
         ]
--- a/packages/modelviewer.dev/data/docs.json
+++ b/packages/modelviewer.dev/data/docs.json
@@ -1122,9 +1122,9 @@
         ]
       },
       {
-        "name": "createVideoTexture(uri)",
+        "name": "createVideoTexture(uri, options?)",
         "htmlName": "createVideoTexture",
-        "description": "Returns a Texture object for use with the <code>setTexture</code> method on <code>TextureInfo</code> objects in the Materials API. The video element created is available through <code>texture.source.element</code> for controlling playback.",
+        "description": "Returns a Texture object for use with the <code>setTexture</code> method on <code>TextureInfo</code> objects in the Materials API. The video element created is available through <code>texture.source.element</code> for controlling playback. for cross origin resources, you can set `options.crossOrigin` to `anonymous`.",
         "links": [
           "<a href=\"../examples/scenegraph/#animatedTexturesExample\">Create animated textures example</a>"
         ]
--- a/packages/model-viewer/src/features/scene-graph.ts
+++ b/packages/model-viewer/src/features/scene-graph.ts
@@ -42,6 +42,10 @@ interface SceneExportOptions {
       forceIndices?: boolean
 }
 
+interface CreateVideoTextureOptions {
+  crossOrigin?: string
+}
+
 export interface SceneGraphInterface {
   readonly model?: Model;
   variantName: string|null;
@@ -53,7 +57,7 @@ export interface SceneGraphInterface {
   createTexture(uri: string, type?: string): Promise<ModelViewerTexture|null>;
   createLottieTexture(uri: string, quality?: number):
       Promise<ModelViewerTexture|null>;
-  createVideoTexture(uri: string): ModelViewerTexture;
+  createVideoTexture(uri: string, options?: CreateVideoTextureOptions): ModelViewerTexture;
   createCanvasTexture(): ModelViewerTexture;
   /**
    * Intersects a ray with the scene and returns a list of materials who's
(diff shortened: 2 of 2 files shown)


elalish:
Yes, please add this info to the `packages/modelviewer.dev/data/docs.json` entry for this function. This looks great, thanks!

navidadelpour:
@elalish Thanks for reviewing. Docs updated. Let me know if anything else is needed.

packages/modelviewer.dev/data/docs.json
```diff
@@ -1122,9 +1122,9 @@
         ]
       },
       {
-        "name": "createVideoTexture(uri)",
+        "name": "createVideoTexture(uri, options?)",
         "htmlName": "createVideoTexture",
-        "description": "Returns a Texture object for use with the <code>setTexture</code> method on <code>TextureInfo</code> objects in the Materials API. The video element created is available through <code>texture.source.element</code> for controlling playback.",
+        "description": "Returns a Texture object for use with the <code>setTexture</code> method on <code>TextureInfo</code> objects in the Materials API. The video element created is available through <code>texture.source.element</code> for controlling playback. for cross origin resources, you can set `options.crossOrigin` to `anonymous`.",
```
elalish, line 1127: Looking at this more closely, I wonder if we need a new parameter at all. We already have an attribute `withCredentials` that is used for all of our image loaders. In three.js it's doing: `credentials: this.withCredentials ? 'include' : 'same-origin'`. Could we just do `video.crossOrigin = this.withCredentials ? 'with-credentials' : 'anonymous'` to be consistent?

elalish: https://modelviewer.dev/docs/index.html#entrydocs-loading-attributes-withCredentials

navidadelpour: This is what I understand from your comment:
```html
<model-viewer with-credentials src="https://modelviewer.dev/shared-assets/models/sphere.glb" ... >
```
```js
videoTexture = modelViewerAnimated.createVideoTexture("https://cdn.glitch.global/68b42b4e-5a36-47f5-ab00-09026a7a73d0/lottie-logo.mp4?v=1705067796906");
// video.crossOrigin would be 'anonymous' as you said
```

Testing it locally gave me this error:
```
Access to fetch at 'https://modelviewer.dev/shared-assets/models/sphere.glb' from origin 'http://localhost:8002' has been blocked by CORS policy: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.
```

Which addresses the problem in a scenario that the glb file is on the same origin but the video is not.

navidadelpour: @elalish For improving the consistency in the code, we can also change:
```
interface CreateVideoTextureOptions {
  withCredentials?: boolean
}
```

elalish: Hmm, how common do you think it is that different assets are coming from different origins for MV users? I do see your point about generality, but the fact is that we've already got this single `with-credentials` switch that affects `src`, `createTexture`, and `lottieLoader`, without any complaints yet. I think I'd rather add video to that pattern for now, or else do a complete refactor if it proves to be a user pain point. What do you think?

navidadelpour: > how common do you think it is that different assets are coming from different origins for MV users? 

I'm not really sure as I'm a newbie in this library, but I believe most users would serve their assets alongside their application, and it would be okay for them. However, it may be concerning for users who have an application on origin A and are using a CMS, which may be on origin B, to serve an image, a video, or a Lottie file.

>  I think I'd rather add video to that pattern for now, or else do a complete refactor if it proves to be a user pain point.

I think that each resource having its option to be with credentials or not would be a more bulletproof solution. You're right; it may not be a user pain point now, and for the sake of consistency, we can do as you suggested, which covers the case for users that have all assets on a cross-origin source.

I will update the MR ASAP.

navidadelpour:
@elalish Hi again.
I've Updated the code as you suggested and everything is fine. I Also update the [example](https://glitch.com/edit/#!/stellar-abstracted-verdict?path=index-updated-in-pr.html%3A7%3A56) so we can test it. I've built the code into a file named `model-viewer-updated-in-pr.min.js` and used it inside a file named `index-updated-in-pr.html` so we can test it. But I couldn't find a way to test it with `with-credentials`.

Thanks!

elalish (approved):
Thanks! Agreed regarding a more general solution - let's see how it goes. Out of curiosity, what is stopping your testing?

navidadelpour:
@elalish I appreciate that :smiley::star_struck:  . Actually, I wasn't be able to find a cross origin server that serves both glb and video assets with the right cors headers and credential checking which I can be able to use `with-credentials` option. my only try was the previous one where we got the 
```
Access to fetch at 'https://modelviewer.dev/shared-assets/models/sphere.glb' from origin 'http://localhost:8002' has been blocked by CORS policy: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.
```
error (which tells us to change the cors configurations on the server that I don't have access to).

Not that I'm thinking maybe there would be some way for me to configure a new server in my localhost and use it as the assets server and configure the cors headers so that it require credentials for serving assets, making it like a cros origin server...

But I Also would be glad if you had any idea and guide me through how it should be tested.

Thanks for your amazing work!

elalish:
I see - I figured you would just test with a local server. Any generic file server should work - a Google Cloud bucket or Amazon S3, etc. Thanks!

Saliiih:
Helped a lot 
Thanks

PR #4643, opened 2024-01-16
Draft: Add model dragging in XR headsets by controllers or hands

FluorescentHallucinogen:
This pull request fixes a problem (described by me a year ago in the https://github.com/google/model-viewer/discussions/3969#discussioncomment-4753669) that users can't move and rotate models in the Meta Quest headsets.

### DIFF
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -13,8 +13,9 @@
  * limitations under the License.
  */
 
-import {Event as ThreeEvent, EventDispatcher, Matrix4, PerspectiveCamera, Vector3, WebGLRenderer} from 'three';
+import {Event as ThreeEvent, EventDispatcher, Matrix4, PerspectiveCamera, Vector3, WebGLRenderer, Line, Raycaster, BufferGeometry} from 'three';
 import {XREstimatedLight} from 'three/examples/jsm/webxr/XREstimatedLight.js';
+import {XRControllerModelFactory} from 'three/examples/jsm/webxr/XRControllerModelFactory.js';
 
 import {CameraChangeDetails, ControlsInterface} from '../features/controls.js';
 import {$currentBackground, $currentEnvironmentMap} from '../features/environment.js';
@@ -117,12 +118,42 @@ export class ARRenderer extends EventDispatcher<
   private yawDamper = new Damper();
   private scaleDamper = new Damper();
 
+  private controller1: any;
+  private controller2: any;
+  private controllerGrip1: any;
+  private controllerGrip2: any;
+  private intersected: any[] = [];
+
   private onExitWebXRButtonContainerClick = () => this.stopPresenting();
 
   constructor(private renderer: Renderer) {
     super();
     this.threeRenderer = renderer.threeRenderer;
     this.threeRenderer.xr.enabled = true;
+
+    this.controller1 = this.threeRenderer.xr.getController(0);
+    this.controller1.addEventListener('selectstart', (e: any) => this.onControllerSelectStart(e));
+    this.controller1.addEventListener('selectend', (e: any) => this.onControllerSelectEnd(e));
+
+    this.controller2 = this.threeRenderer.xr.getController(1);
+    this.controller2.addEventListener('selectstart', (e: any) => this.onControllerSelectStart(e));
+    this.controller2.addEventListener('selectend', (e: any) => this.onControllerSelectEnd(e));
+
+    const controllerModelFactory = new XRControllerModelFactory();
+
+    this.controllerGrip1 = this.threeRenderer.xr.getControllerGrip(0);
+    this.controllerGrip1.add(controllerModelFactory.createControllerModel(this.controllerGrip1));
+
+    this.controllerGrip2 = this.threeRenderer.xr.getControllerGrip(1);
+    this.controllerGrip2.add(controllerModelFactory.createControllerModel(this.controllerGrip2));
+
+    const geometry = new BufferGeometry().setFromPoints([new Vector3(0, 0, 0), new Vector3(0, 0, -1)]);
+    const line = new Line(geometry);
+    line.name = 'line';
+    line.scale.z = 5;
+
+    this.controller1.add(line.clone());
+    this.controller2.add(line.clone());
   }
 
   async resolveARSession(): Promise<XRSession> {
@@ -175,6 +206,11 @@ export class ARRenderer extends EventDispatcher<
     if (this.isPresenting) {
       console.warn('Cannot present while a model is already presenting');
     }
+
+    scene.add(this.controller1);
+    scene.add(this.controller2);
+    scene.add(this.controllerGrip1);
+    scene.add(this.controllerGrip2);
 
     let waitForAnimationFrame = new Promise<void>((resolve, _reject) => {
       requestAnimationFrame(() => resolve());
@@ -254,6 +290,84 @@ export class ARRenderer extends EventDispatcher<
 
     this.lastTick = performance.now();
     this.dispatchEvent({type: 'status', status: ARStatus.SESSION_STARTED});
+  }
+
+  private intersectObjects(controller: any) {
+    // Do not highlight in mobile-ar
+    if (controller.userData.targetRayMode === 'screen') return;
+
+    // Do not highlight when already selected
+    if (controller.userData.selected !== undefined) return;
+
+    const line = controller.getObjectByName('line');
+    const intersections = this.getIntersections(controller);
+
+    if (intersections.length > 0) {
+      const intersection = intersections[0];
+
+      const object = intersection.object;
+      // @ts-ignore
+      object.material.emissive.setHex(0x333333);
+
+      this.intersected.push(object);
+
+      line.scale.z = intersection.distance;
+    } else {
+      line.scale.z = 5;
+    }
+  }
+
+  private getIntersections(controller: any) {
+    controller.updateMatrixWorld();
+
+    const tempMatrix = new Matrix4();
+
+    tempMatrix.identity().extractRotation(controller.matrixWorld);
+
+    const raycaster: Raycaster = new Raycaster();
+
+    raycaster.ray.origin.setFromMatrixPosition(controller.matrixWorld);
+    raycaster.ray.direction.set(0, 0, -1).applyMatrix4(tempMatrix);
+
+    const group = this.presentedScene!.model;
+    // @ts-ignore
+    return raycaster.intersectObjects(group.children, false);
+  }
+
+  private cleanIntersected() {
+    while (this.intersected.length) {
+      const object = this.intersected.pop();
+      object.material.emissive.setHex(0x000000);
+    }
+  }
+
+  private onControllerSelectStart(event: any) {
+    const controller = event.target;
+
+    const intersections = this.getIntersections(controller);
+
+    if (intersections.length > 0) {
+      const intersection = intersections[0];
+      const object = intersection.object;
+      controller.attach(object);
+      controller.userData.selected = object;
+    }
+
+    controller.userData.targetRayMode = event.data.targetRayMode;
+  }
+
+  private onControllerSelectEnd(event: any) {
+    const controller = event.target;
+
+    if (controller.userData.selected !== undefined) {
+      const object = controller.userData.selected;
+
+      const group = this.presentedScene!.model;
+      // @ts-ignore
+      group.attach(object);
+
+      controller.userData.selected = undefined;
+    }
   }
 
   /**
@@ -734,6 +848,10 @@ export class ARRenderer extends EventDispatcher<
    * Only public to make it testable.
    */
   public onWebXRFrame(time: number, frame: XRFrame) {
+    this.cleanIntersected();
+    this.intersectObjects(this.controller1);
+    this.intersectObjects(this.controller2);
+
     this.frame = frame;
     ++this.frames;
     const refSpace = this.threeRenderer.xr.getReferenceSpace()!;


packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -256,6 +292,84 @@ export class ARRenderer extends EventDispatcher<
...
+
+    tempMatrix.identity().extractRotation(controller.matrixWorld);
+
+    const raycaster: Raycaster = new Raycaster();
+
+    raycaster.ray.origin.setFromMatrixPosition(controller.matrixWorld);
+    raycaster.ray.direction.set(0, 0, -1).applyMatrix4(tempMatrix);
+
+    const group = this.presentedScene!.model;
+    // @ts-ignore
```
elalish, line 333: Can you try to replace these `@ts-ignore`s by casting, e.g. `const group = this.presentedScene!.model as Group;` or something? That's a lot more specific and helps me to understand the problem.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -256,6 +292,84 @@ export class ARRenderer extends EventDispatcher<
...
+
+    const line = controller.getObjectByName('line');
+    const intersections = this.getIntersections(controller);
+
+    if (intersections.length > 0) {
+      const intersection = intersections[0];
+
+      const object = intersection.object;
+      // @ts-ignore
+      object.material.emissive.setHex(0x333333);
```
elalish, line 310: The material might already have an emissive factor - can you save this in `userData` and set back to the original instead?

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -117,12 +118,42 @@ export class ARRenderer extends EventDispatcher<
   private yawDamper = new Damper();
   private scaleDamper = new Damper();
 
+  private controller1: any;
+  private controller2: any;
+  private controllerGrip1: any;
+  private controllerGrip2: any;
+  private intersected: any[] = [];
```
elalish, line 125: Are these new Three.js APIs? Otherwise we should have types for them.

elalish (changes requested):
Thanks, this is exciting! Now I just need to hunt down a Quest so I can test this personally...

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -176,6 +207,11 @@ export class ARRenderer extends EventDispatcher<
       console.warn('Cannot present while a model is already presenting');
     }
 
+    scene.add(this.controller1);
+    scene.add(this.controller2);
+    scene.add(this.controllerGrip1);
+    scene.add(this.controllerGrip2);
```
elalish, line 213: Regarding controllers of the wrong size, you can see further down this file that scene.scale is changed when the object is scaled by pinch on a phone. We may need to add another level of object hierarchy to the AR scene to fix this. However, I believe that scale starts at 1, so if you're seeing a problem from the beginning, it may be a different issue. Can you add a screen recording by chance?

elalish: Okay, I just saw your video on the other thread, thanks. Interesting, that does seem like a different problem than scene scale. Position maybe? Let's move discussion from the thread to this PR, FYI @cabanier.

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: This is to highlight (brighten) the 3D model when the controller ray intersects it. Please check out the video again. ;) This highlight is removed back in the `cleanIntersected()` method by calling the `code.object.material.emissive.setHex(0x000000)`.

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: Ok, I'll try adding types everywhere again.

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: Ah, I think I get it, you mean that the 3D model can have the emissive property set to a value other than 0x000000 before the controller ray intersects the 3D model? Then a question: can there be more than one 3D model in a scene at the same time?

BTW, highlighting (brightening) the 3D model on the hover is an optional improvement. I mean, we can opt out of that and simplify the implementation.

elalish: Exactly, but this caused me to notice a potentially bigger issue. We have one 3D model, but it can be composed of an arbitrary hierarchy of objects. We don't expose those sub-meshes in our public API, so they probably should not be independently selectable. Instead we should move everything when anything is selected. And we already have a highlighting concept in WebXR: the PlacementBox - maybe best to just use that. 

I'm interested in what UX you're going for here - it might be nice to make it as similar to our existing AR UX as possible, given controllers. I'm open to other ideas, but I'd like to know how you're envisioning the experience.

FluorescentHallucinogen: Could you please share links to an example where there is more than one 3D model in the one scene at the same time, and an example where one complex 3D model consists of multiple 3D models (parts)?

elalish: MV only supports one 3D model, so we only have complex 3D models that consist of multiple parts. That's pretty common, e.g. https://modelviewer.dev/examples/animation/index.html

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: Yes, it's a position problem, not a scale problem. The controllers seems small because they are far away from the user. E.g. I've found that initially the controller model is in the right place, but then the 3D model drops to the floor and the controller gets lower.

Changing the `isFirstView` variable to `false`:

```ts
// WebXR may return multiple views, i.e. for headset AR. This
// isn't really supported at this point, but make a best-effort
// attempt to render other views also, using the first view
// as the main viewpoint.
let isFirstView: boolean = true;
for (const view of pose.views) {
  this.updateView(view);

  if (isFirstView) {
    this.moveToFloor(frame);

    this.processInput(frame);

    const delta = time - this.lastTick!;
    this.moveScene(delta);
    this.renderer.preRender(scene, time, delta);
    this.lastTick = time;

    scene.renderShadow(this.threeRenderer);
  }

  this.threeRenderer.render(scene, scene.getCamera());
  isFirstView = false;
}
```

and commenting/removing the following code:

```ts
const {theta, radius} =
    (element as ModelViewerElementBase & ControlsInterface)
        .getCameraOrbit();
// Orient model to match the 3D camera view
const cameraDirection = xrCamera.getWorldDirection(vector3);
scene.yaw = Math.atan2(-cameraDirection.x, -cameraDirection.z) - theta;
this.goalYaw = scene.yaw;

position.copy(xrCamera.position)
    .add(cameraDirection.multiplyScalar(radius));

this.updateTarget();
const target = scene.getTarget();
position.add(target).sub(this.oldTarget);

this.goalPosition.copy(position);
```

fixes the problem.

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: The highlighting (brightening) the 3D model on the hover is only needed for visual feedback.

> And we already have a highlighting concept in WebXR: the PlacementBox - maybe best to just use that.

 Just tried to add `this.placementBox!.show = true` to `intersectObjects()` and `this.placementBox!.show = false` to `cleanIntersected()`. Currently, the placement box actually doesn't look like a box (cube) around the 3D model, but like a white rectangle, which is always placed horizontally at the bottom of the 3D model. It is used to select a place on the horizontal surface where to place the 3D model. When you use Meta Quest headset and select a 3D model using the ray, you are free to rotate it with your hand or controller as you want. For example, put it on its side or turn it upside down. This is something you currently can't do in AR using a smartphone and your fingers. So in its current form, the placement box is not a good fit.

FluorescentHallucinogen:
@elalish What are the chances of you getting your hands on Meta Quest? I thought it was a must have for the model-viewer maintainer. :)

elalish:
Well, now that we're looking to get serious about adding headset support, I'd agree. I'm going to get one ordered now.

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: > When you use Meta Quest headset and select a 3D model using the ray, you are free to rotate it with your hand or controller as you want. For example, put it on its side or turn it upside down. This is something you currently can't do in AR using a smartphone and your fingers. 

Yes, but this is less a technical constraint than a UX decision. I think for simplicity and consistency I might prefer to limit the interactivity to Y-up as in smartphone AR. What do you think? Has anyone done UX studies on your side showing that more general interaction is valuable? The key with MV is we're not looking for "cool demo", but "easy for lay people to buy products".

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: Do you understand why this is? Any recommendation for a better approach?

FluorescentHallucinogen: Controllers are added to the scene. The position of the scene changes, the controllers move with it.

elalish: Okay, then it sounds like adding an extra node in the scene hierarchy is the way to go.

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: > showing that more general interaction is valuable

> easy for lay people to buy products

Instead of a thousand words. 😜

https://github.com/google/model-viewer/assets/7892779/0f45dc5d-29a2-45b8-8d0a-98175a6a1151

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: Could you please help with that? 🙏 And I'll focus on other aspects and bugs.

FluorescentHallucinogen:
> we're looking to get serious about adding headset support

That's great news!

> I'm going to get one ordered now.

Please let me know once you order the headset and which model (Meta Quest 3 or Meta Quest Pro). And when you get it in your hands. ;)

I expect when we add support for XR headsets and test the code on Meta Quest, we will automatically have Apple Vision Pro support since it's the same WebXR standard here and there. :)

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: As you can see in the video, a little highlighting (brightening) of the 3D model on the hover is not really needed at all. :)

elalish: I like your style 😄. Fair point, though it makes a bit less sense with furniture. I need to think about this. Will this interaction work the same way with hands? Will we only support moving objects within arm's reach? I wish we had a UX person dedicated to thinking these use cases through, but I'm afraid it's just us for now.

FluorescentHallucinogen: > Will this interaction work the same way with hands?

Yes, hands are also controllers in WebXR. Moreover, on Meta Quest you can switch between controllers and hands within the same XR session. You can put the controllers down and after a few seconds the controller will go to sleep and there will be an automatic switch to the hands. If you pick up the controller, it will wake up and switch back.

> Will we only support moving objects within arm's reach?

No, this approach is already essentially standard, it works with both small and large objects, both close and distant objects (the ray is long enough).

I'll try to record other videos for you to make it clearer. But it would be better if you try it yourself (use different apps and games for a few days) when you get the headset. That will clear up most of the questions. :)

> I wish we had a UX person dedicated to thinking these use cases through, but I'm afraid it's just us for now.

It's been already researched. Check out e.g. the Mixed Reality Design Language by Microsoft. ;)

Please also note that currently the user has no way to interact with the 3D model in the XR headset at all.

So my goal in this pull request is to add the simplest basic implementation. This will already improve the user experience significantly!

Then we can improve the implementation. Let's move step by step. ;)

FluorescentHallucinogen: See https://learn.microsoft.com/en-us/windows/mixed-reality/design/point-and-commit. 😉

elalish: Thanks a lot, that's quite helpful!

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: @elalish Any progress?

elalish: Sadly no, I've been slammed. We did finally get a Quest 3, but no one's even had time to open the box yet...

MickGerrit: Hi @FluorescentHallucinogen, amazing work for the Google modelviewer! I will try to test it on my Quest 3 some time.

MickGerrit: Yesterday I tried getting it working, but no luck. Any way to build the model-viewer module from your fork or can I test already?

FluorescentHallucinogen: @MickGerrit Hi! Thanks!

The only difference in my fork is in one single `/packages/model-viewer/src/three-components/ARRenderer.ts` file, everything else is the same as the original project. Have you tried to build the original project as described in https://github.com/google/model-viewer?tab=readme-ov-file#development?

FluorescentHallucinogen: @MickGerrit, I've deployed 3 demos for you. :wink:

https://model-viewer-xr-drag.web.app/fixed/ — it uses the code from this pull request + the hack described in the https://github.com/google/model-viewer/pull/4643#discussion_r1463845714

https://model-viewer-xr-drag.web.app/unfixed/ — it uses the code from this pull request (without the hack described in the https://github.com/google/model-viewer/pull/4643#discussion_r1463845714)

https://model-viewer-xr-drag.web.app/original/ — it uses the original code

Just point the ray at the 3D model and press and hold the trigger on the controller to select and rotate the model as you want using the controller.

@elalish Please check them too. :wink:

MickGerrit: I tried out the demo. Very cool! The fixed version is already a big improvement for viewing on a Meta Quest.

sprengerst: @FluorescentHallucinogen  👍  great job so far.

sprengerst: I tried your solution with other models, the "drag and drop" doesn't seem to work for all GLBs.

sprengerst: `return raycaster.intersectObjects(group.children, **true**);`
        
This should be set to **true**

FluorescentHallucinogen: @sprengerst But why? From what I can see, it's `true` by default. See https://threejs.org/docs/#api/en/core/Raycaster.intersectObjects.

elalish:
@adamwolf UX design, research, or just musings/examples would all help at this stage. It's harder to figure out what to build than to build it.

elalish:
I know this has been on the backburner for way too long, but I do finally have a Quest 3 on my desk and I just tested this PR in it. Good start! Got a ways to go, but no problem there. Now to do some actual coding...

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: I fixed this by adding a pivot so that we no longer set the position/orientation of the top-level scene object.

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: done.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -256,6 +305,88 @@ export class ARRenderer extends EventDispatcher<
...
   }
 
+  private intersectObjects(controller: XRTargetRaySpace) {
+    // Do not highlight in mobile-ar
+    if (controller.userData.targetRayMode === 'screen')
+      return;
+
+    // Do not highlight when already selected
+    if (controller.userData.selected !== undefined)
+      return;
```
elalish, line 315: I noticed the frame rate is quite low when no object is selected, but when I have an object in my grip the frame rate is fine. I believe this early return is why: the raycast against the whole GLB is slow. We should replace this with a bounding box cube, maybe the placement box, to fix this.

elalish: Fixed, though haven't updated from hit plane to box yet.

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: Fixed - using the placement box as an indicator as in phone AR mode now instead of emissive materials.

elalish:
I'm actually decently happy with this UX now - would love to hear some feedback from you all. There are two main things I dislike still:
1) The model's initial placement is too far away. This should be an easy fix; it's simply because the FoV of a headset is way wider than a phone camera.
2) It's difficult to turn the model around its vertical axis. This is mostly just an XR controller ergonomics problem, but we need a way to work around it since that's the primary way you want to reorient, say, furniture. I'm hoping some tricky matrix math can improve this.

packages/model-viewer/src/three-components/ARRenderer.ts
FluorescentHallucinogen: Thanks!

FluorescentHallucinogen:
@elalish Thanks for your help! Just tested. LGTM. I've removed the "Draft" label. Let's merge it. Further UX improvements can be done by separate pull requests.

FluorescentHallucinogen:
> The model's initial placement is too far away.

It seems this depends on the 3D model (its size?). E.g. small models https://modelviewer.dev/assets/ShopifyModels/Mixer.glb and https://modelviewer.dev/assets/ShopifyModels/GeoPlanter.glb models are initially pretty close, but big https://modelviewer.dev/shared-assets/models/NeilArmstrong.glb and https://modelviewer.dev/assets/ShopifyModels/Chair.glb models are initially far from the user.

> It's difficult to turn the model around its vertical axis. This is mostly just an XR controller ergonomics problem

This is a current implementation problem. I see 2 problems:

1. You can only select the model only using the bottom plate (transparent rectangle with white edges, which becomes visible only by pointing the ray at the bottom of the model). It's not very intuitive. I originally tried to grab an random point at the top of the model. And it didn't work. And at this time, the bottom plate was not visible at all.

2. It's impossible to rotate and leave the model the way I want it. E.g., you cannot turn the model upside down or lay it on its side. It keeps turning bottom down.

elalish:
>You can only select the model only using the bottom plate

Agreed, forgot to mention I'll switch this to grabbing the whole bounding volume. 

>It's impossible to rotate and leave the model the way I want it. E.g., you cannot turn the model upside down or lay it on its side. It keeps turning bottom down.

This is by design - before it was impossible to get it to sit upright on the floor once selected, which I think is the #1 use case. This is also why I'm going to work on this a bit more before merging. Once someone is used to one UX, any change (that's not a pure addition) is seen as worse, due to anchoring bias. 

I'm pretty happy with inspectability, especially now that we can pass the object back and forth between the two controllers/hands. If you have a specific idea for improving the UX, I'll be happy to give it a try.

FluorescentHallucinogen:
> it was impossible to get it to sit upright on the floor once selected, which I think is the primary use case
> I'm pretty happy with inspectability

I'm not sure that's really the *primary* use case.

Look at all these cool demos/promos for Apple Vision Pro. E.g., kids leave in the air some 3D model such as a human heart or a complex animated mechanism with gears *in their desired position*, and they can walk around it and look at it from all sides, including from the bottom.

In the current implementation:

1. the 3D model is constantly turns bottom down
2. the 3D model constantly falls to the floor

For me, the current implementation is only good for large objects like furniture.

It's not very convenient to view small 3D models like a mixer if they fall on the floor. :-D

Yes, you can still view small 3D models by twisting your arm (sometimes into an unnatural position), but as soon as you stop holding the trigger on the controller, the 3D model will fall to the floor.

In other words, all this is just not good for *inspectability*.

elalish:
One other question (@cabanier maybe you know?), since there's no DOM overlay in these headsets, is there a recommendation on how to exit a WebXR experience? I've seen ugly floating rectangles used for this, but that's about it. And when I use the home button and click "quit" on the background WebXR experience, I've been seeing some crashes - does that actually call the webxr session end event, or is it more of a force-quit?

FluorescentHallucinogen:
> And I feel like the idea of an object dropping to the ground when it is let go of makes some intuitive sense, vs floating forever. Both modes would probably be nice

As an idea, what about adding new property for changing modes?

FluorescentHallucinogen:
As for exiting the WebXR session (immersive mode), there is a https://immersive-web.github.io/webxr-samples/tests/exit-button.html, but it doesn't look like best practice. I haven't seen any games/apps that have an exit button in immersive mode.

But I've seen floating overlays for the controller buttons, that can be used to guide the user which button to use to exit, etc. Something like:

controller-overlay

elalish:
Thanks - I asked around and most folks are happy just relying on the browser's built-in exit functionality, which is certainly the path of least resistance, so I think we'll just stick with that for now.

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -256,6 +290,86 @@ export class ARRenderer extends EventDispatcher<
...
+    // this.controllerGrip1.add(
+    //     controllerModelFactory.createControllerModel(this.controllerGrip1));
+
+    // this.controllerGrip2 = this.threeRenderer.xr.getControllerGrip(1);
+    // this.controllerGrip2.add(
+    //     controllerModelFactory.createControllerModel(this.controllerGrip2));
+
+    const line = new Line(lineGeometry);
+    line.name = 'line';
+    line.scale.z = 5;
```
diegoteran, line 320: What is this 5 constant? maybe add a comment for this one

packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -256,6 +290,86 @@ export class ARRenderer extends EventDispatcher<
...
+    this.controller2 = this.threeRenderer.xr.getController(1);
+    this.controller2.addEventListener(
+        'selectstart',
+        (e: XRControllerEvent) => this.onControllerSelectStart(e));
+    this.controller2.addEventListener(
+        'selectend', (e: XRControllerEvent) => this.onControllerSelectEnd(e));
+
+    // const controllerModelFactory = new XRControllerModelFactory();
+
+    // this.controllerGrip1 = this.threeRenderer.xr.getControllerGrip(0);
```
diegoteran, line 310: This file has a lot of commented out code. remove before merge

packages/model-viewer/src/three-components/PlacementBox.ts
```diff
@@ -105,6 +106,12 @@ export class PlacementBox extends Mesh {
     (this.hitPlane.material as Material).side = DoubleSide;
     this.add(this.hitPlane);
 
+    this.hitBox = new Mesh(new BoxGeometry(
+        size.x + 2 * RADIUS, size.y + RADIUS, size.z + 2 * RADIUS));
```
diegoteran, line 110: I didn't understand these values for the box. Maybe a comment or function would help

diegoteran (dismissed):
Just a few nits for future readability.

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: Yeah, or even better, a named constant. Thanks.

packages/model-viewer/src/three-components/ARRenderer.ts
elalish: Yeah, @FluorescentHallucinogen put this in to render the controllers/hands. I wanted to make sure you saw it in case we decide we want it down the road. For now I'm going with small binary size and simplicity of just pointer lines.

packages/model-viewer/src/three-components/PlacementBox.ts
elalish: Fair, that's a bit confusing.

elalish:
Okay, I added scaling and in-place rotation. Could use some polish and UXR, but this is probably mergeable now.

FluorescentHallucinogen:
@elalish

> UX design, research, or just musings/examples would all help

For inspiration, try Layout, Meta's new mixed reality app:

https://www.meta.com/experiences/layout/9298251876913852/

https://youtu.be/vJKuy6NT0JQ

elalish:
>For inspiration, try Layout, Meta's new mixed reality app:

Thanks, though I should probably write another [blog post](https://elalish.blogspot.com/2022/04/3d-interaction.html) about 3D UX, this time in headsets. Part of what irritates me about headsets is that so much of their 3D UX is like what you show above: despite having 6DOF inputs, we're still using a bunch of 1D and 2D widgets as though we're still using a mouse, but now with cerebral palsy. 

I'm attempting to point the way toward a 3D UX that actually takes advantage of 6DOF inputs and doesn't require such precise pointing or so many modes and widgets. I had hoped someone had done this before me, but maybe not.

elalish:
Did some more testing, found and fixed a few edges cases. I think this is good to go now.

diegoteran (approved):
Lgtm.

Only small nit for future readability for people working on these files, every time there is a check for `if (this.xrMode !== 'screen-space')`, it could be a good idea to add a comment saying what we are doing in general that is not required on a phone.

FluorescentHallucinogen:
@elalish See also Figmin XR for some UX ideas:

https://www.meta.com/ru-ru/experiences/figmin-xr-mixed-reality/6849182851823457/

I like that the user can **bring closer and further away** (this is something that is very much lacking in the current implementation!), scale and rotate objects using only **one single** controller.

See this video (the left part of screen at 20 second time): https://youtu.be/bXpQz_lR9lA?t=20

This can be very useful in terms of accessibility (for those people missing one arm). And for the new [Logitech MX Ink Mixed Reality Stylus (3D pen)](https://www.meta.com/quest/accessories/logitech-mixed-reality-stylus/) that is already supported in WebXR by Meta Quest Browser. See https://x.com/rcabanier/status/1841896955091550612. :wink:

PR #4686, opened 2024-02-21
Upgrade three-gpu-pathtracer, screenshots

gkjohnson:
I've recently gotten three-gpu-pathtracer working consistently again after a slew of browser regressions on a few platforms. So this PR upgrades the three-gpu-pathtracer version to latest and associated screenshots. There have been a number of sampling and geometry handling improvements over those versions, as well.

Some other notes:
- Samples have been lowered from 100 to 64 since puppeteer was timing out and I couldn't figure out how to increase the time. It shouldn't be taking longer than before but it seems the timeout has gotten shorter?
- Two screenshots have been removed because they don't look correct, likely have some recent material changes in the sample GLB files. I'll need to investigate and can add them back later.
- Note that some of the simple models are darker because they have a high metalness value which incurs a lot of energy loss with rough surfaces in the pathtracer at the moment which is a common issue, I understand. It's on the list to be looked at.

Once this is merged I'll make a new PR that enables some other screenshots that for new models that now work and have been added to the config.

cc @bhouston


elalish (approved):
This looks great, thanks! FYI, in the coming weeks I'm going to be migrating this fidelity test suite (include your code) over to Khronos, who have graciously decided to adopt and promote it. Ideally your work will be done before I start, but I'll let you know when I'm beginning in case. Of course we'll all still be contributing just as we were before, just in a different repo (using git LFS finally!).

Issue #4688, opened 2024-02-29
Module not found: Error: Can't resolve './styles/evaluators', and many more


Pratik610:
![image](https://github.com/google/model-viewer/assets/60422361/6dd9f55d-2bab-4929-8c14-e13a3eaab02e)

elalish:
There's no way anyone can help without some background on what you're trying to do. It appears you're trying to build MV into your own TS project of some kind? Are you using a particular framework for that? There must be something wrong with how the dependency or import is specified, so maybe start by showing those from your code?

Pratik610:
i am using React for this. as usual i have imported the MV in the App.js file import `@google/model-viewer`  and using it across the app

elalish:
"as usual" makes it sound like you've done this before and it's worked with model-viewer? Did something change recently? We don't have any React experience on this team, but others on our forums do and we know they've used it successfully, so it's probably some kind of import magic. Have you searched our discussions yet?

chamby:
I assume @Pratik610 is using Webpack because if so version 5 of Webpack has a breaking change that [requires imports to have a file extension](https://webpack.js.org/guides/ecma-script-modules/). The relevant section is below. There is a flag to use the legacy behavior. I will try the flag as I am having the same issue. Ideally the file extensions could be added to model-viewer as well.

> Imports in ESM are resolved more strictly. Relative requests must include a filename and file extension (e.g. *.js or *.mjs) unless you have the behaviour disabled with [fullySpecified=false](https://webpack.js.org/configuration/module/#resolvefullyspecified).

elalish:
@chamby Thanks! PRs welcome if you find a change that will simplify this for users.

PR #4707, opened 2024-03-11
Add explicit file extensions to relative imports of model-viewer and model-viewer-effects

chamby:
### Reference Issue
Fixes: #4688


elalish (approved):
Thanks! Not sure why we had our imports half and half - consistency is good!

PR #4715, opened 2024-03-14
A11y implementation for WCAG compliance.

cybmiper:


packages/modelviewer.dev/assets/a11y.json
```diff
@@ -0,0 +1,14 @@
+{
+  "front": "The front of the 3D model",
```
elalish, line 2: Is the idea that you have a separate JSON file for each locality?

packages/modelviewer.dev/examples/wcag/index.html
```diff
@@ -0,0 +1,168 @@
...
+                  <model-viewer
+                    id="reveal"
+                    loading="eager"
+                    camera-controls
+                    touch-action="pan-y"
+                    auto-rotate
+                    src="../../shared-assets/models/Astronaut.glb"
+                    shadow-intensity="1"
+                    alt="A 3D model of a shishkebab"
+                    a11y="../../assets/a11y.json"
```
elalish, line 80: Can you expand this example to demonstrate changing the locality dynamically, since that's probably what any user would be trying to do?

packages/model-viewer/src/features/wcag.ts
```diff
@@ -0,0 +1,128 @@
...
+        'camera-change', this[$onCameraChange] as EventListener);
+  }
+
+  [$onCameraChange] = (event: CustomEvent<CameraChangeDetails>) => {
+    if (event.detail.source === 'user-interaction') {
+      this[$updateAria]();
+    }
+  };
+
+  [$updateAria]() {
```
elalish, line 90: Considering other aria functionality is still left behind in controls.ts, perhaps we should just merge this file in there? I'm open to persuasion, but in general I prefer not to add new mixins if possible.

packages/modelviewer.dev/examples/wcag/index.html
```diff
@@ -0,0 +1,168 @@
...
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+-->
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>&lt;model-viewer&gt; WCAG</title>
```
elalish, line 20: A whole new page with one example seems a bit much - might be better to merge this into one of the existing example categories.

packages/modelviewer.dev/assets/a11y.json
cybmiper: Found it easier to test just adding a json file that was loaded in. Will implement so that it sends in the json structure directly. I agree that it would be preferable to do that. Not sure if we should keep that you can load in a json structure?

elalish: Hmm, that wasn't quite what I was referring to. I meant, this change is supposedly to allow multi-language support, right? But the example is all in English. I would like at least two different languages with a button to switch, so that all the code is there for a complete usage of the feature. Probably switch the alt text at the same time. Unless I'm missing something about how localization is generally accomplished on websites?

cybmiper: Will add so that there is a examples of multiple languages, that you can switch between on the example page.

packages/modelviewer.dev/examples/loading/index.html
```diff
@@ -465,6 +465,104 @@ <h4></h4>
...
+                    back: 'The back of an 3D modeled astronaut showing its backpack with environmental system', 
+                    left: 'The left side of an 3D modeled astronaut showing mission patch', 
+                    right: 'The right side of an 3D modeled astronaut usally has the flag of the country of the astronaut',
+                    'interaction-prompt': 'Use mouse, touch or arrow keys to move'
+                  }, 'A 3D model of a astronaut')">English <span class="a11y-icon">🇺🇸</span></button>
+                  <button class="a11y-btn" onclick="switchA11y(this, {
+                    front: 'Framsidan av den 3D-modellerade astronauten, med portar för olika utrustningar och några kontroller för dräkten', 
+                    back: 'Baksidan av den 3D-modellerad astronauten som visar sin ryggsäck med miljösystemet', 
+                    left: 'Den vänstra sidan av den 3D-modellerade astronauten, som visar uppdragslappen', 
+                    right: 'Den högra sidan av den 3D-modellerade astronauten, har vanligtvis flaggan för astronautens land',
```
elalish, line 506: I think this will result in untranslated default strings for the upper and lower views, right? Can we go ahead and include those too?

elalish (changes requested):
Looking good, thanks! Just a few details to clear up.

packages/model-viewer/src/features/controls.ts
```diff
@@ -809,23 +835,34 @@ export const ControlsMixin = <T extends Constructor<ModelViewerElementBase>>(
...
       const azimuthalQuadrantLabel =
           AZIMUTHAL_QUADRANT_LABELS[azimuthalQuadrant];
       const polarTrientLabel = POLAR_TRIENT_LABELS[polarTrient];
-
-      this[$updateStatus](
-          `View from stage ${polarTrientLabel}${azimuthalQuadrantLabel}`);
+      const position = `${polarTrientLabel}${azimuthalQuadrantLabel}`;
+  
+      const key = position as keyof A11yTranslationsInterface;
+      if (key in this[$a11y]) {
```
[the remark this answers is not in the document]
cybmiper: Yes, thought that maybe you only wanted a better description on a certain case and could use it for this.

elalish (approved):
This looks excellent - thank you for the tests!

PR #4716, opened 2024-03-15
Make Khronos PBR Neutral invertable

elalish:
Feedback from Autodesk was that the ability to analytically invert a tone mapping function has a lot of important use cases through the industry. It turns out my tone mapper was quite difficult to invert, but a very small and almost imperceptible change made it trivial to invert. The key is making the desaturation step not change the compressed brightness value, so instead of mixing toward pure white `[1, 1, 1]`, instead go toward `peak * [1, 1, 1]`. Anywhere the mix is significant, `peak` is already very close to 1, so this is a minor change, but it helps enormously with invertibility since these asymptotes become so sensitive when inverted. 

I've included an analytical inverse function in `lut-writer.mjs` and verified the round-trip relative error is less than `2e-10`.

### DIFF
--- a/packages/model-viewer/src/three-components/Renderer.ts
+++ b/packages/model-viewer/src/three-components/Renderer.ts
@@ -158,7 +158,7 @@ export class Renderer extends
         color *= newPeak / peak;
 
         float g = 1. - 1. / (desaturation * (peak - newPeak) + 1.);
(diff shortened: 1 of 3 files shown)


elalish:
See if you can tell the difference.
Previous:
image

Updated:
image

PR #4745, opened 2024-04-12
Fix Postprocessing for r163

elalish:
Needed to bump the postprocessing version to keep up with the removal of deprecated encodings in Three.js. However, now we have a new problem: 
```
postprocessing.js:274 Uncaught TypeError: Cannot read properties of undefined (reading 'Camera')
```
At this example: https://modelviewer.dev/examples/postprocessing/index.html#custom-effects

@Beilinson would you mind taking a look? I'm guessing this is related to your comment, which might not be true anymore:
>`// The camera is set automatically when added to the <effect-composer>`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2842,9 +2842,9 @@
       }
     },
     "node_modules/@types/express-serve-static-core": {
-      "version": "4.17.43",
-      "resolved": "https://registry.npmjs.org/@types/express-serve-static-core/-/express-serve-static-core-4.17.43.tgz",
-      "integrity": "sha512-oaYtiBirUOPQGSWNGPWnzyAFJ0BP3cwvN4oWZQY+zUBwpVIGsKUkpBpSztp74drYcjavs7SKFZ4DX1V2QeN8rg==",
+      "version": "4.19.0",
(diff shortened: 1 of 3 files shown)


Beilinson:
Sure, I'll look into it

Beilinson:
seems at some point `three.js` stopped putting a global `window.THREE` variable, which postprocessing still expects when imported using a non-esm build. solutions:
1. import from `https://cdn.jsdelivr.net/npm/postprocessing@6.35.3/build/index.js` instead (which is esm built so it runs `import ... from "three"` as expected), however it isn't minified and is about 2x as large as `postprocessing.min.js`
2. set the global THREE variable manually before the postprocessing import:
```ts
import * as THREE from "three";
window.THREE = THREE;
```
however this seems to cause all sorts of other problems, mainly that the minified file doesn't seem to export anything anymore...

elalish:
> 1. import from `https://cdn.jsdelivr.net/npm/postprocessing@6.35.3/build/index.js` instead (which is esm built so it runs `import ... from "three"` as expected), however it isn't minified and is about 2x as large as `postprocessing.min.js`

This sounds good to me - I'm not concerned about the library size for the example, as someone using it for real would likely make their own bundle anyway. I'll update next week. Thanks!

PR #4748, opened 2024-04-13
 parse_hotspot_config.ts DOM text reinterpreted as HTML 

Shivam7-1:


elalish (changes requested):
It looks like TS is complaining because it's not sure if the query you did is returning the right kind of element. Should be a simple casting fix.

Shivam7-1:
Hi @elalish
 Thanks For Reviewing 
I had done changes to it accordingly now it will work fine
Thanks

elalish (approved):
Thanks!

PR #4752, opened 2024-04-15
fix grounded skybox rotation

elalish:
Unlike the normal skybox, the grounded skybox was rotating with the model during `auto-rotate`, which is inconsistent and caused the background to be not in the same place as the applied environment lighting, since that was correctly not rotating. 

This now means the model moves relative to the ground when panning or changing the turntable rotation. I don't totally love that, but it's the only way to be consistent with our existing API.


ArchieBattley:
I understand why this had to be done to keep consistency. However, in the future will the auto-rotate and pan functions be separated. eg. could there be an option to stick the model to the ground plane when panning? Or would it involve a rewrite of the current camera functionality.

elalish:
The real problem is that pan operates by affecting `camera-target` - which is what makes pan emulatable via JS. That's an important feature, but `camera-target` is about centering the object as well, which should certainly keep the skybox centered. I'm somewhat regretting now that these things are so linked, but this also only affects a pretty narrow use case. I think I'm going to leave it as is.

PR #4756, opened 2024-04-17
Update sidebar.ts DOM text reinterpreted as HTML

Shivam7-1:

And Also Added HTMLElement to Recognise  innertext Element in it


Shivam7-1:
Hi @elalish Could You Please Review This PR 
Regards

Shivam7-1:
Hi @elalish Thanks For Reply 
I think it won't cause any test Failure 
Could you Check is there anything else Required 
Regards

elalish:
>I think it won't cause any test Failure

It is *literally* causing a build failure - it can't even get as far as running the tests. Do you see the Red X below? Please click on the details, understand the problem, then fix it.

PR #4757, opened 2024-04-17
Update create-html.ts DOM text reinterpreted as HTML

Shivam7-1:


Shivam7-1:
Hi @elalish Could You Please Review This PR 
Regards

elalish:
Have you tested the docs pages with this change to verify nothing has regressed?

Shivam7-1:
Hi @elalish Thanks For Replying 
I think it shouldn't cause any issue 
Regards

elalish:
>I think it shouldn't cause any issue

This kind of sentiment makes me very nervous. Please provide screenshots.

Shivam7-1:
Hi @elalish Thanks For Reviewing 
 It Passes All Test As i check Here Unit test and fidelity test also
Regards 
image
image

elalish:
Sorry, I may have been imprecise - the docs pages don't have automated tests - you have to manually look at them. What I mean is `npm run serve` and look at them to ensure they aren't broken. e.g.
image

NeilFraser:
This PR would break this product since raw HTML would be printed on the user's screen.

PR #4787, opened 2024-05-21
Upgrade to latest threejs (r164)

### LINKED ISSUE #4786: Upgrade to latest version of threejs (r164.1)
### Description

Upgrade to the latest version of ThreeJS

### Version

- model-viewer: v3.5.0

### Browser Affected

- [x] Chrome, version: xx.x.xxxx.xx
- [x] Edge
- [x] Firefox
- [x] IE
- [x] Safari

### OS

- [x] Android
- [x] iOS
- [x] Linux
- [x] MacOS
- [x] Windows

### AR

- [ ] WebXR
- [ ] SceneViewer
- [ ] QuickLook

aardvarkk:
### Reference Issue

Fixes #4786

### DIFF
--- a/packages/model-viewer/src/features/ar.ts
+++ b/packages/model-viewer/src/features/ar.ts
@@ -446,7 +446,7 @@ configuration or device capabilities');
       model.position.copy(target.position);
       model.updateWorldMatrix(false, true);
 
-      const arraybuffer = await exporter.parse(model);
+      const arraybuffer = await exporter.parseAsync(model);
 
       model.position.set(0, 0, 0);
       target.add(model);
--- a/package-lock.json
+++ b/package-lock.json
@@ -3123,9 +3123,9 @@
       }
     },
     "node_modules/@types/three": {
-      "version": "0.163.0",
-      "resolved": "https://registry.npmjs.org/@types/three/-/three-0.163.0.tgz",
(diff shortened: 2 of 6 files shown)


elalish (approved):
Thanks!

packages/model-viewer/src/features/ar.ts
```diff
@@ -446,7 +446,7 @@ configuration or device capabilities');
       model.position.copy(target.position);
       model.updateWorldMatrix(false, true);
 
-      const arraybuffer = await exporter.parse(model);
+      const arraybuffer = await exporter.parseAsync(model);
```
elalish, line 449: Actually, I believe there's another `exporter.parse` in `scene-graph.ts` as well - could you update that as well for consistency?

aardvarkk: Yes, I'll take a look.

aardvarkk: @elalish I don't believe the call in `scene-graph.ts` needs to be altered as it's using the synchronous callback-based version of `parse`. `parse` previously used to support both a callback and Promise-based overload, but now the asynchronous version is called `parseAsync`. `ar.ts` await-ed the return, so it needed to be adjusted to use the new async version, but the call to `parse` in `scene-graph.ts` is callback-based.

elalish: okay, thanks.

PR #4824, opened 2024-06-27
Fix: tone mapping dropdown used the wrong value (5: Custom instead of 7: Neutral)

hybridherbst:
Fixes https://github.com/google/model-viewer/issues/4804


elalish (approved):
Thank you! I read your bug too fast and repro'd on the wrong page.

PR #4851, opened 2024-08-07
Fixes code example

Hamidrzash:
Fixed code example for pickMaterialExample to work in webXR

About #4847

### DIFF
--- a/packages/modelviewer.dev/examples/scenegraph/index.html
+++ b/packages/modelviewer.dev/examples/scenegraph/index.html
@@ -759,10 +759,12 @@ const modelViewer = document.querySelector("model-viewer#pickMaterial");
 
 modelViewer.addEventListener("load", () => {
   const changeColor = (event) => {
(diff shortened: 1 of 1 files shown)


packages/modelviewer.dev/examples/scenegraph/index.html
```diff
@@ -766,7 +766,7 @@ <h2 class="demo-title">Materials API</h2>
     }
   };
 
-  modelViewer.addEventListener("click", changeColor);
+  document.addEventListener("click", changeColor);
```
elalish, line 769: Have you tested this whole page? I'm guessing since there are multiple examples, this might have odd behavior. Maybe add a check in `changeColor` to ensure the event came through the right element?

Hamidrzash: > add a check in `changeColor` to ensure the event came through the right element?

You are right. i think my new commit fixes this

elalish (approved):
Thanks!

PR #4890, opened 2024-09-19
Set the aces value in the tone-mapping item

rsax:
The “aces” option in the tone mapping selector of the editor is missing a specified value.


elalish (approved):
Good catch, thank you! Would you mind signing the CLA? Should be automatic for an individual.

elalish:
@googlebot I fixed it.

elalish:
@rsax did you have any trouble signing the CLA? I can't take your code until you do. It's by the email associated with your commit: `@rsax <rs*****al​@plasmout.com>`.

rsax:
You’re right, I signed with a different email. Please revalidate.

PR #4898, opened 2024-09-24
update render-fidelity dependencies

elalish:


packages/render-fidelity-tools/test/config.json
```diff
@@ -174,7 +174,10 @@
       "dimensions": {
         "height": 800,
         "width": 550
-      }
+      },
+      "exclude": [
+        "model-viewer"
```
elalish, line 179: These two tests reliably hit Puppeteer's protocol timeout on the CI, but work just fine locally. I assume this must be some kind of bizarre puppeteer regression, but since I can't find any way around it, this is the best I can do for now.

PR #4916, opened 2024-10-23
allow top and bottom view by default

elalish:
As this updates our defaults, it is technically a breaking change, though very minor. Our default camera limits used to not allow a pure top or bottom view - they had to be manually extended to get the full 0-180 range of camera elevation. Now that is our default, but you can always set them back to how they were if you desire the legacy behavior.


packages/modelviewer.dev/data/docs.json
```diff
@@ -627,7 +627,7 @@
         "htmlName": "minCameraOrbit",
         "description": "Set the minimum orbital values of the camera. Note \"Infinity\" is not an accepted keyword, but the default can still be obtained by passing \"auto\". The radius value for \"auto\" is a conservative value to ensure the camera never enters the model, so be careful when setting this to another value.",
         "default": {
-          "default": "-Infinity 22.5deg auto",
```
elalish, line 630: These are the legacy defaults, if anyone needs the old behavior.

PR #4939, opened 2024-11-13
[WIP] switch to WebGPU

elalish:
Now that Three.js supports WebGPU, and better yet, `WebGPURenderer` automatically falls back to `WebGL2Renderer` on unsupported platforms, it's probably about time we started migrating. Amazingly, it sort of already works! Main problems I've noticed so far are with our custom shaders: 

1) custom blur shader for the generated environment. I think we can delete all this code and start using `getTextureLevel` instead, but that means we'll also have to transition from `scene.environment` to scene.environmentNode`. 

2) custom shadow shader. Just rewrite this in TSL - shouldn't be too complex.

We'll see how the rest goes... FYI @mrdoob.

### DIFF
--- a/packages/model-viewer/src/three-components/TextureUtils.ts
+++ b/packages/model-viewer/src/three-components/TextureUtils.ts
@@ -16,6 +16,7 @@
 import {GainMapDecoderMaterial, HDRJPGLoader, QuadRenderer} from '@monogrid/gainmap-js';
 import {BackSide, BoxGeometry, CubeCamera, CubeTexture, DataTexture, EquirectangularReflectionMapping, HalfFloatType, LinearSRGBColorSpace, Loader, Mesh, NoBlending, NoToneMapping, RGBAFormat, Scene, ShaderMaterial, SRGBColorSpace, Texture, TextureLoader, Vector3, WebGLCubeRenderTarget, WebGLRenderer} from 'three';
 import {RGBELoader} from 'three/examples/jsm/loaders/RGBELoader.js';
+import {WebGPURenderer} from 'three/webgpu';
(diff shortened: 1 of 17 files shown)


CDerrell-ADT:
Does it help if I say I don't care about shaders much (for now) for my use cases? Out of curiosity

CDerrell-ADT:
Ahh, okay got you, very important then. I'll hold off until then, was
expecting a uniformly lit model without it.

*Christopher Derrell*


*Head of Web Development*

<https://adtelligent.net/>

4 Sylvan Avenue, Kingston 5, Jamaica

(W) 1 (876) 616 8756 | (C) 1 (876) 567 7884

packages/model-viewer/src/three-components/TextureUtils.ts
```diff
@@ -222,29 +219,35 @@ export default class TextureUtils {
...
     renderer.outputColorSpace = LinearSRGBColorSpace;
 
-    cubeCamera.update(renderer, scene);
+    cubeCamera.update(renderer as unknown as WebGLRenderer, scene);
 
-    this.blurCubemap(cubeTarget, GENERATED_SIGMA);
+    const cubeTarget = new CubeRenderTarget(256);
+    (cubeTarget as any)
+        .fromCubeTexture(
+            this.threeRenderer, tmpTarget.texture, GENERATED_SIGMA);
```
elalish, line 241: Here I'm relying on a WIP three.js PR: https://github.com/mrdoob/three.js/pull/30165. There's still some problem, probably in that PR, because randomly the generated environment either throws or creates a black texture.

Issue #4943, opened 2024-11-16
Non-manifold mesh shadow issues


swift502:
If a mesh doesn't have a bottom, it won't cast a shadow, which at first glance doesn't make any sense.

![Image](https://github.com/user-attachments/assets/c9454ff1-e19d-4668-850c-2887d5de25a6)

It seems shadows are only cast from walls facing away from the light source.

![Image](https://github.com/user-attachments/assets/99826bf4-9d15-46cc-877f-a6ca71c19af5)

Is this intentional?
Could an option for "FrontSide" shadows be introduced?

Thanks.

GLB for testing here. [lamp.zip](https://github.com/user-attachments/files/17784581/lamp.zip)

swift502:
We'd just have to change `side` of the depthMaterial somewhere [around here](https://github.com/google/model-viewer/blob/dd354f6b4ce66419db0423bd62df1972596d26fa/packages/model-viewer/src/three-components/Shadow.ts#L105C5-L105C23). Either Back or Both.

```js
this.depthMaterial.side = THREE.BackSide;
```

Please correct me if I'm wrong. But it makes sense to me that if the light source is up, faces pointing _towards_ it should cast shadows, not faces pointing _away_ from it.

PR #4950, opened 2024-11-19
Improve shadow robustness

swift502:
### Reference Issue
#4943

Make objects cast shadows from top facing geometry. Specifically, this improves the appearance of hard shadows cast from non-manifold geometry, which has holes on the bottom.

The softer shadows become, the less effect this change has.

Before, after:
image
image
image

# Testing

~~I didn't actually manage to get the build setup on windows, so I have no idea if the code works. Let me know if there's an easy way to build/deploy so I can confirm functionality.~~

### DIFF
--- a/packages/model-viewer/src/three-components/Shadow.ts
+++ b/packages/model-viewer/src/three-components/Shadow.ts
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


swift502:
I'll try to get a linux VM and test the changes

swift502:
I tested in a Linux VM and seems to work 👌

image

elalish (approved):
Good call, thanks!

Issue #4942, opened 2024-11-16
Documentation improvements


swift502:
Hi, long time model viewer user. I recently updated to v4 and faced some frustrations. Maybe we can improve on it.

I'd be happy to make PRs on things you'd approve of.

## NPM badge

I struggled to find any npm project link in the readme.

The existing links go to npmjs.org, which feels pointless.

![Image](https://github.com/user-attachments/assets/ef2ce3d6-46e3-44f6-9c57-3c087c4e4957)

- Could we add an NPM badge in the readme?
- Could we add an NPM badge in https://modelviewer.dev?

## NPM install instructions are incomplete

I couldn't find the information that this line has to be included anywhere. Readme, Docs, Examples.

```js
import '@google/model-viewer';
```

I tried `import ModelViewer from '@google/model-viewer';` and `import { Loader } from '@google/model-viewer';` before chatgpt told me to try the corrent solution.

- Can we include this instruction?

## NPM install instructions are misleading?

Instructions say I have to install three peer dependency. It was added here https://github.com/google/model-viewer/issues/4197
But in my testing, this is **not** necessary. My model viewer works without three.

- Is this instruction still correct?

## Naming consistency

Why is documentation sometimes called documentation and other times API Reference?

![Image](https://github.com/user-attachments/assets/db72b67e-eccd-47be-8f10-c75206f2369d)

![Image](https://github.com/user-attachments/assets/c430877f-fa08-46b6-8f52-86a5cfaec477)

I'm always confused by this and struggle to find what I'm looking for.

- Can we unify the two names?

## Some examples only include JS setup

I found the tonemapping example

https://modelviewer.dev/examples/lightingandenv/#toneMapping

But it doesn't say how to use model-viewer HTML attributes to set the property. I had to find the equivalent page in the documentation (API Reference) to learn that it's `tone-mapping="aces"` and not `toneMapping="aces"`

- Is this desirable?
- Could we maybe just provide an API Reference link to every example?

## Focus outline

After 5 years, I've come to accept that the outline on keypress is probably here to stay?

https://github.com/user-attachments/assets/4d586ecd-a471-47f1-8098-ed244ac49c7a

- Can we explain somewhere that this **_isn't_** a bug?
- Can we explain the accessibility, or whatever it is, reasons on why this has to exist?
- Can we include the simple workaround, which removes the outline, in the docs?

```js
// Disable focus outline on keypress
const modelViewer = document.querySelector('model-viewer');

const style = document.createElement('style');
[...]

PR #4951, opened 2024-11-19
Docs: Add NPM badges for <model-viewer>

swift502:
### Reference Issue
#4942

### NPM Badges

I believe these badges will better communicate to users that using model-viewer through NPM is a viable option.


elalish (approved):
Agreed, thanks!

PR #4952, opened 2024-11-19
Docs: Improve install instructions

swift502:
### Reference Issue
#4942

Show how to import the NPM package in a project. Added NPM and CDN headings for clarity.

There's still lots of information missing, when comparing the instructions to [those of three.js](https://threejs.org/docs/#manual/en/introduction/Installation). They have an entire page dedicated to every involved concept.

But I think this is a small improvement.
Let me know what you think.


elalish (approved):
Sure, that looks reasonable.

PR #5020, opened 2025-04-10
✨ New Features & Improvements:

samaneh-kazemi:
Update WebXR functionalities. 

**1. Single-controller rotation**
When a single controller intersects with the model (excluding the footprint) and moves horizontally, the model now rotates in the same direction as the controller movement.

**2. Single-controller translation**
When a controller intersects with the footprint, the user can now translate/move the model across the surface.

**3. Dual-controller scaling**
When both controllers intersect the model, the user can now scale the model using a pinch-like gesture (distance-based scaling).

### DIFF
--- a/packages/model-viewer/src/three-components/ARRenderer.ts
+++ b/packages/model-viewer/src/three-components/ARRenderer.ts
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import {BoxGeometry, BufferGeometry, Event as ThreeEvent, EventDispatcher, Line, Matrix4, Mesh, PerspectiveCamera, Quaternion, Vector3, WebGLRenderer, XRControllerEventType, XRTargetRaySpace} from 'three';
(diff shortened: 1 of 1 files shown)


packages/model-viewer/src/three-components/ARRenderer.ts
```diff
@@ -279,6 +287,7 @@ export class ARRenderer extends EventDispatcher<
         });
 
     if (this.xrMode !== 'screen-space') {
+      console.log("present(", scene);
```
diegoteran, line 290: nit: Delete console log.

samaneh-kazemi:
Hi @elalish, hope you're well. I noticed our recent commits to main cause [deploy_github_pages](https://github.com/google/model-viewer/actions/runs/14451903551/job/40596270052#logs) to fail. Could it be that Diego and I are not the owner? Seems like I needed to add my classic token to the repo, but I'm not seeing the Settings tab.

PR #5025, opened 2025-04-15
feat: make options for appendAnimation and detachAnimation optional

Frank3K:
# Description

This PR makes all parameters of `AppendAnimationOptions` and `DetachAnimationOptions` optional.

Such that the following is possible:

```ts
// Example: Only specify repetitions
appendAnimation('some-animation', { repetitions: 5 });

// Example: Only specify fade for detach
detachAnimation('some-animation', { fade: 200 });
```

Previously, you had to pass _all_ options, or no options at all (in case the default was used). Current code in `master`:

https://github.com/google/model-viewer/blob/4efaf8eebb8a9b48080b87faa39faa544a728f92/packages/model-viewer/src/features/animation.ts#L33-L41

@mohammadbaghaei Would you be so kind to do the review? I tried to use appendAnimation, but then I was a bit surprised I needed to pass all parameters. I have no real experience with these methods, so it would be nice if you can take a look and functionally test.

### DIFF
--- a/packages/model-viewer-effects/src/test/effect-composer-spec.ts
+++ b/packages/model-viewer-effects/src/test/effect-composer-spec.ts
@@ -18,6 +18,8 @@ import {expect} from 'chai';
 import {DotScreenEffect, Effect, EffectPass, GridEffect} from 'postprocessing';
 import {Camera} from 'three';
 
+import {Renderer} from '@google/model-viewer/lib/three-components/Renderer.js';
+
 import {$effectComposer, $normalPass, $renderPass, $scene} from '../effect-composer.js';
 import {EffectComposer} from '../model-viewer-effects.js';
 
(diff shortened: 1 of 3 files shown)


packages/model-viewer/src/features/animation.ts
```diff
@@ -260,27 +266,28 @@ export const AnimationMixin = <T extends Constructor<ModelViewerElementBase>>(
...
+[$appendAnimation](
+    animationName: string = '',
+    options: AppendAnimationOptions = {}) {
+  const opts = {...DEFAULT_APPEND_OPTIONS, ...options};
+  const repetitions = opts.repetitions ?? Infinity;
+  const mode = opts.pingpong ?
+      LoopPingPong :
+      (repetitions === 1 ? LoopOnce : LoopRepeat);
+
+  const needsToStop = repetitions !== Infinity || mode !== LoopPingPong;
```
Frank3K, line 278: @mohammadbaghaei I am particularly not 100% sure about this line.

diegoteran:
Thanks for requesting these changes, I'll need to test them before submitting

diegoteran:
I believe these are all optional already. 

If you look at the example in https://modelviewer.dev/examples/animation/index.html#appendAnimation

Do you mind checking again @Frank3K ? In case this change is not necessary.

Issue #5026, opened 2025-04-15
Backwards animation does not work anymore since model-viewer 4.1.0 release


Frank3K:
### Description
Since the upgrade to model-viewer 4.1.0, animations with a negative `timeScale` (playing in reverse) no longer function correctly. This worked properly in version 4.0.0.

### Expected Behavior
The animation should play in reverse (from end to start).

### Actual Behavior
The animation does not play when `timeScale` is negative.

### Reproduction
I've created a JSFiddle that demonstrates the issue: https://jsfiddle.net/d8r2bmqL/. You can change the model-viewer version from 4.1.0 to 4.0.0 to see that backwards animation worked correctly in the previous version.

The demo uses the `RobotExpressive.glb` model with the "Death" animation. When clicking the play button, it should alternate between playing forward and backward.

### Possible Cause
I believe the issue may be related to the following code:

https://github.com/google/model-viewer/blob/4efaf8eebb8a9b48080b87faa39faa544a728f92/packages/model-viewer/src/three-components/ModelScene.ts#L756-L759

Commenting out this code makes the backwards animation work again. I'm not entirely sure why this code was implemented in the first place, but it appears to be causing the reverse animation functionality to break in version 4.1.0.

@mohammadbaghaei  - Since you worked on the animation code for 4.1.0. Would you have time to take a look at this issue? Your expertise would be appreciated in understanding why this behavior changed and how we can restore the backward animation functionality.

### Environment
- model-viewer version: 4.1.0
- Browser: Firefox 137.0.1 (also reproduced on Chrome)
- Operating System: macOS (also reproduced on Windows)

PR #5028, opened 2025-04-16
fix reverse animation play and improve code structure

mohammadbaghaei:


packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -935,7 +894,29 @@ export class ModelScene extends Scene {
...
+        return { shouldFade: true, duration: defaultDuration };
+      } else if (fade.toLowerCase().trim() === 'false') {
+        return { shouldFade: false, duration: 0 };
+      } else if (!isNaN(parseFloat(fade))) {
+        const duration = parseFloat(fade);
+        return { shouldFade: duration > 0, duration: Math.max(duration, 0) };
+      }
+    }
+    console.warn(`Invalid fade value, fade is set to ${defaultValue}`);
+    return { shouldFade: defaultValue, duration: defaultValue ? defaultDuration : 0 };
```
samaneh-kazemi, line 916: Once you return from an if block, there’s no need to continue chaining else ifs. Maybe update it to sth like this which is more readable. 

``` typescript
private parseFadeValue(
  fade: boolean | number | string,
  defaultValue: boolean = true,
  defaultDuration: number = 1.5
): { shouldFade: boolean; duration: number } {
  const normalizeString = (str: string) => str.toLowerCase().trim();

  if (typeof fade === 'boolean') {
    return { shouldFade: fade, duration: fade ? defaultDuration : 0 };
  }

  if (typeof fade === 'number') {
    const duration = Math.max(fade, 0);
    return { shouldFade: duration > 0, duration };
  }

  if (typeof fade === 'string') {
    const normalized = normalizeString(fade);

    if (normalized === 'true') {
      return { shouldFade: true, duration: defaultDuration };
    }

    if (normalized === 'false') {
      return { shouldFade: false, duration: 0 };
    }

    const parsed = parseFloat(normalized);
    if (!isNaN(parsed)) {
      const duration = Math.max(parsed, 0);
      return { shouldFade: duration > 0, duration };
    }
  }

  console.warn(`Invalid fade value: ${fade}. Using default: ${defaultValue}`);
  return {
    shouldFade: defaultValue,
    duration: defaultValue ? defaultDuration : 0
  };
}
```

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -787,9 +782,9 @@ export class ModelScene extends Scene {
   appendAnimation(
       name: string = '', loopMode: AnimationActionLoopStyles = LoopRepeat,
       repetitionCount: number = Infinity, weight: number = 1,
-      timeScale: number = 1, fade: boolean|number = false,
-      warp: boolean|number = false, relativeWarp: boolean = true,
-      time: null|number = null, needsToStop: boolean = false) {
+      timeScale: number = 1, fade: boolean|number|string = false,
```
samaneh-kazemi, line 785: Could you share a bit more about why you think this change will resolve the issue? Was this type removed in an earlier PR?

packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -787,9 +782,9 @@ export class ModelScene extends Scene {
   appendAnimation(
```
samaneh-kazemi, line 782: It'd be great if you could add unit test for the logic you're changing. 

packages/model-viewer/src/test/three-components/ModelScene-spec.ts

packages/model-viewer/src/three-components/ModelScene.ts
mohammadbaghaei: That was a good point, thank you.

packages/model-viewer/src/three-components/ModelScene.ts
mohammadbaghaei: Actually, the issue wasn’t with this part. The function is designed to be type-flexible to prevent errors caused by unintentional input mistakes. I think it would be helpful to mention that it can accept different input types as well.

Additionally, @Frank3K correctly identified the underlying cause in issue #5026, which turned out to be a mistake on my part. I appreciate the accurate observation.

packages/model-viewer/src/three-components/ModelScene.ts
mohammadbaghaei: Sure, I will add it soon.

mohammadbaghaei:
By the way, I’d appreciate it if you could take a look at the deployment issue with [modelviewer.dev](https://modelviewer.dev/) — it seems like it hasn’t been working recently.

samaneh-kazemi:
> Also, I’d appreciate it if you could take a look at the GitHub Pages deployment for modelviewer.dev. It seems that new changes aren’t being deployed properly due to a token or permission issue.

Thanks for your patience. We’re actively reviewing what’s changed since the previous handoff and will update you shortly.

mohammadbaghaei:
> Hi, thanks for contributing to model-viewer. I noticed each of your commits address a different topic. I'd recommend to have separate PRs for each of those. Having separate PRs make it easier to review, test, and revert them individually.

Hi! Thanks for pointing that out — it's a great catch. I'll definitely keep it in mind for future contributions.
That said, I feel the overall review and PR merging process could be a bit faster. It would really help with keeping things moving smoothly.

davide-granello:
Hello @mohammadbaghaei i had the same issue as #5026 so i've tried using your fork but i still have the same issue.
When i play an animation backward, it seems the animation skips directly to the end (or the start of the forward animation).

This is how i play it forward and backward using a condition:
```ts
modelViewer.appendAnimation(animation.name, {repetitions: 1, timeScale: playForward ? 1 : -1})
```

samaneh-kazemi:
Hi @mohammadbaghaei, thanks for your thoughtful reply and for being open to feedback. I completely understand how a faster review and merging process can make a big difference for contributors, and I appreciate your patience.

This PR addresses both the reverse animation fix and some broader code structure improvements. To help with review and make merging easier, would you be able to split this into smaller PRs? For example, one PR could focus solely on fixing the reverse animation play issue, and another on the code refactoring. This way, each change can be reviewed and tested independently, which reduces risk and usually helps things move through the review process more quickly.

Also, when you’ve updated the PRs or have new ones ready for review, please feel free to ping me, @-mention me, or bump the thread so I know you’re ready for feedback. I’ll do my best to prioritize the review.

Thanks again for your contributions and for sharing your perspective!

davide-granello:
Hey @mohammadbaghaei i actually solved it in the meantime, i made some tests and found that i needed to set the time to `animation.duration` when playing it backwards.
So now my code looks like this:

```ts
modelViewer.value.appendAnimation(animation.name, {
    repetitions: 1,
    time: playForward ? 0 : animation.duration,
    timeScale: playForward ? 1 : -1
  })
```

Everything works fine.
Thank you

mohammadbaghaei:
Split into multiple PRs as requested. Closing this one.

PR #5049, opened 2025-06-05
Update xrpanel opacity to match SVXR

samaneh-kazemi:


StillOnMyWay:
How can I see a screenshot of what this will look like? The XR Panel


*Christopher Derrell*

samaneh-kazemi:
Here is an example. You can also try opening model-viewer in an XR device like Quest to test it on your end. 

Screenshot 2025-06-05 at 10 42 20 AM

PR #5101, opened 2025-11-11
fix #5026 issue and refactor ModelScene.ts

### LINKED ISSUE #5026: Backwards animation does not work anymore since model-viewer 4.1.0 release

mohammadbaghaei:


packages/model-viewer/src/three-components/ModelScene.ts
```diff
@@ -13,22 +13,22 @@
...
+import ModelViewerElementBase, { $renderer, $scene, EffectComposerInterface, RendererInterface } from '../model-viewer-base.js';
+import { ModelViewerElement } from '../model-viewer.js';
+import { normalizeUnit } from '../styles/conversions.js';
+import { NumberNode, parseExpressions } from '../styles/parsers.js';
+
+import { Damper, SETTLING_TIME } from './Damper.js';
+import { ModelViewerGLTFInstance } from './gltf-instance/ModelViewerGLTFInstance.js';
+import { GroundedSkybox } from './GroundedSkybox.js';
+import { Hotspot } from './Hotspot.js';
+import { Shadow } from './Shadow.js';
```
diegoteran: extra spaces everywhere

PR #5102, opened 2025-11-11
add version sync script

mohammadbaghaei:


diegoteran:
@mohammadbaghaei is this a script that will need to be run manually? there is no description at all for this PR

PR #5106, opened 2025-11-11
Upgrade dependencies

mohammadbaghaei:


Beilinson:
~The documentation for `model-viewer-effects` used to update automatically when the three+postprocessing versions changed to reflect that in the examples, quick check on the page it looks to be out of date so maybe that needs to be manually updated: https://modelviewer.dev/examples/postprocessing/index.html~

Nvm just saw #5102 😅

diegoteran:
I updated the dependencies in a separate PR

PR #5141, opened 2026-04-02
feat: support KTX2 textures in createTexture()

douglasholm:
### Reference Issue

Addresses Discussion #4002 — `createTexture()` does not support KTX2 textures.

### Summary

`createTexture()` currently uses Three.js's `TextureLoader` internally, which only handles browser-native image formats (PNG, JPEG, WebP). KTX2 textures — which use Basis Universal compression and are already supported for textures embedded in glTF/GLB files via `KHR_texture_basisu` — fail silently when passed to `createTexture()`.

This PR adds KTX2 support by routing `.ktx2` URLs through the existing `KTX2Loader` instance (already initialized in `CachingGLTFLoader` for glTF loading). The shared instance avoids duplicate WASM transcoder initialization.

### Changes

- **`TextureUtils.ts`**: Added `loadKTX2()` method and URL detection in `loadImage()` to route `.ktx2` URLs through `KTX2Loader`
- **`scene-graph.ts`**: KTX2 textures are loaded correctly via `loadImage()` URL detection; `userData.mimeType` is kept as `image/png` (the export format) since `GLTFExporter` cannot encode KTX2
- **`CachingGLTFLoader.ts`**: Exported the existing `ktx2Loader` instance for reuse
- **Tests**: Added test cases in both `TextureUtils-spec.ts` and `texture-spec.ts`
- **Fixture**: Added `CesiumLogoFlat.ktx2` test asset

### Known Limitations

Scenes containing KTX2-created textures cannot currently be exported via `exportScene()`. This is because Three.js's `GLTFExporter` needs `setTextureUtils()` to be called in order to decompress `CompressedTexture` instances before export, and model-viewer doesn't currently configure this when creating the exporter. Worth noting that this is a pre-existing gap that affects any `CompressedTexture` in the scene — it's not introduced by this change.

### Testing

- All 121 Chromium tests pass (0 failures)
- Pre-existing Firefox/WebKit timeouts are unrelated to this change

cc @diegoteran @samaneh-kazemi


diegoteran (approved):
Thank you, confirmed this works

PR #5143, opened 2026-04-03
Rename pivot to pivotEl to not conflict with new Object3D.pivot in three r183

vincentfretin:
Rename pivot to pivotEl to not conflict with new Object3D.pivot in three r183
r183 have that code now in Object3D.updateMatrix
https://github.com/mrdoob/three.js/blob/r183/src/core/Object3D.js#L1137-L1148
CSS2DObject inherits from Object3D
so the calculated hotspot position was [NaN, NaN, NaN] because pivot.x is undefined, pivot being a div element and not a Vector3.


diegoteran (approved):
thank you, Vincent

PR #5177, opened 2026-06-08
fix: apply orientation attribute to <extra-model> elements

Frank3K:
## Description

The multi-model feature introduced in #5134 is a really nice feature — this PR fixes a small gap in the initial implementation.

`<extra-model orientation="...">` had no effect. The `updateModelTransforms` method in `ModelScene.ts` accepted an `orientation` parameter but never applied it — it was prefixed `_orientation`, signalling intentional non-use. Only `offset` and `scale` were wired up, leaving models always in their default rotation regardless of the declared attribute.

## Root cause

`updateModelTransforms` ([source](https://github.com/Frank3K/model-viewer/blob/master/packages/model-viewer/src/three-components/ModelScene.ts#L325-L327)) received the orientation string from the `extra-model-changed` event but did not parse or apply it to the model's quaternion.

A secondary issue: the multimodel example used bare numbers (`orientation="0 30 0"`). 

## Changes

**`packages/model-viewer/src/three-components/ModelScene.ts`**
- Renamed `_orientation` → `orientation` and implemented parsing using the existing `parseExpressions` / `normalizeUnit` utilities — the same infrastructure used by `applyTransform`.
- Applies the result using the documented `$roll $pitch $yaw` convention: `model.quaternion.setFromEuler(new Euler(pitch, yaw, roll, 'YXZ'))`, consistent with the main `orientation` attribute behaviour.

**`packages/modelviewer.dev/examples/multimodel/index.html`**
- Updated orientation attribute values to use explicit `deg` units and correct roll/pitch/yaw ordering. A 30° yaw (Y-axis) rotation is now `"0deg 0deg 30deg"` (third term), matching the documented `$roll $pitch $yaw` format.

### Before

image

### After
 
image

Robot and Astronaut now rotate around their Y-axis as declared, facing slightly toward the scene center.

## Testing

Verified visually in the multimodel example page. The robot (left, `orientation="0deg 0deg 30deg"`) and the astronaut (right, `orientation="0deg 0deg -30deg"`) each rotate 30° around the Y-axis toward the center.

### DIFF
--- a/packages/model-viewer/src/test/features/extra-model-spec.ts
+++ b/packages/model-viewer/src/test/features/extra-model-spec.ts
@@ -78,6 +78,32 @@ suite('ExtraModel', () => {
       expect(scene._models[1].position.x).to.equal(5);
     });
 
+    test('applies orientation as rotation on the model quaternion', async () => {
+      element.loading = 'eager';
+      element.src = CUBE_GLB_PATH;
+
+      const extra = document.createElement('extra-model');
+      extra.setAttribute('src', CUBE_GLB_PATH);
+      // yaw = 90deg (third term in roll/pitch/yaw) → Euler(0, π/2, 0, 'YXZ')
+      // → quaternion.y ≈ 0.707, quaternion.w ≈ 0.707
+      extra.setAttribute('orientation', '0deg 0deg 90deg');
+      element.appendChild(extra);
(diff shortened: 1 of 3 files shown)


diegoteran (approved):
Thanks!

